PR #7, opened 2014-10-21
fixed compilation with clang for maces

bgaifullin:


steaphan-fb-com:
I am not sure of some of these changes.  For example, I don't see where you actually use "ssize_t len", and you seem to turn on O_DIRECT for some cases where it wasn't used before.

But, most of these changes look like good fixes (one of which seems like it's actually a bug fix).

I will include these portions of your fixes in our internal branch, so you should see them show up in our public branch within a few weeks.

Thanks!

bgaifullin:
I see, there is my glitch, I did not properly handle case with O_DIRECT, where I can fix it?

bgaifullin:
"ssize_t len" is used in macros MYSQL_NET_WRITE_START(len);

steaphan-fb-com:
Sorry, I was on leave for the last week.

These fixes actually interact with a fix that is being pushed internally right now.  It should be in the public branch too in a few days.  Once that is out, I will comment here, and then you can create an updated PR based on the newer version of this branch, and I will re-review that PR.

bgaifullin:
Ok

steaphan-fb-com:
This branch has just been updated, so try it out now and open a PR if you still have to make changes to build.  Thanks!

PR #14, opened 2015-02-07
Adding missing includes + renaming parameters to avoid compile errors with -Werror=shadow turned on

StanislavGlebik:
Some parameters name are the same as methods name. It can produce compile error with -Werror=shadow turned on.


tianx:
Thank you for your contribution! We have cherry-picked your pull request into our branch here:

https://github.com/facebook/mysql-5.6/commit/79fa5c2ebb3f4212cc26dda5dea9f182c551f457

PR #97, opened 2015-11-07
Fixes to give a clean git status after an unmodified build.

gdbentley:
These changes make it so that "git status" still reports clean after a build. I've removed a file that is auto-generated (and diffs from the version in the repo).


jkedgar:
@gdbentley, after reviewing your pull request, I can't approve it as it is.  The include/probes_mysql_nodtrace.h file is an Oracle file (upstream from us) and so it should be fixed by Oracle.  I would recommend you submit that part of the pull request to them.  Once they release it we would automatically get the change once we switched to whatever version they release it in.

If we start making these kinds of changes to MySQL we make our own rebases down the line more complicated.

The other change (adding '*.o' to the .gitignore) file would be fine and if you want to create a separate pull request for that I'll approve it.

gdbentley:
@jkedgar I've changed the pull request to only include the .gitignore changes.

jkedgar:
@gdbentley, I just noticed that your pull request went into 5.6.24.97.  We are no longer actively developing on that branch.  You may want to re-issue your pull request on 5.6.27.75.

PR #309, opened 2016-09-01
[regex/regexp.c] fix the memory leak

bryonglodencissp:
Greetings. On line number `336` of '[regex/regexp.c](https://github.com/facebook/mysql-5.6/blob/webscalesql-5.6.27.75/regex/regexp.c#L336)' there is a memory leak, which is an error. Memory is a finite resource within our running programs. Sure in very short running simple programs, failing to free memory won't have a noticeable effect. However on long running programs, failing to free memory means we will be consuming a finite resource without replenishing it. Eventually it will run out and our program will abruptly crash. This is why we must free memory.

### DIFF
--- a/regex/regexp.c
+++ b/regex/regexp.c
@@ -332,8 +332,10 @@ int		excompat;	/* \( \) operators like in unix ex */
     regnpar = 1;
     regcode = r->program;
     regc(MAGIC);
-    if (reg(0, &flags) == NULL)
+    if (reg(0, &flags) == NULL) {
+	free(r);
 	return ((regexp *) NULL);
+    }
 
     /* Dig out information for optimizations. */
     r->regstart = '\0';		/* Worst-case defaults. */


ghost:
Thank you for signing our Contributor License Agreement. We can now accept your code for this (and any) Facebook open source project. Thanks!

gunnarku:
Thank you for your contribution!

The `regex/regexp.c`file is an Oracle file (upstream from us) and so it should be fixed by Oracle. I would recommend you submit that part of the pull request to them. Once they release it we would automatically get the change once we switched to whatever version they release it in.

If we start making these kinds of changes to MySQL we make our own rebases down the line more complicated.

bryongloden:
Greetings @gunnarku! How are you? When we committed this patch to Facebook we were mindful that it may be part of a library that's upstream to you, but it didn't appear it was, so we believed it was maintained by Facebook. Do you know the code part of this repository that is maintained by Facebook, so we can submit bug fixes to the proper company? Thank you.

PR #318, opened 2016-09-19
Additional SHIP_ASSERT to validate internal consistency

gunnarku:
Test pull request. Not meant to be accepted.


storage/rocksdb/ha_rocksdb.cc
```diff
@@ -2966,6 +2966,8 @@ static int rocksdb_init_func(void *p)
 #endif
 
   rocksdb_hton= (handlerton *)p;
+  SHIP_ASSERT(rocksdb_hton != nullptr);
```
gunnarku: Do we actually need this?

gunnarku: We probably do.

PR #362, opened 2016-10-27
Making D65469 to exclude real Deadlock, adding more test coverage

yoshinorim:
Summary: D65469 should have excluded real deadlock (s.IsDeadlock() == true)
at ha_rocksdb::get_for_update(). This diff fixes it. This diff does
not actually change user behavior, but it's easier to understand.

This diff also increases test coverage for rocksdb_deadlock_\* test cases
-- testing both Read Committed and Repeatable Read.

Test Plan: mtr


storage/rocksdb/ha_rocksdb.cc
```diff
@@ -1539,8 +1539,7 @@ class Rdb_transaction
                                                 false /* just statement */);
       return HA_ERR_LOCK_DEADLOCK;
     }
-
-    if (s.IsBusy())
+    else if (s.IsBusy())
```
jkedgar, line 1542: For my information, why do you prefer this 'else' here?

yoshinorim: IsBusy() includes IsDeadlock(). On real deadlock (IsDeadlock == true), it's not a snapshot conflict error. So "(s.Isbusy() && !s.IsDeadlock())" or "else if (s.IsBusy())" is more accurate.

jkedgar: Right, but if isDeadlock() returns true then we would have returned already and never gotten to the `if (s.IsBusy())`.  I'm fine if you want the `else` - I'm just trying to understand why so if there is a good reason I can add it in the future.

yoshinorim: I wanted to use else to make sure it would never treat deadlock as snapshot conflict. Better fix would be introducing a function like IsSnapshotConflict() in rocksdb.

PR #366, opened 2016-10-28
Making D65469 to exclude real Deadlock, adding more test coverage

yoshinorim:
Summary:
D65469 should have excluded real deadlock (s.IsDeadlock() == true)
at ha_rocksdb::get_for_update(). This diff fixes it. This diff does
not actually change user behavior, but it's easier to understand.

This diff also increases test coverage for rocksdb_deadlock_\* test
cases
-- testing both Read Committed and Repeatable Read.

Squash with: https://reviews.facebook.net/D65469

Test Plan: mtr


yoshinorim:
Replaced with #372

PR #371, opened 2016-10-29
Replacement of volatile sig_atomic_t to atomic

anirbanr-fb:
Summary:
4 changes.
1. Shadow testing revealed that I was misled by sig_atomic_t
   which is only atomic between signal handlers. This leads to our
   global thread_count being not multi-thread safe. Fixed it with
   std:atomic.
2. Changed the code layout of block_until_new_connection making it
   a halflock pattern. this is to ensure that I don't cheat in the
   shutdown routine, by assuming that unlocked conditional variable
   notify will get caught. You need to hold the lock, although there
   are places in mysqld which this cheating happens.
3. The initialization of ShardedLocks and sharded lock init was
   happening too soon. It needs to happen after get_options, otherwise
   you cannot increase num_sharded_locks > 4.
4. made thread_created an atomic

Test Plan: sandcastle and connection rate benchmarking.

Reviewers: jkedgar

Subscribers: webscalesql-eng

Differential Revision: https://reviews.facebook.net/D65031

### DIFF
--- a/include/mpsc.h
+++ b/include/mpsc.h
@@ -89,8 +89,8 @@ public:
   mpsc_queue_t() :
     _head(reinterpret_cast<buffer_node_t*>(new buffer_node_aligned_t)),
     _tail(_head.load(std::memory_order_relaxed))
-#ifndef DBUG_OFF
-    , m_single_consumer_check(false)
+#if !defined(DBUG_OFF)
+    , dequeue_in_use(false)
 #endif
   {
     buffer_node_t* front = _head.load(std::memory_order_relaxed);
@@ -105,6 +105,7 @@ public:
     delete front;
   }
 
+  // multi -producer single consumer queue where enqueue is thread safe
   void enqueue(const T& input)
   {
     buffer_node_t* node =
@@ -116,33 +117,27 @@ public:
     prev_head->next.store(node, std::memory_order_release);
   }
 
+  // only 1 thread should dequeu this queue. We break the assumption only in
+  // the destructor where we expect all consumers have flushed out
+  // hence the assert.
   bool dequeue(T& output)
   {
-#ifndef DBUG_OFF
-    // this is a best effort check that deque happens
-    // in single thread to obey MPSC restrictions.
-    // There is a race condition which we ignore when
-    // the first time dequeue happens from multiple threads
-    // and the threads are trying to initialize m_single_consumer_check
-    if (!m_single_consumer_check)
-    {
-      thread_id = my_thread_dbug_id();
-      m_single_consumer_check = true;
-    } else
-      DBUG_ASSERT(thread_id == my_thread_dbug_id());
-#endif
+    // Make sure it was false before
+    DBUG_ASSERT(!dequeue_in_use.exchange(true));
 
     buffer_node_t* tail = _tail.load(std::memory_order_relaxed);
     buffer_node_t* next = tail->next.load(std::memory_order_acquire);
 
-    if (next == nullptr) {
-      return false;
+    bool result = false;
+    if (next != nullptr) {
+      output = next->data;
+      _tail.store(next, std::memory_order_release);
+      delete tail;
+      result = true;
     }
 
-    output = next->data;
-    _tail.store(next, std::memory_order_release);
-    delete tail;
-    return true;
+    DBUG_ASSERT(dequeue_in_use.exchange(false)); // Make sure it was true before
+    return result;
   }
 
 private:
@@ -156,9 +151,8 @@ private:
   mpsc_queue_t(const mpsc_queue_t&)=delete;
   void operator=(const mpsc_queue_t&)=delete;
 
-#ifndef DBUG_OFF
-  bool m_single_consumer_check;
-  my_thread_id thread_id;
+#if !defined(DBUG_OFF)
+  std::atomic<bool> dequeue_in_use;
 #endif
 };
 
--- a/mysql-test/include/rpl_stop_server.inc
+++ b/mysql-test/include/rpl_stop_server.inc
@@ -50,8 +50,8 @@ if ($rpl_debug)
 --exec echo "wait" > $MYSQLTEST_VARDIR/tmp/mysqld.$rpl_server_number.expect
 
 # Send shutdown to the connected server and give
-# it 14 seconds to die before zapping it
---let $shutdown_server_timeout= 14
+# it 30 seconds to die before zapping it
+--let $shutdown_server_timeout= 30
 if ($rpl_force_stop)
 {
    # if force_stop is set, shutdown server immediately(equal to killing the server)
--- a/sql/mysqld.cc
+++ b/sql/mysqld.cc
@@ -418,7 +418,8 @@ volatile bool ready_to_exit;
 static my_bool opt_debugging= 0, opt_external_locking= 0, opt_console= 0;
 my_bool opt_allow_multiple_engines= 0;
 static my_bool opt_short_log_format= 0;
-static uint kill_blocked_pthreads_flag, wake_pthread;
+static std::atomic<bool> kill_blocked_pthreads_flag(false);
+static uint wake_pthread;
 static ulong killed_threads;
        ulong max_used_connections;
 static char *mysqld_user, *mysqld_chroot;
@@ -631,7 +632,7 @@ uint lower_case_table_names;
 ulong tc_heuristic_recover= 0;
 int32 num_thread_running;
 int32 thread_binlog_client;
-ulong thread_created;
+std::atomic<unsigned long> thread_created;
 ulong back_log, connect_timeout, concurrency, server_id;
 ulong table_cache_size, table_def_size;
 ulong table_cache_instances;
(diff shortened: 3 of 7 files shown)


mysql-test/include/rpl_stop_server.inc
```diff
@@ -51,7 +51,7 @@ if ($rpl_debug)
 
 # Send shutdown to the connected server and give
 # it 14 seconds to die before zapping it
```
jkedgar: Update the comment here to '40' as well?

mysql-test/suite/perfschema/t/socket_instances_func.test
```diff
@@ -309,9 +309,9 @@ WHERE THREAD_ID = @thread_id;
 
 --echo #Compare server listener socket thread ids
 --disable_query_log ONCE
-eval SELECT ($server_tcpip_thread_id = $server_unix_thread_id) into @match_thread_id;
+#eval SELECT ($server_tcpip_thread_id = $server_unix_thread_id) into @match_thread_id;
```
jkedgar: If we don't need these two SELECT statements, can we just remove them?

sql/mysqld.cc
```diff
@@ -1000,7 +1002,7 @@ void deinit_rsa_keys(void);
 int show_rsa_public_key(THD *thd, SHOW_VAR *var, char *buff);
 #endif
 
-static volatile sig_atomic_t global_thread_count= 0;
+static std::atomic<uint> global_thread_count(0);
```
jkedgar, line 1005: Nit - you are not being consistent on using std::atomic<T> vs std::atomic_T (i.e. `thread_created` was `std::atomic_ulong` and `kill_blocked_pthreads_flag` was `std::atomic_bool` but this variable is `std::atomic<uint>`).  It's not a big deal - I just like consistency.  :)

sql/mysqld.cc
```diff
@@ -2279,6 +2280,14 @@ void clean_up(bool print_message)
   have_statement_timeout= SHOW_OPTION_DISABLED;
 #endif
 
+#if !defined(EMBEDDED_LIBRARY)
+  if (separate_conn_handling_thread) {
+    for (auto ii : sockets_list) {
+      if (ii)
```
jkedgar: I believe `delete` will automatically check for null so you can get rid of the check here.

sql/mysqld.cc
```diff
@@ -3106,11 +3117,13 @@ static bool block_until_new_connection()
     DBUG_ASSERT( ! _db_is_pushed_());
 
     // Block pthread
-    while (!abort_loop && !wake_pthread && !kill_blocked_pthreads_flag)
+    bool kbpf = false;
+    while (!(kbpf = kill_blocked_pthreads_flag.load(std::memory_order_acquire))
+      && !abort_loop && !wake_pthread)
```
jkedgar: Nit - because of this command now taking two lines can you add brackets for readability?

sql/mysqld.cc
```diff
@@ -3195,11 +3210,16 @@ bool one_thread_per_connection_end(THD *thd, bool block_pthread)
...
-  // Decrement connection count while holding LOCK_thread_count
+  mysql_mutex_lock(&LOCK_thread_count);
+  if (!global_thread_count.load(std::memory_order_relaxed)) {
+    mysql_cond_broadcast(&COND_thread_count);
+  }
+  dec_connection_count_locked();
+
   if (block_pthread)
-    block_pthread= block_until_new_connection();
+    block_pthread= block_until_new_connection_halflock();
```
jkedgar, line 3245: Nit - I personally don't like having a function release a lock that it did not acquire.  I think it leads to less readable code.  Can we change `block_until_new_connection_halflock()` to not release the lock and this code to the following?

```
if (block_pthread)
  block_pthread= block_until_new_connection();

mysql_mutex_unlock(&LOCK_thread_count);
```

include/mpsc.h
```diff
@@ -128,8 +128,12 @@ class mpsc_queue_t
     {
       thread_id = my_thread_dbug_id();
       m_single_consumer_check = true;
-    } else
-      DBUG_ASSERT(thread_id == my_thread_dbug_id());
+    } else {
+      // turning off this assert as it might be necessary to
+      // dequeue from different thread and it
+      // seems my_thread_init is getting called in ways I don't understand
+      //DBUG_ASSERT(thread_id == my_thread_dbug_id());
```
jkedgar: I'm concerned that this being commented out it implies that you have multiple consumer threads.  How do you know that there is never more than one consumer at a time?  A possible way would be something like this:

```
 class mpsc_queue_t
 {
 private:
#if !defined(DBUG_OFF)
  std::atomic<bool> dequeue_in_use;
#endif
...
  bool dequeue(T& output)
  {
    DBUG_ASSERT(!dequeue_in_use.exchange(true));  // Make sure it was false before

    buffer_node_t* tail = _tail.load(std::memory_order_relaxed);
    buffer_node_t* next = tail->next.load(std::memory_order_acquire);

    bool result;
    if (next == nullptr)
    {
      result = false;
    }
    else
    {
      output = next->data;
      _tail.store(next, std::memory_order_release);
      delete tail;
      result = true;
    }

    DBUG_ASSERT(dequeue_in_use.exchange(false));  // Make sure it was true before
    return result;
  }
```

mysql-test/suite/sys_vars/r/num_conn_handling_threads_basic.result
```diff
@@ -1,13 +1,13 @@
 SELECT @@GLOBAL.num_conn_handling_threads;
 @@GLOBAL.num_conn_handling_threads
-1
+4
 1 Expected
```
jkedgar, line 4: Do we need to update this line in the test file?  Should it be "4 expected"?  And at line 11 and 24.

mysql-test/suite/sys_vars/r/num_sharded_listen_sockets_basic.result
```diff
@@ -1,13 +1,13 @@
 SELECT @@GLOBAL.num_sharded_listen_sockets;
 @@GLOBAL.num_sharded_listen_sockets
-1
+4
 1 Expected
```
jkedgar, line 4: Do we need to update this line in the test file?  Should it be "4 expected"?  And at line 11 and 24.

mysql-test/suite/sys_vars/r/separate_conn_handling_thread_basic.result
```diff
@@ -1,13 +1,13 @@
 SELECT @@GLOBAL.separate_conn_handling_thread;
 @@GLOBAL.separate_conn_handling_thread
-0
+1
 0 Expected
```
jkedgar, line 4: Do we need to update this line in the test file?  Should it be "1 expected"?  And at line 11 and 24.

mysql-test/suite/sys_vars/r/use_lock_sharding_basic.result
```diff
@@ -1,13 +1,13 @@
 SELECT @@GLOBAL.use_lock_sharding;
 @@GLOBAL.use_lock_sharding
-0
+1
 0 Expected
```
jkedgar, line 4: Do we need to update this line in the test file?  Should it be "0 expected"?  And at line 11 and 24.

mysql-test/suite/sys_vars/r/use_socket_sharding_basic.result
```diff
@@ -1,13 +1,13 @@
 SELECT @@GLOBAL.use_socket_sharding;
 @@GLOBAL.use_socket_sharding
-0
+1
 0 Expected
```
jkedgar, line 4: Do we need to update this line in the test file?  Should it be "0 expected"?  And at line 11 and 24.

jkedgar (approved):
Overall looks great.

sql/mysqld.cc
anirbanr-fb: yes deleting a null ptr is actually ok, but people still end up doing this. will remove.

sql/mysqld.cc
anirbanr-fb: yep. will do.

mysql-test/include/rpl_stop_server.inc
anirbanr-fb: this is not required. I will take it out.

mysql-test/suite/perfschema/t/socket_instances_func.test
anirbanr-fb: all these files will come back to normal, when i flip the default back. maybe I should create a diff with all this clutter removed.

include/mpsc.h
anirbanr-fb: The only other consumer is the main-thread which is using dequeue to remove unremoved
members. The issue is not the intent of the check, but how I have implemented it. For some reason I am finding threads within mysqld to have their thread id changed . I like your approach and used it.

sql/mysqld.cc
anirbanr-fb: this one is tricky. The release of the lock happens midway within that function (for efficiency reasons ) and not at the very end.

jkedgar: You are right.  I was misled by not checking for multiple returns within the function.  :)

anirbanr-fb:
this has already been merged using PR-361. This was only for easy review.

PR #373, opened 2016-10-30
Fix race condition in bulk load and fast secondary key creation

lth:
Summary:
There is one special condition where alter table will not call
finalize_bulk_load on the handler that bulk loaded the primary key. Detect
this condition and call finalize_bulk_load before starting secondary key
creation.

Squash with: https://reviews.facebook.net/D61779

Test Plan: mtr

### DIFF
--- a/storage/rocksdb/ha_rocksdb.cc
+++ b/storage/rocksdb/ha_rocksdb.cc
@@ -9811,7 +9811,45 @@ int ha_rocksdb::inplace_populate_sk(const TABLE* new_table_arg,
   const bool hidden_pk_exists = has_hidden_pk(table);
 
   int res= 0;
-  Rdb_transaction *tx;
+  Rdb_transaction *tx= get_or_create_tx(table->in_use);
+
+  /*
+    There is one specific scenario where m_sst_info may not be nullptr. This
+    happens if the handler we're using happens to be the handler where the PK
+    bulk load was done on. The sequence of events that lead to this is as
+    follows (T1 is PK bulk load, T2 is SK alter table):
+
+    T1: Execute last INSERT statement
+    T1: Return TABLE and handler object back to Table_cache_manager
+    T1: Close connection
+    T2: Execute ALTER statement
+    T2: Take same TABLE/handler from Table_cache_manager
+    T2: Call closefrm which will call finalize_bulk_load on every other open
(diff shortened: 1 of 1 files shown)


storage/rocksdb/ha_rocksdb.cc
```diff
@@ -9811,7 +9811,45 @@ int ha_rocksdb::inplace_populate_sk(const TABLE* new_table_arg,
...
+    the destructor on the handler it is holding. Thus, its m_sst_info will not
+    be nullptr.
+
+    At this point, it is safe to refresh the snapshot because we know all other
+    open handlers have been closed at this point, and the one we're on is the
+    only one left.
+  */
+  if (m_sst_info != nullptr)
+  {
+    if ((res= finalize_bulk_load()))
```
vasilep, line 9846: NIT: It's usually cleaner when assignment and test are done on separates lines.
NIT: One extra unneeded pair of ().

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -9811,7 +9811,45 @@ int ha_rocksdb::inplace_populate_sk(const TABLE* new_table_arg,
   const bool hidden_pk_exists = has_hidden_pk(table);
 
   int res= 0;
-  Rdb_transaction *tx;
+  Rdb_transaction *tx= get_or_create_tx(table->in_use);
+
+  /*
```
vasilep, line 9816: I <3 the level of detail explaining WHY we are doing it.

storage/rocksdb/ha_rocksdb.cc
alxyang: @vasilep we already have this assignment/test pattern all over the codebase in ha_rocksdb.cc, so might be best to remain consistent.  Also, the extra pair of () is necessary I believe, it's really doing
`if ((res= finalize_bulk_load()) != 0)`

alxyang:
I'm good to accept.  The sooner this is in the better, so I can cherry-pick it into the next release (it's the last blocker).  @yoshinorim I remember you were saying you ran some long-executing tests this weekend, have they all passed?

It's passed all the cp_sql and data_correctness tests I've executed.

alxyang (approved):
@yoshinorim confirmed his long-running tests passed.

PR #379, opened 2016-10-31
Adding innodb_force_index_records_in_range session variable

lth:
Summary:
Add a variable to skip records_in_range when force index is being used.

Test Plan:
mtr


midom:
do it.

mdcallag:
what value should I use for this with linkbench and other benchmarks?

midom:
10 could be fine

PR #381, opened 2016-10-31
Group purging of rows by table id

santoshbanda:
Summary:
The current assignment of purge records to workers may lead to lock contention on dict_index_t::lock. Fix this by assigning record of same table to one purge thread.

This is a back-port from https://github.com/mysql/mysql-server/commit/7e43fd218aa1a63cc74d2d706e84adac93ac6905. I used simpler
logic to group records by table id. The refactoring in the upstream change is not ported here.

Test Plan: mtr

### DIFF
--- a/storage/innobase/trx/trx0purge.cc
+++ b/storage/innobase/trx/trx0purge.cc
@@ -45,6 +45,8 @@ Created 3/26/1996 Heikki Tuuri
 #include "os0thread.h"
 #include "srv0mon.h"
 #include "mtr0log.h"
+#include <vector>
+#include <unordered_map>
 
 /** Maximum allowable purge history length.  <=0 means 'infinite'. */
 UNIV_INTERN ulong		srv_max_purge_lag = 0;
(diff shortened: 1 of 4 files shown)


storage/innobase/trx/trx0purge.cc
```diff
@@ -1031,10 +1029,26 @@ trx_purge_attach_undo_recs(
...
+				table_thr.emplace(table_id, thr);
+			} else {
+				thr = it->second;
+			}
+
+			ut_a(thr != NULL && !thr->is_active);
+			/* Get the purge node. */
+			node = (purge_node_t*) thr->child;
+			ut_a(que_node_get_type(node) == QUE_NODE_PURGE);
+
```
tianx, line 1051: The simplified logic looks legit to me. Did you get any response from Sunny about your questions?

tianx (commented):
Also the upstream doesn't have unit tests :(. In your test plan, which mtr tests cover the code change?

storage/innobase/trx/trx0purge.cc
santoshbanda: I didn't get any response. Almost all the innodb stress tests cover the purge code path.

PR #385, opened 2016-11-02
Log legacy user connections in slow log

tianx:
Summary:
Introduced two global system variables for legacy user logging:
  - log_legacy_user: log legacy user connections. default is turned off.
  - legacy_user_name_pattern: regex pattern of a legacy user name.

If log_legacy_user is turned on, a legacy user connection will be logged in the
slow log, which is prefixed with `LEGACY_USER: `.

Test Plan: Added new mtr tests.

### DIFF
--- a/mysql-test/t/slow_log_legacy_user.test
+++ b/mysql-test/t/slow_log_legacy_user.test
@@ -1,0 +1,72 @@
+#
+# Test log_legacy_user
+#
+
+--source include/have_fullregex.inc
+--source include/have_innodb.inc
+
+# Just in case slow_log_wait_time-slow.log already exists, delete it first.
+--remove_files_wildcard $MYSQLTEST_VARDIR/mysqld.1/data slow_log_legacy_user-slow.log
+
+set @my_slow_logname = @@global.slow_query_log_file;
+set global slow_query_log_file = "slow_log_legacy_user-slow.log";
+
+set @my_log_legacy_user = @@global.log_legacy_user;
+set global log_legacy_user = TRUE;
+set @my_legacy_user_name_pattern = @@global.legacy_user_name_pattern;
+
+# set the name pattern to indicate the name doesn't contain ':'
+set global legacy_user_name_pattern = '^[^:]+$';
+
+--echo # legacy_user1
+create user legacy_user1@localhost;
+--echo # connection conn1
+connect (conn1, localhost, legacy_user1,,);
+
+--echo # connection default
+connection default;
+
+# set the name pattern to indicate the name doesn't contain '%'
+--echo # change legacy_user_name_pattern
+set global legacy_user_name_pattern = '^[^%]+$';
+--echo # legacy%user2
+create user `legacy_user2`@localhost;
+--echo # connection conn2
+connect (conn2, localhost, legacy_user2,,);
+
+--echo # connection default
+connection default;
+
+--echo
+--echo # Check the slow log result.
+--perl
+my $found = 0;
+open FILE, "$ENV{'MYSQLTEST_VARDIR'}/mysqld.1/data/slow_log_legacy_user-slow.log" or die;
+my @lines = <FILE>;
+foreach $line (@lines) {
+  if ($line =~ /^LEGACY_USER:/) {
+    $found = 1;
+    print $line
+  }
+}
+# (test failure) print the log if we didn't find any legacy user
+if (!$found) {
+  print "**Legacy user not found in slow log**\n";
+  seek FILE, 0, 0;
+  while (<FILE>) {
+    print $_;
+  }
+}
+close FILE
+EOF
+
+--echo
+--echo # clean up
+disconnect conn1;
+drop user `legacy_user1`@localhost;
+disconnect conn2;
+drop user `legacy_user2`@localhost;
+
+set @@global.slow_query_log_file = @my_slow_logname;
+set @@global.legacy_user_name_pattern = @my_legacy_user_name_pattern;
+set @@global.log_legacy_user = @my_log_legacy_user;
--- a/sql/mysqld.cc
+++ b/sql/mysqld.cc
@@ -474,6 +474,10 @@ ulong log_warnings;
 uint host_cache_size;
 ulonglong tmp_table_rpl_max_file_size;
 
+my_bool log_legacy_user;
+const char *opt_legacy_user_name_pattern;
+Regex_list_handler *legacy_user_name_pattern;
+
 #if defined(_WIN32) && !defined(EMBEDDED_LIBRARY)
 ulong slow_start_timeout;
 #endif
@@ -2282,6 +2286,7 @@ void clean_up(bool print_message)
   free_list(opt_plugin_load_list_ptr);
 
   delete gap_lock_exceptions;
+  delete legacy_user_name_pattern;
 
   if (THR_THD)
     (void) pthread_key_delete(THR_THD);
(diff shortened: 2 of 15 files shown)


mysql-test/suite/sys_vars/t/legacy_user_identifier.test
```diff
@@ -0,0 +1,20 @@
+--source include/not_embedded.inc
+--source include/load_sysvars.inc
+
+set @@global.legacy_user_identifier = default;
+select @@global.legacy_user_identifier;
+set @saved_legacy_user_identifier = @@global.legacy_user_identifier;
```
jkedgar: Doesn't it make sense to save the original value _before_ you set it in line 4?  Sure - you are likely running this with the default legacy_user_identifier, but it still looks weird.

mysql-test/t/slow_log_legacy_user.test
```diff
@@ -0,0 +1,71 @@
+#
+# Test log_legacy_user
+#
+
+--source include/have_innodb.inc
+
+# Just in case slow_log_wait_time-slow.log already exists, delete it first.
+--perl
```
jkedgar: If you wanted, I think you could simplify this by removing the perl script and using this:

```
--remove_files_wildcard $MYSQLTEST_VARDIR/mysqld.1/data slow_log_wait_time-slow.log
```

sql/sql_acl.cc
```diff
@@ -11166,6 +11166,21 @@ acl_authenticate(THD *thd, uint com_change_user_pkt_len)
...
+        strstr(mpvio.auth_info.user_name, opt_legacy_user_identifier))
+    {
+      /* Log the legacy user */
+      char buffer[256];
+      snprintf(buffer, 256, "LEGACY_USER: %s@%s on %s",
+               mpvio.auth_info.user_name,
+               mpvio.auth_info.host_or_ip,
+               mpvio.db.str ? mpvio.db.str : (char*) "");
+      bool save_enable_slow_log = thd->enable_slow_log;
+      thd->enable_slow_log = true;
```
jkedgar: We are writing to the slow log even if the user doesn't have the slow log enabled?  That seems odd.  Is the slow log guaranteed to have a valid file name?

sql/mysqld.cc
```diff
@@ -444,6 +444,9 @@ ulong log_warnings;
 uint host_cache_size;
 ulonglong tmp_table_rpl_max_file_size;
 
+my_bool log_legacy_user;
+char *opt_legacy_user_identifier;
```
jkedgar: Does MySQL allow you to mark this buffer as `const char *`?  If so, that would be preferable.

sql/sql_acl.cc
```diff
@@ -11166,6 +11166,21 @@ acl_authenticate(THD *thd, uint com_change_user_pkt_len)
       general_log_print(thd, command, (char*) "%s@%s on %s",
                         mpvio.auth_info.user_name, mpvio.auth_info.host_or_ip,
                         mpvio.db.str ? mpvio.db.str : (char*) "");
+
+    if (log_legacy_user && opt_legacy_user_identifier &&
+        strstr(mpvio.auth_info.user_name, opt_legacy_user_identifier))
+    {
+      /* Log the legacy user */
+      char buffer[256];
+      snprintf(buffer, 256, "LEGACY_USER: %s@%s on %s",
```
jkedgar: If the database is not assigned for a user, the output string will look like `LEGACY_USER: <user>@<location> on`.  This looks weird.  One solution, that also avoids the static buffer is this:

```
std::string output= std::string("LEGACY_USER: ") + mpvio.autho_info.user_name +
                   "@" + mpvio.autho_info.host_or_ip;
if (mpvio.db.str != nullptr)
{
  output += std::string(" on ") + mpvio.db.str;
}
...
slow_log_print(thd, output.c_str(), output.size(), &thd->status_var);
...
```

sql/sys_vars.cc
```diff
@@ -5579,3 +5579,15 @@ static Sys_var_charptr Sys_read_only_error_msg_extra(
...
+
+static Sys_var_mybool Sys_log_legacy_user(
+       "log_legacy_user",
+       "Log legacy user names in slow query log.",
+       GLOBAL_VAR(log_legacy_user),
+       CMD_LINE(OPT_ARG), DEFAULT(FALSE));
+
+static Sys_var_charptr Sys_legacy_user_identifier(
+       "legacy_user_identifier",
+       "String that will identify the user name as a legacy user",
```
jkedgar: Can you update this description to indicate that if the string is found _anywhere_ in the user name it will mark the user name as a legacy user?  It is unclear where this string needs to be as it is.  An example would be "Identifier which if found anywhere in the user name will identify the user name as a legacy user".

jkedgar (commented):
Minor nits

mysql-test/t/slow_log_legacy_user.test
tianx: Changed. Thanks!

sql/sql_acl.cc
tianx: Changed. Previously I followed how we log the similar info to audit log before this code snippet, but agree that it would look nicer.

sql/sql_acl.cc
tianx: Yeah that seems weird. We are actually piggybacking the slow-log for logging the legacy user during connections. The `enable_slow_log` is set in the `dispatch_command` as it is associated with queries. So during connection auth, the variable isn't set yet. The file path has a default value. Whether the file path is valid or not applies to all slow logging scenarios.

sql/mysqld.cc
tianx: Changed.

sql/sys_vars.cc
tianx: This is obsolete now. We use regex instead.

tianx:
Addressed comments from Jay @jkedgar

mysql-test/suite/sys_vars/t/legacy_user_identifier.test
tianx: changed too.

sql/sql_acl.cc
```diff
@@ -11166,6 +11167,24 @@ acl_authenticate(THD *thd, uint com_change_user_pkt_len)
       general_log_print(thd, command, (char*) "%s@%s on %s",
                         mpvio.auth_info.user_name, mpvio.auth_info.host_or_ip,
                         mpvio.db.str ? mpvio.db.str : (char*) "");
+
+    if (log_legacy_user && opt_legacy_user_name_pattern)
+    {
+      std::regex pattern(opt_legacy_user_name_pattern);
```
jkedgar: Creating the std::regex with an invalid pattern (such as having an opening bracket but no closing bracket) will throw an exception which will cause the server to crash.  I've added a class that will handle this case for you.  See `Regex_list_handler` (sql/handler.*).  It supports a list of regex patterns and returns an error if you try to set a bad pattern and even returns a list of the bad patterns so you can output them in the log and then continue.  You can see the use cases by searching for gap_lock_exceptions (sql/handler.cc) and rdb_collation_exceptions (storage/rocksdb/ha_rocksdb.cc).

Another benefit is that you can create the Regex_list_handler once, and then only update the pattern if `opt_legacy_user_name_pattern` changes, getting better performance than the current implementation that creates a new std::regex with every user login.

tianx: Fixed.

tianx:
Test updated.

tianx:
Rebased and ping.

sql/sql_acl.cc
```diff
@@ -53,6 +53,7 @@
 #include <mysql/plugin_validate_password.h>
 #include "password.h"
 #include "crypt_genhash_impl.h"
+#include <regex>
```
jkedgar: nit - you probably don't need this anymore

sql/sql_acl.cc
```diff
@@ -11166,6 +11167,23 @@ acl_authenticate(THD *thd, uint com_change_user_pkt_len)
...
+
+    if (log_legacy_user && opt_legacy_user_name_pattern)
+    {
+      if (legacy_user_name_pattern->matches(mpvio.auth_info.user_name))
+      {
+        /* Log the legacy user */
+        std::string output= std::string("LEGACY_USER: ") +
+          mpvio.auth_info.user_name + "@" + mpvio.auth_info.host_or_ip;
+        if (mpvio.db.str != nullptr)
+            output += std::string(" on ") + mpvio.db.str;
```
jkedgar: nit - indenting is off by two here.

jkedgar (approved):
A couple of minor nits.  Otherwise this looks great.

PR #388, opened 2016-11-02
Another attempt at fixing the flakiness in rocksdb.type_decimal

jkedgar:
Summary: The rocksdb.type_decimal test is still failing.  I thought running ANYLIZE TABLE would solve it but was still getting problems.  I'm now trying to force a memtable flush to see if that solves it.

Test Plan: MTR

Reviewers: gunnarku, mung

Subscribers:

Tasks: D13556119

Blame Revision:


lth:
I looked at the test, and it seems like we're just testing for the absence of "Using index" in the explain output.

Instead of trying to get stats to play nice, would it be easier just to fix the test instead? You could also just disable the ICP optimization: https://dev.mysql.com/doc/refman/5.6/en/server-system-variables.html#sysvar_optimizer_switch

We also have rocksdb_debug_optimizer_n_rows if the non-determinism is coming from GetApproximateSizes.

PR #389, opened 2016-11-03
Add global variable `max_nonsuper_connections`

tianx:
Summary:
The new global variable `max_nonsuper_connections` will enforce the limit for
the sum of all non-admin connections. This will be useful to limit regular user
connections while still allow super users to connect to server.

Test Plan:
mtr max_nonsuper_connections.test and sys_vars tests.


tianx:
This PR is ready for review.

tianx:
Make it explicit that the limit is enforced on nonsuper users.

PR #396, opened 2016-11-04
Making default to 1 shard when sharding is off, so that no change in …

anirbanr-fb:
…behavior.

Summary: In the default case, where use_lock_sharding is off,
we want no change in behavior and #shards = 1

Test Plan: Sandcastle

Reviewers: jkedgar


jkedgar (approved):
Looks good

tianx:
@anirbanr-fb several perfschema mtr tests are failing due to this commit..

PR #398, opened 2016-11-04
Supporting dynamic scheduling in MTS

abhinav04sharma:
Summary: Statically assigning shards to slave workers can cause imbalance if a few shards are hotter than others. We should be able to check the imbalance among slave workers and dynamically reassign shards.

Test Plan: Check if all test cases pass and ascertain if the code kicks off when imbalance occurs


sql/sql_show.cc
```diff
@@ -8838,6 +8881,17 @@ ST_FIELD_INFO authinfo_fields_info[]=
   {0, 0, MYSQL_TYPE_STRING, 0, 0, 0, SKIP_OPEN_TABLE}
 };
 
+ST_FIELD_INFO slave_db_load_fields_info[]=
+{
+  {"DB", NAME_LEN, MYSQL_TYPE_STRING, 0, 0, "Database",
+   SKIP_OPEN_TABLE},
+  {"WORKER", 21, MYSQL_TYPE_LONGLONG, 0, MY_I_S_UNSIGNED, "Worker ID",
+   SKIP_OPEN_TABLE},
+  {"DB_LOAD", 21, MYSQL_TYPE_LONGLONG, 0, MY_I_S_UNSIGNED, "DB Load",
```
abhinav04sharma, line 8949: This denotes the load on the DB since the last checkpoint, not the total load since the server started.

PR #400, opened 2016-11-05
Fixes 2 race conditions based on results from nightly contrun

anirbanr-fb:
Summary: 1. we have to create a connection thread, holding
the sharded lock, otherwise disconnect of connection might
trigger thread removal, before thread addition.

2. when we do a find on global sharded thread list, we cannot
return the global iterator end, because that would access
the end from a different shard, while not holding its lock.

Test Plan: Sandcastle and connection rate benchmarks ran again.

Put a sleep after thread creation to consistently trigger race
condition. Checked it to be fixed after change.

Connection rate benchmarks didn't show any significant reduction
in conn rate, although the critical section increased.

Reviewers: jkedgar

### DIFF
--- a/sql/mysqld.cc
+++ b/sql/mysqld.cc
@@ -1083,8 +1083,7 @@ void add_global_thread(THD *thd)
 {
   DBUG_PRINT("info", ("add_global_thread %p", thd));
   mutex_assert_owner_shard(SHARDED(&LOCK_thread_count), thd);
-  const bool have_thread=
-    global_thread_list->find(thd) != global_thread_list->end();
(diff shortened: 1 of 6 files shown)


lth (approved):

sql/mysqld.cc
```diff
@@ -6970,6 +6969,7 @@ int win_main(int argc, char **argv)
 int mysqld_main(int argc, char **argv)
 #endif
 {
+  //sleep(40);
```
lth: You probably don't need this.

lth (approved):

PR #410, opened 2016-11-10
Recreate iterator after committing in the middle

### LINKED ISSUE #411: Setting rocksdb_commit_in_the_middle commits transaction without releasing iterator
This leaves the iterator in an invalid state.

To reproduce:

```
CREATE TABLE t1 (id1 bigint(20),
                 id2 bigint(20),
                 id3 bigint(20),
                 PRIMARY KEY (id1, id2, id3))
                 DEFAULT CHARSET=latin1;

CREATE TABLE t2 (id1 bigint(20),
                 id2 bigint(20),
                 PRIMARY KEY (id1, id2))
                 DEFAULT CHARSET=latin1;


set rocksdb_commit_in_the_middle=1;
set rocksdb_bulk_load_size = 100;

--disable_query_log
let $j = 10000;
while ($j)
{
  --eval insert into t1 (id1, id2, id3) values (0, $j, 0);
  --eval insert into t2 (id1, id2) values (0, $j);
  dec $j;
}
--enable_query_log

explain DELETE t2, t1 FROM t2 LEFT JOIN t1 ON t2.id2 = t1.id2 AND t2.id1 = t1.id1 WHERE t2.id1 = 0;
DELETE t2, t1 FROM t2 LEFT JOIN t1 ON t2.id2 = t1.id2 AND t2.id1 = t1.id1 WHERE t2.id1 = 0;

drop table t1, t2;
```

lth:
It is not safe to keep using an iterator after the associated transaction has committed. This happens when rocksdb_commit_in_the_middle is set. To fix, we recreate an iterator from the new transaction and seek to the last key we were at.

Closes #411

### DIFF
--- a/mysql-test/suite/rocksdb/r/rocksdb.result
+++ b/mysql-test/suite/rocksdb/r/rocksdb.result
@@ -2468,7 +2468,10 @@ id2 bigint(20),
 PRIMARY KEY (id1, id2))
 DEFAULT CHARSET=latin1;
 set rocksdb_commit_in_the_middle=1;
+SET @save_rocksdb_bulk_load_size= @@rocksdb_bulk_load_size;
 set rocksdb_bulk_load_size = 100;
 DELETE t2, t1 FROM t2 LEFT JOIN t1 ON t2.id2 = t1.id2 AND t2.id1 = t1.id1 WHERE t2.id1 = 0;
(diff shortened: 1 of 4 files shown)


alxyang (approved):
lgtm

PR #412, opened 2016-11-11
MyRocks landing page

zardra:
Would really like to have the Installation and Migration sections reviewed to ensure they are correct technically.


_config.yml
```diff
@@ -1,9 +1,9 @@
...
 permalink: /blog/:year/:month/:day/:title.html
-title: ProjectName
-tagline: This is a brief one-liner about the project
+title: MyRocks
+tagline: A RocksDB storage engine with MySQL
 description: >
-  This is a longer description of the project, a bit of a sales-pitch with a bit of technical information included.
+  Benefit from all the features of MySQL while using RocksDB as backend storage
 fbappid: "1615782811974223"
 gacode: "UA-12345-6"
```
JoelMarcey: We are going to need to add Google Analytics for this.

_config.yml
```diff
@@ -34,7 +34,7 @@ ghrepo: "facebook/Site-Templates"
 # Use these color settings to determine your colour scheme for the site.
 color:
   # primary should be a vivid color that reflects the project's brand
-  primary: "#3b5998"
+  primary: "#0f2647"
```
JoelMarcey, line 38: Ugh -- can't make a comment on line 21 :/

Anyway, for `url`, it should be:

`url: http://facebook.github.io/`

and on line 15, you should add a `baseurl` that is:

`baseurl: mysql-5.6`

JoelMarcey, line 38: ...and line 32 should be changed to `facebook/mysql-5.6` for the repo

static/logo1.svg
```diff
@@ -0,0 +1,560 @@
+<?xml version="1.0" encoding="utf-8"?>
```
JoelMarcey: What is `logo1.svg`?

zardra: I think that's from me messing around with images.  Easy enough to delete.

_config.yml
zardra: Are these suppose to have quotes? The example in the file does...

zardra: Done

_config.yml
zardra: Is there something special I need to do for that?

And what about the fbappid?

JoelMarcey: We can do the GA id together. I will need to add something to our internal GA page, I think.

I believe the fbappid stays the same across all things using this template.

PR #416, opened 2016-11-11
[mysql-cli] Fix handle_kill_signal to use SSL

jrahman-zz:
Summary: Currently the kill signal handler does not use SSL, even when
the correct options are passed. This causes it to fail to kill a running
query that was started on an SSL connection because it does not have a
password to connect with and does not perform an SSL connection with the
certificate.

Test Plan: Run mysql CLI with an SSL connection, execute SELECT
SLEEP(500); then Ctrl-C the process

### DIFF
--- a/client/mysql.cc
+++ b/client/mysql.cc
@@ -240,7 +240,7 @@ static my_bool execute_buffer_conversion_done= 0;
   will be multiplied to 0 and we'll get 0 as a result.
   The same is true for stderr.
 */
-static uint win_is_console_cache= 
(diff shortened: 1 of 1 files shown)


lth:
You need to undo your whitespace changes.

jrahman-zz:
Ok, will do. Nuclide was overzealous.

jrahman-zz:
Abandoning in favor of https://github.com/facebook/mysql-5.6/pull/419

PR #418, opened 2016-11-11
Change SET GLOBAL [super_]read_only to not acquire the global read lock

jkedgar:
Summary: Currently calling SET GLOBAL [super_]read_only acquires the global read lock before setting the flag and this will block on a long running DML statement.  We think it is safe in our environment to not acquire this lock.

Test Plan: MTR

### DIFF
--- a/sql/lock.cc
+++ b/sql/lock.cc
@@ -1068,16 +1068,21 @@ void Global_read_lock::unlock_global_read_lock(THD *thd)
 {
   DBUG_ENTER("unlock_global_read_lock");
 
-  DBUG_ASSERT(m_mdl_global_shared_lock && m_state);
+  DBUG_ASSERT(m_state);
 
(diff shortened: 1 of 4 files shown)


sql/lock.cc
```diff
@@ -1106,7 +1111,7 @@ bool Global_read_lock::make_global_read_lock_block_commit(THD *thd)
     If we didn't succeed lock_global_read_lock(), or if we already suceeded
```
tianx, line 1111: The original comment/assumption is no longer accurate. Same on Line 1095. (github won't allow me to comment on a line outside the change context...)

tianx (commented):
Minor comment.

lth:
https://github.com/facebook/mysql-5.6/commit/a3f808d711fd421515ca26fc676f68cca070703e

PR #419, opened 2016-11-12
Fix signal handler to perform SSL connection to kill query

jrahman-zz:
Summary: The signal handler does not correctly use SSL when connecting
to the server in order to kill any outstanding query. Fix that.

Test Plan: Run mysql; execute SELECT SLEEP(500); Ctrl-C the process

### DIFF
--- a/client/mysql.cc
+++ b/client/mysql.cc
@@ -2195,7 +2195,7 @@ static int read_and_execute(bool interactive)
 	&& !ml_comment && !in_string && (com= find_command(line)))
     {
       if ((*com->func)(&glob_buffer,line) > 0)
-       break;
+	break;
       if (glob_buffer.is_empty())		// If buffer was emptied
 	in_string=0;
 #ifdef HAVE_READLINE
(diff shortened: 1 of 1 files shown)


client/mysql.cc
```diff
@@ -2192,8 +2194,9 @@ static int read_and_execute(bool interactive)
     if ((named_cmds || glob_buffer.is_empty())
 	&& !ml_comment && !in_string && (com= find_command(line)))
     {
-      if ((*com->func)(&glob_buffer,line) > 0)
-	break;
+      if ((*com->func)(&glob_buffer,line) > 0) {
```
lth: We usually try not to change more than what's needed.

lth (commented):
File a bug upstream if you want.

client/mysql.cc
```diff
@@ -4908,6 +4897,26 @@ sql_real_connect(char *host,char *database,char *user,char *password,
   return 0;
 }
 
+static void
+configure_ssl(MYSQL* mysql) {
+  #if defined(HAVE_OPENSSL) && !defined(EMBEDDED_LIBRARY)
```
lth: Can we keep the same indentation as before? The #if should not be indented.

lth (approved):

jrahman-zz:
Fixed indentation as requested. Imported and re-running tests now.

jrahman-zz:
Fixed last indentation issue, should be ready

PR #421, opened 2016-11-13
Fixing how to get Iterator with commit_in_the_middle()

yoshinorim:
Summary: NewIterator() needs CF handle, otherwise it returns
an iterator for default column family, which doesn't return
expected rows.

Squash with: https://phabricator.intern.facebook.com/D4159222

Test Plan: mtr

### DIFF
--- a/storage/rocksdb/ha_rocksdb.cc
+++ b/storage/rocksdb/ha_rocksdb.cc
@@ -8149,8 +8149,9 @@ void ha_rocksdb::setup_scan_iterator(
       m_scan_it_snapshot= rdb->GetSnapshot();
 
       auto read_opts = rocksdb::ReadOptions();
+      read_opts.total_order_seek = true; // TODO: set based on WHERE conditions
       read_opts.snapshot= m_scan_it_snapshot;
-      m_scan_it= rdb->NewIterator(read_opts);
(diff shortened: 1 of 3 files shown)


lth (approved):

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -8150,7 +8150,7 @@ void ha_rocksdb::setup_scan_iterator(
 
       auto read_opts = rocksdb::ReadOptions();
       read_opts.snapshot= m_scan_it_snapshot;
-      m_scan_it= rdb->NewIterator(read_opts);
+      m_scan_it= rdb->NewIterator(read_opts, kd.get_cf());
```
lth: I just realized that I'm not setting the read_opts from `skip_bloom` and `fill_cache` in this case. Do you think this is necessary?

lth (approved):

alxyang:
the test case you wrote doesn't seem to repro the issue properly on my machine. 

I think the pk needs to be created with a different column family, and you can also keep commit_in_the_middle in the on so it still tries to use `NewIterator` when it scans the SK w/ SELECT, though not necessary.

yoshinorim:
- Updated the test to use non-default CF for both primary key and secondary key
- Set total_order_seek=true (disabling using bloom filter)

PR #424, opened 2016-11-14
Fix a race condition in mysqlimport

tianx:
Summary:
When running `mysqlimport` with multiple threads (`--use-threads=N`), we ran
into a race condition where the threads try to (de-)initialize the client
plugin (e.g. the initialization code path
`db_connect->mysql_init->mysql_server_init->mysql_client_plugin_init`).  This
causes test failure in mysqldump in the test section `use-thread option`, with
the following symptons:

  safe_mutex: Trying to lock unitialized mutex at client_plugin.c, line 275
  safe_mutex: Error in thread libray: Got mutex at client_plugin.c, line 275 more than 1 time

To make mysqlimport thread safe, this patch initializes client (`mysql_init`)
once in the main thread. So subsequent `mysql_init` calls in worker threads
will be thread safe (similar code pattern in mysqlslap.c).

Test Plan:
Hard to reproduce, but I ran `mtr --valgrind --async-client --parallel=16 mysqldump` cleanly.

### DIFF
--- a/client/mysqlimport.c
+++ b/client/mysqlimport.c
@@ -410,14 +410,12 @@ static void lock_table(MYSQL *mysql, int tablecount, char **raw_tablename)
 
 
 
-static MYSQL *db_connect(char *host, char *database,
-                         char *user, char *passwd)
+static void db_connect(const char *host, const char *database,
+                       const char *user, const char *passwd,
+                       MYSQL *mysql)
 {
-  MYSQL *mysql;
   if (verbose)
     fprintf(stdout, "Connecting to %s\n", host ? host : "localhost");
-  if (!(mysql= mysql_init(NULL)))
-    return 0;
   if (opt_compress)
     mysql_options(mysql,MYSQL_OPT_COMPRESS,NullS);
   if (opt_local_file)
@@ -470,7 +468,6 @@ static MYSQL *db_connect(char *host, char *database,
     ignore_errors=0;
     db_error(mysql);
   }
-  return mysql;
 }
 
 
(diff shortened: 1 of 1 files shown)


jkedgar (commented):
While this probably solves the problem it seems like overkill; most of the code in `db_connect()` and `db_disconnect` is already thread-safe.  It looks to me like `mysql_server_init()` and `mysql_server_end()` are where the problem is.  If you add this mutex here you will solve the problem for mysqlimport (at the expense of some performance as each connect and disconnect will block all others), but won't solve it for any other utility that could call `mysql_server_init()` from multiple threads.

tianx:
Jay, thanks for the comment. I am not too comfortable to add a mutex to the mysql_server_init() code path, which is called quite widely through mysql_init(). I looked through other tools that call into mysql_init. It looks to me the code pattern is to call mysql_init() once (e.g. in main function), and subsequent calls to mysql_init will be thread safe. An example is `mysqlslap.c`.

The code in mysqlimport.c is definitely ill-designed w.r.t multi-threading. While adding a mutex (like what I did now) is simplest to do but like you said, somewhat overkill, I think a good fix will be what I described above. I will update with a revision.

client/mysqlimport.c
```diff
@@ -410,14 +410,12 @@ static void lock_table(MYSQL *mysql, int tablecount, char **raw_tablename)
 
 
 
-static MYSQL *db_connect(char *host, char *database,
-                         char *user, char *passwd)
+static void db_connect(char *host, char *database,
```
jkedgar: nit - since you are changing db_connect anyway.. can we have each of the `char*` parameters be changed to `const char*`?

jkedgar (approved):
This looks great.  I like it much better than adding a mutex.
One comment - you could simplify the change slightly if you wanted.  Your change pulled the `mysql_init()` out of the `db_connect()`, but you could leave it in and call `db_connect()` at the start of `main()` instead of `mysql_init()`.  It would mean you did an extra connect when in multi-threaded mode, but that isn't a big deal.
Either way is just fine and I'm approving this diff so you can choose which way you like.

client/mysqlimport.c
tianx: Changed to `const char*`. On your other comment, I actually prefer to pull `mysql_init()` out :). `mysql_init()` and `db_connect()` look to me two separate logic.

PR #430, opened 2016-11-16
Change linter to use local cpplint.py instead of a copy in a submodule

jkedgar:
Summary: One of the linters was using a python script in the submodule.  Sandcastle was not loading this submodule before attempting to run the linters via 'arc diff'.  Change the linting code to use a local version of the cpplint.py.

Test Plan: N/A


gunnarku:
LGTM

gunnarku (approved):
LGTM

PR #442, opened 2016-11-21
Increasing default RocksDB write buffer and block cache size

yoshinorim:
Summary: This diff increases RocksDB default write buffer size
from 4MB to 64MB, and block cache size from 8MB to 512MB.

Test Plan: mtr


gunnarku (approved):
LGTM

Issue #444, opened 2016-11-21
compression_zstd test fails when the server is compiled without ZStandard


spetrunia:
(Copy-pasting from myrocks-dev group post):
Currently, compression_zstd test fails when the server is compiled without ZStd:

rocksdb.compression_zstd                 w3 [ fail ]
        Test ended at 2016-11-14 13:26:21

Server log is:
[Note] Global variable initial value (from config file) @@open_files_limit = 1024
...
2016-11-14 13:26:20 8734 [ERROR] RocksDB: Error opening instance: Invalid argument: Compression type ZSTD is not linked with the binary.
2016-11-14 13:26:20 8734 [ERROR] Plugin 'ROCKSDB' init function returned error.
2016-11-14 13:26:20 8734 [ERROR] Plugin 'ROCKSDB' registration as a STORAGE ENGINE failed.
2016-11-14 13:26:20 8734 [Warning] /home/psergey/dev-git/mysql-5.6-rocksdb-look2/sql/mysqld: unknown option '--loose-enable-performance-schema'
2016-11-14 13:26:20 8734 [Warning] /home/psergey/dev-git/mysql-5.6-rocksdb-look2/sql/mysqld: unknown variable 'loose-performance-schema-accounts-size=100'
2016-11-14 13:26:20 8734 [Warning] /home/psergey/dev-git/mysql-5.6-rocksdb-look2/sql/mysqld: unknown variable 'loose-performance-schema-digests-size=200'
...
2016-11-14 13:26:20 8734 [ERROR] Unknown/unsupported storage engine: rocksdb
2016-11-14 13:26:20 8734 [ERROR] Aborting

I think ZStandard is not a requirement for MyRocks to work (if it is, then CMake script should refuse to compile without ZStandard).

I didn't want the test to fail, so for MariaRocks I've made this change:

https://github.com/MariaDB/server/commit/e370d0a9ae2f440b4f5ae100906c70c9e9e6b186

the patch
* Adds @@rocksdb_supported_compression_types variable which show what compression is compiled in
* Makes compression_zstd test skip itself when the server is not compiled with ZStandard.

Any opinions, should I submit this patch for MyRocks too?

spetrunia:
https://github.com/facebook/mysql-5.6/pull/446

PR #446, opened 2016-11-21
Issue #444: compression_zstd test fails when the server is compiled w…

spetrunia:
…ithout ZStandard

- Introduce @@rocksdb_supported_compression_types read-only variable.
  It has a comma-separated list of compiled-in compression algorithms.

- Make rocksdb.compression_zstd test skip itself when ZSTD support
  is not compiled in

### DIFF
--- a/storage/rocksdb/rdb_utils.cc
+++ b/storage/rocksdb/rdb_utils.cc
@@ -27,6 +27,17 @@
 /* MyRocks header files */
 #include "./ha_rocksdb.h"
 
+/*
+  Both innobase/include/ut0counter.h and rocksdb/port/port_posix.h define
+  CACHE_LINE_SIZE.
+*/
+#ifdef CACHE_LINE_SIZE
+#  undef CACHE_LINE_SIZE
+#endif
+
+/* RocksDB header files */
+#include "util/compression.h"
+
 namespace myrocks {
 
 /*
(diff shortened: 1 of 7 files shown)


spetrunia:
One thing to check: the patch does
```
#include "util/compression.h"
```
which, as far as I understand, includes RocksDB's internal files (I assume the interface is all inside  rocksdb/include/rocksdb ). Is this ok?

storage/rocksdb/rdb_utils.cc
```diff
@@ -27,6 +27,17 @@
...
+/*
+  Both innobase/include/ut0counter.h and rocksdb/port/port_posix.h define
+  CACHE_LINE_SIZE.
+*/
+#ifdef CACHE_LINE_SIZE
+#  undef CACHE_LINE_SIZE
+#endif
+
+/* RocksDB header files */
+#include "util/compression.h"
```
hermanlee, line 39: These header files should be internal to the rocksdb submodule though and not meant for public consumption.

hermanlee:
@IslamAbdelRahman: Maybe the compression types can be part of the include/rocksdb headers?

IslamAbdelRahman:
If we don't care about experimental/old compression types (ex: kZSTDNotFinalCompression).

I can have a value like 
```
const CompressionType kLastStableCompression = CompressionType::kZSTD;
```

So you can do a loop from 0 -> kLastStableCompression to get a list of known compressions.
will that be good enough ?

spetrunia:
@IslamAbdelRahman, this variable would be useful, but it's not what I need the most. Please find a detailed description of what this patch needs below:


The patch makes use of:

1. Members of enum CompressionType, like rocksdb::kBZip2Compression, rocksdb::kSnappyCompression, etc
2. CompressionTypeSupported()
3. CompressionTypeToString()


I have no issue with #1. enum CompressionType is defined in include/rocksdb/options.h, it's a part of the public API.

Currently the list of possible compression methods is hard-coded in this patch (see get_rocksdb_supported_compression_types(),. https://github.com/facebook/mysql-5.6/commit/af75edd1b95e421200d845f37f050f9b5c3f086d#diff-50c3a295aed7723b0dc222d54b8863ceR335 )

kLastStableCompression would let me avoid hard-coding the list of compression types. It's not a high priority, though. Compression types are not added often, it's feasible to maintain the list manually.
Also
- I will also need names of compression types, see #3 below.
- Should it be 'KLastStableCompression'? What if somebody wants to try MyRocks with an unstable compression type? It will be nice that ALL possible compression types that are (or could be) supported by the binary are visible.

2. CompressionTypeSupported(), defined in util/compression.h, only internally.

This is a function which tells whether a particular compression type is compiled in. This is what I need the most and cannot do without. 
I would like to have this exposed through the RocksDB's public API.

3. CompressionTypeToString(), defined in util/compression.h, only internally.

This is nice to have but not essential. If I maintain the list of compression types, I may as well maintain their names.  If I stop doing that and start using kLastStableCompression constant instead, I'll also need 
CompressionTypeToString exposed through the public API.

IslamAbdelRahman:
1) We can have a function that returns a std::vector of supported compressions
2) let me expose it in public API
3) in `include/rocksdb/convenience.h` we have `GetStringFromCompressionType` is that good enought ?

spetrunia:
@IslamAbdelRahman , 
> We can have a function that returns a std::vector of supported compressions
> let me expose it in public API

Ok.

> in include/rocksdb/convenience.h we have GetStringFromCompressionType is that good enought ?

Yes.  This and the one above will meet all of my needs. 

Thanks!

IslamAbdelRahman:
Sorry @spetrunia  for the long delay, I am adding GetSupportedCompressions() function in this PR 
https://github.com/facebook/rocksdb/pull/1747

PR #450, opened 2016-11-22
Remove rocksdb_sequence_number

santoshbanda:
Summary:
This is removed in rocksdb here https://github.com/facebook/rocksdb/commit/36e4762ce0d1799021c1ce00a1c1b52edab3be02

Test Plan: mtr

Reviewers: tianx


gunnarku:
This is already getting fixed by https://github.com/facebook/mysql-5.6/pull/448. Test cases need to be fixed as well.

PR #454, opened 2016-11-28
Fix errors in TC_LOG_MMAP

jkedgar:
Summary: MySQL 5.6 has some errors in the implementation of TC_LOG_MMAP that were resolved in 5.7.  These errors show up when attempting to alter a table to change its storage engine from InnoDB to MyRocks or vice versa.  See also: https://www.percona.com/blog/2013/10/23/problems-with-multiple-xa-storage-engines-in-mysql-5-6/

This diff (at least the code change pieces) should be removed if/when MyRocks is ported to 5.7 or later.

This diff also contains the beginnings of a new suite of tests that intentionally run with two transactional engines at the same time (InnoDB an MyRocks).

Test Plan: New tests where both InnoDB and MyRocks are enabled simultaneously


mysql-test/suite/dual_engine/t/alter_table_between_engines_with_binlog.test
```diff
@@ -0,0 +1,20 @@
+--source include/have_innodb_and_rocksdb.inc
+--source include/have_log_bin.inc
+
+# Alter a table to move it from InnoDB to RocksDB
```
hermanlee: Maybe refactor the test code being shared by these two files.

hermanlee (commented):
Changes look good to me. Instead of dual-engine, maybe another name for the new test suite (multiengine, rocksdb-innodb, etc)? The engines subdirectory is probably not a good fit for this test.

lth (approved):

PR #460, opened 2016-12-01
Fix test to account for aio requests after restart

lgalanis:
Summary:
With performance schema enabled, sometimes after restart we see
non-zero submitted aio requests, which is not expected by the test. This
fix acounts for this small possibility.

Test Plan:
innodb_logical_read_ahead.test passes with new expected result

Reviewers:
tianx

Subscribers:

Tasks:
10401691

Blame Revision:


lgalanis:
moved internally

PR #463, opened 2016-12-06
Add support for LOCK IN SHARE MODE

### LINKED ISSUE #336: Support SELECT ... LOCK IN SHARE MODE in MyRocks

lth:
Add support for LOCK IN SHARE MODE. We already track what lock level we want, but on rocksdb, we always obtain it exclusively. We just need to pass in the right flag.

Closes #336

@update-submodule: rocksdb


reidHoruff (approved):
LGTM

PR #464, opened 2016-12-06
Adding MyRocks savepoint test case

yoshinorim:
Summary: SAVEPOINT + ROLLBACK TO SAVEPOINT is a common technique
to avoid MDL. This diff verifies it works.

Test Plan: mtr


mysql-test/suite/rocksdb/t/rollback_savepoint.test
```diff
@@ -0,0 +1,37 @@
+--source include/have_rocksdb.inc
+
+--disable_warnings
+DROP TABLE IF EXISTS t1, t2;
+--enable_warnings
+
+--enable_connect_log
+
+--source include/count_sessions.inc
```
lth: Probably not needed.

lth (approved):

PR #467, opened 2016-12-07
Yield the background thread when not necessary (#465)

### LINKED ISSUE #465: myrocks uses 1/4 to 1/2 of a CPU core doing nothing
server uses 4.0.9 kernel and MyRocks build is from Nov21:
git_date        2016-11-23T09:56:32-08:00
git_hash        e3d26b4d9e0317026d15f98695f4713a176d8c40
rocksdb_git_date        2016-11-21T21:24:11-08:00
rocksdb_git_hash        734e4acafbcea77934d5164c424b5084f0422c82

top shows mysqld CPU between 25% and 50%

Results from perf shows this is from the use of pthread_cond_timedwait. Not sure if it is from both uses or one as the one from void Rdb_drop_index_thread::run() has at least a 60 second timeout and the one from Rdb_background_thread::run() has a one second timeout

https://gist.github.com/mdcallag/05a020211058518530df20a4b3df1820

gunnarku:
These changes are the proposed fix for
https://github.com/facebook/mysql-5.6/issues/465. The intent of the
background thread is to wake up once per second and call SyncWAL() if
all the preconditions are met.

Previous implementation ended up constantly looping because of a logic
error. This solution removes the unnecessary checks and uses a call to
`mysql_cond_timedwait()` (ends up being `pthread_cond_timedwait()`) to
make sure that thread yields its quantum and OS scheduler can do its
thing while we're waiting for the timer to expire.

We still loop, but for the majority of the time we end up sleeping and
not using the CPU.

Basic testing:

  - Verify that function body is called once per second. If we need then
    we can tune it to be more precise and use what approximate amount of
    ms-s or ns-s we desire.
  - Before the fix the code path was executed hundreds of time per
    second and 25-49% of CPU was utilized. After the fix `htop` show
    marginal (0.0%) utilization.
  - `perf` records multiple orders of magnitude less events than before.

Fixes #465.

### DIFF
--- a/storage/rocksdb/ha_rocksdb.cc
+++ b/storage/rocksdb/ha_rocksdb.cc
@@ -10600,17 +10600,23 @@ static SHOW_VAR rocksdb_status_vars[]= {
 
 void Rdb_background_thread::run()
 {
+  // How many seconds to wait till flushing the WAL next time.
+  const int WAKE_UP_INTERVAL = 1;
+
   timespec ts_next_sync;
   clock_gettime(CLOCK_REALTIME, &ts_next_sync);
-  ts_next_sync.tv_sec++;
+  ts_next_sync.tv_sec += WAKE_UP_INTERVAL;
 
   for (;;)
   {
-    // wait for 1 second or until we received a condition to stop the thread
+    // Wait until the next timeout or until we receive a signal to stop the
+    // thread. Request to stop the thread should only be triggered when the
+    // storage engine is being unloaded.
     mysql_mutex_lock(&m_signal_mutex);
     auto ret __attribute__((__unused__)) = mysql_cond_timedwait(
-        &m_signal_cond, &m_signal_mutex, &ts_next_sync);
-    // make sure that no program error is returned
+      &m_signal_cond, &m_signal_mutex, &ts_next_sync);
+
+    // Check that we receive only the expected error codes.
     DBUG_ASSERT(ret == 0 || ret == ETIMEDOUT);
     bool local_stop= m_stop;
     bool local_save_stats= m_save_stats;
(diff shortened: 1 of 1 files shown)


storage/rocksdb/ha_rocksdb.cc
```diff
@@ -10619,29 +10625,37 @@ void Rdb_background_thread::run()
...
     }
 
-    // Flush the WAL if need be but don't do it more frequent
-    // than once per second
     timespec ts;
     clock_gettime(CLOCK_REALTIME, &ts);
-    if (ts.tv_sec - ts_next_sync.tv_sec >= 1)
+
+    // Flush the WAL.
+    if (rdb && rocksdb_background_sync)
```
hermanlee, line 10647: The rdb_bg_thread is used to both sync the wal file in the background and flush out the cached stats. rdb_bg_thread.request_save_stats() is called randomly based on compaction, memtable flushes. If it happens frequently, then on each loop, we call SyncWal(). The 1 second SyncWal behavior was meant to emulate Innodb's flush_log_at_trx_commit=2 behavior: http://dev.mysql.com/doc/refman/5.7/en/innodb-parameters.html#sysvar_innodb_flush_log_at_trx_commit.

While this code is functional and correct, we might end up calling SyncWal() multiple times in a second.

An alternative would be to ignore the signals from rdb_queue_save_stats_request(). In that case, this thread would only wake up once a second to perform syncs and stat flushes.

hermanlee (approved):

storage/rocksdb/ha_rocksdb.cc
gunnarku: Discussed in person. There are no signals generated from ` rdb_queue_save_stats_request()` so the current solution will still work.

PR #471, opened 2016-12-13
rename rocksdb_disable_2pc to rocksdb_enable_2pc, default true

reidHoruff:
This diff renames the flag rocksdb_disable_2pc to rocksdb_enable_2pc. The default for this value remains true, enabling 2pc by default.


hermanlee (approved):
Changes look good to me. You can even shorten the variable from rocksdb_enable_2pc to rocksdb_2pc, since the ON/OFF value implies "enable".

PR #478, opened 2016-12-16
const fix for storage/rocksdb (first half)

sidigit1205:
First harf of the const fix for all of the files under storage/rocksdb folder.

This pull request includes the changes from the previous two requests that are canceled.

### DIFF
--- a/storage/rocksdb/rdb_threads.cc
+++ b/storage/rocksdb/rdb_threads.cc
@@ -24,10 +24,10 @@
 
 namespace myrocks {
 
-void* Rdb_thread::thread_func(void* thread_ptr)
+void* Rdb_thread::thread_func(void* const thread_ptr)
 {
   DBUG_ASSERT(thread_ptr != nullptr);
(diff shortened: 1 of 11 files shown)


gunnarku (approved):

gunnarku (approved):

storage/rocksdb/rdb_threads.cc
```diff
@@ -68,7 +68,7 @@ int Rdb_thread::create_thread(
 }
 
 
-void Rdb_thread::signal(bool stop_thread)
+void Rdb_thread::signal(const bool &stop_thread)
```
laurynas-biveinis, line 71: This and other patches in the constification series replaced several pass-by-value args with pass-by-const references, where the arguments are small enough to be passed by value efficiently - like bool here. Has this been done intentionally?

george-lorch: It comes from upstream, and it was an intentional change, but I have no idea why. I agree 100%, there is no need to introduce the extra needed dereferencing when by-value is so minimal.

PR #483, opened 2016-12-21
Remove format version checking code

### LINKED ISSUE #296: Eventually remove old varchar code based on PRIMARY_FORMAT_VERSION_INITIAL and SECONDARY_FORMAT_VERSION_INITIAL
With the new varchar fixes and format going into mysqld, the old format will eventually need to be removed from the code once the new format is used everywhere. Filing an issue to remind us this needs to be done.

lth:
Now that the new format is deployed, we don't need to check for old formats anymore. Remove the checks as well as associated tests that exercises the format upgrade path.

Closes #296.


hermanlee (approved):
Red diffs look great. Thanks for removing this code.

PR #496, opened 2017-01-04
Fix prefixed unique secondary keys

### LINKED ISSUE #491: 'corrupt ' table with prefix blob primary + unique key
**Problem:
----------**
Version: '5.6.27-log'  socket: '/tmp/mysql.sock'  port: 3306  Source distribution
```
2016-12-31 22:48:06 16633 [Note] CHECKTABLE t: Checking table t
2016-12-31 22:48:06 16633 [Note] CHECKTABLE t:   Checking index a
2016-12-31 22:48:06 16633 [ERROR] CHECKTABLE t:   .. row 0: failed to fetch row by rowid

mysql> check table t extended;
+--------+-------+----------+----------+
| Table  | Op    | Msg_type | Msg_text |
+--------+-------+----------+----------+
| test.t | check | error    | Corrupt  |
+--------+-------+----------+----------+
```

```
mysql> show status like '%hash%';
+------------------+------------------------------------------+
| Variable_name    | Value                                    |
+------------------+------------------------------------------+
| git_hash         | f6ed777697db4ad7aee1e98c53243dced2b5891c |
| rocksdb_git_hash | 335981d47371cef0c48ebb58af53f2c8b8d8bd35 |
+------------------+------------------------------------------+
```

**How To Repeat;
----------------**
```
set sql_mode="";
drop table if exists t;
create table t (
  a blob,
  primary key(a(2)),
  unique (a(1))
) engine=rocksdb;
insert into t values('a');
check table t extended;
```

lth:
Unique secondary keys with prefixes do not work with extended keys, because the check to see if a key part in the unique index is equivalent to a key part in the primary key only looks at the field index, and does not check if it is prefixed or not. If it is prefixed, then in general, you cannot use it as part of the rowid since the primary key part may be longer than the secondary key part.

The fix is to check for key part length.

Closes #491

### DIFF
--- a/mysql-test/suite/rocksdb/r/unique_sec.result
+++ b/mysql-test/suite/rocksdb/r/unique_sec.result
@@ -183,3 +183,24 @@ ERROR 23000: Duplicate entry '1-1' for key 'PRIMARY'
 INSERT INTO t2 VALUES (2,1);
 ERROR 23000: Duplicate entry '1' for key 'a'
 DROP TABLE t2;
+#
+# Issue #491 (https://github.com/facebook/mysql-5.6/issues/491)
+#
+CREATE TABLE t (a BLOB, PRIMARY KEY(a(2)), UNIQUE KEY (a(1))) engine=rocksdb;
+INSERT INTO t VALUES('a');
+CHECK TABLE t EXTENDED;
+Table	Op	Msg_type	Msg_text
+test.t	check	status	OK
+DROP TABLE t;
+CREATE TABLE t (a VARCHAR(255), PRIMARY KEY(a), UNIQUE KEY (a(1))) engine=rocksdb;
+INSERT INTO t VALUES('a');
+CHECK TABLE t EXTENDED;
+Table	Op	Msg_type	Msg_text
+test.t	check	status	OK
+DROP TABLE t;
+CREATE TABLE t (a VARCHAR(255), PRIMARY KEY(a(2)), UNIQUE KEY (a(1))) engine=rocksdb;
+INSERT INTO t VALUES('a');
+CHECK TABLE t EXTENDED;
+Table	Op	Msg_type	Msg_text
+test.t	check	status	OK
+DROP TABLE t;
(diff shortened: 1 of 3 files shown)


hermanlee (commented):
I think the changes look good. If I understand the change correctly, since we did not check for prefix length in the past, we assume the field found in the secondary key part was equivalent to the primary key field. However, since it's not, the secondary key would not have the full primary key. With this change, the secondary key would store both the prefix and the full primary key field.

What condition causes check table to return back corrupt?

lth:
Check table will access the primary key row from secondary keys. The error from check tables is "failed to fetch row by rowid", which occurs because the key part it is using it is wrong.

storage/rocksdb/rdb_datadic.h
```diff
@@ -387,7 +387,11 @@ class Rdb_key_def
     SECONDARY_FORMAT_VERSION_INITIAL= 10,
     // This change the SK format to include unpack_info.
     SECONDARY_FORMAT_VERSION_UPDATE1= 11,
-    SECONDARY_FORMAT_VERSION_LATEST= SECONDARY_FORMAT_VERSION_UPDATE1,
```
hermanlee: Since indices with prefix columns are already broken and cannot be used correctly by queries, it might be fine to fix the code without updating the key version. This would simplify the diff and testcase. Backward compatibility shouldn't be needed since there are probably no existing installations relying on it because it never worked correctly in the first place.

spetrunia: What if there is an affected index somewhere that happens not to be used for reads, and so there are no complaints?
The data in the index is corrupted, and will remain so if not modified. 

Then, somebody might eventually write a query that uses the affected index. They will get an error.  If this patch bumps the index number, it will be easier to catch that case such case (by old index number).

I'm not sure whether the above reasoning has much merit in production

spetrunia (commented):
Aside from the question of whether to bump the format number, looks good to me.

storage/rocksdb/rdb_datadic.h
lth: So, I think there is a possibility that this could break something that was already working. I ended up removing it because it's not clear to me when we'd be able to remove backwards compatibility code, if we don't know who uses it.

We should probably decide together what our stance is towards backwards compatibility.

hermanlee:
Running check table with the fix on a database containing the old index format would report an error, so we can detect the problem. Bumping the version number might help us identify this issue more easily if it exists in some someone's setup, but I suspect there are few enough cases that they can be identified relatively quickly when the issue is raised. Hence, I'm in favor of the simpler change/test.

PR #498, opened 2017-01-06
add persistent cache support for MyRocks

reidHoruff:
This diff adds support for BlockBasedTableOptions.persistent_cache in MyRocks.

### DIFF
--- a/mysql-test/suite/rocksdb/r/rocksdb.result
+++ b/mysql-test/suite/rocksdb/r/rocksdb.result
@@ -925,6 +925,7 @@ rocksdb_paranoid_checks	ON
 rocksdb_pause_background_work	ON
 rocksdb_perf_context_level	0
 rocksdb_persistent_cache_path	
+rocksdb_persistent_cache_size	0
 rocksdb_pin_l0_filter_and_index_blocks_in_cache	ON
 rocksdb_print_snapshot_conflict_queries	OFF
 rocksdb_rate_limiter_bytes_per_sec	0
--- a/mysql-test/suite/rocksdb/r/rocksdb.result
+++ b/mysql-test/suite/rocksdb/r/rocksdb.result
@@ -924,6 +924,7 @@ rocksdb_override_cf_options
 rocksdb_paranoid_checks	ON
 rocksdb_pause_background_work	ON
 rocksdb_perf_context_level	0
+rocksdb_persistent_cache_path	
 rocksdb_pin_l0_filter_and_index_blocks_in_cache	ON
 rocksdb_print_snapshot_conflict_queries	OFF
 rocksdb_rate_limiter_bytes_per_sec	0
--- a/mysql-test/suite/rocksdb/t/persistent_cache.test
+++ b/mysql-test/suite/rocksdb/t/persistent_cache.test
@@ -6,11 +6,12 @@ DROP TABLE IF EXISTS t1;
 
 --let $_server_id= `SELECT @@server_id`
 --let $_expect_file_name= $MYSQLTEST_VARDIR/tmp/mysqld.$_server_id.expect
+--let $_cache_file_name= $MYSQLTEST_VARDIR/tmp/persistent_cache
 --exec echo "wait" >$_expect_file_name
 
 # restart server with correct parameters
 shutdown_server 10;
---exec echo "restart:--rocksdb_persistent_cache_path=/tmp/persistent_cache --rocksdb_persistent_cache_size=1000000000" >$_expect_file_name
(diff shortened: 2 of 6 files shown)


storage/rocksdb/ha_rocksdb.cc
```diff
@@ -1197,6 +1214,7 @@ static struct st_mysql_sys_var* rocksdb_system_variables[]= {
   MYSQL_SYSVAR(disabledatasync),
   MYSQL_SYSVAR(use_fsync),
   MYSQL_SYSVAR(wal_dir),
+  MYSQL_SYSVAR(persistent_cache_path),
```
hermanlee, line 1217: This seems to be missing the persistent_cache_size.

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -707,6 +711,19 @@ static MYSQL_SYSVAR_STR(wal_dir, rocksdb_wal_dir,
...
 
+static MYSQL_SYSVAR_STR(persistent_cache_path,
+  rocksdb_persistent_cache_path,
+  PLUGIN_VAR_RQCMDARG | PLUGIN_VAR_READONLY,
+  "Path for BlockBasedTableOptions::persistent_cache for RocksDB",
+  nullptr, nullptr, "");
+
+static MYSQL_SYSVAR_ULONG(persistent_cache_size,
+  rocksdb_persistent_cache_size,
+  PLUGIN_VAR_RQCMDARG,
```
hermanlee: Can make this parameter readonly too.

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -3697,6 +3715,12 @@ static int rocksdb_init_func(void* const p)
     mysql_mutex_unlock(&rdb_sysvars_mutex);
   }
 
+  if (rocksdb_persistent_cache_size > 0) {
```
hermanlee, line 3719: Should also verify the path is valid or fail bringing up the server.

hermanlee (changes requested):
Can you add a unittest for this?

storage/rocksdb/ha_rocksdb.cc
reidHoruff: I think it is best to let rocksdb handle that, no?

reidHoruff: To expand, it is also no trivial to determine if the cache size is valid so to save messy logic of user specifying one option and not the other (and I don't think its wise to check that user specified both size and path because then the cache could fail silently) I use the size as the flag, send values in and let rdb validate everything.

hermanlee: Ah, okay. RocksDB checks to see if the path is empty and fails the setup. That would be fine. It wasn't clear what would happen when the empty string was passed in and if RocksDB would create a default one.

mysql-test/suite/rocksdb/t/persistent_cache.test
```diff
@@ -0,0 +1,40 @@
...
+DROP TABLE IF EXISTS t1;
+--enable_warnings
+
+--let $_server_id= `SELECT @@server_id`
+--let $_expect_file_name= $MYSQLTEST_VARDIR/tmp/mysqld.$_server_id.expect
+--exec echo "wait" >$_expect_file_name
+
+# restart server with correct parameters
+shutdown_server 10;
+--exec echo "restart:--rocksdb_persistent_cache_path=/tmp/persistent_cache --rocksdb_persistent_cache_size=1000000000" >$_expect_file_name
```
hermanlee: This should use the $MYSQL_VARDIR path instead of /tmp. Otherwise, running tests in parallel could cause conflict with /tmp/persistent_cache.

storage/rocksdb/ha_rocksdb.cc
hermanlee: Yes, that is fine since it looks like RocksDB checks for empty string and invalidates. It wasn't clear what the default behavior was.

hermanlee (changes requested):

mysql-test/suite/rocksdb/r/rocksdb.result
```diff
@@ -924,6 +924,7 @@ rocksdb_override_cf_options
 rocksdb_paranoid_checks	ON
 rocksdb_pause_background_work	ON
 rocksdb_perf_context_level	0
+rocksdb_persistent_cache_path
```
hermanlee, line 927: Also surprised rocksdb_persistent_cache_size is not showing up here in these test results. Do you need to re-record them?

hermanlee (approved):
Looks good!

PR #501, opened 2017-01-09
Fix alloc_key_buffers() usage to avoid resource leaks

gunnarku:
Function `alloc_key_buffers()` has a problematic contract - if it fails
to allocate any memory then it doesn't clean up before returning the
error code. The caller would have to either call `free_key_buffers()`
in that case or leak some resources.

Fixes here:

  - Make the behavior of `alloc_key_buffers()` more explicit. We either
    succeed with all the allocations and return a success code or clean
    up after ourselves and return a failure code.

  - In all the places where `alloc_key_buffers()` is used, make sure that
    subsequent exit paths will call `free_key_buffers()` to avoid resource
    leaks.

  - Initialize all member variables properly to `nullptr`.

Discovered all of this while stepping through MyRocks initialization
logic.

### DIFF
--- a/storage/rocksdb/ha_rocksdb.cc
+++ b/storage/rocksdb/ha_rocksdb.cc
@@ -4245,7 +4245,7 @@ ha_rocksdb::ha_rocksdb(my_core::handlerton* const hton,
     m_pk_can_be_decoded(false),
     m_maybe_unpack_info(false),
     m_pk_tuple(nullptr), m_pk_packed_tuple(nullptr),
-    m_sk_packed_tuple(nullptr),
+    m_sk_packed_tuple(nullptr), m_end_key_packed_tuple(nullptr),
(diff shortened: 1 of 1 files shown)


alxyang (commented):
Correct me if I am mistaken, but `free_key_buffers` is already called during the course of `close`, which covers all exit paths where `alloc_key_buffers` is called.

So these extra calls to free_key_buffers() are redundant (it will be called on close anyways)

gunnarku:
Can you please tell me exactly which calls to `free_key_buffers()` you think are redundant? They are needed in case of error path to clean up because `alloc_key_buffers` would have allocated resources earlier. If `open()` fails then `close()` won't be called.

In addition, `open()` can be called multiple times in succession depending on what error code it returns. Therefore it's responsible for cleaning up the state in the beginning as well.

lth (approved):
looks good to me

PR #502, opened 2017-01-10
avoid calling GetOptions() destructor per query

alxyang:
We currently call GetOptions() for every query by MyRocks, to determine if we can use a bloom filter.  GetOptions() constructs an object, which we can avoid doing if we cache the prefix_extractor (which is immutable for the lifetime of the db) on startup.


lth (changes requested):
As discussed, unordered_map isn't actually thread safe...

PR #503, opened 2017-01-10
avoid calling GetOptions() destructor per query

alxyang:
We currently call GetOptions() for every query by MyRocks, to determine if we can use a bloom filter. GetOptions() constructs an object, which we can avoid doing if we cache the prefix_extractor (which is immutable for the lifetime of the db) when setting up key descriptors.

### DIFF
--- a/storage/rocksdb/ha_rocksdb.cc
+++ b/storage/rocksdb/ha_rocksdb.cc
@@ -10778,8 +10778,8 @@ bool can_use_bloom_filter(THD *thd,
     return can_use;
   }
 
-  rocksdb::Options opt = rdb->GetOptions(kd.get_cf());
-  if (opt.prefix_extractor)
+  const rocksdb::SliceTransform* prefix_extractor= kd.get_extractor();
+  if (prefix_extractor)
   {
     /*
       This is an optimized use case for CappedPrefixTransform.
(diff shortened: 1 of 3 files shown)


storage/rocksdb/ha_rocksdb.cc
```diff
@@ -4876,7 +4876,7 @@ int ha_rocksdb::alloc_key_buffers(const TABLE* const table_arg,
   }
 
   // move this into get_table_handler() ??
-  m_pk_descr->setup(table_arg, tbl_def_arg);
+  m_pk_descr->setup(rdb, table_arg, tbl_def_arg);
```
lth: I think you can just use `rdb_get_rocksdb_db` instead of passing rdb.

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -10778,8 +10778,8 @@ bool can_use_bloom_filter(THD *thd,
     return can_use;
   }
 
-  rocksdb::Options opt = rdb->GetOptions(kd.get_cf());
-  if (opt.prefix_extractor)
+  auto prefix_extractor= kd.get_extractor();
```
lth: I'm pretty sure this will create a copy of the shared_ptr, which will inc/dec reference counts.

lth (approved):
I wish tests weren't broken!

PR #505, opened 2017-01-10
Record missing test for persistent cache variables

reidHoruff:
Recent tests also show that rocksdb.persistent_cache was also failing in sandcastle but I was not able to reproduce with: mysqltest.sh --clean --jenkins --testset=RocksDBBig --asan --async-client --workers=16 --copybuild=/tmp/mysql_sandcastle rocksdb.persistent_cache


lth:
This is not the right fix. You need to add the sys var test under rocksdb_sys_vars.

alxyang:
I can repro `rocksdb.persistent_cache` failure by running the test suite as a whole, with `--suite=rocksdb`

reidHoruff:
https://github.com/facebook/mysql-5.6/pull/509

PR #510, opened 2017-01-12
Name MyRocks threads correctly to enhance debugging

gunnarku:
Currently in `gdb` the MyRocks threads show up as "my-func". That's
because how setting thread names is implemented in MySQL (using the
substring of a callback function name).

Given how `Rdb_thread` class creates threads, the names all end up the
same. Therefore we need to reset them ourselves.


alxyang (approved):
lgtm

PR #512, opened 2017-01-18
change persistent_cache_size to persistent_cache_size_mb

reidHoruff:
Changes the name and unit of rocskdb_persistent_cache_size from bytes to megabytes.


reidHoruff:
unable to import pull request to phabricator so I am aborting this diff and using internal workflow.

PR #513, opened 2017-01-18
rename missing rocksdb_disable_2pc to rocksdb_enable_2pc

reidHoruff:
rocksdb_disable_2pc was recently renamed to rocksdb_enable_2pc and not all files were updated. This updates a new test that was missed.


reidHoruff:
Unable to import pull request into phabricator so I am aborting and using internal work flow. A task has been filed.

Issue #427, opened 2016-11-15
Flush memtable and compact L0 files for read only workloads


mdcallag:
The state of the LSM tree (data in memtable, #files in L0) is a significant source of variance for performance on read-only and read-heavy tests. This is much worse for read-only than for read-mostly workloads. Data in the memtable and files in L0 are overhead for read-only/mostly tests. When the memtable isn't full and when there aren't too many files in L0 then the LSM can stay in that state for a long time because there are no writes to trigger it to change. This is an issue for real workloads and for benchmarks.

I lost a few days debugging this problem recently. I know our users will also be confused by this and some will be disappointed by MyRocks performance. Part of the problem is that synthetic benchmarks are synthetic, but we can't get the world to stop running sysbench.

Can we solve this in two parts?
1. Give me an option to flush memtable and then remove all files from L0. This must be dynamic. Perhaps there is an option to flush the memtable today. Temporarily setting the L0 compaction trigger to 0 lets me flush the L0, but I doubt that is supported today.
2. Most users won't know to do things I list in step 1, and they shouldn't have to know them. Can we make MyRocks and/or RocksDB adaptive and figure out it should flush the memtable and L0 when a test is read-only and sometimes when it is extremely read-heavy?

Someone suggested that CompactRange(nullptr, nullptr) flushes memtables and compacts L0. For step 1 we need a way to trigger that via SQL.

mdcallag:
With sysbench read-only range query tests I get 5% to 20% more QPS when the memtable and level 0 is flushed. Details at https://gist.github.com/mdcallag/68052cdd36fe122354bc23ec337fb986

IslamAbdelRahman:
@siying suggested that MyRocks can for now issues a Flush() and compact all L0 files using CompactFiles() API.
but a better solution will be to bring back read triggered compactions from leveldb

yoshinorim:
Siying pointed out that for MyRocks to know list of L0/L1 files from a CF, DB::GetColumnFamilyMetaData() can be used. Find files from there and call DB::CompactFiles() with those files.

yoshinorim:
We have rocksdb_compact_cf variable to rebuild specific CF. We can add another command variable to flush MemTable and L0/L1 (or SQL function).

zhangjinpeng87:
Since RocksDB has block-cache, is it necessary to flush memtable and compact L0 sst files for read-only queries? Because there are only at most 4 sst files in L0.

[...]

PR #520, opened 2017-02-03
add capability to manually flush memtable and l0

alxyang:
This provides an option to flush memtable and remove all files from L0, which can be useful for lessening variance in benchmarking on read-only and read-heavy tests, in cases where the memtable isn't full and/or there aren't many files in L0, and no writes are occurring.

See more details at https://github.com/facebook/mysql-5.6/issues/427.

@update-submodule: rocksdb

### DIFF
--- a/storage/rocksdb/ha_rocksdb.cc
+++ b/storage/rocksdb/ha_rocksdb.cc
@@ -283,8 +283,44 @@ static void rocksdb_force_flush_memtable_now_stub(
 static int rocksdb_force_flush_memtable_now(
     THD *const thd, struct st_mysql_sys_var *const var, void *const var_ptr,
     struct st_mysql_value *const value) {
-  sql_print_information("RocksDB: Manual memtable flush\n");
+  sql_print_information("RocksDB: Manual memtable flush.");
   rocksdb_flush_all_memtables();
+  return HA_EXIT_SUCCESS;
+}
+
+static void rocksdb_force_flush_memtable_and_lzero_now_stub(
+    THD *const thd, struct st_mysql_sys_var *const var, void *const var_ptr,
+    const void *const save) {}
+
+static int rocksdb_force_flush_memtable_and_lzero_now(
+    THD *const thd, struct st_mysql_sys_var *const var, void *const var_ptr,
+    struct st_mysql_value *const value) {
+  sql_print_information("RocksDB: Manual memtable and L0 flush.");
+  rocksdb_flush_all_memtables();
+
+  const Rdb_cf_manager &cf_manager = rdb_get_cf_manager();
+  const rocksdb::CompactionOptions c_options = rocksdb::CompactionOptions();
+  for (const auto &cf_handle : cf_manager.get_all_cf()) {
+    rocksdb::ColumnFamilyMetaData metadata;
+    rdb->GetColumnFamilyMetaData(cf_handle, &metadata);
+
+    DBUG_ASSERT(metadata.levels[0].level == 0);
+    std::vector<std::string> file_names;
+    for (auto &file : metadata.levels[0].files) {
+      file_names.emplace_back(file.db_path + file.name);
+    }
+
+    if (!file_names.empty()) {
+      rocksdb::Status s;
+      s = rdb->CompactFiles(c_options, cf_handle, file_names, 1);
+
+      if (!s.ok()) {
+        rdb_handle_io_error(s, RDB_IO_ERROR_GENERAL);
+        return HA_EXIT_FAILURE;
+      }
+    }
+  }
+
   return HA_EXIT_SUCCESS;
 }
 
@@ -385,6 +421,7 @@ static my_bool rocksdb_enable_2pc = 0;
 static char *rocksdb_strict_collation_exceptions;
 static my_bool rocksdb_collect_sst_properties = 1;
 static my_bool rocksdb_force_flush_memtable_now_var = 0;
+static my_bool rocksdb_force_flush_memtable_and_lzero_now_var = 0;
 static uint64_t rocksdb_number_stat_computes = 0;
 static uint32_t rocksdb_seconds_between_stat_computes = 3600;
 static long long rocksdb_compaction_sequential_deletes = 0l;
@@ -1086,6 +1123,13 @@ static MYSQL_SYSVAR_BOOL(
     rocksdb_force_flush_memtable_now, rocksdb_force_flush_memtable_now_stub,
     FALSE);
 
+static MYSQL_SYSVAR_BOOL(
+    force_flush_memtable_and_lzero_now,
+    rocksdb_force_flush_memtable_and_lzero_now_var, PLUGIN_VAR_RQCMDARG,
+    "Acts similar to force_flush_memtable_now, but also compacts all L0 files.",
+    rocksdb_force_flush_memtable_and_lzero_now,
+    rocksdb_force_flush_memtable_and_lzero_now_stub, FALSE);
+
 static MYSQL_THDVAR_BOOL(
     flush_memtable_on_analyze, PLUGIN_VAR_RQCMDARG,
     "Forces memtable flush on ANALZYE table to get accurate cardinality",
@@ -1299,6 +1343,7 @@ static struct st_mysql_sys_var *rocksdb_system_variables[] = {
     MYSQL_SYSVAR(strict_collation_exceptions),
     MYSQL_SYSVAR(collect_sst_properties),
     MYSQL_SYSVAR(force_flush_memtable_now),
+    MYSQL_SYSVAR(force_flush_memtable_and_lzero_now),
     MYSQL_SYSVAR(flush_memtable_on_analyze),
     MYSQL_SYSVAR(seconds_between_stat_computes),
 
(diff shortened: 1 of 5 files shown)


storage/rocksdb/ha_rocksdb.cc
```diff
@@ -290,6 +290,42 @@ static int rocksdb_force_flush_memtable_now(
...
+
+  const Rdb_cf_manager &cf_manager = rdb_get_cf_manager();
+  for (const auto &cf_handle : cf_manager.get_all_cf()) {
+    rocksdb::ColumnFamilyMetaData metadata;
+    rdb->GetColumnFamilyMetaData(cf_handle, &metadata);
+
+    std::vector<std::string> file_names;
+    for (auto &file : metadata.levels[0].files) {
+      DBUG_ASSERT(metadata.levels[0].level == 0);
+      file_names.emplace_back(file.db_path + "/" + file.name);
```
hermanlee: It seems odd that db_path doesn't end in '/', but I do see code in rocksdb adding '/' between path and filename in some cases. I assume the "/" is required to assemble the full path?

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -290,6 +290,42 @@ static int rocksdb_force_flush_memtable_now(
...
+    }
+
+    if (file_names.size() > 0) {
+      rocksdb::Status s;
+      s = rdb->CompactFiles(rocksdb::CompactionOptions(), cf_handle, file_names,
+          metadata.levels.size() - 1);
```
hermanlee: Should this target level 1 for the output, instead of the bottommost level?

storage/rocksdb/ha_rocksdb.cc
alxyang: Yes, it seems so.  I used https://github.com/facebook/rocksdb/blob/master/db/compact_files_test.cc#L105 as my reference which is doing pretty much the exact same thing we want here.

storage/rocksdb/ha_rocksdb.cc
alxyang: Ok, no problem.

storage/rocksdb/ha_rocksdb.cc
alxyang: ok, I stepped through in GDB and found that actually the "/" isn't required (even though the test case above uses it).

It parses the number from the sst file during CompactFiles() using https://github.com/facebook/rocksdb/blob/master/db/filename.cc#L99-L108.  

So I'll remove the extra "/" even though it doesn't hurt really

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -292,6 +292,42 @@ static int rocksdb_force_flush_memtable_now(
...
+    rocksdb::ColumnFamilyMetaData metadata;
+    rdb->GetColumnFamilyMetaData(cf_handle, &metadata);
+
+    std::vector<std::string> file_names;
+    for (auto &file : metadata.levels[0].files) {
+      DBUG_ASSERT(metadata.levels[0].level == 0);
+      file_names.emplace_back(file.db_path + file.name);
+    }
+
+    if (file_names.size() > 0) {
```
jkedgar: Nit - I have small preference for `if (!file_names.empty()) {`.  Not for performance - I just think it is easier to read.

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -292,6 +292,42 @@ static int rocksdb_force_flush_memtable_now(
...
+
+  const Rdb_cf_manager &cf_manager = rdb_get_cf_manager();
+  for (const auto &cf_handle : cf_manager.get_all_cf()) {
+    rocksdb::ColumnFamilyMetaData metadata;
+    rdb->GetColumnFamilyMetaData(cf_handle, &metadata);
+
+    std::vector<std::string> file_names;
+    for (auto &file : metadata.levels[0].files) {
+      DBUG_ASSERT(metadata.levels[0].level == 0);
+      file_names.emplace_back(file.db_path + file.name);
```
jkedgar: Is `file.db_path` guaranteed to end with a '/'?  The reason I ask is RocksDB's own tests add a '/' in the middle.  https://github.com/facebook/rocksdb/blob/master/db/compact_files_test.cc#L104-L107

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -292,6 +292,42 @@ static int rocksdb_force_flush_memtable_now(
...
 }
 
+static void rocksdb_force_flush_memtable_and_lzero_now_stub(
+    THD *const thd, struct st_mysql_sys_var *const var, void *const var_ptr,
+    const void *const save) {}
+
+static int rocksdb_force_flush_memtable_and_lzero_now(
+    THD *const thd, struct st_mysql_sys_var *const var, void *const var_ptr,
+    struct st_mysql_value *const value) {
+  sql_print_information("RocksDB: Manual memtable and L0 flush\n");
```
jkedgar: The `sql_print_information()` function automatically adds a '\n' to the end of the string that you pass it.  Unless you are intentionally trying to add an empty line in the log file after this line you should remove the '\n' here.  Note that `rocksdb_force_flush_memtable_now()` should probably be changed as well.

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -292,6 +292,42 @@ static int rocksdb_force_flush_memtable_now(
...
+  rocksdb_flush_all_memtables();
+
+  const Rdb_cf_manager &cf_manager = rdb_get_cf_manager();
+  for (const auto &cf_handle : cf_manager.get_all_cf()) {
+    rocksdb::ColumnFamilyMetaData metadata;
+    rdb->GetColumnFamilyMetaData(cf_handle, &metadata);
+
+    std::vector<std::string> file_names;
+    for (auto &file : metadata.levels[0].files) {
+      DBUG_ASSERT(metadata.levels[0].level == 0);
```
jkedgar: Why is this assert inside the inner for loop?  I would expect that checking this once in the outer for loop would be sufficient.

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -292,6 +292,42 @@ static int rocksdb_force_flush_memtable_now(
...
+      file_names.emplace_back(file.db_path + file.name);
+    }
+
+    if (file_names.size() > 0) {
+      rocksdb::Status s;
+      s = rdb->CompactFiles(rocksdb::CompactionOptions(), cf_handle, file_names,
```
jkedgar: I suspect that `rocksdb::CompactionOptions()` will not change throughout both the inner and the outer loop.  Can we put this into a local variable to avoid calling it over and over?

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -292,6 +292,42 @@ static int rocksdb_force_flush_memtable_now(
...
+static int rocksdb_force_flush_memtable_and_lzero_now(
+    THD *const thd, struct st_mysql_sys_var *const var, void *const var_ptr,
+    struct st_mysql_value *const value) {
+  sql_print_information("RocksDB: Manual memtable and L0 flush\n");
+  rocksdb_flush_all_memtables();
+
+  const Rdb_cf_manager &cf_manager = rdb_get_cf_manager();
+  for (const auto &cf_handle : cf_manager.get_all_cf()) {
+    rocksdb::ColumnFamilyMetaData metadata;
+    rdb->GetColumnFamilyMetaData(cf_handle, &metadata);
```
jkedgar: What is guaranteeing that the data inside `metadata` is not being changed while we are reading it?  Is this known to be safe?

storage/rocksdb/ha_rocksdb.cc
alxyang: Originally I did put a `/` there.  @hermanlee commented above asking if it was necessary, I stepped through GDB and found that it was not.  See my comment above for more details.

See https://github.com/facebook/rocksdb/blob/master/db/filename.cc#L99-L108 for how the sstfile name is parsed.

storage/rocksdb/ha_rocksdb.cc
reidHoruff: When calling into GetColumnFamilyMetaData() the current SuperVersion is referenced and the metadata is copied over from that. When the function exits the SuperVersion is de-referenced. A SuperVersion represents an immutable state of the database (when the manifest is updated). It is possible, however, that we are not the first to request that a given L0 file be compacted. This would cause CompactFiles to error out.

It is also possible that by the time we request to have a file flushed the file has been deleted because we have de-referenced the SuperVersion from which the metadata was copied. To solve this we should hold our own superversion reference and copy over the metadata ourselves.

This still doesn't seem like an issue to me considering this is to be used for performance testing/debugging. There is no copy-race condition and CompactFiles() should gracefully handle all sanity checks.

alxyang: I tested a few things, like passing in an SST file that no longer exists, or trying to compact a level that doesnt exist, and in both cases `CompactFiles` handled it gracefully.


It'll write something to the logs like so:
```
2017-03-20 18:26:50 2306114 [Warning] MyRocks: failed to read/write in RocksDB. Error type = RDB_IO_ERROR_GENERAL, stat
us code = 4, status = Invalid argument: Specified compaction input file /000099.sst does not exist in column family def
ault.
```

jkedgar: Right - metadata is a copy.  I should have seen that.

storage/rocksdb/ha_rocksdb.cc
jkedgar: No problem.  As long as it has been checked.

hermanlee (approved):

PR #531, opened 2017-02-15
fix cardinality for secondary keys created via inplace alter

alxyang:
Cardinality requires the keydef during the sst file creation to be calculated. Without the keydef, the collector doesn't store cardinality stats. This adds a member to the ddl_manager so that the properties collector can find the keydef ongoing creation before it is persisted to the data dictionary.


storage/rocksdb/rdb_datadic.cc
```diff
@@ -2692,6 +2692,20 @@ void Rdb_ddl_manager::erase_index_num(const GL_INDEX_ID &gl_index_id) {
   m_index_num_to_keydef.erase(gl_index_id);
```
alxyang, line 2692: why aren't we locking here? std::map isn't thread safe, and couldn't this be called from multiple threads?

alxyang:
putting this up so it can start being reviewed, but i plan to run through a few bulk load runs via cp_sql for further testing, in addition to the test case i added.

storage/rocksdb/rdb_datadic.cc
```diff
@@ -2692,6 +2692,24 @@ void Rdb_ddl_manager::erase_index_num(const GL_INDEX_ID &gl_index_id) {
...
+void Rdb_ddl_manager::add_uncommitted_keydefs(
+    const std::unordered_set<std::shared_ptr<Rdb_key_def>> &indexes) {
+  mysql_rwlock_wrlock(&m_rwlock);
+  for (const auto &index : indexes) {
+    m_index_num_to_uncommitted_keydef[index->get_gl_index_id()] = index;
+  }
+  mysql_rwlock_unlock(&m_rwlock);
+}
+
+void Rdb_ddl_manager::erase_uncommitted_keydefs(
```
hermanlee: Preference to call this remove_uncommitted_keydefs()

storage/rocksdb/rdb_datadic.h
```diff
@@ -742,6 +742,10 @@ class Rdb_ddl_manager {
   my_core::HASH m_ddl_hash; // Contains Rdb_tbl_def elements
   // maps index id to <table_name, index number>
   std::map<GL_INDEX_ID, std::pair<std::string, uint>> m_index_num_to_keydef;
+
+  //maps index id to key definitons not yet committed to data dictionary
```
hermanlee: nit typo: definitions and space between // and map.

Could also mentioned this is mainly used by alter table.

storage/rocksdb/rdb_datadic.cc
hermanlee: It's only called during the ~Rdb_tbl_def().

mysql-test/suite/rocksdb/t/add_index_inplace.test
```diff
@@ -342,4 +342,28 @@ while ($i <= $max) {
 
 DROP TABLE t1;
 
+# Test that fast secondary index creation updates cardinality properly
+CREATE TABLE t1 (i INT PRIMARY KEY, j INT) ENGINE = ROCKSDB;
+INSERT INTO t1 VALUES (1,2), (2,4), (3,6), (4,8), (5,10);
+ALTER TABLE t1 ADD INDEX kj(j), ALGORITHM=INPLACE;
```
hermanlee, line 348: This alter table completes before the flush memtable is called. Can you modify this test so that flush memtable is called while the alter table is running? Basically add a debug_sync_point in the alter path so that it waits on the commit function. Data is in the memtable and the temp key_def is in the hashtable, but the new key definition has not been made permanent in the data dictionary. Trigger a memtable flush at this point. You will need to use multiple connections to do this.

hermanlee (changes requested):

mysql-test/suite/rocksdb/t/add_index_inplace.test
alxyang: as discussed offline, @hermanlee will make a separate diff including this test case to be squashed into this diff later

storage/rocksdb/rdb_datadic.cc
alxyang: actually, I was wrong.  There is a lock around this code https://github.com/facebook/mysql-5.6/blob/fb-mysql-5.6.35/storage/rocksdb/rdb_datadic.cc#L3229-L3249

hermanlee (approved):

PR #535, opened 2017-02-20
Remove bashism in t/slocket_listen.sh

grooverdan:
cases a test failure when sh isn't bash.

### DIFF
--- a/mysql-test/t/slocket_listen.sh
+++ b/mysql-test/t/slocket_listen.sh
@@ -14,7 +14,7 @@ dirname=`dirname "$0"`
 
 ###########################################################################
 
-function log_debug {
+log_debug() {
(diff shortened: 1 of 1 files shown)


jkedgar:
Which shell were you using that generated the failure?  I checked ksh and it looks like it uses 'function <name>'.

grooverdan:
dash, which is what ubuntu-16.04 uses as the default sh link
<pre>
$ dpkg -l | grep dash
ii  dash                                0.5.8-2.1ubuntu2                    ppc64el      POSIX-compliant shell
$ ls -la /bin/sh
lrwxrwxrwx 1 root root 4 Feb 18  2016 /bin/sh -> dash
</pre>

grooverdan:
thanks @jkedgar

PR #536, opened 2017-02-21
Ignore all socket files not just log sockets

jkedgar:
Summary: A change was added recently to ignore specific socket files in the data directory, but other users may create socket files there as well.  We don't need to specify which socket files to ignore as attempting to access any socket file will generate an error.  We simply need to skip all socket files.

Test Plan: TBD

Reviewers: hermanlee4

Subscribers:


hermanlee (approved):
Can you squash this with the changes that added checks for log sockets?

PR #539, opened 2017-02-22
Improve add_index_alter_cardinality test

hermanlee:
Summary:

Split add_index_inplace_cardinality test out and add a debug_sync point
to it so that the flush of the memtable occurs while the alter is
running.

Test Plan:

Run unit test


alxyang (approved):
thanks! :)

PR #540, opened 2017-02-22
Don't assume buffer is relative to table->record[0]

jkedgar:
Summary:  Some of the functions that unpack and pack a record assume that the buffer passed in is a part of table->record[0].  This appears to be false when using partitions.  This diff updates this code to not make that assumption.  To do this, the field->ptr value is temporarily updated to point into the provided buffer and then restored.  In order to avoid potential problems with this (i.e. field->ptr left pointing into the buffer accidentally), code was refactored to keep the location where field->ptr was reset and where it was restored close together.

Test Plan: MTR plus new test for Issue 495

### DIFF
--- a/mysql-test/suite/rocksdb/t/issue495.test
+++ b/mysql-test/suite/rocksdb/t/issue495.test
@@ -1,0 +1,29 @@
+drop table if exists t;
+create table t (
+  a int,
+  b int,
+  c varchar(12249) collate latin1_bin,
+  d datetime,
+  e int,
+  f int,
+  g blob,
+  h int,
+  i int,
+  key (b,e),
+  key (h,b)
+) engine=rocksdb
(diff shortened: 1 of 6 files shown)


alxyang (commented):
The refactor looks good to me overall.  There are a lot of changes though so I'll let someone else (maybe @hermanlee  or @lth) look through as well.

I'd like this to be landed ASAP as some of the TTL functionality I'm working on can probably reuse some of these functions.

mysql-test/suite/rocksdb/t/issue495.test
```diff
@@ -0,0 +1,35 @@
+set sql_mode='';
```
lth: You can probably reproduce the issue without this line.

lth (approved):

PR #541, opened 2017-02-22
Have the created thread set its own name

jkedgar:
Summary: On some scenarios, having the creating thread set the created thread's name gets an errno 13 (EACCES).  Change the code so that the created thread sets its own name solves the problem.

Test Plan:  MTR


alxyang (approved):
thanks! and welcome back :)

PR #550, opened 2017-02-27
Fix rocksdb.bulk_load on slower machines

jkedgar:
Summary: The rocksdb.bulk_load test creates and accesses some very large tables and doesn't finish in time on a standard Debug build on a devserver.  Reduce the size of the tables so that it will finish.

I reduced the size of the tables from 10,000,000 rows to 5,000,000 rows.  This should still provide sufficient size to validate what we are trying to test without taking so long.

Test Plan: MTR


hermanlee (approved):
Should probably remove devservers from the title.

PR #552, opened 2017-02-27
Improve feature (embedded server, storage engines) selection

aloiret:
Summary:
RocksDB was built by default using the DEFAULT option from MYSQL_ADD_PLUGIN.
Use the FEATURE_SET mechanism instead, which still builds it by default but
makes it possible to disable it.

Also, allow to set features as a semicolon-separated list of features.
Possibles values are: embedded, archive, blackhole, example, federated,
innobase, partition and rocksdb.

Signed-off-by: Arthur Loiret <arthur.loiret@adeline.mobi>

### DIFF
--- a/cmake/build_configurations/feature_set.cmake
+++ b/cmake/build_configurations/feature_set.cmake
@@ -13,66 +13,88 @@
 # along with this program; if not, write to the Free Software
 # Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 
-SET(FEATURE_SET "community" CACHE STRING 
+SET(FEATURE_SET "community" CACHE STRING
 " Selection of features. Options are
- - xsmall : 
- - small: embedded
- - classic: embedded + archive + federated + blackhole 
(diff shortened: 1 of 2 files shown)


jkedgar:
@aloiret, I'm sorry this is taking so long.  When I first imported this into our test framework a bunch of tests failed, but it looked like flaky tests - it didn't initially appear to be related to your changes.  I tried to initiate the tests again, but that isn't succeeding.  I've alerted the appropriate people and they have confirmed that something is broken related to restarting tests.  Unfortunately they have been busy and haven't fixed it yet, so I can't re-run the tests automatically.

However, we could get the tests re-run if you make any modification to your Pull Request and push it back up.  You can even make a change, push it, undo the change and push that.  I'll reimport your PR and the tests should automatically re-run in that scenario.

jkedgar:
We want to have RocksDB to be built by default.  Changing this would affect us as well as MariaDB and Percona.  Also I understand that FEATURE_SET is disabled in 8.0 as we think about moving forward.

PR #554, opened 2017-02-28
Fix compile error when building unittest/gunit

jkedgar:
Multiple changes to fix problems compiling and running merge_large_test-t and merge_small_test-t.


alxyang (approved):

hermanlee:
Can we have more descriptive title for the diff?

PR #556, opened 2017-03-03
Correctly handle setting auto increment to largest value

jkedgar:
Summary: The auto increment code in MyRocks allowed a wrap-around, yield incorrect results

Test Plan: MTR and new rocksdb.issue255 test

### DIFF
--- a/storage/rocksdb/ha_rocksdb.cc
+++ b/storage/rocksdb/ha_rocksdb.cc
@@ -25,6 +25,7 @@
 
 /* C++ standard header files */
 #include <algorithm>
+#include <limits>
 #include <queue>
 #include <set>
 #include <string>
@@ -3740,7 +3741,12 @@ void ha_rocksdb::update_auto_incr_val() {
 
   my_bitmap_map *const old_map =
       dbug_tmp_use_all_columns(table, table->read_set);
-  new_val = field->val_int() + 1;
+  new_val = field->val_int();
+  // don't increment if we would wrap around
+  if (new_val != std::numeric_limits<longlong>::max()) {
+    new_val++;
+  }
+
   dbug_tmp_restore_column_map(table->read_set, old_map);
 
   longlong auto_incr_val = m_tbl_def->m_auto_incr_val;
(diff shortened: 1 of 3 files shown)


storage/rocksdb/ha_rocksdb.cc
```diff
@@ -3819,7 +3820,11 @@ void ha_rocksdb::update_auto_incr_val() {
 
   my_bitmap_map *const old_map =
       dbug_tmp_use_all_columns(table, table->read_set);
-  new_val = field->val_int() + 1;
+  new_val = field->val_int();
+  if (new_val + 1 > new_val) {  // don't increment if we would wrap around
```
laurynas-biveinis: Isn't this relying on the undefined behavior of signed integer overflow?

laurynas-biveinis: ...whereas the next snippet checks against "std::numeric_limits<longlong>::max()" without UB

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -9062,6 +9070,23 @@ void ha_rocksdb::get_auto_increment(ulonglong off, ulonglong inc,
       // For further information please visit:
       // http://dev.mysql.com/doc/refman/5.7/en/replication-options-master.html
       new_val = ((last_val + (inc - off) - 1) / inc) * inc + off;
```
laurynas-biveinis: AFAICT this a (pre-existing) undefined behavior too

jkedgar:
Good calls @laurynas-biveinis.  I was getting lazy.  There's a difference between using code that you know will work on 99.999% of the servers in existence and using code that you know will always work.  I'll fix all of the mentioned issues.

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -9047,21 +9053,48 @@ void ha_rocksdb::get_auto_increment(ulonglong off, ulonglong inc,
   if (inc == 1 && off == 1) {
     // Optimization for the standard case where we are always simply
     // incrementing from the last position
-    *first_value = auto_incr++;
+    if ((*first_value= auto_incr) != std::numeric_limits<longlong>::max()) {
+      // Make sure we don't wrap around
+      auto_incr++;
```
hermanlee: This change breaks auto-increment and two concurrent calls can end up returning the same "first_value". If you want to check before the increment, I think you need to switch this to a cas call.

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -9047,21 +9053,48 @@ void ha_rocksdb::get_auto_increment(ulonglong off, ulonglong inc,
...
+        //   1) set the new value to the last possible number in our sequence
+        //      as described above.  The problem with this is that this
+        //      number could be smaller than a value in an existing row.
+        //   2) set the new value to the largest possible number.  This number
+        //      may not be in our sequence, but it is guaranteed to be equal
+        //      to or larger than any other value already inserted.
+        //
+        //  For now I'm going to take option @2.
+        new_val = std::numeric_limits<longlong>::max();
+        auto_incr = new_val; // Store the largest value into auto_incr
```
hermanlee: Blindly setting auto_incr to max value can cause auto_incr to overflow when combined with concurrent calls to auto_incr++. Again, I think you need to perform a cas here to do this correctly.

hermanlee (changes requested):

storage/rocksdb/ha_rocksdb.cc
jkedgar: You are absolutely correct.  I'm getting lazy.  :(

storage/rocksdb/ha_rocksdb.cc
jkedgar: I agree that the way it is in the other one (with auto_incr++) this has the possibility of overflow, but once I fix that to use a cas I don't think I need to change this here.

hermanlee (approved):
Thanks for fixing this!

PR #558, opened 2017-03-06
The srv_monitor_event event needs to be created even in read-only mode

jkedgar:
Summary:
In a past diff a change was made to use srv_monitor_event more often.  This means that it needs to be created even if the server is in read-only mode.

Squash with: https://reviews.facebook.net/D17673

Test Plan: MTR


alxyang (approved):
lgtm

PR #571, opened 2017-03-16
Support TTL for tables with primary keys in MyRocks

alxyang:
This changeset adds a TTL (Time to live) functionality for tables with primary keys in MyRocks.   At a high level, here are the core mechanisms behind how the feature works.

Firstly, the user currently has two methods of specifying TTL.  (explicit vs implicit)
The implicit method is to specify `ttl_duration=NUM_SECONDS` inside the table comment.
```
CREATE TABLE t1 (
a INT PRIMARY KEY (a)
) ENGINE=rocksdb
COMMENT='ttl_duration=1000;';
```
This causes MyRocks to generate an implicit 8-byte timestamp (in unixtime seconds), which shall be placed at the start of each record.  (this happens in `convert_record_to_storage_format`).  During compaction, the compaction filter will parse the timestamp in the front of each record and compare it to the current time (again in unixtime seconds).  If the difference in time is longer than the duration specified, the record shall be 'filtered' out.

The explicit method is to specify a TTL column as well in the table comment, using `ttl_col=COLUMN_NAME;`.  If this is specified, MyRocks will search for a column within the table that matches `COLUMN_NAME` and attempt to use the value inside as the TTL timestamp.  It currently does this by copying the value to the front of the record inside `convert_record_to_storage_format`.  (so it uses extra space.  We are exploring a few ideas to mitigate this which may be implemented in the near future.)  Note that the TTL column currently must be an 8-byte unsigned integer type e.g. BIGINT, LONGLONG, etc.  It is still compared against the current unix time in seconds.
```
CREATE TABLE t1 (
a INT PRIMARY KEY (a), 
ts BIGINT
) ENGINE=rocksdb
COMMENT='ttl_duration=1000;ttl_col=ts;';
```

Note that TTL based on partitioned tables is also supported.
```
...
) ENGINE=ROCKSDB
PARTITION BY LIST(c1) (
    PARTITION p0 VALUES IN (1, 4, 7),
    PARTITION p1 VALUES IN (2, 5, 8),
    PARTITION p2 VALUES IN (3, 6, 9)
);
COMMENT='p0_ttl_duration=42;p0_ttl_col=name;p2_ttl_duration=5000;p2_ttl_col=age'
```

Miscellaneous points:
When TTL records need to be decoded, we determine that the record is from a TTL table inside `convert_record_from_storage_format` and skip the first 8 bytes.

As it is currently implemented, users of this feature must understand that expired data can disappear from out under them, even while they have a snapshot open, as the compaction filter has `IgnoreSnapshots()` on.

There is no support for secondary keys yet.


alxyang:
there are still a few more tests I want to add, but putting this up for initial review...

PR #572, opened 2017-03-16
Added missing have_rocksdb.inc

george-lorch:


mysql-test/suite/rocksdb_sys_vars/t/all_vars.test
```diff
@@ -1,3 +1,4 @@
+--source include/have_rocksdb.inc
```
laurynas-biveinis, line 1: FWIW this testcase does not actually need MyRocks

george-lorch: I was thinking the same thing when I added the test but really it seemed correct to not test existence of rocks tests if rocks is not an actual engine and just skip.

Issue #565, opened 2017-03-13
Need to add/update bloomfilter test cases


yoshinorim:
bloomfilter.test does not have enough test cases with reverse order column family. Patching https://gist.github.com/yoshinorim/ea93242f6cf771017863344fe36fd5e4 will give better coverage. The additional test case revealed the followings.

- Prior to 2eed684599a131cd06a26a48282d049d6ef37fe1, MyRocks had a bug that it returned wrong result set with reverse CF. The bug happened if bloom filter length was 4 (which is very uncommon in practice though) and entering myrocks::ha_rocksdb::index_first -> myrocks::ha_rocksdb::index_last_intern code path. One example was "select count(id1) from r0" and r0 had rev cf. This was because Prev() was called after calling Seek() with bloom filter enabled.
- After 2eed684599a131cd06a26a48282d049d6ef37fe1, the above bug no longer happened because relevant code was changed. But bloomfilter.result was not updated. This ended up with bloomfilter.test failures, because when bloom filter was enabled had changed with the diff.

alxyang:
closed by https://github.com/facebook/mysql-5.6/commit/639b92350f25de340c3d859db267e151b49f4512

PR #573, opened 2017-03-16
Adding more bloomfilter tests (#565)

yoshinorim:
Summary: This diff adds more test coverages for bloomfilter
tests with reverse order column family

Test Plan: mtr


alxyang (changes requested):
I think rocksdb.bloomfilter_skip must be re-recorded

alxyang (approved):
lgtm

Issue #575, opened 2017-03-17
missing sysvars rocksdb_enable_2pc test


george-lorch:
From all_vars.result:
> There should be *no* long test name listed below:
> select variable_name as `There should be *no* variables listed below:` from t2
> left join t1 on variable_name=test_name where test_name is null ORDER BY variable_name;
> There should be *no* variables listed below:
> rocksdb_enable_2pc
> rocksdb_enable_2pc

PR #576, opened 2017-03-17
re-record enable_2pc test to fix rocksdb_sys_vars.all_vars

alxyang:
Fixes https://github.com/facebook/mysql-5.6/issues/575


reidHoruff (approved):

reidHoruff:
How did this happen?

alxyang:
what I think happened is at one point we renamed `disable_2pc` to `enable_2pc`.
But at that time, we didn't update the corresponding test names in the rocksdb_sys_vars suite.

Which then means `rocksdb_sys_vars.all_vars` is going to have additional output, and then it was mistakenly recorded over.

PR #580, opened 2017-03-24
test: slocket_listen.py /bin/env -> /usr/bin/env

grooverdan:
Ubuntu-16.04 doesn't have /bin/env, only /usr/bin/env. Centos7.3 has
both installed (and not hard/soft linked). So change to the compatible
instance.


alxyang (approved):

grooverdan:
Thanks @alxyang

PR #598, opened 2017-04-08
gcc-7 -Wint-in-bool-contex fix for rdb_datadic.cc

grooverdan:
The warning (observed in MariaDB rocksDB at an identical line was:)
warning: '*' in boolean context, suggest '&&' instead [-Wint-in-bool-context]
     if (real_val_size % Rdb_key_def::PACKED_SIZE * 2) {
         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~


grooverdan:
thanks @gunnarku

PR #599, opened 2017-04-10
Bigendian - compilers cannot unwrap the static_assert

### LINKED ISSUE #597: storage/rocksdb/rdb_datadic.cc:1585
From [MariaDB -MDEV-12469](https://jira.mariadb.org/browse/MDEV-12469?focusedCommentId=93902&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-93902) where the code is the same (except on line 1306).

gcc-7 on big endian isn't smart enough to realise the swap_func is always nullptr

grooverdan:
Replace this with a DBUG_ASSERT. Closes #597


grooverdan:
Thanks @gunnarku

Issue #602, opened 2017-04-10
alter table drop key cause mysqld crash


chuckcai:
**Description:**
./bin/mysqld(my_print_stacktrace+0x47)[0x13c145b]
./bin/mysqld(handle_fatal_signal+0x727)[0xcfac4f]
/lib64/libpthread.so.0(+0xf500)[0x7f95efd0a500]
/lib64/libc.so.6(gsignal+0x35)[0x310b0328a5]
/lib64/libc.so.6(abort+0x175)[0x310b034085]
./bin/mysqld(_ZN7myrocks16Rdb_dict_manager20log_start_drop_indexENS_14_gl_index_id_sEPKc+0xe9)[0x14931c7]
./bin/mysqld(_ZN7myrocks16Rdb_dict_manager14add_drop_indexERKSt13unordered_setINS_14_gl_index_id_sESt4hashIS2_ESt8equal_toIS2_ESaIS2_EEPN7rocksdb10WriteBatchE+0xc3)[0x149242f]
./bin/mysqld(_ZN7myrocks10ha_rocksdb26commit_inplace_alter_tableEP5TABLEP18Alter_inplace_infob+0x68a)[0x14148c0]
./bin/mysqld(_ZN7handler29ha_commit_inplace_alter_tableEP5TABLEP18Alter_inplace_infob+0x115)[0xb058a7]
./bin/mysqld[0xf42087]
./bin/mysqld(_Z17mysql_alter_tableP3THDPcS1_P24st_ha_create_informationP10TABLE_LISTP10Alter_infojP8st_orderb+0x2e53)[0xf4a766]
./bin/mysqld(_ZN19Sql_cmd_alter_table7executeEP3THD+0x8e3)[0x119dbd3]
./bin/mysqld(_Z21mysql_execute_commandP3THD+0x10338)[0xe70511]
./bin/mysqld(_Z11mysql_parseP3THDPcjP12Parser_state+0x99d)[0xe77c7d]
./bin/mysqld(_Z16dispatch_command19enum_server_commandP3THDPcj+0x1aa3)[0xe5abd7]
./bin/mysqld(_Z10do_commandP3THD+0x773)[0xe57856]
./bin/mysqld(_Z24do_handle_one_connectionP3THD+0x4e8)[0xde8f56]
./bin/mysqld(handle_one_connection+0x57)[0xde852a]
/lib64/libpthread.so.0(+0x7851)[0x7f95efd02851]
[...]

PR #603, opened 2017-04-10
fix cases where optimizing out dropping/creating the same index in on…

alxyang:
…e DDL statement improperly removed the index from data dictionary

Fixes https://github.com/facebook/mysql-5.6/issues/602.


alxyang:
actually, this isn't ready for review yet. there are some corner cases I need to cover, and need to add tests.

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -5422,9 +5423,32 @@ std::unordered_map<std::string, uint> ha_rocksdb::get_old_key_positions(
...
-    if (!compare_key_parts(old_key, new_key)) {
-      old_key_pos[old_key->name] = i;
+    /* If index algorithms are different we need to rebuild. */
+    if (old_key->algorithm != new_key->algorithm) {
+      continue;
+    }
+
+    /*
+      Check that the key is identical between old and new tables.
+      If so, we still need to create a new index.
```
hermanlee: Should say "If not, we still need to create a new index."

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -5422,9 +5423,32 @@ std::unordered_map<std::string, uint> ha_rocksdb::get_old_key_positions(
...
+
+      The exception is if there is an index changed from unique to non-unique,
+      in these cases we don't need to rebuild as they are stored the same way in
+      RocksDB.
+    */
+    bool unique_to_non_unique =
+        (old_key->flags ^ new_key->flags) == HA_NOSAME &&
+        (old_key->flags & HA_NOSAME);
+    if (!unique_to_non_unique && ((old_key->flags & HA_KEYFLAG_MASK) !=
```
hermanlee: If the mask is needed, then the mask should apply to the old_key->flags & new_key->flags calculation. Logically, your code works too though.

hermanlee (commented):
What does InnoDB do in the case where the same key is dropped and added? Does it skip over it, or does it drop the index and re-create it, even though it is the same?

alxyang:
@hermanlee if an index is exactly the same (meaning all the key flags match) and the drop key + add key is combined in one alter statement, then it's actually optimized out at the SQL layer.  So yes, it skips over it (does not recreate).

PR #606, opened 2017-04-11
add global variable to disable ttl and add global counters for TTL ro…

alxyang:
…ws expired

### DIFF
--- a/storage/rocksdb/ha_rocksdb.cc
+++ b/storage/rocksdb/ha_rocksdb.cc
@@ -349,6 +349,18 @@ static void rocksdb_set_pause_background_work(
   RDB_MUTEX_UNLOCK_CHECK(rdb_sysvars_mutex);
 }
 
+static void rocksdb_set_enable_ttl(
+    my_core::THD *const thd,
+    my_core::st_mysql_sys_var *const var MY_ATTRIBUTE((__unused__)),
+    void *const var_ptr, const void *const save) {
+  DBUG_ASSERT(save != nullptr);
+  RDB_MUTEX_LOCK_CHECK(rdb_sysvars_mutex);
+
+  *static_cast<bool *>(var_ptr) = *static_cast<const bool *>(save);
+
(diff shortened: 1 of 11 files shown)


storage/rocksdb/ha_rocksdb.cc
```diff
@@ -1098,6 +1111,11 @@ static MYSQL_SYSVAR_BOOL(pause_background_work, rocksdb_pause_background_work,
                          "Disable all rocksdb background operations", nullptr,
                          rocksdb_set_pause_background_work, FALSE);
 
+static MYSQL_SYSVAR_BOOL(
+    disable_ttl, rocksdb_disable_ttl, PLUGIN_VAR_RQCMDARG,
```
hermanlee: Make this enable_ttl and set the default to TRUE.

storage/rocksdb/rdb_compact_filter.h
```diff
@@ -76,6 +76,7 @@ class Rdb_compact_filter : public rocksdb::CompactionFilter {
     } else if (m_ttl_duration > 0 &&
                should_filter_ttl_rec(key, existing_value)) {
       m_num_expired++;
+      global_stats->rows[ROWS_EXPIRED].inc();
```
hermanlee: It might be more efficient to upload these stats during deconstruction of the Rdb_compact_filter object to the global stats. Maybe add an api to the ha_rocksdb_proto.h that lets you update a specific counter?

hermanlee (changes requested):

storage/rocksdb/rdb_compact_filter.h
alxyang: @hermanlee  can you elaborate what you mean by 

> add an api to ha_rocksdb_proto.h that lets you update a specific counter?

By exposing global_stats you can already update whatever operation type.  
I have moved the stats upload to the destructor.

alxyang:
addressed comments

PR #615, opened 2017-04-25
Aborting myrocks_hotbackup if *.frm is changed

yoshinorim:
Summary: If table definition (frm) has changed during myrocks_hotbackup,
destination instance may get inconsistent -- having new frm file,
but MyRocks sst files are based on snapshots, and their data might be
older -- based on old frm file. To prevent such inconsistencies,
myrocks_hotbackup checks frm file midification timestamp, and
aborts if it is newer than backup starting time.

Test Plan: xbstream.test

### DIFF
--- a/mysql-test/suite/rocksdb_hotbackup/include/stream_run.sh
+++ b/mysql-test/suite/rocksdb_hotbackup/include/stream_run.sh
@@ -1,3 +1,5 @@
+#!/bin/bash
+
 if [ "$STREAM_TYPE" == 'wdt' ]; then
   which wdt >/dev/null 2>&1
   if [ $? -ne 0 ]; then
@@ -7,6 +9,7 @@ if [ "$STREAM_TYPE" == 'wdt' ]; then
 fi
 
 set -e
+set -o pipefail
 
 # Takes a full backup from server_1 to server_2
 # using myrocks_hotbackup streaming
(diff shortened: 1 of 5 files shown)


scripts/myrocks_hotbackup
```diff
@@ -77,6 +81,15 @@ class MiscFilesProcessor():
     # do nothing
     pass
 
+  def check_frm_timestamp(self, fname, path):
+    if not self.skip_check_frm_timestamp and fname.endswith('.frm'):
+      if os.path.getmtime(path) > self.start_backup_time:
+        logger.error('FRM file %s was updated after starting backups! '
```
hermanlee: Can you explain further why this is bad: "FRM file %s was updated after starting backups, schema could have changed and the resulting copy may not be valid. Aborting!"

hermanlee (commented):
While the sleep works in the unit test, it might occasionally fail depending on scheduling during the test harness run. Would it be possible to change this to a signal model? Perhaps the hotbackup tool can take a parameter to wait/poll for a file to be created instead of sleeping some number of seconds, and you can create this file after issuing the "create table call", or use the .frm file as the file to poll for.

scripts/myrocks_hotbackup
```diff
@@ -329,17 +343,19 @@ class MySQLUtil:
 
 class BackupRunner:
   datadir = None
+  start_timestamp = None
```
alxyang: I think this can be removed, isn't being used anywhere

alxyang (commented):
mostly LGTM

mysql-test/suite/rocksdb_hotbackup/include/stream_run.sh
```diff
@@ -29,36 +32,42 @@ rm -rf $dest_data_dir/
 mkdir $dest_data_dir
 
 COPY_LOG="${MYSQL_TMP_DIR}/myrocks_hotbackup_copy_log"
+SIGNAL_CONDITION=""
+SIGNAL_FILE=${MYSQL_TMP_DIR}/myrocks_hotbackup_signal
```
hermanlee, line 36: Probably just need to add a line to ensure the COPY_LOG and SIGNAL_FILE don't exist before the test starts. If there was anything left over from an old run, it could still result in a timing issue.

hermanlee (approved):

PR #617, opened 2017-05-03
implement query read filtering for tables with TTL enabled (primary k…

alxyang:
…ey only)

This diff adds support for query read filtering for TTL records.  This is needed for a few reasons.  The first is that the compaction filter does not respect snapshots when filtering out records.  The second is that records stay around even after they have expired, until compaction happens.

An example describing the problems with this scenario is shown below:

```
CREATE TABLE t1 (a INT PRIMARY KEY) Engine=ROCKSDB COMMENT='ttl_duration=100';
# some records are inserted here...

BEGIN; # opens snapshot
# at this time, some records expire
select * from t1; # <= returns all records
# at this time compaction runs, and removes records
select * from t1; # <= some records have been removed, even from within the same snapshot!
```

So as you can see this breaks repeatable read behavior, which can give the user unexpected results.  Also, in the future when we support TTL with secondary keys, we need to avoid race conditions where compaction has removed the primary key, but not the secondary key.  This changeset lays the foundation for that as well.

In MyRocks, during index reads or table scans, there is now logic to check if the table has TTL enabled.  If so, it will see if it needs to filter out the record based on the current transaction start time.  This means that anything that SHOULD be expired from the POV of the transaction when it starts will be filtered out of the results. 

In the compaction filter, we check the oldest snapshot time from RocksDB.  Nothing after the oldest snapshot time is dropped.  This prevents keys that expire during long-running transactions to disappear from out under you.  

The combination of the above two fixes should lead to consistent behaviour from the POV of the server.

### DIFF
--- a/mysql-test/suite/rocksdb/t/ttl_primary.test
+++ b/mysql-test/suite/rocksdb/t/ttl_primary.test
@@ -1,13 +1,5 @@
+--source include/have_debug.inc
 --source include/have_rocksdb.inc
-
---disable_warnings
-DROP TABLE IF EXISTS t1;
---enable_warnings
-
-SET @@global.rocksdb_update_cf_options = 'default={disable_auto_compactions=true;};';
-
-# Disable read filtering for the duration of this test.
-set global rocksdb_enable_ttl_read_filtering=0;
 
 # Basic TTL test
 CREATE TABLE t1 (
@@ -20,11 +12,11 @@ PRIMARY KEY (`b`,`a`,`c`)
 ) ENGINE=ROCKSDB DEFAULT CHARSET=latin1
 COMMENT='ttl_duration=1;ttl_col=ts;';
 
+set global rocksdb_debug_ttl_rec_ts = -100;
 INSERT INTO t1 values ('a', 'b', 'c', UNIX_TIMESTAMP(), 'd');
 INSERT INTO t1 values ('d', 'e', 'f', UNIX_TIMESTAMP(), 'g');
+set global rocksdb_debug_ttl_rec_ts = 0;
 SELECT COUNT(*) FROM t1;
-
---real_sleep 3
 set global rocksdb_force_flush_memtable_now=1;
 set global rocksdb_compact_cf='default';
 
@@ -42,11 +34,12 @@ CREATE TABLE t1 (
 ) ENGINE=rocksdb
 COMMENT='ttl_duration=1;ttl_col=ts;';
 
+set global rocksdb_debug_ttl_rec_ts = -100;
 INSERT INTO t1 values (1, 3, UNIX_TIMESTAMP(), 5);
 INSERT INTO t1 values (2, 4, UNIX_TIMESTAMP(), 6);
+set global rocksdb_debug_ttl_rec_ts = 0;
 SELECT COUNT(*) FROM t1;
 
---real_sleep 3
 set global rocksdb_force_flush_memtable_now=1;
 set global rocksdb_compact_cf='default';
 
@@ -64,11 +57,12 @@ CREATE TABLE t1 (
 ) ENGINE=rocksdb
 COMMENT='ttl_duration=1;ttl_col=ts;';
 
+set global rocksdb_debug_ttl_rec_ts = -100;
 INSERT INTO t1 values (1, 3, 5, UNIX_TIMESTAMP());
 INSERT INTO t1 values (2, 4, 6, UNIX_TIMESTAMP());
+set global rocksdb_debug_ttl_rec_ts = 0;
 SELECT COUNT(*) FROM t1;
 
---real_sleep 3
 set global rocksdb_force_flush_memtable_now=1;
 set global rocksdb_compact_cf='default';
 
@@ -86,11 +80,12 @@ CREATE TABLE t1 (
 ) ENGINE=rocksdb
 COMMENT='ttl_duration=1;ttl_col=ts;';
 
+set global rocksdb_debug_ttl_rec_ts = -100;
 INSERT INTO t1 values (1, 3, 5, UNIX_TIMESTAMP());
 INSERT INTO t1 values (2, 4, 6, UNIX_TIMESTAMP());
+set global rocksdb_debug_ttl_rec_ts = 0;
 SELECT COUNT(*) FROM t1;
 
---real_sleep 3
 set global rocksdb_force_flush_memtable_now=1;
 set global rocksdb_compact_cf='default';
 
@@ -108,11 +103,12 @@ CREATE TABLE t1 (
 ) ENGINE=rocksdb
 COMMENT='ttl_duration=1;ttl_col=ts;';
 
+set global rocksdb_debug_ttl_rec_ts = -100;
 INSERT INTO t1 values (1, NULL, NULL, UNIX_TIMESTAMP());
 INSERT INTO t1 values (2, NULL, NULL, UNIX_TIMESTAMP());
+set global rocksdb_debug_ttl_rec_ts = 0;
 SELECT COUNT(*) FROM t1;
 
---real_sleep 3
 set global rocksdb_force_flush_memtable_now=1;
 set global rocksdb_compact_cf='default';
 
@@ -131,11 +127,12 @@ PRIMARY KEY (`a`)
 ) ENGINE=ROCKSDB DEFAULT CHARSET=latin1
 COMMENT='ttl_duration=1;ttl_col=ts;';
 
+set global rocksdb_debug_ttl_rec_ts = -100;
 INSERT INTO t1 values ('a', NULL, 'bc', UNIX_TIMESTAMP(), 'd');
 INSERT INTO t1 values ('d', 'efghijk', NULL, UNIX_TIMESTAMP(), 'l');
+set global rocksdb_debug_ttl_rec_ts = 0;
 SELECT COUNT(*) FROM t1;
 
---real_sleep 3
 set global rocksdb_force_flush_memtable_now=1;
 set global rocksdb_compact_cf='default';
 
@@ -152,11 +149,12 @@ CREATE TABLE t1 (
 ) ENGINE=rocksdb
 COMMENT='ttl_duration=1;';
 
+set global rocksdb_debug_ttl_rec_ts = -100;
 INSERT INTO t1 values (1, 3, 5);
 INSERT INTO t1 values (2, 4, 6);
+set global rocksdb_debug_ttl_rec_ts = 0;
 SELECT COUNT(*) FROM t1;
 
---real_sleep 3
 set global rocksdb_force_flush_memtable_now=1;
 set global rocksdb_compact_cf='default';
 
@@ -166,23 +164,26 @@ DROP TABLE t1;
 
 # TTL field as the PK
 CREATE TABLE t1 (
+	a int,
   ts bigint(20) UNSIGNED NOT NULL,
-  PRIMARY KEY (ts)
+  PRIMARY KEY (a, ts)
 ) ENGINE=rocksdb
 COMMENT='ttl_duration=5;ttl_col=ts;';
 
-INSERT INTO t1 values (UNIX_TIMESTAMP());
---real_sleep 1
-INSERT INTO t1 values (UNIX_TIMESTAMP());
+INSERT INTO t1 values (1, UNIX_TIMESTAMP());
+INSERT INTO t1 values (2, UNIX_TIMESTAMP());
 SELECT COUNT(*) FROM t1;
 
+set global rocksdb_debug_ttl_snapshot_ts = -10;
 set global rocksdb_force_flush_memtable_now=1;
 set global rocksdb_compact_cf='default';
+set global rocksdb_debug_ttl_snapshot_ts = 0;
 # should all still be there..
 SELECT COUNT(*) FROM t1;
 
---real_sleep 5
+set global rocksdb_debug_ttl_snapshot_ts = 10;
 set global rocksdb_compact_cf='default';
+set global rocksdb_debug_ttl_snapshot_ts = 0;
 # should have filtered the rows out since ttl is passed in compaction filter
 SELECT COUNT(*) FROM t1;
 DROP TABLE t1;
@@ -198,11 +199,12 @@ CREATE TABLE t1 (
 ) ENGINE=rocksdb
 COMMENT='ttl_duration=1;ttl_col=ts;';
 
+set global rocksdb_debug_ttl_rec_ts = -100;
 INSERT INTO t1 values (1, 3, UNIX_TIMESTAMP(), 5);
 INSERT INTO t1 values (2, 4, UNIX_TIMESTAMP(), 6);
+set global rocksdb_debug_ttl_rec_ts = 0;
 SELECT COUNT(*) FROM t1;
 
---real_sleep 3
 set global rocksdb_force_flush_memtable_now=1;
 set global rocksdb_compact_cf='default';
 
@@ -221,11 +223,12 @@ PRIMARY KEY (`a`, `ts`)
 ) ENGINE=ROCKSDB DEFAULT CHARSET=latin1
 COMMENT='ttl_duration=1;ttl_col=ts;';
 
+set global rocksdb_debug_ttl_rec_ts = -100;
 INSERT INTO t1 values ('a', NULL, 'bc', UNIX_TIMESTAMP(), 'd');
 INSERT INTO t1 values ('de', 'fghijk', NULL, UNIX_TIMESTAMP(), 'l');
+set global rocksdb_debug_ttl_rec_ts = 0;
 SELECT COUNT(*) FROM t1;
 
---real_sleep 3
 set global rocksdb_force_flush_memtable_now=1;
 set global rocksdb_compact_cf='default';
 
@@ -245,26 +248,31 @@ PRIMARY KEY (b,a,c)
 ) ENGINE=ROCKSDB DEFAULT CHARSET=latin1
 COMMENT='ttl_duration=10;ttl_col=ts;';
 
+set global rocksdb_debug_ttl_rec_ts = -300;
 INSERT INTO t1 values (1, 'b', 'c', UNIX_TIMESTAMP(), 'd');
 INSERT INTO t1 values (2, 'e', 'f', UNIX_TIMESTAMP(), 'g');
-
---real_sleep 5
-
+set global rocksdb_debug_ttl_rec_ts = 300;
 INSERT INTO t1 values (3, 'i', 'j', UNIX_TIMESTAMP(), 'k');
 INSERT INTO t1 values (4, 'm', 'n', UNIX_TIMESTAMP(), 'o');
+set global rocksdb_debug_ttl_rec_ts = 0;
 
+# Nothing should get removed here.
+set global rocksdb_debug_ttl_snapshot_ts = -3600;
 set global rocksdb_force_flush_memtable_now=1;
 set global rocksdb_compact_cf='default';
+set global rocksdb_debug_ttl_snapshot_ts = 0;
 --sorted_result
 SELECT a FROM t1;
 
---real_sleep 6
+# 1 and 2 should get removed here.
 set global rocksdb_compact_cf='default';
 --sorted_result
 SELECT a FROM t1;
 
---real_sleep 5
+# 3 and 4 should get removed here.
+set global rocksdb_debug_ttl_snapshot_ts = 3600;
 set global rocksdb_compact_cf='default';
+set global rocksdb_debug_ttl_snapshot_ts = 0;
 --sorted_result
 SELECT a FROM t1;
 
@@ -331,8 +339,6 @@ COMMENT='ttl_duration=500;';
 
 INSERT INTO t1 values (1);
 SELECT COUNT(*) FROM t1;
-
---real_sleep 1
 set global rocksdb_force_flush_memtable_now=1;
 set global rocksdb_compact_cf='default';
 
@@ -349,10 +355,10 @@ COMMENT='ttl_duration=100;';
 INSERT INTO t1 values (1);
 SELECT * FROM t1;
 
+set global rocksdb_debug_ttl_rec_ts = -300;
 ALTER TABLE t1 COMMENT = 'ttl_duration=1';
+set global rocksdb_debug_ttl_rec_ts = 0;
 SHOW CREATE TABLE t1;
-
---real_sleep 3
 set global rocksdb_force_flush_memtable_now=1;
 set global rocksdb_compact_cf='default';
 
@@ -384,14 +390,18 @@ INSERT INTO t1 VALUES (1,1);
 INSERT INTO t1 VALUES (2,2);
 
 ALTER TABLE t1 DROP PRIMARY KEY, ADD PRIMARY KEY(b);
+set global rocksdb_debug_ttl_snapshot_ts = -3600;
 set global rocksdb_force_flush_memtable_now=1;
---real_sleep 1
 set @@global.rocksdb_compact_cf = 'default';
+set global rocksdb_debug_ttl_snapshot_ts = 0;
+
 --sorted_result
 SELECT COUNT(*) FROM t1;
 
---real_sleep 5
+set global rocksdb_debug_ttl_snapshot_ts = 3600;
 set @@global.rocksdb_compact_cf = 'default';
+set global rocksdb_debug_ttl_snapshot_ts = 0;
+
 --sorted_result
 SELECT COUNT(*) FROM t1;
 
@@ -401,27 +411,35 @@ DROP TABLE t1;
 # (basically, it needs semicolon before and after)
 CREATE TABLE t1 (
   a bigint(20) UNSIGNED NOT NULL,
-  PRIMARY KEY (a)
+  b int,
+  PRIMARY KEY (a,b)
 ) ENGINE=rocksdb
 COMMENT='asdadfasdfsadfadf ;ttl_duration=1; asfasdfasdfadfa';
-INSERT INTO t1 values (UNIX_TIMESTAMP());
+INSERT INTO t1 values (UNIX_TIMESTAMP(), 1);
 SELECT COUNT(*) FROM t1;
 
---real_sleep 3
+set global rocksdb_debug_ttl_snapshot_ts = 3600;
 set global rocksdb_force_flush_memtable_now=1;
 set global rocksdb_compact_cf='default';
+set global rocksdb_debug_ttl_snapshot_ts = 0;
 
 SELECT COUNT(*) FROM t1;
 
 ALTER TABLE t1 COMMENT = 'adsf;;ttl_duration=5;asfasdfa;ttl_col=a;asdfasdf;';
-INSERT INTO t1 values (UNIX_TIMESTAMP());
+
+set global rocksdb_debug_ttl_rec_ts = 300;
+INSERT INTO t1 values (UNIX_TIMESTAMP(), 2);
+set global rocksdb_debug_ttl_rec_ts = 0;
 set global rocksdb_force_flush_memtable_now=1;
 
---real_sleep 1
+# nothing removed here
 set global rocksdb_compact_cf='default';
 SELECT COUNT(*) FROM t1;
---real_sleep 5
+
+# all removed here
+set global rocksdb_debug_ttl_snapshot_ts = 3600;
 set global rocksdb_compact_cf='default';
+set global rocksdb_debug_ttl_snapshot_ts = 0;
 SELECT COUNT(*) FROM t1;
 
 DROP TABLE t1;
@@ -433,12 +451,14 @@ CREATE TABLE t1 (
 ) ENGINE=rocksdb
 COMMENT='ttl_duration=5;';
 
+set global rocksdb_debug_ttl_rec_ts = -300;
 INSERT INTO t1 values (1);
 INSERT INTO t1 values (3);
 INSERT INTO t1 values (5);
---real_sleep 6
+set global rocksdb_debug_ttl_rec_ts = 300;
 INSERT INTO t1 values (7);
 INSERT INTO t1 values (9);
+set global rocksdb_debug_ttl_rec_ts = 0;
 
 UPDATE t1 SET a=a+1;
 --sorted_result
@@ -447,6 +467,7 @@ SELECT * FROM t1;
 set global rocksdb_force_flush_memtable_now=1;
 set global rocksdb_compact_cf='default';
 
+# 1,3,5 should be dropped
 --sorted_result
 SELECT * FROM t1;
 DROP TABLE t1;
@@ -459,19 +480,23 @@ CREATE TABLE t1 (
 ) ENGINE=rocksdb
 COMMENT='ttl_duration=5;ttl_col=b;';
 
+set global rocksdb_debug_ttl_rec_ts = -300;
 INSERT INTO t1 values (1, UNIX_TIMESTAMP());
 INSERT INTO t1 values (3, UNIX_TIMESTAMP());
 INSERT INTO t1 values (5, UNIX_TIMESTAMP());
 INSERT INTO t1 values (7, UNIX_TIMESTAMP());
---real_sleep 5
 
+set global rocksdb_debug_ttl_rec_ts = 300;
 UPDATE t1 SET b=UNIX_TIMESTAMP() WHERE a < 4;
+set global rocksdb_debug_ttl_rec_ts = 0;
+
 --sorted_result
 SELECT a FROM t1;
 
 set global rocksdb_force_flush_memtable_now=1;
 set global rocksdb_compact_cf='default';
 
+# 5 and 7 should be gone here
 --sorted_result
 SELECT a FROM t1;
 DROP TABLE t1;
@@ -483,11 +508,12 @@ CREATE TABLE t1 (
 ) ENGINE=rocksdb
 COMMENT='ttl_duration=1;';
 
+set global rocksdb_debug_ttl_rec_ts = -100;
 INSERT INTO t1 values (1);
 INSERT INTO t1 values (2);
 INSERT INTO t1 values (3);
+set global rocksdb_debug_ttl_rec_ts = 0;
 
---real_sleep 3
 set global rocksdb_enable_ttl=0;
 set global rocksdb_force_flush_memtable_now=1;
 set global rocksdb_compact_cf='default';
@@ -517,10 +543,5 @@ select variable_value into @c from information_schema.global_status where variab
 set global rocksdb_force_flush_memtable_now=1;
 set global rocksdb_compact_cf='default';
 select variable_value-@c from information_schema.global_status where variable_name='rocksdb_rows_expired';
+
 DROP TABLE t1;
-
-SET @@global.rocksdb_update_cf_options = 'default={disable_auto_compactions=false;};';
-SET @@global.rocksdb_update_cf_options = '';
-set global rocksdb_enable_ttl_read_filtering=1;
-
-
--- a/mysql-test/suite/rocksdb/t/ttl_primary.test
+++ b/mysql-test/suite/rocksdb/t/ttl_primary.test
@@ -6,6 +6,9 @@ DROP TABLE IF EXISTS t1;
 
 SET @@global.rocksdb_update_cf_options = 'default={disable_auto_compactions=true;};';
 
+# Disable read filtering for the duration of this test.
+set global rocksdb_enable_ttl_read_filtering=0;
+
 # Basic TTL test
 CREATE TABLE t1 (
 `a` binary(8) NOT NULL,
@@ -518,6 +521,6 @@ DROP TABLE t1;
 
 SET @@global.rocksdb_update_cf_options = 'default={disable_auto_compactions=false;};';
 SET @@global.rocksdb_update_cf_options = '';
-
+set global rocksdb_enable_ttl_read_filtering=1;
 
 
--- a/mysql-test/suite/rocksdb/t/ttl_primary_read_filtering.test
+++ b/mysql-test/suite/rocksdb/t/ttl_primary_read_filtering.test
@@ -1,0 +1,245 @@
+--source include/have_rocksdb.inc
+--source include/have_debug.inc
+
+--disable_warnings
+DROP TABLE IF EXISTS t1;
+--enable_warnings
+
+SET @@global.rocksdb_update_cf_options = 'default={disable_auto_compactions=true;};';
+
+# The purpose of read filtering for tables with TTL is to ensure that during a
+# transaction a key which has expired already but not removed by compaction
+# yet, is not returned to the user.
+#
+# Without this the user might be hit with problems such as disappearing rows
+# within a transaction, etc, because the compaction filter ignores snapshots
+# when filtering keys.
+
+# Basic read filtering test
+CREATE TABLE t1 (
+  a int PRIMARY KEY
+) ENGINE=rocksdb
+COMMENT='ttl_duration=1;';
+
+INSERT INTO t1 values (1);
+INSERT INTO t1 values (2);
+set global rocksdb_force_flush_memtable_now=1;
+
+--sleep 1
+--sorted_result
+SELECT * FROM t1;
+
+select variable_value-@c from information_schema.global_status where variable_name='rocksdb_rows_expired';
+set global rocksdb_compact_cf='default';
+select variable_value-@c from information_schema.global_status where variable_name='rocksdb_rows_expired';
+
+DROP TABLE t1;
+
+# Test that some rows are hidden but others aren't...
+CREATE TABLE t1 (
+  a int PRIMARY KEY,
+  b BIGINT UNSIGNED NOT NULL
+) ENGINE=rocksdb
+COMMENT='ttl_duration=5;';
+
+INSERT INTO t1 values (1, UNIX_TIMESTAMP());
+--sleep 5
+INSERT INTO t1 values (2, UNIX_TIMESTAMP());
+INSERT INTO t1 values (3, UNIX_TIMESTAMP());
+
+set global rocksdb_force_flush_memtable_now=1;
+
+# 1 should be hidden even though compaction hasn't run.
+--sorted_result
+SELECT a FROM t1;
+
+set global rocksdb_compact_cf='default';
+
+# none should be hidden yet, compaction runs but records arent expired
+--sorted_result
+SELECT a FROM t1;
+
+# all should be hidden now, even though compaction hasnt run again
+--sleep 5
+--sorted_result
+SELECT a FROM t1;
+
+DROP TABLE t1;
+
+# Read filtering index scan tests (None of these queries should return any results)
+CREATE TABLE t1 (
+  a int,
+  b int,
+  c int,
+  PRIMARY KEY (a,b,c)
+) ENGINE=rocksdb
+COMMENT='ttl_duration=1;';
+
+INSERT INTO t1 values (0,0,0);
+INSERT INTO t1 values (0,0,1);
+INSERT INTO t1 values (0,1,0);
+INSERT INTO t1 values (0,1,1);
+INSERT INTO t1 values (1,1,2);
+INSERT INTO t1 values (1,2,1);
+INSERT INTO t1 values (1,2,2);
+INSERT INTO t1 values (1,2,3);
+
+select variable_value into @c from information_schema.global_status where variable_name='rocksdb_rows_expired';
+
+set global rocksdb_force_flush_memtable_now=1;
+
+--sleep 1
+# HA_READ_KEY_EXACT, using full key
+SELECT * FROM t1 WHERE a=1 AND b=2 AND c=2;
+
+# HA_READ_KEY_EXACT, not using full key
+SELECT * FROM t1 WHERE a = 1;
+
+# HA_READ_BEFORE_KEY, not using full key
+SELECT max(a) from t1 where a < 3;
+
+#HA_READ_BEFORE_KEY, using full key
+SELECT max(a) from t1 where a < 2 AND b = 1 AND c < 3;
+
+# HA_READ_KEY_OR_NEXT
+SELECT min(a) from t1 where a >= 1;
+
+# HA_READ_AFTER_KEY,              /* Find next rec. after key-record */
+SELECT min(a) from t1 where a > 1;
+
+# HA_READ_PREFIX_LAST,            /* Last key with the same prefix */
+select * from t1 where a=1 and b in (1) order by c desc;
+
+# HA_READ_PREFIX_LAST_OR_PREV,    /* Last or prev key with the same prefix */
+select max(a) from t1 where a <=10;
+
+# need to test read_range_first()
+# calls into read_range_next() and uses compare_keys() to see if its out of
+# range
+select a from t1 where a > 0 and a <= 2;
+
+select variable_value-@c from information_schema.global_status where variable_name='rocksdb_rows_expired';
+set global rocksdb_compact_cf='default';
+select variable_value-@c from information_schema.global_status where variable_name='rocksdb_rows_expired';
+DROP TABLE t1;
+
+
+# duplicate PK value attempt to be inserted when old one is expired...
+# in this case, since it's still inside rocksdb, we don't allow it to be
+# inserted.
+CREATE TABLE t1 (
+  a int PRIMARY KEY
+) ENGINE=rocksdb
+COMMENT='ttl_duration=1;';
+INSERT INTO t1 values (1);
+set global rocksdb_force_flush_memtable_now=1;
+--sleep 1
+SELECT * FROM t1;
+--error 1062
+INSERT INTO t1 values (1);
+DROP TABLE t1;
+
+# Attempt to update expired value, should filter out
+CREATE TABLE t1 (
+  a int PRIMARY KEY
+) ENGINE=rocksdb
+COMMENT='ttl_duration=1;';
+INSERT INTO t1 values (1);
+--sleep 1
+
+--sorted_result
+SELECT * FROM t1;
+
+# No error is thrown here, under the hood rnd_next_with_direction is
+# filtering out the record from being seen in the first place.
+UPDATE t1 set a = 1;
+DROP TABLE t1;
+
+##
+## More tests on update behaviour with expired keys.
+##
+
+CREATE TABLE t1 (
+  a int PRIMARY KEY
+) ENGINE=rocksdb
+COMMENT='ttl_duration=4;';
+INSERT INTO t1 values (1);
+INSERT INTO t1 values (3);
+--sleep 3
+
+INSERT INTO t1 values (5);
+INSERT INTO t1 values (7);
+--sleep 1
+
+# Expect error because expired key (1) is still around under the hood, but
+# this time rnd_next_with_direction finds other non-expired keys (5 and 7). So the
+# execution flow in the SQL layer moves onto update_write_row, where it then
+# finds the dupicate key (1).
+--error 1062
+UPDATE t1 set a = 1;
+
+# get_row_by_rowid tested here via index_read_map_impl
+UPDATE t1 set a = 999 where a = 5;
+--sorted_result
+SELECT * FROM t1;
+
+
+UPDATE t1 set a = a - 1;
+--sorted_result
+SELECT * FROM t1;
+
+DROP TABLE t1;
+
+# Ensure no rows can disappear in the middle of long-running transactions
+# Also ensure repeatable-read works as expected
+connect (con1,localhost,root,,);
+connect (con2,localhost,root,,);
+connection con1;
+
+CREATE TABLE t1 (
+  a int PRIMARY KEY
+) ENGINE=rocksdb
+COMMENT='ttl_duration=3;';
+
+INSERT INTO t1 values (1);
+INSERT INTO t1 values (3);
+INSERT INTO t1 values (5);
+INSERT INTO t1 values (7);
+
+--echo # Creating Snapshot (start transaction)
+BEGIN;
+--sorted_result
+SELECT * FROM t1;
+
+--sleep 3
+
+--sorted_result
+SELECT * FROM t1; # <= shouldn't be filtered out here
+
+--echo # Switching to connection 2
+connection con2;
+set global rocksdb_force_flush_memtable_now=1;
+set global rocksdb_compact_cf='default';
+# filtered out, because on a different connection, on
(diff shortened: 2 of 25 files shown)


storage/rocksdb/ha_rocksdb.cc
```diff
@@ -6969,10 +7103,20 @@ int ha_rocksdb::get_row_by_rowid(uchar *const buf, const char *const rowid,
 
   table->status = STATUS_NOT_FOUND;
   if (found) {
+    /* If we found the record, but it's expired, pretend we didn't find it.  */
+    if (!skip_ttl_check && m_pk_descr->has_ttl() &&
+        should_hide_ttl_rec(rocksdb::Slice(&m_retrieved_record.front(),
+                                           m_retrieved_record.size()),
+                            tx->m_snapshot_timestamp)) {
+      DBUG_RETURN(HA_ERR_KEY_NOT_FOUND);
```
hermanlee, line 7278: In the case of get_for_update(), this needs to free the lock on the row. Otherwise, other scans can potentially hit the lock and stall waiting for it to timeout.

Can you add test cases which cover this scenario? First transaction issues select via get_for_update on keys that have expired. Second transaction tries to do the same and should not block.

storage/rocksdb/rdb_compact_filter.h
```diff
@@ -65,6 +65,17 @@ class Rdb_compact_filter : public rocksdb::CompactionFilter {
 
       if (!m_should_delete) {
         get_ttl_duration(gl_index_id, &m_ttl_duration);
+
+        /*
+          For efficiency reasons, we lazily call GetIntProperty to get the
+          oldest snapshot time, each time we find a new index to be processed.
+        */
+        rocksdb::DB *const rdb = rdb_get_rocksdb_db();
```
hermanlee: This call can be skipped in the case where m_ttl_duration == 0 or m_snapshot_timestamp > 0.

mysql-test/suite/rocksdb/t/ttl_primary_read_filtering.test
```diff
@@ -0,0 +1,245 @@
...
+SELECT * FROM t1; # <= shouldn't be filtered out here
+
+COMMIT;
+
+--sorted_result # <= filtered out here because time has past.  This is correct.
+SELECT * FROM t1;
+
+DROP TABLE t1;
+
+SET @@global.rocksdb_update_cf_options = '';
```
laurynas-biveinis: The testcase does not disconnect con1 and con2?

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -1146,8 +1159,17 @@ static MYSQL_SYSVAR_BOOL(pause_background_work, rocksdb_pause_background_work,
...
+    "Enable expired TTL records to be dropped during compaction.", nullptr,
+    rocksdb_set_enable_ttl, TRUE);
+
+static MYSQL_SYSVAR_BOOL(
+    enable_ttl_read_filtering, rocksdb_enable_ttl_read_filtering,
+    PLUGIN_VAR_RQCMDARG,
+    "For tables with TTL, expired records are skipped/filtered out during "
+    "processing and in query results. Disabling this will allow these records "
+    "to be seen, but as a result rows may disappear in the middle of "
+    "transactions as they are dropped during compaction. Use with caution. ",
```
hermanlee: nit: remove space at the end of "caution. "

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -4164,6 +4196,52 @@ bool ha_rocksdb::init_with_fields() {
...
+  DBUG_ASSERT(m_pk_descr->has_ttl());
+
+  if (!rdb_is_ttl_read_filtering_enabled() || !rdb_is_ttl_enabled()) {
+    return false;
+  }
+
+  Rdb_string_reader reader(&ttl_rec_val);
+
+  uint64 ts;
+  if (reader.read_uint64(&ts)) {
```
hermanlee: So all places that read the TTL field in the value should go through a common function that, when given the function and the field, determines from the keydef what fields are present, goes to the right offset and returns it. This way, when other fields are added to the, this code doesn't need to change. Right now, this code explicitly reads the first 8 bytes and assumes it's the TTL field. But that might not be true in the future, so let's set this up so that future changes don't have to refactor this code.

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -6044,35 +6158,53 @@ int ha_rocksdb::read_after_key(const Rdb_key_def &kd,
     with HA_READ_KEY_OR_NEXT, $GT = '>='
   */
   rocksdb_smart_seek(kd.m_is_reverse_cf, m_scan_it, key_slice);
-  return m_scan_it->Valid() ? HA_EXIT_SUCCESS : HA_ERR_KEY_NOT_FOUND;
+
+  for (;;) {
```
hermanlee: Would it be simpler to do something like:

```
while (m_scan_it->Valid() && kd.has_ttl() && should_hide_ttl_rec()) {
  rocksdb_smart_next();
}

return m_scan_it->Valid() ? HA_EXIT_SUCCESS : HA_ERR_KEY_NOT_FOUND;
```

mysql-test/suite/rocksdb/t/ttl_primary_read_filtering.test
alxyang: good catch!

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -6003,37 +6082,72 @@ int ha_rocksdb::read_key_exact(const Rdb_key_def &kd,
...
   rocksdb_smart_seek(!kd.m_is_reverse_cf, m_scan_it, key_slice);
-  if (m_scan_it->Valid() && full_key_match &&
-      kd.value_matches_prefix(m_scan_it->key(), key_slice)) {
-    /* We are using full key and we've hit an exact match */
-    if (kd.m_is_reverse_cf) {
-      m_scan_it->Next();
-    } else {
-      m_scan_it->Prev();
+
+  for (;;) {
```
hermanlee: Maybe

```
while (m_scan_it->Valid()) {
  if ((full_key_match && kd.value_matches_prefix()) || (kd.has_ttl() && should_hide_ttl_rec())) {
    rocksdb_smart_next();
    continue;
  }
  return m_scan_it->Valid() ? HA_EXIT_SUCCESS : HA_ERR_KEY_NOT_FOUND;
}
return HA_ERR_KEY_NOT_FOUND;
```

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -6003,37 +6082,72 @@ int ha_rocksdb::read_key_exact(const Rdb_key_def &kd,
     lookup_tuple may be a prefix of the index.
   */
   rocksdb_smart_seek(kd.m_is_reverse_cf, iter, key_slice);
-  if (!iter->Valid() || !kd.value_matches_prefix(iter->key(), key_slice)) {
+
+  for (;;) {
```
hermanlee: I think this loop can be rewritten more efficiently. Look at the below examples...

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -6044,35 +6158,53 @@ int ha_rocksdb::read_after_key(const Rdb_key_def &kd,
...
 
-int ha_rocksdb::position_to_correct_key(const Rdb_key_def &kd,
-                                        const enum ha_rkey_function &find_flag,
-                                        const bool &full_key_match,
-                                        const uchar *const key,
-                                        const key_part_map &keypart_map,
-                                        const rocksdb::Slice &key_slice,
-                                        bool *const move_forward) {
+int ha_rocksdb::position_to_correct_key(
+    const Rdb_key_def &kd, Rdb_transaction *const tx,
```
hermanlee: Instead of passing in the Rdb_transaction, can we pass in the filter time to be explicit in how the transaction is being used?

storage/rocksdb/ha_rocksdb.cc
hermanlee: This should be a common function that figure out how to read the TTL field from the value based on fields present in the keydef. Right now, it assumes TTL is the first 8 bytes, but that might not always be the case. Consolidating this in one function makes it easy for us to add fields to be read later on.

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -4164,6 +4196,52 @@ bool ha_rocksdb::init_with_fields() {
...
+    // NO_LINT_DEBUG
+    sql_print_error("Decoding ttl from PK value failed, "
+                    "for index (%u,%u), val: %s",
+                    gl_index_id.cf_id, gl_index_id.index_id, buf.c_str());
+    DBUG_ASSERT(0);
+    return false;
+  }
+
+  /* Hide record if it has expired before the current snapshot time. */
+  uint64 time_diff = std::difftime(curr_ts ? curr_ts : std::time(nullptr), ts);
```
hermanlee: If curr_ts is 0, then the time comparison constantly changes relative to the current time. It could be possible for some rows to disappear (i.e. secondary key passes, but the next read to the primary key returns nothing because it expired). Can you somehow guarantee the there is always a non-zero time passed in that is preserved for the duration of the query?

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -6003,37 +6082,72 @@ int ha_rocksdb::read_key_exact(const Rdb_key_def &kd,
     lookup_tuple may be a prefix of the index.
   */
   rocksdb_smart_seek(kd.m_is_reverse_cf, iter, key_slice);
-  if (!iter->Valid() || !kd.value_matches_prefix(iter->key(), key_slice)) {
+
+  for (;;) {
+    if (!iter->Valid() || !kd.value_matches_prefix(iter->key(), key_slice)) {
```
hermanlee: Maybe

```
while (iter->valid()) {
  matches_prefix = kd.value_matches_prefix();
  if (matches_prefix && && kd.has_ttl() && should_hide_ttl_rec()) {
    rocksdb_smart_next();
    continue;
  }
  return matches_prefix ? HA_EXIT_SUCCESS : HA_ERR_KEY_NOT_FOUND;
}
return HA_ERR_KEY_NOT_FOUND;
```

storage/rocksdb/ha_rocksdb.cc
hermanlee: Maybe

```
while (iter->valid()) {
  matches_prefix = kd.value_matches_prefix();
  if (matches_prefix && && kd.has_ttl() && should_hide_ttl_rec()) {
    rocksdb_smart_next();
    continue;
  }
  return matches_prefix ? HA_EXIT_SUCCESS : HA_ERR_KEY_NOT_FOUND;
}
return HA_ERR_KEY_NOT_FOUND;
```

hermanlee: Maybe

```
while (iter->valid()) {
  matches_prefix = kd.value_matches_prefix();
  if (matches_prefix && && kd.has_ttl() && should_hide_ttl_rec()) {
    rocksdb_smart_next();
    continue;
  }
  return matches_prefix ? HA_EXIT_SUCCESS : HA_ERR_KEY_NOT_FOUND;
}
return HA_ERR_KEY_NOT_FOUND;
```

storage/rocksdb/ha_rocksdb.cc
hermanlee: Maybe

```
while (m_scan_it->Valid()) {
  if ((full_key_match && kd.value_matches_prefix()) || (kd.has_ttl() && should_hide_ttl_rec()) {
    rocksdb_smark_next();
    continue;
  }
  return m_scan_it->Valid() ? HA_EXIST_SUCCESS : HA_ERR_KEY_NOT_FOUND;
}
return HA_ERR_KEY_NOT_FOUND;
```

storage/rocksdb/ha_rocksdb.cc
hermanlee: Maybe

```
while (m_scan_it->Valid() && kd.has_ttl() && should_hide_ttl_rec()) {
  rocksdb_smart_next();
}
return m_scan_it->Valid() ? HA_EXIT_SUCCESS : HA_ERR_KEY_NOT_FOUND;
```

storage/rocksdb/ha_rocksdb.cc
hermanlee: Instead of passing in the transaction, can we explicitly pass in the timestamp to indicate this is used for TTL filtering?

mysql-test/suite/rocksdb/t/ttl_primary_read_filtering.test
```diff
@@ -0,0 +1,245 @@
+--source include/have_rocksdb.inc
+--source include/have_debug.inc
+
+--disable_warnings
+DROP TABLE IF EXISTS t1;
+--enable_warnings
+
+SET @@global.rocksdb_update_cf_options = 'default={disable_auto_compactions=true;};';
```
hermanlee: Store the original value in a session variable and restore it at the end. If it changes, you don't need to update this code.

mysql-test/suite/rocksdb/t/ttl_primary_read_filtering.test
```diff
@@ -0,0 +1,245 @@
...
+
+# Basic read filtering test
+CREATE TABLE t1 (
+  a int PRIMARY KEY
+) ENGINE=rocksdb
+COMMENT='ttl_duration=1;';
+
+INSERT INTO t1 values (1);
+INSERT INTO t1 values (2);
+set global rocksdb_force_flush_memtable_now=1;
```
hermanlee: There should be some extra test cases. Instead of flushing, you can leave the keys in the memtable perhaps to guarantee they are not removed through the compaction filter.

Then you can do the following: set ttl_duration to 1 second. sleep 2 seconds. select * should return nothing. Disable filtering. select * should return everything. enable filtering, select * should return nothing. This explicitly tests the filtering code and you know the rows exist.

Alternatively, you can do the following:

transaction 1, create a snapshot and select * => returns nothing.
transaction 2, insert into table, flush
transaction 1, select * => returns nothing, but the snapshot should prevent the compaction code from removing the rows, no matter what the ttl duration is.
transaction 2, select * -> sees nothing, disable filter, select * -> sees everything, enable filter, select * -> sees nothing.

mysql-test/suite/rocksdb/t/ttl_primary_read_filtering.test
```diff
@@ -0,0 +1,245 @@
...
+# range
+select a from t1 where a > 0 and a <= 2;
+
+select variable_value-@c from information_schema.global_status where variable_name='rocksdb_rows_expired';
+set global rocksdb_compact_cf='default';
+select variable_value-@c from information_schema.global_status where variable_name='rocksdb_rows_expired';
+DROP TABLE t1;
+
+
+# duplicate PK value attempt to be inserted when old one is expired...
```
hermanlee: Nice test case. This is unexpected behavior for the caller and should be documented. Read filtering gets us into a situation where the old keys cannot be deleted because they've been filtered out, but inserts won't work because they trigger a duplicate key. I think the write path should also filter on keys too.

hermanlee (changes requested):
mysql-test/suite/rocksdb/t/ttl_primary_read_filtering-master.opt should be removed.

alxyang:
awesome comments @hermanlee. will address them all.

storage/rocksdb/ha_rocksdb.cc
alxyang: This is a good point.  I think what we can do is to add another member in the `Rdb_transaction` class that stores the current time (let's call it m_ttl_filter_timestamp) if we see that `m_snapshot_timestamp` is 0.  So within the same query you will be able to reuse this time through multiple `should_hide_ttl_rec` calls.

mysql-test/suite/rocksdb/t/ttl_primary_read_filtering.test
alxyang: As discussed offline, going to investigate the ability to just overwrite the key if we determine that its a TTL key that's expired already.

storage/rocksdb/rdb_compact_filter.h
alxyang: Regarding the check to `m_snapshot_timestamp > 0`.  I was thinking to purposely check for a 'newer' snapshot between every index change, not once per compaction.  

Are you suggesting that we use the same `m_snapshot_timestamp` for the duration of the entire compaction? (I'm fine with this, just making sure that this is what you meant)

storage/rocksdb/ha_rocksdb.cc
alxyang: Nice, I took the idea and came out with this instead, which I think is even cleaner

```  
while (iter->Valid() && kd.value_matches_prefix(iter->key(), key_slice)) {
   
    if (kd.has_ttl() &&
        should_hide_ttl_rec(iter->value(), tx->m_snapshot_timestamp)) {
      rocksdb_smart_next(kd.m_is_reverse_cf, iter);
      continue;
    }

    return HA_EXIT_SUCCESS;
 }

return HA_ERR_KEY_NOT_FOUND;
```

storage/rocksdb/rdb_compact_filter.h
hermanlee: Yes, keep the same m_snapshot_timestamp for the whole compaction run. No need to update it because a slightly later time will unlikely change the outcome, so it'd be easier to keep it simple with less work done.

storage/rocksdb/ha_rocksdb.cc
alxyang: as discussed offline, decided to not release the lock to stay consistent with current behaviour.

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -4238,6 +4276,68 @@ bool ha_rocksdb::init_with_fields() {
...
+  /*
+    Curr_ts can only be 0 if there are no snapshots open.
+    should_hide_ttl_rec can only be called when there is >=1 snapshots, unless
+    we are filtering on the write path (single INSERT/UPDATE) in which case
+    we are passed in the current time as tx->m_ttl_filter_timestamp.
+  */
+  if (curr_ts == 0) {
+    sql_print_error("MyRocks: No snapshot open or timestamp passed in call to "
+                    "should_hide_ttl_rec, Aborting.");
+    abort_with_stack_traces();
```
hermanlee: It might be better to allow passing in 0, and treating it as now(). Then you won't need m_ttl_filter_timestamp. The downside though is if 0 is passed in a lot here, we call std::time() on each record, but the code is more resilient and doesn't crash with unexpected cases where 0 is passed to this function.

storage/rocksdb/rdb_datadic.cc
```diff
@@ -3137,6 +3145,19 @@ bool Rdb_tbl_def::put_dict(Rdb_dict_manager *const dict,
...
 }
 
+uint32 Rdb_key_def::generate_index_flags_bitmap(const Rdb_key_def &kd) {
+  return (kd.has_ttl() ? Rdb_key_def::TTL_FLAG : 0);
+}
+
+uint32 Rdb_key_def::calculate_index_flag_offset(uint32 index_flags,
+                                                enum INDEX_FLAG flag) {
+  /*
+    TODO(alexyang): Always return 0 for now as TTL is the only index flag,
```
hermanlee: This is coming in a separate diff?

storage/rocksdb/rdb_datadic.cc
```diff
@@ -4250,12 +4270,29 @@ bool Rdb_dict_manager::get_index_info(const GL_INDEX_ID &gl_index_id,
...
-    *kv_version = 0;
-    *m_index_type = 0;
-    *ttl_duration = 0;
+    index_info->m_index_dict_version = rdb_netbuf_to_uint16(val);
     ptr += RDB_SIZEOF_INDEX_INFO_VERSION;
-    switch (*m_index_dict_version) {
+
+    switch (index_info->m_index_dict_version) {
+    case Rdb_key_def::INDEX_INFO_VERSION_FIELD_FLAGS:
+      /* Sanity check to prevent reading bogus into TTL record. */
```
hermanlee: nit:  remove into

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -7841,6 +8001,20 @@ int ha_rocksdb::check_uniqueness_and_lock(
...
+    /*
+      If the pk key has ttl, we may need to pretend the row wasn't
+      found if it is already expired. The pk record is read into
+      m_retrieved_record by check_and_lock_unique_pk().
+    */
+    if (is_pk(key_id, table, m_tbl_def) && found && m_pk_descr->has_ttl() &&
+        should_hide_ttl_rec(rocksdb::Slice(&m_retrieved_record.front(),
+                                           m_retrieved_record.size()),
+                            (row_info.tx->m_snapshot_timestamp
+                                 ? row_info.tx->m_snapshot_timestamp
```
hermanlee, line 8089: This seems to be the only place you check for m_snapshot_timestamp being 0, but all of the other ones assume it's not. All of the other callsites always have a snapshot involved?

storage/rocksdb/rdb_datadic.cc
```diff
@@ -3137,6 +3145,19 @@ bool Rdb_tbl_def::put_dict(Rdb_dict_manager *const dict,
   return false;
 }
 
+uint32 Rdb_key_def::generate_index_flags_bitmap(const Rdb_key_def &kd) {
+  return (kd.has_ttl() ? Rdb_key_def::TTL_FLAG : 0);
```
hermanlee: Might be better to have this in a separate line for future additions.

```
  if (kd.has_ttl()) {
    flags |= Rdb_key_def::TTL_FLAG;
  }
  return flags;
```

storage/rocksdb/rdb_datadic.cc
alxyang: I left it out for now since TTL is the only index flag, it will always return 0.  But I can add some skeleton logic in there :)

storage/rocksdb/ha_rocksdb.cc
alxyang: Yes, that is my logic.  All the other callsites of `should_hide_ttl_rec` are on the read path, and I believe that they should all have snapshots.  That's also why I have the logic to abort in `should_hide_ttl_rec` if curr_ts is 0.  I want to catch any cases where that may happen to verify that my logic is correct.   

This particular callsite is on the write path filtering side.  Here it's possible not to have a snapshot (for inserts).

storage/rocksdb/ha_rocksdb.cc
alxyang: I created `m_ttl_filter_timestamp` keeping code for the future (secondary keys) in mind.  I think we will need it anyways.  In the event that an insert does not have a snapshot open, we still need some consistent time to filter by throughout the TX or we will get race conditions between pk and SK. If we use std::time() then it's possible that different times will be used in `should_hide_ttl_rec` calls in the write path between pk and sk.

Though, I can't immediately think of a case where that would cause a problem off the top of my head.  Will think some more about it.

alxyang: I think you are correct and we can make do without `m_ttl_filter_timestamp`.

in `check_uniqueness_and_lock()`
   PK found, and it's not expired => return HA_ERR_DUP_KEY
   PK found, and it IS expired => pretend not found, overwrite it
   SK found, then PK must also be expired because it was checked first => pretend not found, overwrite

So I don't think there will be any problem with using `std::time()`.  I'd still like to keep the `curr_ts == 0` check though, to make sure there aren't any unexpected cases where there aren't any snapshots open, think that's ok?

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -376,6 +376,18 @@ static void rocksdb_set_enable_ttl(
   RDB_MUTEX_UNLOCK_CHECK(rdb_sysvars_mutex);
 }
 
+static void rocksdb_set_enable_ttl_read_filtering(
```
hermanlee: I realize we missed this with set_enable_ttl, but this function isn't necessary since you're not doing anything more than setting the variable in memory. The lock is needed if there is more action that needs to be taken in addition to seeing the variable (e.g. calling down into RocksDB to do something, or resetting data structures in MySQL). Having it doesn't hurt though.

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -5018,6 +5115,9 @@ int ha_rocksdb::open(const char *const name, int mode, uint test_if_locked) {
     DBUG_RETURN(HA_ERR_OUT_OF_MEM);
   }
 
+  m_ttl_bytes =
```
hermanlee: ROCKSDB_SIZEOF_TTL_RECORD is 8 bytes. The size of the m_ttl_bytes is also 8 bytes. Why not just store ttl_bytes directly in m_ttl_bytes instead of allocating an array and having m_ttl_bytes point to it?

storage/rocksdb/rdb_datadic.cc
```diff
@@ -3137,6 +3147,29 @@ bool Rdb_tbl_def::put_dict(Rdb_dict_manager *const dict,
...
+}
+
+uint32 Rdb_key_def::calculate_index_flag_offset(uint32 index_flags,
+                                               enum INDEX_FLAG flag) {
+  uint offset = 0;
+  uint flag_len = 0;
+
+  if (index_flags & Rdb_key_def::TTL_FLAG) {
+    offset += ROCKSDB_SIZEOF_TTL_RECORD;
+    if (flag == Rdb_key_def::TTL_FLAG) {
```
hermanlee: Why is flag_len needed? Would the following work?

```
  enum INDEX_FLAG {
    TTL_FLAG,
    FEATURE2_FLAG,
    FEATURE3_FLAG,
  };
  
  uint offset = 0;
  switch (flag) {
    case Rdb_key_def::FEATURE3_FLAG:
        offset += ROCKSDB_SIZEOF_FEATURE2_RECORD;
        /* fall through */
    case Rdb_key_def::FEATURE2_FLAG:
        offset += ROCKSDB_SIZEOF_TTL_RECORD;
        /* fall through */
    case Rdb_key_def::TTL_FLAG:
       break;
  }
  return offset;
```

storage/rocksdb/rdb_datadic.cc
```diff
@@ -3125,9 +3128,16 @@ bool Rdb_tbl_def::put_dict(Rdb_dict_manager *const dict,
...
-    dict->add_or_update_index_cf_mapping(
-        batch, kd.m_index_type, kd.m_kv_format_version, kd.m_index_number,
-        cf_id, kd.m_ttl_duration);
+
+    struct Rdb_index_info index_info;
+    index_info.m_gl_index_id = {cf_id, kd.m_index_number};
+    index_info.m_index_dict_version = Rdb_key_def::INDEX_INFO_VERSION_LATEST;
+    index_info.m_index_type = kd.m_index_type;
+    index_info.m_kv_version = kd.m_kv_format_version;
+    index_info.m_index_flags = Rdb_key_def::generate_index_flags_bitmap(kd);
```
hermanlee: Doesn't Rdb_key_def now m_index_flags_bitmap?

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -5738,11 +5840,15 @@ int ha_rocksdb::create_key_def(const TABLE *const table_arg, const uint &i,
     kv_version = Rdb_key_def::PRIMARY_FORMAT_VERSION_UPDATE1;
   });
 
+  uint32 index_flags = (ttl_duration > 0 ? Rdb_key_def::TTL_FLAG : 0);
```
hermanlee, line 5919: This is where the index_flags are generated. Then they are stored. Do we still need generate_index_flags_bitmap()? Effectively the two lines of code perform the same function. Can we consolidate it down to one callsite?

mysql-test/suite/rocksdb/t/ttl_primary_read_filtering.test
```diff
@@ -0,0 +1,341 @@
...
+
+set global rocksdb_force_flush_memtable_now=1;
+
+# 1 should be hidden even though compaction hasn't run.
+--sorted_result
+SELECT a FROM t1;
+
+set global rocksdb_compact_cf='default';
+
+# none should be hidden yet, compaction runs but records arent expired
```
hermanlee: nit: aren't

mysql-test/suite/rocksdb/t/ttl_primary_read_filtering.test
```diff
@@ -0,0 +1,341 @@
...
+--sorted_result
+SELECT a FROM t1;
+
+# all should be hidden now, even though compaction hasnt run again
```
hermanlee: nit: hasn't

mysql-test/suite/rocksdb/t/ttl_primary_read_filtering.test
```diff
@@ -0,0 +1,341 @@
...
+CREATE TABLE t1 (
+  a int PRIMARY KEY
+) ENGINE=rocksdb
+COMMENT='ttl_duration=1;';
+
+INSERT INTO t1 values (1);
+INSERT INTO t1 values (3);
+INSERT INTO t1 values (5);
+INSERT INTO t1 values (7);
+--real_sleep 2
```
hermanlee: Are these times long enough for filtering to kick in?

mysql-test/suite/rocksdb/t/ttl_primary_read_filtering.test
```diff
@@ -0,0 +1,341 @@
...
+COMMENT='ttl_duration=3;';
+
+INSERT INTO t1 values (1);
+INSERT INTO t1 values (3);
+INSERT INTO t1 values (5);
+INSERT INTO t1 values (7);
+
+connection con1;
+--echo # Creating Snapshot (start transaction)
+BEGIN;
```
hermanlee: If for some reason, this begin starts 3 seconds too late, the filtering could have kicked in. Maybe this is good enough for now, but it's possible this test could become flaky given all of the timing requirements. I'm not sure if a better solution is a debug point that filters or not depending on what the test is trying to accomplish.

hermanlee (changes requested):

mysql-test/suite/rocksdb/t/ttl_primary_read_filtering.test
alxyang: yes, I believe so.  If the ttl_duration is 1 second, and we sleep for 2 seconds, there's no possible way it won't be filtered during on the read path.  The snapshot opened by the select will be guaranteed to be 2 seconds after the last insert.  Even if real_sleep 2 doesn't sleep for a full 2 seconds for any reason, it at least sleeps for 1000ms.

mysql-test/suite/rocksdb/t/ttl_primary_read_filtering.test
alxyang: I have run these tests on repeat=100 parallel=10 quite a few times and have seen no flaky failures thus far. 

I'll discuss with you offline on how we can maybe figure out something less timing intensive, I don't currently see how we would be able to utilize a debug point.  Perhaps it can be added in a future diff.

storage/rocksdb/rdb_datadic.cc
alxyang: In the example code above, what happens if TTL and feature 3 is turned on for an index, but feature 2 is turned off? I think there may be a miscalculation in these cases.

Let's say `ROCKSDB_SIZEOF_FEATURE3_RECORD == 4`, `ROCKSDB_SIZEOF_FEATURE2_RECORD == 5`, and `ROCKSDB_SIZEOF_TTL_RECORD == 8`.

Now you want where Feature 3 starts in the index record.
Then you will get  `5 + 8 == 13`, even though what we actually want is just `8`, since only the TTL field exists before feature 3.

storage/rocksdb/ha_rocksdb.cc
alxyang: ill remove it for now since its unneeded.

storage/rocksdb/ha_rocksdb.cc
alxyang: hmmm, I know we talked about this offline but I think I misunderstood. `ttl_bytes` is the result of an `Rdb_string_reader.read` call, which returns pointer to the start of the bytes read.  The bytes read in this case are from `m_retrieved_record`.  However, in the code it is now possible for `m_retrieved_record` to change between the `convert_record_from_storage_format` (where `m_ttl_bytes` is populated) and `convert_record_to_storage_format` calls, this potentially happens on the write path when finding a duplicate key conflict.  This means if we store `ttl_bytes` into `m_ttl_bytes` directly (aka if they point to the same place), it can change from out under us by the time `convert_record_to_storage_format` is called, which we don't want.

So either way, we need to copy the content `ttl_bytes` somewhere.  Am I missing something?  I changed it to use `std::string` assign instead but it's still allocating a buffer under the hood.

mysql-test/suite/rocksdb/t/ttl_primary_read_filtering.test
laurynas-biveinis: Have you considered introducing a MySQL status variable or I_S table that would count e.g. the number of filtered records? This would work with both release and debug builds and then real_sleeps could be replaced by condition waits. A complication could remain that the execution of BEGIN can be delayed too, but that can also be detected in the testcase, with possible handling of skipping the testcase.

alxyang: @laurynas-biveinis We already have this, it's called `rocksdb_rows_expired`.  But in this case, we are talking about 'read filtering', not the compaction filter.

On the read paths, we need to hide records from the user that have expired but not yet removed by compaction.  Here I don't see currently how a status variable/I_S table would help.

That is a good idea for the rest of the tests though (ttl_primary and ttl_primary_with_partitions), where read filtering is disabled.

laurynas-biveinis: Right, I misunderstood the problem at hand. Now, looking at this snippet, the following should guarantee the desired snapshot contents on testing hosts that insert arbitrary pauses between statements:

```
INSERT INTO t1 values (1);
INSERT INTO t1 values (3);
INSERT INTO t1 values (5);
INSERT INTO t1 values (7);

BEGIN
--let $snapshot_size= `SELECT COUNT(*) FROM t1`
--let $i= 0
while ($snapshot_size != 4)
{
  $i++;
  ROLLBACK;
  .. insert ...
  BEGIN;
  --let $snapshot_size= `SELECT COUNT(*) FROM t1`
}

if ($snapshot_size != 4 && $i == 1000)
{
  --die Your testing host is too slow for reasonable TTL testing
}
```

This should remove any flakiness on reasonable hosts, but cumbersome, still would fail (but at least explicitly) on the worst machines and would have to be applied not only there but even for the code above like 

```
COMMENT='ttl_duration=5;';

INSERT INTO t1 values (1, UNIX_TIMESTAMP());
--real_sleep 6
INSERT INTO t1 values (2, UNIX_TIMESTAMP());
INSERT INTO t1 values (3, UNIX_TIMESTAMP());
```

where 1 might get TTL-filtered too on a slow host.

Perhaps the way to proceed would be to accept the current code as-is (as you report no flakiness failures on your testing), and wait for us to submit patches, if any

alxyang: @laurynas-biveinis I have figured out a way to make the tests completely deterministic using debug variables, via manually overriding the generated TTL time and the snapshot time compaction uses to filter TTL records.  An update is coming soon., stay tuned :)

alxyang:
cc @hermanlee @laurynas-biveinis OK, i have updated the PR.  All the tests should be completely deterministic now and not flaky.  Also all comments have either been addressed or I responded here regarding why they haven't been.

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -4238,6 +4277,76 @@ bool ha_rocksdb::init_with_fields() {
...
+    should_hide_ttl_rec can only be called when there is >=1 snapshots, unless
+    we are filtering on the write path (single INSERT/UPDATE) in which case
+    we are passed in the current time as curr_ts.
+
+    In the event curr_ts is 0, we always decide not to filter the record. We
+    also log a warning and increment a diagnostic counter.
+  */
+  if (curr_ts == 0) {
+    // NO_LINT_DEBUG
+    sql_print_warning("MyRocks: No snapshot or timestamp passed in call to "
```
hermanlee: Hopefully, this isn't too verbose.

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -4238,6 +4277,76 @@ bool ha_rocksdb::init_with_fields() {
...
+    sql_print_error("Decoding ttl from PK value failed, "
+                    "for index (%u,%u), val: %s",
+                    gl_index_id.cf_id, gl_index_id.index_id, buf.c_str());
+    DBUG_ASSERT(0);
+    return false;
+  }
+
+  /* Hide record if it has expired before the current snapshot time. */
+#ifndef NDEBUG
+  return ts + m_pk_descr->m_ttl_duration + rdb_dbug_set_ttl_read_filter_ts() <=
```
hermanlee: It might be better to have one return statement and let the read_filter_ts() value be a variable set in NDEBUG:

```
  uint64 read_filter_ts = 0;
#ifndef NDEBUG
  read_filter_ts = rdb_dbug_set_ttl_read_filter_ts();
#endif

  return ts + m_pk_descr->m_ttl_duration + read_filter_ts <= static_cast<uint64>(curr_ts);
```

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -4285,8 +4394,14 @@ int ha_rocksdb::convert_record_to_storage_format(
...
       char *const data = const_cast<char *>(m_storage_record.ptr());
+#ifndef NDEBUG
+      rdb_netbuf_store_uint64(
+          reinterpret_cast<uchar *>(data),
+          rdb_netbuf_to_uint64(reinterpret_cast<const uchar *>(ts)) +
+              rdb_dbug_set_ttl_rec_ts());
+#else
       memcpy(data, ts, ROCKSDB_SIZEOF_TTL_RECORD);
-
+#endif
```
hermanlee: Maybe something like

```
  memcpy(data, ts, ROCKSDB_SIZEOF_TTL_RECORD);
#ifndef NDEBUG
  // Adjust for test case if needed
  rdb_netbuf_store_uint64(reinterpret_cast<uchar *>(data),  rdb_netbuf_to_uint64(reinterpret_cast<const uchar *>(data)) + rdb_dbug_set_ttl_rec_ts());
#endif
```

storage/rocksdb/ha_rocksdb_proto.h
```diff
@@ -71,6 +71,12 @@ Rdb_cf_manager &rdb_get_cf_manager();
 
 const rocksdb::BlockBasedTableOptions &rdb_get_table_options();
 bool rdb_is_ttl_enabled();
+bool rdb_is_ttl_read_filtering_enabled();
+#ifndef NDEBUG
```
laurynas-biveinis, line 75: MySQL uses DBUG_OFF

storage/rocksdb/rdb_compact_filter.h
```diff
@@ -64,7 +64,29 @@ class Rdb_compact_filter : public rocksdb::CompactionFilter {
...
+            oldest snapshot time (occurs once per compaction).
+          */
+          rocksdb::DB *const rdb = rdb_get_rocksdb_db();
+          if (!rdb->GetIntProperty(rocksdb::DB::Properties::kOldestSnapshotTime,
+                                   &m_snapshot_timestamp) ||
+              m_snapshot_timestamp == 0) {
+            m_snapshot_timestamp = static_cast<uint64_t>(std::time(nullptr));
+          }
+
+#ifndef NDEBUG
```
laurynas-biveinis, line 82: MySQL uses DBUG_OFF

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -4238,6 +4277,76 @@ bool ha_rocksdb::init_with_fields() {
...
+    // NO_LINT_DEBUG
+    sql_print_error("Decoding ttl from PK value failed, "
+                    "for index (%u,%u), val: %s",
+                    gl_index_id.cf_id, gl_index_id.index_id, buf.c_str());
+    DBUG_ASSERT(0);
+    return false;
+  }
+
+  /* Hide record if it has expired before the current snapshot time. */
+#ifndef NDEBUG
```
laurynas-biveinis: MySQL uses DBUG_OFF

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -4238,6 +4277,76 @@ bool ha_rocksdb::init_with_fields() {
...
+
+  The purpose of read filtering for tables with TTL is to ensure that
+  during a transaction a key which has expired already but not removed by
+  compaction yet is not returned to the user.
+
+  Without this the user might be hit with problems such as disappearing
+  rows within a transaction, etc, because the compaction filter ignores
+  snapshots when filtering keys.
+*/
+int ha_rocksdb::should_hide_ttl_rec(const rocksdb::Slice &ttl_rec_val,
```
laurynas-biveinis: "should" and all the returns returning bool expressions suggest this method should return bool, not int

mysql-test/suite/rocksdb/t/ttl_primary.test
```diff
@@ -1,11 +1,10 @@
+--source include/have_debug.inc
```
laurynas-biveinis, line 1: Release build testing for TTL to the extent possible still might be a good idea, I'd suggest to rename this one to "ttl_primary_debug.test", create a separate "ttl_primary.test" without a have_debug.inc, and move the the timing-insensitive tests (i.e. insert records with TTL=1, real_sleep 2, observe empty snapshot) there. For those tests setting negative debug_ttl values shouldn't be required.

mysql-test/suite/rocksdb/t/ttl_primary_read_filtering.test
```diff
@@ -0,0 +1,370 @@
+--source include/have_rocksdb.inc
+
+--disable_warnings
```
laurynas-biveinis: Obsolete idiom

mysql-test/suite/rocksdb/t/ttl_primary_read_filtering.test
```diff
@@ -0,0 +1,370 @@
+--source include/have_rocksdb.inc
```
laurynas-biveinis, line 2: Needs --source include/have_debug.inc as the test is using those debug TTL variables

laurynas-biveinis, line 2: Same comment re. release/debug test splitting as for ttl_primary.test

mysql-test/suite/rocksdb/t/ttl_primary_read_filtering.test
```diff
@@ -0,0 +1,370 @@
...
+--sorted_result
+SELECT * FROM t1;
+
+UPDATE t1 set a = a - 1;
+--sorted_result
+SELECT * FROM t1;
+
+DROP TABLE t1;
+
+# Ensure no rows can disappear in the middle of long-running transactions
```
laurynas-biveinis, line 241: Add --source include/count_sessions.inc here and ...

mysql-test/suite/rocksdb/t/ttl_primary_read_filtering.test
```diff
@@ -0,0 +1,370 @@
...
+SELECT * FROM t1;
+set global rocksdb_enable_ttl_read_filtering=0;
+--sorted_result
+SELECT * FROM t1;
+set global rocksdb_enable_ttl_read_filtering=1;
+
+disconnect con2;
+disconnect con1;
+connection default;
+DROP TABLE t1;
```
laurynas-biveinis: ... --source/wait_until_count_sessions.inc here. The server processes disconnects asynchronously, and so it's actually possible for the next test in the run to see con1/con2 at its start, resulting in flaky tests if they do i.e. SHOW PROCESSLIST or otherwise depend on the open connections.

mysql-test/suite/rocksdb/t/ttl_primary_read_filtering.test
```diff
@@ -0,0 +1,370 @@
...
+--sorted_result
+SELECT * FROM t1; # <= shouldn't be filtered out here
+
+UPDATE t1 set a = a + 1;
+--sorted_result
+SELECT * FROM t1; # <= shouldn't be filtered out here
+
+COMMIT;
+
+--sorted_result # <= filtered out here because time has past.  This is correct.
```
laurynas-biveinis: s/past/passed ?

mysql-test/suite/rocksdb/t/ttl_primary_with_partitions.test
```diff
@@ -5,8 +5,6 @@ DROP TABLE IF EXISTS t1;
 DROP TABLE IF EXISTS t2;
```
laurynas-biveinis: Needs have_debug.inc and same comment re. release/debug build testing as for ttl_primary.test

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -1157,8 +1148,43 @@ static MYSQL_SYSVAR_BOOL(pause_background_work, rocksdb_pause_background_work,
 
 static MYSQL_SYSVAR_BOOL(
     enable_ttl, rocksdb_enable_ttl, PLUGIN_VAR_RQCMDARG,
-    "Disable expired ttl records from being dropped during compaction.",
-    nullptr, rocksdb_set_enable_ttl, TRUE);
+    "Enable expired TTL records to be dropped during compaction.", nullptr,
+    nullptr, TRUE);
+
+static MYSQL_SYSVAR_BOOL(
```
laurynas-biveinis, line 1174: Have you considered hiding the variables altogether from release builds? This is the usual MySQL practice for debug-only variables, IMHO this would prevent some minor confusion for users seeing them, and the tests need them for debug builds only anyway. The single drawback I can think of is that their use in MTR -master.opt files would require --loose- prefix, introducing uncaught typo risk, but these variables are very unlikely to be used that way.

laurynas-biveinis (commented):
@alxyang, yes this scheme will work. I have left some comments for your consideration

storage/rocksdb/rdb_datadic.cc
```diff
@@ -3137,6 +3147,20 @@ bool Rdb_tbl_def::put_dict(Rdb_dict_manager *const dict,
   return false;
 }
 
+uint32 Rdb_key_def::calculate_index_flag_offset(uint32 index_flags,
+                                                enum INDEX_FLAG flag) {
+  uint offset = 0;
+  uint flag_len = 0;
+  if (index_flags & Rdb_key_def::TTL_FLAG) {
+    offset += ROCKSDB_SIZEOF_TTL_RECORD;
```
hermanlee: Spoke about this offline. Could change this to use an array that has all of the sizes of each of the flag types. Then iterate through index_flags to determine if a particular flag is set. If it is, then add the size of the value to the offset. Return offset when the iteration level == flag. Could assert flag & index_flags somewhere for correctness, but its possible future flag types might not exist in index_flags, so this needs to handle that condition. Returning UINT_MAX for the not found offset is probably fine, but all callers to this function would need to check/assert on it.

storage/rocksdb/rdb_datadic.h
```diff
@@ -326,6 +328,12 @@ class Rdb_key_def {
     PER_PARTITION_CF_FLAG = 4,
   };
 
+  // bit flags which denote myrocks specific fields stored in the record
+  // currently only used for TTL.
+  enum INDEX_FLAG {
+    TTL_FLAG = 1,
```
hermanlee: Spoken earlier, probably need to add a static array for the sizes of each index flag field. Also change this from "TTL_FLAG = 1" to "TTL_FLAG = 1 << 0" so that it's explicit these should be bitfields.

mysql-test/suite/rocksdb/t/ttl_primary_with_partitions.test
```diff
@@ -27,6 +25,7 @@ PARTITION BY LIST(c1) (
     PARTITION custom_p2 VALUES IN (3, 6, 9)
 );
 
+set global rocksdb_debug_ttl_rec_ts = -2;
```
hermanlee: Might as well set these values really large or really small: -3600, or +3600. You probably want it to be two orders of magnitude different so that the changes of that much time passing is very small.

hermanlee (changes requested):

storage/rocksdb/ha_rocksdb_proto.h
alxyang: True. But we are already using `NDEBUG` in other areas of MyRocks.  Also RocksDB uses `NDEBUG` as well.  I think I'll keep this the same for now, we can always change later.

storage/rocksdb/rdb_compact_filter.h
alxyang: see other comment.

storage/rocksdb/ha_rocksdb.cc
alxyang: see other comment.

storage/rocksdb/ha_rocksdb.cc
alxyang: yes, I considered it. Hiding them completely would break `main.mysqld-help-notwin` test.  It will also break `rocksdb.rocksdb` test.  This is why I left them there.  I do explicitly say in the variable description that it is a no-op in non-debug builds.

laurynas-biveinis: I see, something could be done about mysqld--help-notwin as it already filters plugin variables and does not fail for the other debug-only ones, but rocksdb.rocksdb could be enough of a reason to leave them in

storage/rocksdb/ha_rocksdb.cc
alxyang: I'll just remove it.  I think now that we have stats this is mostly a useless warning unless we add more information to it.  We'll forgo it for now.

mysql-test/suite/rocksdb/t/ttl_primary_read_filtering.test
```diff
@@ -0,0 +1,371 @@
...
+--sorted_result
+SELECT * FROM t1;
+
+DROP TABLE t1;
+
+# Ensure no rows can disappear in the middle of long-running transactions
+# Also ensure repeatable-read works as expected
+connect (con1,localhost,root,,);
+connect (con2,localhost,root,,);
+ --source include/count_sessions.inc
```
laurynas-biveinis: Must go before the connects (wait_until_count_sessions has to wait until 1 session, not 3)

storage/rocksdb/rdb_datadic.cc
```diff
@@ -3137,6 +3147,29 @@ bool Rdb_tbl_def::put_dict(Rdb_dict_manager *const dict,
...
 
+// Length that each index flag takes inside the record.
+// Each index in the array maps to the enum INDEX_FLAG
+static const std::array<int, 1> index_flag_lengths = {
+    {ROCKSDB_SIZEOF_TTL_RECORD}};
+
+uint32 Rdb_key_def::calculate_index_flag_offset(uint32 index_flags,
+                                                enum INDEX_FLAG flag) {
+  uint offset = 0;
+  for (size_t bit = 0; bit < sizeof(index_flags); ++bit) {
```
hermanlee: Should be bit < sizeof(index_flags) * BITS_PER_BYTE.

storage/rocksdb/rdb_datadic.cc
```diff
@@ -3137,6 +3147,29 @@ bool Rdb_tbl_def::put_dict(Rdb_dict_manager *const dict,
...
+    int mask = 1 << bit;
+    if (index_flags & mask) {
+      /* Exit once we've reached the proper flag */
+      if (flag == mask) {
```
hermanlee: This check should be outside the index_flags & mask conditional. It should probably be an error perhaps if the index_flag bit isn't set for the flag we're looking for, but this function is calculating the offset of where the data would have started from. I could see us using this function to skip over all of the metatags to get to the data. In which case, maybe INDEX_FLAG should have a max enum. If that is requested, then we return whatever the offset is at that point in time.

hermanlee (dismissed):

storage/rocksdb/rdb_datadic.h
```diff
@@ -326,6 +328,13 @@ class Rdb_key_def {
     PER_PARTITION_CF_FLAG = 4,
   };
 
+  // bit flags which denote myrocks specific fields stored in the record
+  // currently only used for TTL.
+  enum INDEX_FLAG {
+    TTL_FLAG = 1 << 0,
+    MAX_FLAG = TTL_FLAG << 1, // Marks where actual record starts
```
hermanlee: Can you add a comment that this must be last.

storage/rocksdb/rdb_datadic.cc
```diff
@@ -3137,6 +3147,38 @@ bool Rdb_tbl_def::put_dict(Rdb_dict_manager *const dict,
...
+    if (flag == mask) {
+      break;
+    }
+
+    if (index_flags & mask) {
+      offset += index_flag_lengths[bit];
+    }
+  }
+
+	return offset;
```
hermanlee: Fix spacing?

storage/rocksdb/rdb_datadic.cc
```diff
@@ -3137,6 +3147,38 @@ bool Rdb_tbl_def::put_dict(Rdb_dict_manager *const dict,
...
+                                                enum INDEX_FLAG flag) {
+
+  DBUG_ASSERT(Rdb_key_def::has_index_flag(index_flags, flag));
+
+  uint offset = 0;
+  for (size_t bit = 0; bit < sizeof(index_flags) * CHAR_BIT; ++bit) {
+    int mask = 1 << bit;
+
+    /* Exit once we've reached the proper flag */
+    if (flag == mask) {
```
hermanlee: To be safer, flag & mask in case multiple flags are set.

alxyang: Hmm...what do you mean by multiple flags? This is the exit condition where the `flag == mask` the mask should always only have 1 bit set, and the enum `flag` passed is an enum where only 1 bit can be turned on.  I guess it doesn't hurt to turn it into `&`, but that makes it seem like multiple bits can be turned on when only 1 can.

hermanlee (approved):

PR #620, opened 2017-05-09
Implementent mysql_pause for Windows and Power

grooverdan:
Added compile barrier to prevent removal.

Added low priority/resume priority to allow for CPU scheduling
on Power.

### DIFF
--- a/sql/mysqld.cc
+++ b/sql/mysqld.cc
@@ -124,7 +124,6 @@
 #include <poll.h>
 #endif
 
-#include <xmmintrin.h>
 #include "mpsc.h"
 
 #ifdef HAVE_FESETROUND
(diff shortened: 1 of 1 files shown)


gunnarku:
This looks like a change which is better suited for an upstream branch.

grooverdan:
It was added by fb in commit https://github.com/facebook/mysql-5.6/commit/08729515efc16fa412621c7209afee9862da7fd3 in this branch. Which upstream branch are you referring too?

anirbanr-fb (approved):
I am assuming this works on Windows and compiles on Power pc.

grooverdan:
Works on Power(8).

Windows is based on what I've seen elsewhere and that the existing header xmmintrin.h isn't a windows header afaik.

Issue #132, opened 2016-01-03
Trying to lock mutex at sql/table_stats.cc, line 953, when the mutex was already locked at sql/table_stats.cc, line 465


RoelVdP:
<pre>
Version: '5.6.24-debug'  socket: '/sda/FB-02.01.16-mysql-5.6.24-linux-x86_64/socket.sock'  port: 10529  MySQL Community Server (GPL)
safe_mutex: Trying to lock mutex at /git/FB-mysql-5.6/sql/table_stats.cc, line 953, when the mutex was already locked at /git/FB-mysql-5.6/sql/table_stats.cc, line 465 in thread T@4
23:27:43 UTC - mysqld got signal 6 ;

+bt
#0  0x00007f6736c22741 in __pthread_kill (threadid=<optimized out>, signo=6) at ../nptl/sysdeps/unix/sysv/linux/pthread_kill.c:61
#1  0x0000000000e3d22e in my_write_core (sig=6) at /git/FB-mysql-5.6/mysys/stacktrace.c:422
#2  0x0000000000ab6060 in handle_fatal_signal (sig=6) at /git/FB-mysql-5.6/sql/signal_handler.cc:239
#3  signal handler called
#4  0x00007f6734baf5f7 in __GI_raise (sig=sig@entry=6) at ../nptl/sysdeps/unix/sysv/linux/raise.c:56
#5  0x00007f6734bb0ce8 in __GI_abort () at abort.c:90
#6  0x0000000000e40fae in safe_mutex_lock (mp=0x1f85d00 <LOCK_global_table_stats>, try_lock=0 '\000', file=0x14cf740 "/git/FB-mysql-5.6/sql/table_stats.cc", line=953) at /git/FB-mysql-5.6/mysys/thr_mutex.c:124
#7  0x00000000009ac9ef in inline_mysql_mutex_lock (that=0x1f85d00 <LOCK_global_table_stats>, src_file=0x14cf740 "/git/FB-mysql-5.6/sql/table_stats.cc", src_line=953) at /git/FB-mysql-5.6/include/mysql/psi/mysql_thread.h:698
#8  0x00000000009b01fb in table_stats_delete (old_name=0x7f672e47af83 "data/#sql_3a8e_0") at /git/FB-mysql-5.6/sql/table_stats.cc:953
[...]

PR #621, opened 2017-05-12
Change LOCK_global_table_stats to error-checking mutex to avoid deadlock

alxyang:
It's possible for this mutex to be locked twice by one thread when
`ha_myisam::write_row()` errors out during an information schema query.

We use an error checking mutex here so we can handle this situation.
Fixes https://github.com/facebook/mysql-5.6/issues/132.

### DIFF
--- a/mysql-test/include/check-testcase.test
+++ b/mysql-test/include/check-testcase.test
@@ -91,5 +91,5 @@ if (!$tmp) {
   SHOW SLAVE STATUS;
 }
 
-#call mtr.check_testcase();
+call mtr.check_testcase();
 --enable_query_log
--- a/mysql-test/include/check-testcase.test
+++ b/mysql-test/include/check-testcase.test
@@ -91,5 +91,5 @@ if (!$tmp) {
   SHOW SLAVE STATUS;
 }
 
-call mtr.check_testcase();
+#call mtr.check_testcase();
 --enable_query_log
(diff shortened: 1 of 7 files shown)


mysql-test/include/check-testcase.test
```diff
@@ -91,5 +91,5 @@ if (!$tmp) {
   SHOW SLAVE STATUS;
 }
 
-call mtr.check_testcase();
+#call mtr.check_testcase();
```
laurynas-biveinis: Is this change intentional?

storage/myisam/ha_myisam.cc
```diff
@@ -813,6 +813,10 @@ int ha_myisam::write_row(uchar *buf)
 {
   ha_statistic_increment(&SSV::ha_write_count);
 
+  DBUG_EXECUTE_IF("myisam_simulate_crash_in_write_row", {
```
laurynas-biveinis: "myisam_simulate_max_fs_exceeded_in_write_row" or similar?

alxyang: that wouldn't be an entirely accurate name. this issue manifests for any error in write_row, as long as `fill_table_stats()` was called prior.  It's not necessary for max_fs to be exceeded.  Internally, we've run into this deadlock without that being the case.

though, "crash" doesn't seem like a good choice of words either, i'll rename it to `myisam_simulate_err_in_write_row`

mysql-test/include/check-testcase.test
alxyang: nice catch.

PR #625, opened 2017-05-18
Track number of tables being used and autocommit when it hits 0.

hermanlee:
Summary:

There is a bug in myrocks::ha_rocksdb::external_lock where releasing a
lock on the table triggers a commit of the transaction. The query plan
for index_merge also acquires ::external_lock on a table, in addition to
the ::external lock acquired during the update query. However, the
index_merge's lock is released early (when the SQL_SELECT object is
destroyed), which would trigger a commit of the Rdb_transaction before
sql_update calls prepare.

This would result in a crash when sql_update calls rocksdb_prepare.  The
transaction is incorrectly committed and the prepare segfaults on a null
pointer.

The solution is to follow the InnoDB one, reference the number of
external lock calls through n_tables_in_use.

Test Plan:

  mtr unit tests
  copy over index_merge test plan from InnoDB.

### DIFF
--- a/mysql-test/suite/rocksdb/t/index_merge_rocksdb.test
+++ b/mysql-test/suite/rocksdb/t/index_merge_rocksdb.test
@@ -1,0 +1,85 @@
+--source include/have_rocksdb.inc
+
+#
+# index_merge_rocksdb test copied over from index_merge_ror.inc
+#
+# Triggers issue # https://github.com/facebook/mysql-5.6/issues/604
+
+CREATE TABLE t1
+(
+  /* fields/keys for row retrieval tests */
+  key1 INT,
+  key2 INT,
+  key3 INT,
+  key4 INT,
+
+  /* make rows much bigger then keys */
+  filler1 CHAR(200),
+
+  KEY(key1),
+  KEY(key2)
+) ENGINE=ROCKSDB;
+
+# fill table
+CREATE TABLE t0 AS SELECT * FROM t1;
+--disable_query_log
+--echo # Printing of many insert into t0 values (....) disabled.
+let $cnt=100;
+while ($cnt)
+{
+  eval INSERT INTO t0 VALUES (0, 0, 0, 0, 'data1');
+  dec $cnt;
+}
+
+--echo # Printing of many insert into t1 select .... from t0 disabled.
+let $1=4;
+while ($1)
+{
+  let $2=4;
+  while ($2)
+  {
+    let $3=4;
+    while ($3)
+    {
+      eval INSERT INTO t1 SELECT key1, key2, key3, key4, filler1 FROM t0;
+      dec $3;
+    }
+    dec $2;
+  }
+  dec $1;
+}
+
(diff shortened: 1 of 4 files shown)


mysql-test/suite/rocksdb/t/index_merge_rocksdb.test
```diff
@@ -0,0 +1,93 @@
...
+
+SELECT COUNT(*) FROM t1;
+
+-- disable_query_log
+-- disable_result_log
+ANALYZE TABLE t1;
+-- enable_result_log
+-- enable_query_log
+
+SET GLOBAL rocksdb_force_flush_memtable_now = 1;
```
alxyang, line 79: nit: I don't think we need this anymore, it was for debugging purposes to break on something before the crash in gdb. guess it doesnt hurt to leave it in though :)

mysql-test/suite/rocksdb/t/index_merge_rocksdb.test
```diff
@@ -0,0 +1,93 @@
+--source include/have_rocksdb.inc
+# Skiping this test from Valgrind execution as per Bug-14627884
+--source include/not_valgrind.inc
+# Adding big test option for this test.
+#--source include/big_test.inc
```
alxyang: nit: remove?

alxyang (approved):
lgtm!

mysql-test/suite/rocksdb/t/index_merge_rocksdb.test
```diff
@@ -0,0 +1,93 @@
...
+
+#
+# index_merge_rocksdb test copied over from index_merge_ror.inc
+#
+# Triggers issue # https://github.com/facebook/mysql-5.6/issues/604
+
+--disable_warnings
+DROP TABLE IF EXISTS t0,t1;
```
laurynas-biveinis: Nit: this idiom is considered obsolete, we are removing it as we encounter on Percona Server port

mysql-test/suite/rocksdb/t/index_merge_rocksdb.test
```diff
@@ -0,0 +1,93 @@
+--source include/have_rocksdb.inc
+# Skiping this test from Valgrind execution as per Bug-14627884
```
laurynas-biveinis: Without knowing what "14627884" is, I'd try running this test under Valgrind and possibly removing not_valgrind.inc if it appears to work then. Depending on the runtime, it might need not_valgrind_without_big.inc instead

mysql-test/suite/rocksdb/t/index_merge_rocksdb.test
hermanlee: I'll remove these lines. They were left over from the earlier test cases which have been removed.

mysql-test/suite/rocksdb/t/index_merge_rocksdb.test
hermanlee: I'll remove them.

PR #628, opened 2017-05-31
Pass rocksdb logger and trash directory to SstFileManager

IslamAbdelRahman:
We need to pass a trash directory to SstFileManager component to be able to smooth file deletions.


alxyang (approved):
lgtm

PR #630, opened 2017-06-01
Use rocksdb table in type set indexes test

### LINKED ISSUE #595: rocksdb.type_set_indexes appears to be incorrectly creating innodb table as test target
https://github.com/facebook/mysql-5.6/blob/fb-mysql-5.6.35/mysql-test/suite/rocksdb/t/type_set_indexes.test#L75

I think this is supposed to be creating a rocksdb table, I see no logical reason why this would use innodb.

pbjc:
Fixes #595

### DIFF
--- a/mysql-test/suite/rocksdb/r/type_set_indexes.result
+++ b/mysql-test/suite/rocksdb/r/type_set_indexes.result
@@ -77,4 +77,39 @@ a
 Africa,Europe,Asia
 S.America,Europe
 DROP TABLE t1;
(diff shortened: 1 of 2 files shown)


alxyang (changes requested):
let's actually enable this test, since it's currently disabled. 

Make sure to re-record it after it's enabled.
To make it pass you're going to have to add the PK values to the end of each insert statement on lines 86-93.

alxyang (approved):
lgtm!

PR #632, opened 2017-06-02
Initialize default CF flags when initializing data dict

### LINKED ISSUE #589: INFORMATION_SCHEMA.ROCKSDB_GLOBAL_INFO returns uninitialized values when no rocksdb tables have been created yet
Example:
```
   mysql> SELECT * FROM INFORMATION_SCHEMA.ROCKSDB_GLOBAL_INFO;
   +----------+------+-------------------------+
   | TYPE     | NAME | VALUE                   |
   +----------+------+-------------------------+
   | CF_FLAGS | 0    | default [2684988720]    |
   | CF_FLAGS | 1    | __system__ [2684988720] |
   +----------+------+-------------------------+
   2 rows in set (0.00 sec)
```

When data dictionary contains no tables `Rdb_dict_manager::get_cf_flags` returns `false` which means "nothing found", but `rdb_i_s_global_info_fill_table` doesn't check return value and fills I_S table with uninitialized values. `mysqld --bootstrap` always create some RocksDB tables, so `rocksdb.information_schema` test doesn't expose this bug.

pbjc:
Fixes #589. Changes:

- Add short circuit behavior to `Rdb_dict_manager::init` by returning `HA_EXIT_FAILURE` immediately if the system CF handle is null.
- Initialize default CF flags when system CF flags are initialized

Need to investigate how to test for the case where no rdb tables have been created yet since our test setup creates them automatically.

### DIFF
--- a/storage/rocksdb/rdb_datadic.cc
+++ b/storage/rocksdb/rdb_datadic.cc
@@ -4097,6 +4097,13 @@ bool Rdb_dict_manager::init(rocksdb::DB *const rdb_dict,
   m_db = rdb_dict;
 
   m_system_cfh = cf_manager->get_or_create_cf(m_db, DEFAULT_SYSTEM_CF_NAME);
+  rocksdb::ColumnFamilyHandle *default_cfh =
+      cf_manager->get_cf(DEFAULT_CF_NAME);
+
+  // System CF and default CF should be initialized
+  if (m_system_cfh == nullptr || default_cfh == nullptr) {
+    return HA_EXIT_FAILURE;
+  }
 
   rdb_netbuf_store_index(m_key_buf_max_index_id, Rdb_key_def::MAX_INDEX_ID);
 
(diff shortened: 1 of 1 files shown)


storage/rocksdb/rdb_datadic.cc
```diff
@@ -4097,6 +4097,12 @@ bool Rdb_dict_manager::init(rocksdb::DB *const rdb_dict,
   m_db = rdb_dict;
 
   m_system_cfh = cf_manager->get_or_create_cf(m_db, DEFAULT_SYSTEM_CF_NAME);
+  rocksdb::ColumnFamilyHandle *default_cfh = cf_manager->get_or_create_cf(m_db, DEFAULT_CF_NAME);
```
alxyang: let's call `get_cf` here instead of `get_or_create_cf`, it makes it clearer that the default column family should already be created by this point.  Maybe add a comment too.

alxyang (changes requested):
mostly lgtm, one nit. Don't forget to import into phabricator

alxyang:
cc @gunnarku who has touched this code recently

alxyang (approved):

PR #633, opened 2017-06-07
Enable simulated cache in MyRocks

IslamAbdelRahman:
Simulated cache allow us to figure out the hit/miss rate with a specific cache size without changing the real block cache.

Add an option to enable this feature in MyRocks

### DIFF
--- a/storage/rocksdb/ha_rocksdb.cc
+++ b/storage/rocksdb/ha_rocksdb.cc
@@ -57,6 +57,7 @@
 #include "rocksdb/utilities/checkpoint.h"
 #include "rocksdb/utilities/convenience.h"
 #include "rocksdb/utilities/memory_util.h"
+#include "rocksdb/utilities/sim_cache.h"
 #include "util/stop_watch.h"
 
 /* MyRocks includes */
@@ -412,6 +413,7 @@ static void rocksdb_set_max_background_jobs(THD *thd,
 // Options definitions
 //////////////////////////////////////////////////////////////////////////////
 static long long rocksdb_block_cache_size;
+static long long rocksdb_sim_cache_size;
 /* Use unsigned long long instead of uint64_t because of MySQL compatibility */
 static unsigned long long  // NOLINT(runtime/int)
     rocksdb_rate_limiter_bytes_per_sec;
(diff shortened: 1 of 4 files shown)


alxyang (approved):
lgtm, but you need to re-record main.mysqld-notwin and main.mysqld-notwin-profiling

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -3656,8 +3668,17 @@ static int rocksdb_init_func(void *const p) {
       (rocksdb::BlockBasedTableOptions::IndexType)rocksdb_index_type;
 
   if (!rocksdb_tbl_options->no_block_cache) {
-    rocksdb_tbl_options->block_cache =
+    std::shared_ptr<rocksdb::Cache> block_cache =
         rocksdb::NewLRUCache(rocksdb_block_cache_size);
+    if (rocksdb_sim_cache_size > 0) {
```
hermanlee, line 3697: Rather than a new size for simulated block cache, maybe reuse block_cache_size and add an enum that controls the type of block_cache being created? 0 for LRU, 1 for simulated_cache, 2+ are reserved for new block_cache types? Then you don't end up creating a block_cache and throwing it away.

IslamAbdelRahman: @hermanlee, I am not sure I understand your point

The main purpose of simulated cache is to have a different size than block cache, so that we can know what will be the hit rate if we increase/decrease the size without actually changing it

hermanlee: Ah, nevermind. I missed the fact the NewSimCache is constructed from the newly allocated LRU block_cache. I thought it was a standalone cache type, hence, the thought on expanding this to support different types of caches: LRU, FIFO, etc.

IslamAbdelRahman:
@alxyang @hermanlee can you guys take another look ?

IslamAbdelRahman:
@alxyang @hermanlee ping ping :D

hermanlee (approved):

hermanlee:
Looks like there's one more test to update: rocksdb_sys_vars.all_vars.

lth (approved):

PR #634, opened 2017-06-12
Changing default rocksdb_sst_mgr_rate_bytes_per_sec to 0

yoshinorim:
Summary: This diff changes default rocksdb_sst_mgr_rate_bytes_per_sec
from 64MB to 0. Default 64MB has caused some issues for benchmarks (e.g.
filling up space because of trash). And this option is not effective on
HDD.

Test Plan: mtr

Subscribers:

Tasks:

Tags:

Blame Revision:


alxyang (approved):
lgtm

Issue #303, opened 2016-08-16
Support covering secondary key lookups for text fields


hermanlee:
If a table has the following schema:

create table t1 {
  pk int,
  data text,
  primary key (pk),
  key (data(256)),
}

it would be nice to be able to have the data index support covering lookups for data sizes less than 256.

lth:
Basically this:
https://github.com/facebook/mysql-5.6/commit/fd944130

Issue #568, opened 2017-03-14
innodb_prefix_index_cluster_optimization silently returns wrong query results for UTF-8 and other > 1 byte character sets  


laurynas-biveinis:
./mtr --do-test=ctype --mysqld=--innodb_prefix_index_cluster_optimization=ON
will result in a variety of failures showing missing rows in result sets, for example:
main.ctype_utf8mb4_heap                  w4 [ fail ]
        Test ended at 2017-03-14 14:06:37

CURRENT_TEST: main.ctype_utf8mb4_heap
--- /home/laurynas/mysql-server/mysql-test/r/ctype_utf8mb4_heap.result	2017-03-14 14:28:34.018997655 +0300
+++ /home/laurynas/obj-fb-mysql-5.6.35-asan-debug/mysql-test/var/4/log/ctype_utf8mb4_heap.reject	2017-03-14 15:06:37.300334942 +0300
@@ -808,7 +808,6 @@
 alter table t1 engine=innodb;
 select * from t1 where b like 'foob%';
 a	b
-2	foobar
 drop table t1;
 create table t1 (
 a enum('петя','вася','анюта') character set utf8mb4 not null default 'анюта',

mysqltest: Result content mismatch

Further, a debug build will hit asserts on some but not all cases:

main.ctype_utf8mb4                       w5 [ fail ]
        Test ended at 2017-03-14 14:06:52

CURRENT_TEST: main.ctype_utf8mb4
mysqltest: At line 412: query 'select c as c_a from t1 where c='a'' failed: 2013: Lost connection to MySQL server during query
...
2017-03-14 15:06:54 7fefe5e1f700  InnoDB: Assertion failure in thread 140668330702592 in file row0sel.cc line 2730
InnoDB: Failing assertion: len * templ->mbmaxlen >= templ->mysql_col_len || (field_no == templ->icp_rec_field_no && field->prefix_len > 0)
...

Note that the actual data encoding does not have to be encoded with > 1 byte sequences, it is enough to declare column encoding for the bug to trigger.

laurynas-biveinis:
The root cause is the added check "if (record_size >= field->prefix_len)" in row_search_for_mysql. record_size is in bytes, and field->prefix_len is prefix length, in characters, times maximum possible character width in its charset, e.g. VARCHAR(10) in UTF8MB4 will be 40 here, not 10, as the code expects.

[...]

PR #639, opened 2017-06-20
Support covering secondary key lookups for varchars

pbjc:
Addresses #303. Changes:
- Adds a status var called `ROCKSDB_COVERED_SECONDARY_KEY_LOOKUPS` which is incremented when we avoid reading a record via a keyread.
- Modifies `Rdb_key_def::pack_record` to store, for secondary indexes with varchar columns, a 2-byte bitmap in the unpack info after the header. This is the "covered bitmap", where each bit corresponds to a keypart. A bit is set if that key part contains the entire column value. Bits are only used currently for varchar columns, but could later be extended to blobs/text.
- Modifies `unpack_record` to reflect `pack_record` changes.
- Adds a new data tag for unpack_info called `RDB_UNPACK_COVERED_DATA_TAG`. This is used in place of `RDB_UNPACK_DATA_TAG` to indicate a covered bitmap is present in unpack_info.
- Backwards-compatible with older formats.
- Adds `Rdb_key_def::covers_lookup()`, which returns true if all of the columns requested by a query are covered by the secondary key. This is used in `read_row_from_secondary_key` and `secondary_index_read`.

Note: This doesn't support blobs/text yet, like the example in #303, since we currently skip unpacking blobs in unpack_record.

This feature will exist but be disabled in prod until further testing.


mysql-test/suite/rocksdb/t/fast_prefix_index_fetch.test
```diff
@@ -0,0 +1,97 @@
+--source include/have_rocksdb.inc
```
laurynas-biveinis, line 5: This testcase seems to be taken from the sibling feature in InnoDB, and that one is affected by https://github.com/facebook/mysql-5.6/issues/568, which IMHO is rather serious for affected charset users. I don't know if RocksDB implementation is affected by the same - please consider taking https://github.com/percona/percona-server/blob/5.6/mysql-test/suite/innodb/include/prefix_index_only_query_check.inc, https://github.com/percona/percona-server/blob/5.6/mysql-test/suite/innodb/t/percona_fast_prefix_index_fetch.test, https://github.com/percona/percona-server/blob/5.6/mysql-test/suite/innodb/r/percona_fast_prefix_index_fetch.result, which are based on the same testcase but test more charsets and prefix lengths.

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -6327,17 +6326,22 @@ int ha_rocksdb::read_row_from_secondary_key(uchar *const buf,
...
     pk_size =
         kd.get_primary_key_tuple(table, *m_pk_descr, &rkey, m_pk_packed_tuple);
-    const rocksdb::Slice &value = m_scan_it->value();
     if (pk_size == RDB_INVALID_KEY_LEN) {
       rc = HA_ERR_ROCKSDB_CORRUPT_DATA;
     } else {
       rc = kd.unpack_record(table, buf, &rkey, &value,
                             m_verify_row_debug_checksums);
+      if (!rc) {
+        global_stats.covered_secondary_key_lookups.inc();
```
hermanlee: Is this counter used for the error case only? If so, can you name it more appropriately, like covered_sk_lookups_error_cnt?

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -6327,17 +6326,22 @@ int ha_rocksdb::read_row_from_secondary_key(uchar *const buf,
...
   uint pk_size;
 
-  if (m_keyread_only && m_lock_rows == RDB_LOCK_NONE && !has_hidden_pk(table)) {
-    /* Get the key columns and primary key value */
-    const rocksdb::Slice &rkey = m_scan_it->key();
+  /* Get the key columns and primary key value */
+  const rocksdb::Slice &rkey = m_scan_it->key();
+  const rocksdb::Slice &value = m_scan_it->value();
+
+  bool covered_lookup = m_keyread_only || kd.covers_lookup(table, &value);
```
hermanlee: Do we know how much this affects performance of selects via secondary indexes? Could we measure this using sysbench.

storage/rocksdb/rdb_datadic.cc
```diff
@@ -890,8 +941,20 @@ uint Rdb_key_def::pack_record(const TABLE *const tbl, uchar *const pack_buffer,
     unpack_len_pos = unpack_info->get_current_pos();
     // we don't know the total length yet, so write a zero
     unpack_info->write_uint16(0);
+
+    if (m_index_type == INDEX_TYPE_SECONDARY) {
+      // Reserve two bytes for the covered bitmap. This will store for each key
+      // part whether or not the column is covered. A max of 16 columns are
+      // permitted for indexes.
+      covered_bitmap_pos = unpack_info->get_current_pos();
+      unpack_info->write_uint16(0);
```
hermanlee: This changes the unpack_info format for secondary keys. It needs to be backwards compatible, but I don't see the secondary index key versioning being bumped up. Secondary keys before this patch need to continue using the old 16 byte unpack info. New secondary keys created after the patch would use the 32 byte unpack info.

storage/rocksdb/rdb_datadic.cc
```diff
@@ -780,6 +783,53 @@ int Rdb_key_def::successor(uchar *const packed_tuple, const uint &len) {
...
+      fieldnr = table->key_info[table->s->primary_key].key_part[j].fieldnr;
+    } else {
+      break;
+    }
+
+    // Checks two conditions:
+    // - The column corresponding with this key part was requested
+    // - The entire column value is contained in the SK
+    if (bitmap_is_set(table->read_set, fieldnr - 1) &&
+        bitmap_is_set(&covered_bitmap, i)) {
```
hermanlee: If the requested column is not covered, this can probably return false immediately.

storage/rocksdb/rdb_datadic.cc
```diff
@@ -780,6 +783,53 @@ int Rdb_key_def::successor(uchar *const packed_tuple, const uint &len) {
...
+
+  MY_BITMAP covered_bitmap;
+  my_bitmap_map covered_bits;
+  bitmap_init(&covered_bitmap, &covered_bits, 16, false);
+  if (unp_reader.read_uint16(&covered_bits)) {
+    return false;
+  }
+
+  uint covered_columns = 0;
+  KEY key_info = table->key_info[get_keyno()];
```
hermanlee: This calculation seems redundant when we are scanning a secondary index. Once a table is opened for reading by the handler, could we perform the calculation once to determine which fields need to have bits set in the covered_columns bitmap? On reading the unpack_info bitmap, we can immediately check to see if those bits are set, and not need to recalculate this mapping of columns being requested to bit position in the covered_columns bitmap?

storage/rocksdb/rdb_datadic.cc
```diff
@@ -925,13 +988,67 @@ uint Rdb_key_def::pack_record(const TABLE *const tbl, uchar *const pack_buffer,
...
+    if (unpack_info && m_index_type == INDEX_TYPE_SECONDARY) {
+      size_t data_length = field->data_length();
+      uint16 key_length;
+
+      // We always need the PK as a suffix, but in some cases MySQL won't add
+      // the PK suffix to a secondary key, and therefore it won't be included
+      // in key_info. If so, we need to get the key length for key parts from
+      // the PK directly.
+      DBUG_ASSERT(tbl->key_info != nullptr);
+      if (i >= tbl->key_info[get_keyno()].actual_key_parts) {
```
hermanlee: Would this code run into a problem when actual_key_parts is greater than 16? The sk + pk parts could be greater than 16. Should it ignore values greater than 16? For example:

```
CREATE TABLE `t4` (
  `a` int(11) NOT NULL,
  `b` int(11) DEFAULT NULL,
  `c` int(11) DEFAULT NULL,
  `d` int(11) DEFAULT NULL,
  `e` int(11) DEFAULT NULL,
  `f` int(11) DEFAULT NULL,
  `g` int(11) DEFAULT NULL,
  `h` int(11) DEFAULT NULL,
  `i` int(11) DEFAULT NULL,
  `j` int(11) DEFAULT NULL,
  `k` int(11) DEFAULT NULL,
  `l` int(11) DEFAULT NULL,
  `m` int(11) DEFAULT NULL,
  `n` int(11) DEFAULT NULL,
  `o` int(11) DEFAULT NULL,
  `p` int(11) DEFAULT NULL,
  `q` int(11) DEFAULT NULL,
  `r` int(11) DEFAULT NULL,
  `s` int(11) NOT NULL,
  `t` int(11) NOT NULL,
  PRIMARY KEY (`a`,`s`,`t`),
  KEY `b` (`b`,`c`,`d`,`e`,`f`,`g`,`h`,`i`,`j`,`k`,`l`,`m`,`n`,`o`,`p`,`q`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 ;
```

storage/rocksdb/rdb_datadic.cc
```diff
@@ -925,13 +988,67 @@ uint Rdb_key_def::pack_record(const TABLE *const tbl, uchar *const pack_buffer,
...
+        case MYSQL_TYPE_FLOAT:
+        case MYSQL_TYPE_NEWDECIMAL:
+        case MYSQL_TYPE_DATETIME2:
+        case MYSQL_TYPE_TIMESTAMP2:
+        case MYSQL_TYPE_TIME2:
+        case MYSQL_TYPE_YEAR:
+        case MYSQL_TYPE_NEWDATE:
+        case MYSQL_TYPE_VARCHAR:
+        case MYSQL_TYPE_STRING:
+          bitmap_set_bit(&covered_bitmap, i);
```
hermanlee: Most of these types don't need unpack_info to be unpackable. Could they be ignored if they are in secondary keys? We only need the bitmap for varchar type columns.

storage/rocksdb/rdb_datadic.cc
```diff
@@ -925,13 +988,67 @@ uint Rdb_key_def::pack_record(const TABLE *const tbl, uchar *const pack_buffer,
...
+
     // Restore field->ptr and field->null_ptr
     field->move_field(tbl->record[0] + field_offset,
                       maybe_null ? tbl->record[0] + null_offset : nullptr,
                       field->null_bit);
   }
 
   if (unpack_info) {
+    if (m_index_type == INDEX_TYPE_SECONDARY) {
+      unpack_info->write_uint16_at(covered_bitmap_pos, covered_bits);
```
hermanlee: There is an optimization below that tries not to store unpack_info if none of the columns requires it. This always adds an extra 2 bytes to unpack info, which means len != RDB_UNPACK_HEADER_SIZE, so this means we always write out unpack_info for secondary indexes, even if it's not needed (e.g. secondary index consists entire of ints).

hermanlee (changes requested):

mysql-test/suite/rocksdb/t/fast_prefix_index_fetch.test
```diff
@@ -0,0 +1,97 @@
...
+--echo # Baseline sanity check: 0, 0.
+--let $base_count = query_get_value($count_statement, Value, 1)
+select "no-op query";
+--let $count = query_get_value($count_statement, Value, 1)
+eval select $count - $base_count into @covered_lookups_count;
+select @covered_lookups_count = 0 as covered_lookups_count_matched;
+
+--echo # Eligible for optimization.
+--let $base_count = query_get_value($count_statement, Value, 1)
+select id, bigfield from t1 where bigfield = repeat('d', 31);
```
hermanlee: Should also add in force index() to ensure the query plan selected uses this index. Depending on the run, it's possible it might do a full table scan and cause the test to be inconsistent.

mysql-test/suite/rocksdb/t/fast_prefix_index_fetch.test
pbjc: Thanks for the heads up. I used your testcase to clean up and add tests for utf8. I don't think we are affected by this issue but it's good to know.

storage/rocksdb/ha_rocksdb.cc
pbjc: unpack_record returns false for success and true for failure, so this is actually for the success case.

hermanlee: Ah, okay, that sounds good then. We could also bump the count regardless of rc to indicate covered_sk lookup attempts. rc would be non-zero in the case of key/value corruption or I/O error, which should be rare. I realize the row read counters are only incremented on rc == 0. Either way works for me.

pbjc: :+1: I removed the rc check.

storage/rocksdb/rdb_datadic.cc
pbjc: Yes, we could do this. I'm having trouble seeing where to perform and store the calculation in the code. Where is the table opened (and later closed) for reading? Would it be appropriate to store this as a member variable in ha_rocksdb?

storage/rocksdb/rdb_datadic.cc
pbjc: I bumped the sk format version and added checks for m_kv_format_version everywhere the bitmap is being read. Is that sufficient, or should we also check the version when writing?

storage/rocksdb/rdb_datadic.cc
pbjc: :+1: It now stops writing to/reading from the bitmap when the number of columns written/read is 16. This should be fairly rare since bits are only written/read for varchars now.

storage/rocksdb/rdb_datadic.cc
pbjc: :+1: I updated so bits are only used for varchars, so a secondary key with n varchar columns will only use n bits in the bitmap. When there are more than 16 varchars, the rest are ignored and treated as not covered.

storage/rocksdb/rdb_datadic.cc
pbjc: Updated. If no varchars are used or none are covered, then unpack_info is not written.

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -4871,8 +4871,7 @@ void ha_rocksdb::setup_field_converters() {
       If hidden pk exists, we skip this check since the field will never be
       part of the hidden pk.
     */
-    if (!has_hidden_pk(table) &&
-        field->part_of_key.is_set(table->s->primary_key)) {
```
lth, line 4999: Do you why this was removed? It seems like part_of_key is somehow different from the the for loop below?

storage/rocksdb/rdb_datadic.cc
hermanlee: ha_rocksdb::open() is called to construct the handler to the table used for the scan. index_init() is invoked to setup_read_decoders(), so this might be the place for it. You can cache the calculation in ha_rocksdb. Just be sure it gets cleared appropriately before the structure is reused.

Also verify it works with the index_merge query plan. There are some test cases that generate index_merge query plans, but none which use varchar() keys, so it'll be good to ensure index_merge query plans work well with the caching scheme because the query plan also opens up the table for scanning and performs the initialization differently from the regular secondary index scan.

storage/rocksdb/rdb_datadic.cc
hermanlee: You will need to check the version during writing too. Any index that has the old format needs all values to be written using the old format. Otherwise, existing indexes will end up with a mix of key/values with and without the covered index bitmaps, which would break things. So the code below which creates the covered index bitmap would only need to execute on secondary_index versions >= SECONDARY_FORMAT_VERSION_UPDATE3.

storage/rocksdb/rdb_datadic.cc
```diff
@@ -780,6 +783,74 @@ int Rdb_key_def::successor(uchar *const packed_tuple, const uint &len) {
...
+
+  uint covered_columns = 0;
+  for (uint i = 0; i < m_key_parts && curr_bitmap_pos < 16; i++) {
+    if (table_has_hidden_pk(table) && i + 1 == m_key_parts) {
+      continue;
+    }
+
+    Field *const field = m_pack_info[i].get_field_in_table(table);
+    switch (field->real_type()) {
+    // If a varchar was requsted, it must be covered.
```
lth: requsted -> requested

storage/rocksdb/rdb_datadic.cc
hermanlee: It looks like your code covers this now, but let's add a test case for it. Create a table with 20 varchar(64) columns, create a primary key with 4 of the varchar columns using varchar(8), then a secondary key with the rest of the 20 columns using varchar(8). Each key column only uses the first 8 characters of varchar. Then have various rows with varchar strings of < 8 bytes, or > 8 bytes for the primary and secondary columns, and select on those columns to see if the pk is used for covered index or not (should always fail).

i.e.

```
CREATE TABLE T1 (
  P1 VARCHAR(64), P2 VARCHAR(64), P3 VARCHAR(64), P4 VARCHAR(64),
  S1 VARCHAR(64), S2 VARCHAR(64), ..., S16 VARCHAR(64),
  PRIMARY KEY (P1, P2, P3, P4),
  KEY (S1, S2, S3, ..., S16) 
) ENGINE=InnoDB;

INSERT INTO t1 values ('1', '2', '3', '4', '5', '6, '7', ..., '20');
SELECT P4, S2 FROM t1 where S1 = '5'; # Should fail to return P4 as covered because there were not enough bits to cover it.
```

mysql-test/suite/rocksdb/t/add_index_inplace.test
```diff
@@ -255,7 +255,7 @@ SET @prior_rocksdb_merge_buf_size = @@rocksdb_merge_buf_size;
 
 SET global rocksdb_strict_collation_check = off;
 SET session rocksdb_merge_combine_read_size = 566;
-SET session rocksdb_merge_buf_size = 336;
+SET session rocksdb_merge_buf_size = 340;
```
hermanlee, line 258: Why is this change needed?

storage/rocksdb/rdb_datadic.h
```diff
@@ -112,6 +112,7 @@ const char RDB_UNPACK_DATA_TAG = 0x02;
 const size_t RDB_UNPACK_DATA_LEN_SIZE = sizeof(uint16_t);
 const size_t RDB_UNPACK_HEADER_SIZE =
     sizeof(RDB_UNPACK_DATA_TAG) + RDB_UNPACK_DATA_LEN_SIZE;
+const size_t RDB_COVERED_BITMAP_SIZE = sizeof(uint16_t);
```
hermanlee: Should we have a third tag for values that have the covered_bitmap? While the index version of UPDATE3 should be sufficient information to indicate there is possible covered_bitmap for unpack data if there are any varchar columns, I think it would be safer if we used a new tag whenever we have to write out the covered_bitmap too.

Would this complicate the code unnecessarily?

So RDB_UNPACK_DATA_TAG = 2 would be used for all secondary index and primary indexes where there is not covered_bitmap. For any indexes that have a varchar key, then covered_bitmap can exist. The values for those keys should use RDB_UNPACK_COVERED_DATA_TAG = 3, which consists of 5 bytes: 1 for the tag, 2 for the size, 2 for the covered_bitmap. This way, we can consistently verify the DATA_TAG value matches our key columns types and catch problems early on.

Thoughts?

hermanlee (changes requested):

storage/rocksdb/rdb_datadic.h
hermanlee: Also, since your code updates the sk versions, could you provide test cases that verify being able to read/write both formats? You may need to set debug points (see other tests that do this) in the code to allow you to execute as if the index was of one format version or the other, despite the actual latest index format version to be used.

mysql-test/suite/rocksdb/t/add_index_inplace.test
pbjc: cc @alxyang

storage/rocksdb/rdb_datadic.cc
pbjc: Updated. In index_init, the mapping is computed for the active index. In index_end it is cleared. Rdb_key_def::covers_lookup now only needs to read the bitmap from unpack_info and check if the lookup bitmap is a subset.

storage/rocksdb/rdb_datadic.cc
pbjc: Done, caught a bug in Rdb_key_def::get_lookup_bitmap so this was a good call!

storage/rocksdb/rdb_datadic.cc
```diff
@@ -780,6 +783,88 @@ int Rdb_key_def::successor(uchar *const packed_tuple, const uint &len) {
   return changed;
 }
 
+void Rdb_key_def::get_lookup_bitmap(TABLE *const table, MY_BITMAP *map) const {
+  bitmap_init(map, nullptr, 16, false);
```
lth: You can use MAX_REF_PARTS instead of literal 16 here.

storage/rocksdb/rdb_datadic.cc
```diff
@@ -780,6 +783,88 @@ int Rdb_key_def::successor(uchar *const packed_tuple, const uint &len) {
   return changed;
 }
 
+void Rdb_key_def::get_lookup_bitmap(TABLE *const table, MY_BITMAP *map) const {
```
lth: I think you meant `const TABLE *`?

storage/rocksdb/rdb_datadic.h
```diff
@@ -409,7 +432,11 @@ class Rdb_key_def {
     //    an inefficient where data that was a multiple of 8 bytes in length
     //    had an extra 9 bytes of encoded data.
     SECONDARY_FORMAT_VERSION_UPDATE2 = 12,
-    SECONDARY_FORMAT_VERSION_LATEST = SECONDARY_FORMAT_VERSION_UPDATE2,
+    // This change includes support for covering SK lookups for varchars
+    //  - A 2-byte bitmap is added after the header to unpack_info only for
+    //    records which have covered varchar columns.
+    SECONDARY_FORMAT_VERSION_UPDATE3 = 13,
+    SECONDARY_FORMAT_VERSION_LATEST = SECONDARY_FORMAT_VERSION_UPDATE3,
```
lth: I don't know if we want to make the latest version default yet. Maybe we should wait until we at least get blob/text support?

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -4513,6 +4513,23 @@ void ha_rocksdb::setup_read_decoders() {
                         m_decoders_vect.end());
 }
 
+void ha_rocksdb::setup_lookup_bitmaps() {
+  DBUG_ASSERT(m_lookup_bitmap_vect.empty());
+
+  for (uint i = 0; i < m_tbl_def->m_key_count; i++) {
+    MY_BITMAP lookup_bitmap;
+    m_key_descr_arr[i]->get_lookup_bitmap(table, &lookup_bitmap);
```
lth: We can probably skip calling this for PK?

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -6327,17 +6343,23 @@ int ha_rocksdb::read_row_from_secondary_key(uchar *const buf,
...
   uint pk_size;
 
-  if (m_keyread_only && m_lock_rows == RDB_LOCK_NONE && !has_hidden_pk(table)) {
-    /* Get the key columns and primary key value */
-    const rocksdb::Slice &rkey = m_scan_it->key();
+  /* Get the key columns and primary key value */
+  const rocksdb::Slice &rkey = m_scan_it->key();
+  const rocksdb::Slice &value = m_scan_it->value();
+
+  DBUG_ASSERT_IMP(!m_keyread_only, !m_lookup_bitmap_vect.empty());
```
lth: It seems to me like m_lookup_bitmap_vect is never empty since we always fill it up in setup_lookup_bitmaps?

storage/rocksdb/rdb_datadic.cc
```diff
@@ -780,6 +783,88 @@ int Rdb_key_def::successor(uchar *const packed_tuple, const uint &len) {
...
+    case MYSQL_TYPE_TINY:
+    case MYSQL_TYPE_DOUBLE:
+    case MYSQL_TYPE_FLOAT:
+    case MYSQL_TYPE_NEWDECIMAL:
+    case MYSQL_TYPE_DATETIME2:
+    case MYSQL_TYPE_TIMESTAMP2:
+    case MYSQL_TYPE_TIME2:
+    case MYSQL_TYPE_YEAR:
+    case MYSQL_TYPE_NEWDATE:
+      maybe_covered_columns++;
```
lth: Shouldn't this also be guarded by bitmap_is_set read_set?

storage/rocksdb/rdb_datadic.cc
```diff
@@ -780,6 +783,88 @@ int Rdb_key_def::successor(uchar *const packed_tuple, const uint &len) {
   return changed;
 }
 
+void Rdb_key_def::get_lookup_bitmap(TABLE *const table, MY_BITMAP *map) const {
+  bitmap_init(map, nullptr, 16, false);
+  uint maybe_covered_columns = 0;
+  uint curr_bitmap_pos = 0;
+
+  for (uint i = 0; i < m_key_parts; i++) {
```
lth: I think there may be problems in your accounting if you have duplicated columns in your key since maybe_covered_columns will be incremented twice, while requested_columns is derived from read_set.

storage/rocksdb/rdb_datadic.cc
```diff
@@ -1190,6 +1319,18 @@ int Rdb_key_def::unpack_record(TABLE *const table, uchar *const buf,
     return HA_ERR_ROCKSDB_CORRUPT_DATA;
   }
 
+  // Read the covered bitmap
+  MY_BITMAP covered_bitmap;
+  my_bitmap_map covered_bits;
+  uint curr_bitmap_pos = 0;
+  bitmap_init(&covered_bitmap, &covered_bits, 16, false);
```
lth: Same here about literal 16 (and elsewhere).

storage/rocksdb/ha_rocksdb.cc
pbjc: In some testcase, field->part_of_key was set to 0 even though field was a column in the primary key. So get_storage_type wasn't getting called, which made m_maybe_unpack_info false. This field was a varchar, and since there was no unpack_info provided for unpacking (because m_maybe_unpack_info = false) it caused an assertion error.

The loop below already checks for essentially the same condition in O(m_key_count), so I removed the part_of_key check. I'm not sure how to have an O(1) check without part_of_key though.

storage/rocksdb/ha_rocksdb.cc
pbjc: ha_rocksdb::load_auto_incr_value will use secondary_index_read without a call to index_init, so the vector will be empty, but sets m_keyread_only to true before doing so. So in that specific case, it's fine that the vector is empty. I also put the assert in read_row_from_secondary_key as a sanity check, but it's probably not needed since it won't be called like that.

storage/rocksdb/rdb_datadic.cc
lth: Actually, it looks like you can't have duplicated columns in your key, so this should be okay.

mysql-test/suite/rocksdb/t/add_index_inplace.test
alxyang: To understand why this is happening we need to look at the full test case

```
SET global rocksdb_strict_collation_check = off;
SET session rocksdb_merge_combine_read_size = 566;
SET session rocksdb_merge_buf_size = 336;

show variables like '%rocksdb_bulk_load%';
CREATE TABLE t1 (a VARCHAR(80)) ENGINE=RocksDB;
INSERT INTO t1 (a) VALUES (REPEAT("a", 80));
INSERT INTO t1 (a) VALUES (REPEAT("a", 80));
INSERT INTO t1 (a) VALUES (REPEAT("a", 80));
INSERT INTO t1 (a) VALUES (REPEAT("a", 80));
ALTER TABLE t1 ADD INDEX ka(a), ALGORITHM=INPLACE;
SHOW CREATE TABLE t1;
CHECK TABLE t1;
--sorted_result
SELECT * FROM t1 FORCE INDEX(ka) WHERE a > "";
DROP TABLE t1;
```

`rocksdb_combine_read_size` is the total amount of space we have available to allocate for the heap that we use for the combine phase of our external merge sort during inplace alter.  

The relevant calculation is 
https://github.com/facebook/mysql-5.6/blob/fb-mysql-5.6.35/storage/rocksdb/rdb_index_merge.cc#L237-L241

The problem here is that there are now more sort buffers because of the bitmap stored in each record takes extra space. (Note that we should be able to avoid this. if the index is on the full key then its always covering, i don't think we should store the bitmap in these cases).

The end result is that the alter can't continue because there isn't enough space in the chunk to process the merge.  We should also return a better error to the user when this happens (currently it's only written to the error logs).  I'll open a task for it later.

storage/rocksdb/rdb_datadic.cc
```diff
@@ -780,6 +783,95 @@ int Rdb_key_def::successor(uchar *const packed_tuple, const uint &len) {
   return changed;
 }
 
+void Rdb_key_def::get_lookup_bitmap(const TABLE *table, MY_BITMAP *map) const {
+  bitmap_init(map, nullptr, MAX_REF_PARTS, false);
```
hermanlee: Should add DBUG_ASSERT(map->bitmap == nulltpr);

storage/rocksdb/rdb_datadic.h
```diff
@@ -113,6 +113,9 @@ const size_t RDB_UNPACK_DATA_LEN_SIZE = sizeof(uint16_t);
 const size_t RDB_UNPACK_HEADER_SIZE =
     sizeof(RDB_UNPACK_DATA_TAG) + RDB_UNPACK_DATA_LEN_SIZE;
 
+const char RDB_UNPACK_COVERED_DATA_TAG = 0x03;
```
hermanlee: Can you add a comment about the format here? It's a 1 tag-byte plus a two byte length field plus a two byte covered bitmap. Can you also define a constant for RDB_UNPACK_COVERED_HEADER_SIZE that is sizeof(RDB_UNPACK_COVERED_DATA_TAG) + RDB_UNPACK_COVERED_DATA_LEN_SIZE + RDB_COVERED_BITMAP_SIZE and use this constant where you are reading/writing RDB_UNPACK_COVERED_DATA_TAG? Even though there are similarities in the format, it makes the two code paths explicit for header construction and one does not depend on the other.

storage/rocksdb/rdb_datadic.h
hermanlee: I think our discussion was to make this version be debug by setting the upper bit of the enum. Is that still the plan?

storage/rocksdb/rdb_datadic.cc
```diff
@@ -940,9 +1067,17 @@ uint Rdb_key_def::pack_record(const TABLE *const tbl, uchar *const pack_buffer,
...
+    if (m_index_type == Rdb_key_def::INDEX_TYPE_SECONDARY) {
+      size_t empty_unpack_info_len = RDB_UNPACK_HEADER_SIZE;
+      if (use_covered_bitmap_format()) {
+        empty_unpack_info_len += RDB_COVERED_BITMAP_SIZE;
+      }
+      if (len == empty_unpack_info_len && covered_bits == 0) {
+        unpack_info->clear();
+      } else if (use_covered_bitmap_format()) {
+        unpack_info->write_uint16_at(covered_bitmap_pos, covered_bits);
+        unpack_info->write_uint8_at(0, RDB_UNPACK_COVERED_DATA_TAG);
```
hermanlee: I think this code which setups up the unpack info header should go up where the other data tag is written. It would be easier to reason about the code if the code that handles RDB_UNPACK_DATA_TAG is grouped together, and code that handles RDB_UNPACK_COVERED_DATA_TAG is grouped together. This way, if we added a third type, etc., this function can be more easily amended to include it.

This step of the process is mainly recording the length of the total unpack_info for secondary keys and if it's a covered index, recording the covered index bitmap along with the length.

hermanlee (changes requested):

mysql-test/suite/rocksdb/t/add_index_inplace.test
pbjc: Update: The bitmap is now used only for partially covering key parts and is not stored if there are no such key parts.

pbjc:
Benchmark results, plus the scripts I used with sysbench 0.5: https://gist.github.com/pbjc/25c70467c0208b7ce1e2d16670c39f7e

I used [20e993f](https://github.com/facebook/mysql-5.6/commit/20e993f5435391eb828cb653e953f759e2e960d1) as the baseline. Throughput could vary by up to 1 read/sec between runs, but in general the results I put up are representative. Overall there doesn't seem to be a significant difference for this particular benchmark. The disparity would be larger if there were more columns in the table. @hermanlee @alxyang

storage/rocksdb/rdb_datadic.h
```diff
@@ -757,6 +790,13 @@ class Rdb_field_packing {
   // spaces in the upack_info
   bool m_unpack_info_uses_two_bytes;
 
+  /*
+    True implies that an index-only read is always possible for this field.
+    False means an index-only read may be possible depending on the record and
+    field type.
+  */
+  bool covered;
```
hermanlee: Can you rename this to m_covered?

hermanlee (approved):

PR #642, opened 2017-06-26
add test for prefix extractor override

### LINKED ISSUE #641: Support multiple prefix bloom filter size per CF
This will be addressed in RocksDB (https://github.com/facebook/rocksdb/pull/2474). Need to make sure if it does not break in MyRocks.

alxyang:
Add accompanying test to verify prefix extractor override now works in MyRocks.  When DB is opened with new prefix extractor we expect the old files to ignore the old prefix extractor.

See https://github.com/facebook/rocksdb/commit/c7004840d2f4ad5fc1bdce042902b822492f3a0e for more details

Fixes https://github.com/facebook/mysql-5.6/issues/641.
Squash with D5307132.

### DIFF
--- a/mysql-test/suite/rocksdb/t/prefix_extractor_override.test
+++ b/mysql-test/suite/rocksdb/t/prefix_extractor_override.test
@@ -1,0 +1,96 @@
+--source include/have_rocksdb.inc
+
+--disable_warnings
+DROP TABLE IF EXISTS t1;
+--enable_warnings
+
+CREATE TABLE t1 (id1 BIGINT, id2 BIGINT, id3 BIGINT, id4 BIGINT, PRIMARY KEY (id1, id2, id3, id4) comment 'cf1') ENGINE=rocksdb collate latin1_bin;
+--disable_query_log
+let $i = 1;
+while ($i <= 100) {
+  let $insert = INSERT INTO t1 VALUES(1, $i, $i, $i);
+  eval $insert;
+  inc $i;
+}
+--enable_query_log
+set global rocksdb_force_flush_memtable_now = 1;
+
+--echo
+--echo Original Prefix Extractor:
+--echo
+--sorted_result
+SELECT * FROM information_schema.rocksdb_cf_options WHERE option_type like '%prefix_extractor%';
+
+# BF used (4+8+8+8)
+select variable_value into @u from information_schema.global_status where variable_name='rocksdb_bloom_filter_prefix_checked';
+SELECT COUNT(*) FROM t1 WHERE id1=1 AND id2=1 AND id3=1;
+select variable_value-@u from information_schema.global_status where variable_name='rocksdb_bloom_filter_prefix_checked';
+
+--exec echo "" > $MYSQLTEST_VARDIR/log/mysqld.1.err
+--let $_mysqld_option=--rocksdb_override_cf_options=cf1={prefix_extractor=capped:26};
+
+--echo
+--echo Prefix Extractor (after override_cf_options set, should not be changed):
+--echo
+--sorted_result
+SELECT * FROM information_schema.rocksdb_cf_options WHERE option_type like '%prefix_extractor%';
+
+# This should no longer crash. See https://github.com/facebook/mysql-5.6/issues/641
+--echo
+--echo Restarting with new Prefix Extractor...
+--echo
+--source include/restart_mysqld_with_option.inc
+
+--echo
+--echo Changed Prefix Extractor (after restart):
+--echo
+--sorted_result
+SELECT * FROM information_schema.rocksdb_cf_options WHERE option_type like '%prefix_extractor%';
+
+# Satisfies can_use_bloom_filter (4+8+8+8), but can't use because the old SST
+# files have old prefix extractor
+select variable_value into @u from information_schema.global_status where variable_name='rocksdb_bloom_filter_prefix_checked';
+SELECT COUNT(*) FROM t1 WHERE id1=1 AND id2=1 AND id3=1;
+select variable_value-@u from information_schema.global_status where variable_name='rocksdb_bloom_filter_prefix_checked';
+
+# Insert more data into t1, verify it uses new bloom filter
+--disable_query_log
+let $i = 101;
+while ($i <= 200) {
+  let $insert = INSERT INTO t1 VALUES(1, $i, $i, $i);
+  eval $insert;
+  inc $i;
+}
+--enable_query_log
+
+set global rocksdb_force_flush_memtable_now = 1;
+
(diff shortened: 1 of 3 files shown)


mysql-test/suite/rocksdb/t/prefix_extractor_override.test
```diff
@@ -0,0 +1,39 @@
...
+let $i = 1;
+while ($i <= 10000) {
+  let $insert = INSERT INTO t1 VALUES(1, $i, $i, $i);
+  eval $insert;
+  inc $i;
+}
+--enable_query_log
+
+# BF used (4+8+8+8)
+select variable_value into @u from information_schema.global_status where variable_name='rocksdb_bloom_filter_prefix_useful';
```
hermanlee: Don't you need to flush the memtable to ensure all of the data lives in sst files?

mysql-test/suite/rocksdb/t/prefix_extractor_override.test
```diff
@@ -0,0 +1,39 @@
...
+--exec echo "" > $MYSQLTEST_VARDIR/log/mysqld.1.err
+--let $_mysqld_option=--rocksdb_override_cf_options=cf1={prefix_extractor=capped:26};
+
+# This should no longer crash. See https://github.com/facebook/mysql-5.6/issues/641
+--source include/restart_mysqld_with_option.inc
+
+--echo
+--echo Changed Prefix Extractor:
+--echo
+SELECT * FROM information_schema.rocksdb_cf_options WHERE option_type like '%prefix_extractor%';
```
hermanlee: Can you perform this check before restarting the instance too and show cf1 with .26 extractor? Also order by name to prevent mysql from reordering the rows adhoc, causing test consistencies issue.

mysql-test/suite/rocksdb/t/prefix_extractor_override.test
```diff
@@ -0,0 +1,39 @@
...
+
+# Satisfies can_use_bloom_filter (4+8+8+8), but can't use because the old SST
+# files have old prefix extractor
+select variable_value into @u from information_schema.global_status where variable_name='rocksdb_bloom_filter_prefix_useful';
+SELECT COUNT(*) FROM t1 WHERE id1=1 AND id2=1 AND id3=1;
+select case when variable_value-@u > 0 then 'true' else 'false' end from information_schema.global_status where variable_name='rocksdb_bloom_filter_prefix_useful';
+
```
hermanlee: Maybe you can re-insert all of the data into t1 at this point, reissue the select and verify bloomfilters are now used with the newly created sst files? This shows the system can handle new data with the new bloomfilter settings correctly.

hermanlee (dismissed):

mysql-test/suite/rocksdb/t/prefix_extractor_override.test
alxyang: the other bloom filter tests don't do this, I assumed its because 10k rows is always enough to ensure that a compaction happens.  But it can't hurt, I'll add it anyways.

mysql-test/suite/rocksdb/t/prefix_extractor_override.test
```diff
@@ -0,0 +1,73 @@
...
+# Truncate data, re-insert and verify it can use bloom filter
+TRUNCATE TABLE t1;
```
hermanlee: Can you modify this to insert into a new range so that the old sst files are not dropped. Then perform a select over the new range and verify the bloomfilters are being used. Then update a couple of rows in the old range, and perform a select over both ranges to verify we can iterate through different combinations prefix extractors for different sst files.

mysql-test/suite/rocksdb/t/prefix_extractor_override.test
```diff
@@ -0,0 +1,101 @@
+--source include/have_rocksdb.inc
+
+--disable_warnings
+DROP TABLE IF EXISTS t1;
```
hermanlee, line 4: Seems like this pattern is not needed anymore, so this can probably be dropped.

mysql-test/suite/rocksdb/t/prefix_extractor_override.test
```diff
@@ -0,0 +1,101 @@
...
+--enable_warnings
+
+
+SET @@global.rocksdb_update_cf_options = 'default={disable_auto_compactions=true;};';
```
hermanlee: I think you need to unset this value at the end of the test to prevent mtr from complaining about global variables being changed. Or store the original value and then restore it at the end.

mysql-test/suite/rocksdb/t/prefix_extractor_override.test
```diff
@@ -0,0 +1,101 @@
...
+set global rocksdb_force_flush_memtable_now = 1;
+
+# BF used w/ new prefix extractor (4+8+8+8) (still increments once because it
+# needs to check the new SST file, but doesnt increment for SST file with old
+# extractor)
+select variable_value into @u from information_schema.global_status where variable_name='rocksdb_bloom_filter_prefix_checked';
+SELECT COUNT(*) FROM t1 WHERE id1=1 AND id2=1 AND id3=1;
+select variable_value-@u from information_schema.global_status where variable_name='rocksdb_bloom_filter_prefix_checked';
+
+# should have 2 sst files, one with old prefix extractor and one with new
```
hermanlee, line 75: Depending on a fixed number of sst files seems prone to test flakiness, but it's probably fine for now and can be fixed later. I think an alternative might be to keep relative counts, or ensure the value is > 0 or > than previously recorded value, but that would be more complicated. You could record the names of the sst files in a new table, restart the server, insert more data, trigger compaction, figure out the list of new sst files created by joining the stored data with the latest list of sst files. This list would be the number of new sst files created, and give you the expected bloomfilter count increase.

hermanlee (approved):

mysql-test/suite/rocksdb/t/prefix_extractor_override.test
laurynas-biveinis: This is required in general but not in this particular case, because the server is restarted further in the testcase

alxyang: it is explicitly unset later on in the test case. But i'm going to refactor the test to just include the disable in an opt file.

mysql-test/suite/rocksdb/t/prefix_extractor_override.test
alxyang: I think the number of sst files *should* be consistent, because we disable auto compaction.  I'll land this for now, if we see that the test is flaky for any reason then I'll come back and implement this approach. :)

PR #644, opened 2017-06-29
Snapshot conflicts debug message

armishra:
added debug message to differentiate between deadlock and snapshot conflict in error message


alxyang:
hm, do we need to update the submodule here?

alxyang (changes requested):
i think we don't need to update the submodule here

PR #649, opened 2017-07-07
reset detailed error string for deadlock error

armishra:
Changes:
* reset m_detailed_error to empty string
* added tests to verify snapshot conflicts message disappears on consecutive deadlocks


storage/rocksdb/ha_rocksdb.cc
```diff
@@ -1695,6 +1695,7 @@ class Rdb_transaction {
     if (s.IsDeadlock()) {
       my_core::thd_mark_transaction_to_rollback(thd,
                                                 false /* just statement */);
+      m_detailed_error = String("", system_charset_info);
```
lth: Might be easier to have

`m_detailed_error = String();`

lth (approved):

Issue #299, opened 2016-08-08
table_stats for MyRocks


mdcallag:
Can we update table_stats for MyRocks? The compression counters are innodb-only. But the IO metrics can be reused by MyRocks. It might also be nice to track deadlock and lock-wait timeout per table.

```
                    TABLE_SCHEMA: tpcc  
                      TABLE_NAME: orders  
                    TABLE_ENGINE: ROCKSDB  
                   ROWS_INSERTED: 42417447  
                    ROWS_UPDATED: 41972287
                    ROWS_DELETED: 0  
                       ROWS_READ: 92368058  
                  ROWS_REQUESTED: 92368059  
            COMPRESSED_PAGE_SIZE: 0  
                COMPRESS_PADDING: 0  
                    COMPRESS_OPS: 0
                 COMPRESS_OPS_OK: 0  
            COMPRESS_PRIMARY_OPS: 0  
         COMPRESS_PRIMARY_OPS_OK: 0  
                  COMPRESS_USECS: 0  
               COMPRESS_OK_USECS: 0  
          COMPRESS_PRIMARY_USECS: 0
       COMPRESS_PRIMARY_OK_USECS: 0  
                  UNCOMPRESS_OPS: 0  
                UNCOMPRESS_USECS: 0  
                ROWS_INDEX_FIRST: 92368058  
                 ROWS_INDEX_NEXT: 0  
                   IO_READ_BYTES: 0
                IO_READ_REQUESTS: 0  
               IO_READ_SVC_USECS: 0  
           IO_READ_SVC_USECS_MAX: 0  
              IO_READ_WAIT_USECS: 0  
          IO_READ_WAIT_USECS_MAX: 0  
                IO_READ_SLOW_IOS: 0
                  IO_WRITE_BYTES: 0  
               IO_WRITE_REQUESTS: 0  
              IO_WRITE_SVC_USECS: 0  
          IO_WRITE_SVC_USECS_MAX: 0  
             IO_WRITE_WAIT_USECS: 0  
         IO_WRITE_WAIT_USECS_MAX: 0
               IO_WRITE_SLOW_IOS: 0  
              IO_READ_BYTES_BLOB: 0  
           IO_READ_REQUESTS_BLOB: 0  
          IO_READ_SVC_USECS_BLOB: 0  
      IO_READ_SVC_USECS_MAX_BLOB: 0  
         IO_READ_WAIT_USECS_BLOB: 0
[...]

PR #651, opened 2017-07-11
added lock_wait_timeout counter stat to table_statistics

armishra:
Addresses #299. Changes:
* passed in the table_handler to set_status_error to count the number of lock_wait_timeouts
* added simple test case to verify that consecutive queries return the same value for lock_wait_timeouts


mysql-test/suite/rocksdb/t/lock_wait_timeout_stats.test
```diff
@@ -0,0 +1,26 @@
...
+
+--source include/count_sessions.inc
+--connect (con1,localhost,root,,)
+--connection con1
+set @@rocksdb_lock_wait_timeout=1;
+begin;
+--error ER_LOCK_WAIT_TIMEOUT
+insert into t values(0);
+select INNODB_ROW_LOCK_WAIT_TIMEOUTS from information_schema.table_statistics where table_engine="ROCKSDB" and table_name="t";
+select INNODB_ROW_LOCK_WAIT_TIMEOUTS from information_schema.table_statistics where table_engine="ROCKSDB" and table_name="t";
```
alxyang: Now that we are incrementing properly, let's add a few more queries that cause timeouts from different connections and verify the stats go up properly, rather just query twice in a row where the stats don't change like this.

alxyang (commented):
mostly lgtm.  I'll wait for @lth to approve.

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -3451,6 +3454,8 @@ static void rocksdb_update_table_stats(
 
     io_perf_read.bytes = table_handler->m_io_perf_read.bytes.load();
     io_perf_read.requests = table_handler->m_io_perf_read.requests.load();
+    lock_wait_timeout_stats = table_handler->m_lock_wait_timeout_counter.load();
+    table_handler->m_lock_wait_timeout_counter.clear();
```
lth: The only concern is whether this will work well if run concurrently. It seems like there's a chance that we double count the number of lock wait timeouts?

alxyang: good point. Looks like you may use https://en.wikipedia.org/wiki/Compare-and-swap perhaps?

PR #653, opened 2017-07-12
added lock_wait_timeout and deadlock stats to rocksdb

armishra:
Addresses #299. Changes:
* Fixed InnoDB table_stats bug that increments weirdly when consecutively querying for lock_waits
* added rocksdb stat to count lock_wait_timeouts
* added test case that queries this stat in various ways.


mysql-test/suite/rocksdb/t/lock_wait_timeout_stats.test
```diff
@@ -0,0 +1,37 @@
+--source include/have_rocksdb.inc
+
+--enable_connect_log
```
lth: I think we can take this out.

mysql-test/suite/rocksdb/t/lock_wait_timeout_stats.test
```diff
@@ -0,0 +1,37 @@
...
+select INNODB_ROW_LOCK_WAIT_TIMEOUTS from information_schema.table_statistics where table_name="t";
+
+--connection con2
+--error ER_LOCK_WAIT_TIMEOUT
+insert into t values(0);
+select INNODB_ROW_LOCK_WAIT_TIMEOUTS from information_schema.table_statistics where table_name="t";
+select INNODB_ROW_LOCK_WAIT_TIMEOUTS from information_schema.table_statistics where table_name="t";
+
+--disconnect con1
+--disconnect con2
```
lth: It seems like you shouldn't run disconnect on the connection you are currently on (from the mysqltest docs).

mysql-test/suite/rocksdb/t/lock_wait_timeout_stats.test
```diff
@@ -0,0 +1,34 @@
+create table t (a int not null primary key) engine=rocksdb;
```
alxyang: nit: the not null is redundant here since it PRIMARY KEY always have the not null constraint.

mysql-test/suite/rocksdb/t/deadlock_stats.test
```diff
@@ -0,0 +1,43 @@
+CREATE TABLE t1 (
+                   id int primary key, 
+                     value int
+                ) engine=rocksdb;
```
alxyang: for this stat it seems that we have added support for collecting the stats in both `innodb` and `rocksdb`.  But I don't see the tests that verify the `innodb` stats collection works properly.

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -3477,8 +3486,8 @@ static void rocksdb_update_table_stats(
     my_core::filename_to_tablename(tablename.c_str(), tablename_sys,
                                    sizeof(tablename_sys));
     (*cb)(dbname_sys, tablename_sys, is_partition, &io_perf_read, &io_perf,
-          &io_perf, &io_perf, &io_perf, &page_stats, &comp_stats, 0, 0,
-          rocksdb_hton_name);
+          &io_perf, &io_perf, &io_perf, &page_stats, &comp_stats, 0,
+          lock_wait_timeout_stats, deadlock_stats, rocksdb_hton_name);
```
alxyang, line 3490: it seems to me that if we are going to add a new `deadlock_stats` field in the table statistics, we may as well also add a corresponding `rocksdb_row_lock_waits` and `rocksdb_row_lock_wait_timeout`.   I feel that users may be confused when they see `innodb_row_lock_wait_timeout` increasing when using MyRocks and not InnoDB.  Or we could consider renaming the variable to something more general. e.g. (row_lock_waits/row_lock_wait_timeout). Any reason we haven't taken this approach?  cc @lth

alxyang (commented):
see comments

storage/rocksdb/ha_rocksdb.cc
lth: I think renaming would be a good approach.

sql/handler.h
```diff
@@ -1003,6 +1003,7 @@ struct handlerton
                                         comp_stats_t* comp_stats,
                                         int n_lock_wait,
                                         int n_lock_wait_timeout,
+                                        int n_deadlock,
```
lth: Why don't we rename these as well for consistency (and all similar instances).

mysql-test/suite/innodb/t/deadlock_stats.test
```diff
@@ -0,0 +1,10 @@
+let $engine=innodb;
+let $engine_start_txn=start transaction;
```
lth: I'm pretty sure you can use "begin" for both innodb/rocksdb.

mysql-test/include/deadlock_stats.inc
```diff
@@ -0,0 +1,41 @@
...
+eval set global $raw_engine_lwt = 100000;
+eval set global $raw_engine_dld = ON;
+
+connect (con1,localhost,root,,);
+let $con1= `SELECT CONNECTION_ID()`;
+
+connect (con2,localhost,root,,);
+let $con2= `SELECT CONNECTION_ID()`;
+
+connect (con3,localhost,root,,);
```
lth: I don't think you need con3.

mysql-test/suite/innodb/t/deadlock_stats.test
```diff
@@ -0,0 +1,10 @@
...
+let $prior_engine_lwt=@prior_innodb_lock_wait_timeout;
```
lth: I think the "prior" variables aren't strictly needed for correctness. For simplicity, why don't we just use "prior_lock_wait_timeout" as variable name?

mysql-test/include/deadlock_stats.inc
```diff
@@ -0,0 +1,41 @@
+eval set $prior_engine_lwt = $engine_lwt;
+eval set $prior_engine_dld= $engine_dld;
+eval set global $raw_engine_lwt = 100000;
+eval set global $raw_engine_dld = ON;
+
+connect (con1,localhost,root,,);
```
lth: Since we're adding `--source include/count_sessions.inc` in the other tests, might as well add them here too.

sql/sql_base.h
```diff
@@ -294,6 +294,7 @@ typedef void (*table_stats_cb)(const char *db, const char *table,
 			       comp_stats_t* comp_stats,
 			       int n_lock_wait,
 			       int n_lock_wait_timeout,
+			       int n_deadlock,
```
lth: Rename here too.

storage/innobase/fil/fil0fil.cc
```diff
@@ -7413,6 +7427,28 @@ fil_change_lock_wait_timeout_count(
 }
 
 /*************************************************************************
+Changes count of deadlocks on a row for this space. Will lock/unlock
+fil_system->mutex */
+void
+fil_change_deadlock_count(
```
lth: and here.

mysql-test/include/simple_deadlock.inc
```diff
@@ -0,0 +1,31 @@
...
+begin;
+select * from t where i=2 for update;
+
+connection con1;
+--send select * from t where i=2 for update
+
+connection con2;
+if ($engine == "rocksdb"){
+  let $wait_condition=
+  `select concat('select count(*) = 1 from information_schema.rocksdb_trx
```
lth: Doesn't look like we need concat in either cases.

mysql-test/suite/innodb/t/deadlock_stats.test
```diff
@@ -0,0 +1,7 @@
+let $engine=innodb;
+let $engine_lwt=@@innodb_lock_wait_timeout;
```
lth: We don't need the @@ versions.

mysql-test/suite/innodb/t/deadlock_stats.test
```diff
@@ -0,0 +1,7 @@
...
+let $engine_dld=@@innodb_deadlock_detect;
+let $raw_engine_lwt=innodb_lock_wait_timeout;
```
lth: If you want, we can see if we can get rid of these two variables as well.

lth (approved):

PR #662, opened 2017-07-20
Support TTL for tables with secondary keys via compaction filter

pbjc:
* Adds index flags support for secondary indexes.
* Stores an 8-bit TTL timestamp in secondary indexes with TTL enabled. This value always mirrors the TTL timestamp in the corresponding PK record. This is implemented via an index flag field and is considered distinct from `unpack_info`.
* For testing, `rocksdb.ttl_primary` is used as a base. Added secondary keys to all tables and performed compaction with `rocksdb_debug_ignore_pk_ttl` on. Also added some tests for creating indexes in place.

Read filtering will be supported in a future diff.

### DIFF
--- a/mysql-test/suite/rocksdb/t/ttl_secondary.test
+++ b/mysql-test/suite/rocksdb/t/ttl_secondary.test
@@ -1,7 +1,7 @@
 --source include/have_debug.inc
 --source include/have_rocksdb.inc
 
-# Basic TTL test
+# Basic TTL test, pk ignored, no sk
 CREATE TABLE t1 (
 `a` binary(8) NOT NULL,
 `b` varbinary(64) NOT NULL,
@@ -17,11 +17,41 @@ INSERT INTO t1 values ('a', 'b', 'c', UNIX_TIMESTAMP(), 'd');
 INSERT INTO t1 values ('d', 'e', 'f', UNIX_TIMESTAMP(), 'g');
 set global rocksdb_debug_ttl_rec_ts = 0;
 SELECT COUNT(*) FROM t1;
+
+set global rocksdb_debug_ttl_ignore_pk = 1;
 set global rocksdb_force_flush_memtable_now=1;
 set global rocksdb_compact_cf='default';
+set global rocksdb_debug_ttl_ignore_pk = 0;
+
+# no rows should be filtered
+SELECT COUNT(*) FROM t1;
+DROP TABLE t1;
+
+# Basic TTL test
+CREATE TABLE t1 (
+`a` binary(8) NOT NULL,
+`b` varbinary(64) NOT NULL,
+`c` varbinary(256) NOT NULL,
+`ts` bigint(20) UNSIGNED NOT NULL,
+`value` mediumblob NOT NULL,
+PRIMARY KEY (`b`,`a`,`c`),
+KEY kb (`b`)
+) ENGINE=ROCKSDB DEFAULT CHARSET=latin1
+COMMENT='ttl_duration=1;ttl_col=ts;';
+
+set global rocksdb_debug_ttl_rec_ts = -100;
+INSERT INTO t1 values ('a', 'b', 'c', UNIX_TIMESTAMP(), 'd');
+INSERT INTO t1 values ('d', 'e', 'f', UNIX_TIMESTAMP(), 'g');
+set global rocksdb_debug_ttl_rec_ts = 0;
+SELECT COUNT(*) FROM t1 FORCE INDEX (kb);
+
+set global rocksdb_debug_ttl_ignore_pk = 1;
+set global rocksdb_force_flush_memtable_now=1;
+set global rocksdb_compact_cf='default';
+set global rocksdb_debug_ttl_ignore_pk = 0;
 
 # should have filtered the rows out since ttl is passed in compaction filter
-SELECT COUNT(*) FROM t1;
+SELECT COUNT(*) FROM t1 FORCE INDEX (kb);
 DROP TABLE t1;
 
 # column before TTL in value
@@ -30,7 +60,8 @@ CREATE TABLE t1 (
   b int NOT NULL,
   ts bigint(20) UNSIGNED NOT NULL,
   c int NOT NULL,
-  PRIMARY KEY (a)
+  PRIMARY KEY (a),
+  KEY kb (b)
 ) ENGINE=rocksdb
 COMMENT='ttl_duration=1;ttl_col=ts;';
 
@@ -38,13 +69,15 @@ set global rocksdb_debug_ttl_rec_ts = -100;
 INSERT INTO t1 values (1, 3, UNIX_TIMESTAMP(), 5);
 INSERT INTO t1 values (2, 4, UNIX_TIMESTAMP(), 6);
 set global rocksdb_debug_ttl_rec_ts = 0;
-SELECT COUNT(*) FROM t1;
+SELECT COUNT(*) FROM t1 FORCE INDEX (kb);
 
+set global rocksdb_debug_ttl_ignore_pk = 1;
 set global rocksdb_force_flush_memtable_now=1;
 set global rocksdb_compact_cf='default';
+set global rocksdb_debug_ttl_ignore_pk = 0;
 
 # should have filtered the rows out since ttl is passed in compaction filter
-SELECT COUNT(*) FROM t1;
+SELECT COUNT(*) FROM t1 FORCE INDEX (kb);
 DROP TABLE t1;
 
 # multi-part PK w/ TTL
@@ -53,7 +86,8 @@ CREATE TABLE t1 (
   b int NOT NULL,
   c int NOT NULL,
   ts bigint(20) UNSIGNED NOT NULL,
-  PRIMARY KEY (a,c)
+  PRIMARY KEY (a,c),
+  KEY kb (b)
 ) ENGINE=rocksdb
 COMMENT='ttl_duration=1;ttl_col=ts;';
 
@@ -61,36 +95,15 @@ set global rocksdb_debug_ttl_rec_ts = -100;
 INSERT INTO t1 values (1, 3, 5, UNIX_TIMESTAMP());
 INSERT INTO t1 values (2, 4, 6, UNIX_TIMESTAMP());
 set global rocksdb_debug_ttl_rec_ts = 0;
-SELECT COUNT(*) FROM t1;
+SELECT COUNT(*) FROM t1 FORCE INDEX (kb);
 
+set global rocksdb_debug_ttl_ignore_pk=1;
 set global rocksdb_force_flush_memtable_now=1;
 set global rocksdb_compact_cf='default';
+set global rocksdb_debug_ttl_ignore_pk=0;
 
 # should have filtered the rows out since ttl is passed in compaction filter
-SELECT COUNT(*) FROM t1;
-DROP TABLE t1;
-
-# multi-part PK w/ TTL
-CREATE TABLE t1 (
-  a bigint(20) NOT NULL,
-  b int NOT NULL,
-  c int NOT NULL,
-  ts bigint(20) UNSIGNED NOT NULL,
-  PRIMARY KEY (a,c)
-) ENGINE=rocksdb
-COMMENT='ttl_duration=1;ttl_col=ts;';
-
-set global rocksdb_debug_ttl_rec_ts = -100;
-INSERT INTO t1 values (1, 3, 5, UNIX_TIMESTAMP());
-INSERT INTO t1 values (2, 4, 6, UNIX_TIMESTAMP());
-set global rocksdb_debug_ttl_rec_ts = 0;
-SELECT COUNT(*) FROM t1;
-
-set global rocksdb_force_flush_memtable_now=1;
-set global rocksdb_compact_cf='default';
-
-# should have filtered the rows out since ttl is passed in compaction filter
-SELECT COUNT(*) FROM t1;
+SELECT COUNT(*) FROM t1 FORCE INDEX (kb);
 DROP TABLE t1;
 
 # nullable column(s) before TTL
@@ -99,7 +112,8 @@ CREATE TABLE t1 (
   b int,
   c int,
   ts bigint(20) UNSIGNED NOT NULL,
-  PRIMARY KEY (a)
+  PRIMARY KEY (a),
+  KEY kbc (b, c)
 ) ENGINE=rocksdb
 COMMENT='ttl_duration=1;ttl_col=ts;';
 
@@ -107,13 +121,15 @@ set global rocksdb_debug_ttl_rec_ts = -100;
 INSERT INTO t1 values (1, NULL, NULL, UNIX_TIMESTAMP());
 INSERT INTO t1 values (2, NULL, NULL, UNIX_TIMESTAMP());
 set global rocksdb_debug_ttl_rec_ts = 0;
-SELECT COUNT(*) FROM t1;
+SELECT COUNT(*) FROM t1 FORCE INDEX (kb);
 
+set global rocksdb_debug_ttl_ignore_pk=1;
 set global rocksdb_force_flush_memtable_now=1;
 set global rocksdb_compact_cf='default';
+set global rocksdb_debug_ttl_ignore_pk=0;
 
 # should have filtered the rows out since ttl is passed in compaction filter
-SELECT COUNT(*) FROM t1;
+SELECT COUNT(*) FROM t1 FORCE INDEX (kb);
 DROP TABLE t1;
 
 # variable len columns + null column(s) before TTL
@@ -123,7 +139,8 @@ CREATE TABLE t1 (
 `c` varbinary(256),
 `ts` bigint(20) UNSIGNED NOT NULL,
 `value` mediumblob NOT NULL,
-PRIMARY KEY (`a`)
+PRIMARY KEY (`a`),
+KEY kbc (`b`, `c`)
 ) ENGINE=ROCKSDB DEFAULT CHARSET=latin1
 COMMENT='ttl_duration=1;ttl_col=ts;';
 
@@ -131,13 +148,15 @@ set global rocksdb_debug_ttl_rec_ts = -100;
 INSERT INTO t1 values ('a', NULL, 'bc', UNIX_TIMESTAMP(), 'd');
 INSERT INTO t1 values ('d', 'efghijk', NULL, UNIX_TIMESTAMP(), 'l');
 set global rocksdb_debug_ttl_rec_ts = 0;
-SELECT COUNT(*) FROM t1;
+SELECT COUNT(*) FROM t1 FORCE INDEX (kb);
 
+set global rocksdb_debug_ttl_ignore_pk=1;
 set global rocksdb_force_flush_memtable_now=1;
 set global rocksdb_compact_cf='default';
+set global rocksdb_debug_ttl_ignore_pk=0;
 
 # should have filtered the rows out since ttl is passed in compaction filter
-SELECT COUNT(*) FROM t1;
+SELECT COUNT(*) FROM t1 FORCE INDEX (kb);
 DROP TABLE t1;
 
 # TTL implicitly generated (no ttl column)
@@ -145,7 +164,8 @@ CREATE TABLE t1 (
   a bigint(20) NOT NULL,
   b int NOT NULL,
   c int NOT NULL,
-  PRIMARY KEY (a)
+  PRIMARY KEY (a),
+  KEY kb (b)
 ) ENGINE=rocksdb
 COMMENT='ttl_duration=1;';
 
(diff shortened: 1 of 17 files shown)


storage/rocksdb/rdb_datadic.cc
```diff
@@ -3179,6 +3181,17 @@ uint32 Rdb_key_def::calculate_index_flag_offset(uint32 index_flags,
   return offset;
 }
 
+uint32 Rdb_key_def::calculate_total_index_flags_length(uint32 index_flags) {
```
alxyang: perhaps we should just reuse `calculate_index_flag_offset`, but pass in the `MAX_FLAG`.  This can get us the offset to where the actual record starts.  

```
  enum INDEX_FLAG {
    TTL_FLAG = 1 << 0,

    // MAX_FLAG marks where the actual record starts
    // This flag always needs to be set to the last index flag enum.
    MAX_FLAG = TTL_FLAG << 1,
  };
```

storage/rocksdb/ha_rocksdb.h
```diff
@@ -494,6 +494,12 @@ class ha_rocksdb : public my_core::handler {
     Pointer to the original TTL timestamp value (8 bytes) during UPDATE.
   */
   char m_ttl_bytes[ROCKSDB_SIZEOF_TTL_RECORD];
+  /*
+    The TTL timestamp value can change if the explicit TTL column is
+    updated. If we detect this when updating the PK, we indicate it here so
+    we know we must always update any SK's.
```
alxyang, line 511: This is an important perf impact and I think we need to remember to document this later on in a public wiki (in fact we should start writing a wiki for TTL).  This means MySQL is forced to update all SKs even if none of the actual SK fields have changed.

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -10414,15 +10467,33 @@ int ha_rocksdb::inplace_populate_sk(
       }
 
       /* Create new secondary index entry */
+
+      String value_buffer;
```
alxyang: hmm, not sure how I feel about this.  I think it makes the code change simpler, but now we are making a new string for every secondary index entry that's getting added and appending (allocing memory) to it.  

Have we considered, passing the info we need into`pack_record` and attempting to build the record in-place via the `Rdb_string_writer`?

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -8317,6 +8368,9 @@ int ha_rocksdb::update_indexes(const struct update_row_info &row_info,
                                const bool &pk_changed) {
   int rc;
 
+  // If this table has a non-hidden PK, verify it will be updated first for TTL
+  DBUG_ASSERT_IMP(!has_hidden_pk(table), table->s->primary_key == 0);
```
alxyang: while I think this *should* always be true, I still think it's better to explicitly update the pk first even if it makes the code a little less clean.  `DBUG_ASSERT_IMP` are ignored in production so if in some weird scenario it's not true it's going to cause problems (it'll put the wrong ttl into secondary keys? so I don't think it's worth risking data corruption).

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -8301,11 +8350,13 @@ int ha_rocksdb::update_sk(const TABLE *const table_arg, const Rdb_key_def &kd,
                                                          old_key_slice);
   }
 
+  // Write unpack_info
+  value_buffer.append(reinterpret_cast<char *>(m_sk_tails.ptr()),
```
alxyang: see my comment below about trying to push the work here into `pack_record` if possible

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -4574,6 +4584,17 @@ int ha_rocksdb::convert_record_to_storage_format(
   *packed_rec =
       rocksdb::Slice(m_storage_record.ptr(), m_storage_record.length());
 
+  // We've obtained the TTL timestamp for this record. Now store it in
+  // m_ttl_bytes if it wasn't already so we can update secondary indexes with
+  // this value as well.
+  if (has_ttl) {
+    if (!row_info.old_pk_slice.empty()) {
+      m_ttl_bytes_updated = memcmp(m_ttl_bytes, m_storage_record.ptr(),
+                                   ROCKSDB_SIZEOF_TTL_RECORD);
```
alxyang: i think you can optimize further here by also checking if there is a `ttl_column`.  Because without a ttl column, you are guaranteed that the timestamp remains the same.  So now you can skip this memcmp.

alxyang (changes requested):
Here's a first round or review comments.  Overall, this looks really solid already! Well done!

storage/rocksdb/ha_rocksdb.cc
pbjc: `pack_record` will sometimes clear the passed `Rdb_string_writer` if `unpack_info` only contains the header. This makes things tricky since `Rdb_string_writer` doesn't provide any way to clear only a part of the data. One way to avoid using a buffer is to write the index flags, call `pack_record`, and write the index flags again if the writer was cleared by `pack_record`. Thoughts?

mysql-test/suite/rocksdb/t/ttl_secondary.test
```diff
@@ -0,0 +1,780 @@
...
+set global rocksdb_compact_cf='default';
+set global rocksdb_debug_ttl_snapshot_ts = 0;
+set global rocksdb_debug_ignore_pk_ttl=0;
+--sorted_result
+SELECT a FROM t1 FORCE INDEX (kb);
+
+DROP TABLE t1;
+
+# TTL field with nullable ttl column (should fail)
+--error 1948
```
laurynas-biveinis: Can you use an ER_ mnemonic here?

mysql-test/suite/rocksdb/t/ttl_secondary.test
```diff
@@ -0,0 +1,780 @@
...
+  b int NOT NULL,
+  c int NOT NULL,
+  ts bigint(20),
+  PRIMARY KEY (a,c),
+  KEY (b)
+) ENGINE=rocksdb
+COMMENT='ttl_duration=1;ttl_col=ts;';
+
+# TTL field with non 8-bit integer column (should fail)
+--error 1948
```
laurynas-biveinis: Likewise

mysql-test/suite/rocksdb/t/ttl_secondary.test
```diff
@@ -0,0 +1,780 @@
...
+  b int NOT NULL,
+  c int NOT NULL,
+  ts int,
+  PRIMARY KEY (a,c),
+  KEY (b)
+) ENGINE=rocksdb
+COMMENT='ttl_duration=1;ttl_col=ts;';
+
+# TTL duration as some random garbage value
+--error 1949
```
laurynas-biveinis: Likewise

mysql-test/suite/rocksdb/t/ttl_secondary.test
```diff
@@ -0,0 +1,780 @@
...
+  a bigint(20) UNSIGNED NOT NULL,
+  b int NOT NULL,
+  c int NOT NULL,
+  PRIMARY KEY (a,c),
+  KEY (b)
+) ENGINE=rocksdb
+COMMENT='ttl_duration=abc;';
+
+# TTL col is some column outside of the table
+--error 1948
```
laurynas-biveinis: Likewise

mysql-test/suite/rocksdb/t/ttl_secondary.test
```diff
@@ -0,0 +1,780 @@
...
+  a bigint(20) UNSIGNED NOT NULL,
+  b int NOT NULL,
+  c int NOT NULL,
+  PRIMARY KEY (a,c),
+  KEY (b)
+) ENGINE=rocksdb
+COMMENT='ttl_duration=1;ttl_col=abc;';
+
+# TTL col must have accompanying duration
+--error 1948
```
laurynas-biveinis: Likewise

mysql-test/suite/rocksdb/t/ttl_secondary.test
```diff
@@ -0,0 +1,780 @@
...
+
+# Tables with hidden PK disabled
+CREATE TABLE t1 (
+  a INT PRIMARY KEY,
+  b INT,
+  KEY (b)
+) ENGINE=rocksdb
+COMMENT='ttl_duration=100;';
+
+--error 1947
```
laurynas-biveinis: Likewise

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -11104,6 +11174,7 @@ int rdb_dbug_set_ttl_snapshot_ts() { return rocksdb_debug_ttl_snapshot_ts; }
 int rdb_dbug_set_ttl_read_filter_ts() {
   return rocksdb_debug_ttl_read_filter_ts;
 }
+bool rdb_dbug_set_ignore_pk_ttl() { return rocksdb_debug_ignore_pk_ttl; }
```
laurynas-biveinis: const method?

storage/rocksdb/ha_rocksdb.cc
alxyang: Hmm....looking at the code it seems that `Rdb_string_writer` could be modified to truncate data.  `Rdb_string_writer` is essentially just a wrapper around `std::vector<uchar>`.  So I think we can call `resize()` since we know the length of `index_flags` forsure.  

This way we can remove just the header but keep `index_flags`.  This may be another way of approaching it.

storage/rocksdb/ha_rocksdb.cc
pbjc: Not a member function, so no need for const :+1:

laurynas-biveinis: Sorry, shouldn't review things before the 1st coffee :)

storage/rocksdb/ha_rocksdb.cc
pbjc: I added `Rdb_string_writer::truncate` so we no longer need another buffer :+1:

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -8238,16 +8271,39 @@ int ha_rocksdb::update_sk(const TABLE *const table_arg, const Rdb_key_def &kd,
...
+
+  // Insert TTL timestamp
+  if (Rdb_key_def::has_index_flag(kd.m_index_flags_bitmap,
+                                  Rdb_key_def::TTL_FLAG)) {
+    uint32 ttl_offset = Rdb_key_def::calculate_index_flag_offset(
+        kd.m_index_flags_bitmap, Rdb_key_def::TTL_FLAG);
+    memcpy(m_sk_tails.ptr() + ttl_offset, m_ttl_bytes,
+           ROCKSDB_SIZEOF_TTL_RECORD);
+  }
+
```
vasilep: I would suggest moving this code into a helper function.

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -10397,6 +10464,20 @@ int ha_rocksdb::inplace_populate_sk(
...
+      /* Insert TTL timestamp */
+      if (Rdb_key_def::has_index_flag(index->m_index_flags_bitmap,
+                                      Rdb_key_def::TTL_FLAG)) {
+        uint32 ttl_offset = Rdb_key_def::calculate_index_flag_offset(
+            index->m_index_flags_bitmap, Rdb_key_def::TTL_FLAG);
+        memcpy(m_sk_tails.ptr() + ttl_offset, m_ttl_bytes,
+               ROCKSDB_SIZEOF_TTL_RECORD);
+      }
+
```
vasilep: I would suggest moving this code into a helper function. Same helper can be called from update_sk().

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -10397,6 +10464,20 @@ int ha_rocksdb::inplace_populate_sk(
...
+      if (index->m_total_index_flags_length > 0) {
+        m_sk_tails.allocate(0, index->m_total_index_flags_length);
+      }
+
+      /* Insert TTL timestamp */
+      if (Rdb_key_def::has_index_flag(index->m_index_flags_bitmap,
+                                      Rdb_key_def::TTL_FLAG)) {
+        uint32 ttl_offset = Rdb_key_def::calculate_index_flag_offset(
+            index->m_index_flags_bitmap, Rdb_key_def::TTL_FLAG);
+        memcpy(m_sk_tails.ptr() + ttl_offset, m_ttl_bytes,
```
vasilep: I would also add an assertion that we are not writing beyond the buffer's end.

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -4553,6 +4561,19 @@ int ha_rocksdb::convert_record_to_storage_format(
...
+  // m_ttl_bytes if it wasn't already so we can update secondary indexes with
+  // this value as well.
+  if (has_ttl) {
+    if (!row_info.old_pk_slice.empty() && has_ttl_column) {
+      m_ttl_bytes_updated = memcmp(m_ttl_bytes, m_storage_record.ptr(),
+                                   ROCKSDB_SIZEOF_TTL_RECORD);
+    } else {
+      m_ttl_bytes_updated = false;
+    }
+    memcpy(m_ttl_bytes, m_storage_record.ptr(), ROCKSDB_SIZEOF_TTL_RECORD);
```
vasilep: Why are we always doing the mem copy, when there are clearly cases when the data is unchanged?
Maybe adding an `if (m_ttl_bytes_updated)` before `memcpy()` call?

pbjc: Good point! The correct check is a bit more convoluted so I added a comment to clarify things.

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -444,6 +444,7 @@ static my_bool rocksdb_enable_ttl_read_filtering = 1;
 static int rocksdb_debug_ttl_rec_ts = 0;
 static int rocksdb_debug_ttl_snapshot_ts = 0;
 static int rocksdb_debug_ttl_read_filter_ts = 0;
+static my_bool rocksdb_debug_ignore_pk_ttl = 0;
```
hermanlee: preference to rename this rocksdb_debug_ttl_ignore_pk, similar to the other ones.

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -4570,6 +4578,28 @@ int ha_rocksdb::convert_record_to_storage_format(
   *packed_rec =
       rocksdb::Slice(m_storage_record.ptr(), m_storage_record.length());
 
+  // We've obtained the TTL timestamp for this record. Now store it in
```
hermanlee: These checks are already performed at the start of the function. Could this be integrated into the first "if (has_ttl)" conditional near the beginning of ha_rocksdb::convert_record_to_storage_format()?

storage/rocksdb/rdb_datadic.cc
```diff
@@ -72,6 +72,8 @@ Rdb_key_def::Rdb_key_def(uint indexnr_arg, uint keyno_arg,
 {
   mysql_mutex_init(0, &m_mutex, MY_MUTEX_INIT_FAST);
   rdb_netbuf_store_index(m_index_number_storage_form, m_index_number);
+  m_total_index_flags_length =
```
hermanlee, line 78: Can you add a debug assert that this is 0 when the kv_version is <= SECONDARY_FORMAT_VERSION_UPDATE2 && index_type == INDEX_TYPE_SECONDARY. Also add a similar check for index_type == INDEX_TYPE_PRIMARY && kv_version <= PRIMARY_FORMAT_VERSION_UPDATE2.

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -6502,7 +6531,11 @@ int ha_rocksdb::read_row_from_secondary_key(uchar *const buf,
     const rocksdb::Slice &rkey = m_scan_it->key();
     pk_size =
         kd.get_primary_key_tuple(table, *m_pk_descr, &rkey, m_pk_packed_tuple);
-    const rocksdb::Slice &value = m_scan_it->value();
+
+    rocksdb::Slice value = m_scan_it->value();
+    /* Skip index flag fields */
+    value.remove_prefix(kd.m_total_index_flags_length);
```
hermanlee: This can get confusing in the code. In some places, the index flags are stripped, in others they'll be included (i.e. when processing read filtering). Since all of the places that perform this remove_prefix() operation occurs before unpack_record(), it makes more sense for unpack_record() to handle skipping the TTL field directly, similar to convert_from_storage_format() with primary keys. Also, other features later on that add new flags and features may need to perform their operations in unpack_record. Skipping the metadata at this point does not seem like the right thing to do. It's better to pass them to unpack_record() and let unpack_record() figure out what to do.

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -6521,7 +6554,7 @@ int ha_rocksdb::read_row_from_secondary_key(uchar *const buf,
       if (pk_size == RDB_INVALID_KEY_LEN) {
         rc = HA_ERR_ROCKSDB_CORRUPT_DATA;
       } else {
-        rc = get_row_by_rowid(buf, m_pk_packed_tuple, pk_size);
+        rc = get_row_by_rowid(buf, m_pk_packed_tuple, pk_size, false, true);
```
hermanlee: Since the default parameters for skip_lookup and skip_ttl_check, I think these changes can be dropped.

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -6620,10 +6653,12 @@ int ha_rocksdb::secondary_index_read(const int keyno, uchar *const buf) {
       if (m_keyread_only && m_lock_rows == RDB_LOCK_NONE &&
           !has_hidden_pk(table)) {
         rocksdb::Slice value = m_scan_it->value();
+        // Skip index flag fields
+        value.remove_prefix(m_key_descr_arr[keyno]->m_total_index_flags_length);
         rc = m_key_descr_arr[keyno]->unpack_record(
             table, buf, &key, &value, m_verify_row_debug_checksums);
       } else {
-        rc = get_row_by_rowid(buf, m_pk_packed_tuple, size);
+        rc = get_row_by_rowid(buf, m_pk_packed_tuple, size, false, true);
```
hermanlee: Same as above with dropping the false/true.

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -7072,6 +7112,7 @@ int ha_rocksdb::check(THD *const thd, HA_CHECK_OPT *const check_opt) {
         }
 
         /* Check if we get the same secondary key value */
+        m_sk_tails.clear();
```
hermanlee: Why clear here? Isn't this also performed in pack_record()?

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -7847,6 +7888,7 @@ int ha_rocksdb::get_pk_for_update(struct update_row_info *const row_info) {
   if (!has_hidden_pk(table)) {
     row_info->hidden_pk_id = 0;
 
+    m_pk_unpack_info.clear();
```
hermanlee: I think pack_record() also clears this?

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -8260,16 +8302,37 @@ int ha_rocksdb::update_sk(const TABLE *const table_arg, const Rdb_key_def &kd,
 
   const bool store_row_debug_checksums = should_store_row_debug_checksums();
 
+  m_sk_tails.clear();
+  // Reserve space for index flag fields
```
hermanlee: This should happen in pack_record() instead. unpack_info is generated there. The TTL field should also be added there.

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -8260,16 +8302,37 @@ int ha_rocksdb::update_sk(const TABLE *const table_arg, const Rdb_key_def &kd,
...
       &m_sk_tails, store_row_debug_checksums, row_info.hidden_pk_id);
 
   if (row_info.old_data != nullptr) {
+    m_sk_tails_old.clear();
     // The old value
     old_packed_size = kd.pack_record(
         table_arg, m_pack_buffer, row_info.old_data, m_sk_packed_tuple_old,
         &m_sk_tails_old, store_row_debug_checksums, row_info.hidden_pk_id);
 
+    const uchar *old_unpack_info = m_sk_tails_old.ptr();
```
hermanlee: If pack_record() handled recording the ttl value, then this manipulation to adjust the sizes are all taken care of automatically.

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -8322,9 +8385,21 @@ int ha_rocksdb::update_indexes(const struct update_row_info &row_info,
                                const bool &pk_changed) {
   int rc;
 
+  // If this table has a non-hidden PK, update it first.
```
hermanlee: Can you add additional comment that the pk must be updated first to pull out the ttl fields? Also, I don't think you need to special case out has_hidden_pk(). You should be able to handle them both here and let the loop handle sk only.

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -10421,6 +10495,18 @@ int ha_rocksdb::inplace_populate_sk(
...
 
       /* Create new secondary index entry */
+      m_sk_tails.clear();
+      if (index->m_total_index_flags_length > 0) {
+        m_sk_tails.allocate(0, index->m_total_index_flags_length);
+      }
+
+      /* Insert TTL timestamp */
+      if (index->has_ttl()) {
+        index->write_index_flag_field(&m_sk_tails,
```
hermanlee: Again, this code should go into pack_record()

hermanlee (changes requested):

storage/rocksdb/rdb_buff.h
```diff
@@ -359,6 +359,21 @@ class Rdb_string_writer {
...
+    m_data.resize(pos);
+  }
+
+  void allocate(const uchar &val, const size_t &len) {
+    DBUG_ASSERT(len > 0);
+    m_data.resize(m_data.size() + len, val);
+  }
+
+  /*
+    An awful hack to deallocate the buffer without relying on the deconstructor.
```
hermanlee, line 374: Can you add a comment about why this is needed. Valgrind errors showing up, etc.

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -5358,6 +5380,11 @@ int ha_rocksdb::close(void) {
     rdb_open_tables.release_table_handler(m_table_handler);
     m_table_handler = nullptr;
   }
+  m_storage_record.free();
```
hermanlee: Add a comment about free called here to prevent valgrind reported issues.

storage/rocksdb/rdb_buff.h
```diff
@@ -359,6 +359,21 @@ class Rdb_string_writer {
     DBUG_ASSERT(pos < get_current_pos() && (pos + 1) < get_current_pos());
     rdb_netbuf_store_uint16(m_data.data() + pos, new_val);
   }
+
+  void truncate(const size_t &pos) {
+    DBUG_ASSERT(0 <= pos && pos < m_data.size());
+    m_data.resize(pos);
+  }
+
+  void allocate(const uchar &val, const size_t &len) {
```
hermanlee: Probably better to swap these arguments, and have val be 0 by default, so we can just call allocate(len).

hermanlee (approved):

PR #666, opened 2017-07-25
Displayed Deadlock Paths

armishra:
@update-submodule: rocksdb

Changes:
* extracted deadock information buffer from rocksdb
* displayed information of interest to the user by adding a section with information about the latest deadlocks
* annotates the transaction id of the transaction that go the deadlock
* added small test case to verify the changes and recorded engine transaction status tests

### DIFF
--- a/mysql-test/r/mysqld--help-notwin-profiling.result
+++ b/mysql-test/r/mysqld--help-notwin-profiling.result
@@ -1170,6 +1170,9 @@ The following options may be given as the first argument:
  ON, OFF, FORCE (don't start if the plugin fails to load).
  --rocksdb-deadlock-detect 
  Enables deadlock detection
+ --rocksdb-deadlock-detect-depth=# 
+ Number of transactions deadlock detection will traverse
+ through before assuming deadlock
  --rocksdb-debug-optimizer-n-rows=# 
  Test only to override rocksdb estimates of table size in
  a memtable
@@ -1296,6 +1299,8 @@ The following options may be given as the first argument:
  slave.
  --rocksdb-max-background-jobs=# 
  DBOptions::max_background_jobs for RocksDB
+ --rocksdb-max-latest-deadlocks=# 
+ Maximum number of recent deadlocks to store
  --rocksdb-max-log-file-size=# 
  DBOptions::max_log_file_size for RocksDB
  --rocksdb-max-manifest-file-size=# 
@@ -2094,6 +2099,7 @@ rocksdb-db-write-buffer-size 0
 rocksdb-dbstats ON
 rocksdb-ddl ON
 rocksdb-deadlock-detect FALSE
+rocksdb-deadlock-detect-depth 50
 rocksdb-debug-optimizer-n-rows 0
 rocksdb-debug-optimizer-no-zero-cardinality TRUE
 rocksdb-debug-ttl-read-filter-ts 0
@@ -2132,6 +2138,7 @@ rocksdb-manifest-preallocation-size 4194304
 rocksdb-manual-wal-flush TRUE
 rocksdb-master-skip-tx-api FALSE
 rocksdb-max-background-jobs 2
+rocksdb-max-latest-deadlocks 5
 rocksdb-max-log-file-size 0
 rocksdb-max-manifest-file-size 18446744073709551615
 rocksdb-max-open-files -1
--- a/mysql-test/suite/rocksdb/t/deadlock_tracking.test
+++ b/mysql-test/suite/rocksdb/t/deadlock_tracking.test
@@ -1,0 +1,144 @@
+set @prior_lock_wait_timeout = @@rocksdb_lock_wait_timeout; 
+set @prior_deadlock_detect = @@rocksdb_deadlock_detect; 
+set @prior_max_latest_deadlocks = @@rocksdb_max_latest_deadlocks; 
+set global rocksdb_deadlock_detect = on; 
+set global rocksdb_lock_wait_timeout = 10000;
+let $engine = rocksdb;
+
+--source include/count_sessions.inc
+connect (con1,localhost,root,,);
+let $con1= `SELECT CONNECTION_ID()`;
+
+connect (con2,localhost,root,,);
+let $con2= `SELECT CONNECTION_ID()`;
+
+connect (con3,localhost,root,,);
+let $con3= `SELECT CONNECTION_ID()`;
+
+connection default;
+eval create table t (i int primary key) engine=$engine;
+insert into t values (1), (2), (3);
+--replace_regex /[0-9]{4}-[0-9]{2}-[0-9]{2} [0-9]{2}:[0-9]{2}:[0-9]{2}/TIMESTAMP/ /WAITING KEY: [0-9a-f]{16}/KEY/ /TRANSACTIONID: [0-9]*/TXN_ID/
+show engine rocksdb transaction status;
+
+echo Deadlock #1;
+--source include/simple_deadlock.inc
+connection default;
+--replace_regex /[0-9]{4}-[0-9]{2}-[0-9]{2} [0-9]{2}:[0-9]{2}:[0-9]{2}/TIMESTAMP/ /WAITING KEY: [0-9a-f]{16}/KEY/ /TRANSACTIONID: [0-9]*/TXN_ID/
+show engine rocksdb transaction status;
+
+echo Deadlock #2;
+--source include/simple_deadlock.inc
+connection default;
+--replace_regex /[0-9]{4}-[0-9]{2}-[0-9]{2} [0-9]{2}:[0-9]{2}:[0-9]{2}/TIMESTAMP/ /WAITING KEY: [0-9a-f]{16}/KEY/ /TRANSACTIONID: [0-9]*/TXN_ID/
+show engine rocksdb transaction status;
+set global rocksdb_max_latest_deadlocks = 10;
+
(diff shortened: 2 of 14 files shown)


mysql-test/suite/rocksdb/t/deadlock_tracking.test
```diff
@@ -0,0 +1,47 @@
+set @prior_lock_wait_timeout = @@rocksdb_lock_wait_timeout; 
+set @prior_deadlock_detect = @@rocksdb_deadlock_detect; 
+set global rocksdb_deadlock_detect = on; 
+set global rocksdb_lock_wait_timeout = 10000;
+
+let $engine=rocksdb;
+
+let $prior_set_lwt = `select concat('set @prior_lock_wait_timeout = @@',
```
alxyang: I'm not seeing the purpose for these variables being set here.  It seems that we already save the state above. Why not just `set global rocksdb_lock_wait_timeout=100000` directly?

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -3007,6 +3007,47 @@ class Rdb_snapshot_status : public Rdb_tx_list_walker {
           tx->get_delete_count());
     }
   }
+
+  void populateDeadlockBuffer() {
+    uint max_key_size = 50;
+    m_data += "---------\n"
+              "DEADLOCKS\n"
+              "---------\n"
+              "LIST OF DEADLOCKS FOR EACH SESSION:\n";
```
alxyang: If i'm not mistaken the 'list of snapshots for each session' actually shows which snapshots are from which session via the mysql thread id. 

However in your case there isn't such a separation.  It looks like the number next to DEADLOCK is an incremented counter.  So I think maybe we just say "LIST OF DEADLOCKS".

alxyang (changes requested):
I think you are going to have to re-record tests, like `show_engine` for example.  Because output of all `show engine rocksdb transaction status` has now changed.

mysql-test/suite/rocksdb/t/deadlock_tracking.test
armishra: This is meant to be consistent with the simple_deadlock.inc format of setting global variables in deadlock tests

armishra: Also the global vars need to be reset to their original values before the test. The test suite throws an error otherwise

mysql-test/r/mysqld--help-notwin-profiling.result
```diff
@@ -1170,6 +1170,8 @@ The following options may be given as the first argument:
  ON, OFF, FORCE (don't start if the plugin fails to load).
  --rocksdb-deadlock-detect 
  Enables deadlock detection
+ --rocksdb-deadlock-detect-depth=# 
+ Maximum depth a dedlock can have
```
lth: dedlock -> deadlock

A better description for users might be:
Number of transactions deadlock detection will traverse through before assuming deadlock.

storage/rocksdb/rdb_datadic.cc
```diff
@@ -3875,6 +3875,16 @@ Rdb_ddl_manager::safe_find(GL_INDEX_ID gl_index_id) {
   return ret;
 }
 
+const std::string
+Rdb_ddl_manager::get_table_name(const GL_INDEX_ID &gl_index_id) {
+  std::string ret;
+  auto it = m_index_num_to_keydef.find(gl_index_id);
```
hermanlee: Could you move this function after Rdb_ddl_manager::find(), so that safe_find()'s description is more accurate. Also, is mysql_rwlock_rdlock(&m_rwlock); not needed here? If so, there should be a comment to indicate it like the find function. However, given the place you're calling get_table_name() also calls safe_find(), I don't believe the lock is held. If you do acquire the lock, then probably rename this function to safe_get_table_name().

hermanlee (changes requested):

hermanlee (commented):
Most of the code looks fine to me. I'll leave it to Manuel to accept. My only recommendation is to have more complex test cases, like transactions holding multiple locks and verifying the deadlock detector prints out the right set of locks.

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -3072,6 +3093,82 @@ class Rdb_snapshot_status : public Rdb_tx_list_walker {
           tx->get_delete_count());
     }
   }
+
+  void populateDeadlockBuffer() {
+    uint max_key_size = 16;
+    m_data += "---------\n"
+              "DEADLOCKS\n"
+              "---------\n"
+              "LIST OF DEADLOCKS FOR EACH SESSION:\n";
```
lth: I don't think it's accurate to say "for each session".

Maybe this should just say "LATEST DETECTED DEADLOCK" (similar to InnoDB), and consolidate that with the line above that just says "DEADLOCKS".

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -3072,6 +3093,82 @@ class Rdb_snapshot_status : public Rdb_tx_list_walker {
           tx->get_delete_count());
     }
   }
+
+  void populateDeadlockBuffer() {
+    uint max_key_size = 16;
```
lth: Where does this value come from? I don't think a maximum size is needed in this case, and the default argument to rdb_hexdump for max_size is 0, which means that the string will never be truncated.

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -731,6 +744,11 @@ static MYSQL_SYSVAR_ULONGLONG(delayed_write_rate, rocksdb_delayed_write_rate,
                               rocksdb_db_options->delayed_write_rate, 0,
                               UINT64_MAX, 0);
 
+static MYSQL_SYSVAR_UINT(max_deadlocks, rocksdb_max_deadlocks,
+                         PLUGIN_VAR_RQCMDARG, "DBOptions::max_deadlocks",
```
lth: It's not a DBOption, so I don't think this is a good description.

How about we actually describe what this variable tracks?

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -3072,6 +3093,82 @@ class Rdb_snapshot_status : public Rdb_tx_list_walker {
...
+              "---------\n"
+              "LIST OF DEADLOCKS FOR EACH SESSION:\n";
+    std::vector<rocksdb::DeadlockPath> dlock_buffer =
+        rdb->GetDeadlockInfoBuffer();
+    for (auto path_entry : dlock_buffer) {
+      if (!path_entry.limit_exceeded) {
+        m_data += "\n*** DEADLOCK PATH\n"
+                  "=========================================\n";
+        for (auto txn : path_entry.path) {
+          GL_INDEX_ID gl_index_id = {
```
lth: I think we can just do 
`GL_INDEX_ID gl_index_id = { txn.m_cf_id, rdb_netbuf... };`

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -3072,6 +3093,82 @@ class Rdb_snapshot_status : public Rdb_tx_list_walker {
...
+                                              max_key_size)
+                                      .c_str(),
+                                  txn.m_exclusive ? "EXCLUSIVE" : "SHARED",
+                                  idx_name.c_str(), table_name.c_str());
+        }
+      } else {
+        m_data += "-------DEADLOCK EXCEEDED MAX DEPTH-------\n";
+      }
+
+      if (!path_entry.path.empty()) {
```
lth: I think that we should assert this:

`DBUG_ASSERT_IFF(path_entry.limit_exceeded, !path_entry.path.empty());`

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -3072,6 +3093,82 @@ class Rdb_snapshot_status : public Rdb_tx_list_walker {
...
+        if (!path_entry.limit_exceeded) {
+          auto txn = path_entry.path[0];
+          GL_INDEX_ID gl_index_id = {
```
lth: Same comment as above.

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -3072,6 +3093,82 @@ class Rdb_snapshot_status : public Rdb_tx_list_walker {
...
+                                      .c_str(),
+                                  txn.m_exclusive ? "EXCLUSIVE" : "SHARED",
+                                  idx_name.c_str(), table_name.c_str());
+        }
+      } else {
+        m_data += "-------DEADLOCK EXCEEDED MAX DEPTH-------\n";
+      }
+
+      if (!path_entry.path.empty()) {
+        if (!path_entry.limit_exceeded) {
```
lth: Wouldn't it be cleaner to move this code into the `if (!path_entry.limit_exceeded) {` check above.

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -3072,6 +3093,82 @@ class Rdb_snapshot_status : public Rdb_tx_list_walker {
...
+          auto txn = path_entry.path[0];
```
lth: Since this code is duplicated, I wonder if it'd be cleaner if you wrote a helper for this.

lth (changes requested):

lth (approved):

PR #668, opened 2017-07-27
Improve multiclient_2pc test

maysamyabandeh:
1. Remove the sleep 60, and instead properly wait for disconnect/connect events.
2. Reduce the insert size from 1m to 4096 and configure binlog size to ensure that it is rotated with that many insertions
3. Update the test with the newest variable changes in mysql
4. Disable the test for valgrind: the test involves a crash which does not seem to be handled well with mysql-test/lib/My/SafeProcess/my_safe_process under valgrind as it hangs forever. The test did not mean to verify the memory leaks so not much coverage should be missed by not running it under valgrind.

Test Plan:
mysqltest.sh rocksdb_rpl.multiclient_2pc --big-test

### DIFF
--- a/mysql-test/suite/rocksdb_rpl/t/multiclient_2pc.test
+++ b/mysql-test/suite/rocksdb_rpl/t/multiclient_2pc.test
@@ -3,6 +3,11 @@
 --source include/have_debug.inc
 --source include/have_debug_sync.inc
 --source include/big_test.inc
+# The test involves a crash which does not seem to be handled well with
+# mysql-test/lib/My/SafeProcess/my_safe_process under valgrind as it hangs
+# forever. The test did not mean to verify the memory leaks so not much
+# coverage should be missed by not running it under valgrind.
+--source include/not_valgrind.inc
 
 --exec echo > $MYSQLTEST_VARDIR/log/mysqld.1.err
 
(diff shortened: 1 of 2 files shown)


mysql-test/suite/rocksdb_rpl/t/multiclient_2pc.test
```diff
@@ -3,23 +3,30 @@
 --source include/have_debug.inc
 --source include/have_debug_sync.inc
 --source include/big_test.inc
+# The test involves a crash which does not seem to be handled well with
```
laurynas-biveinis, line 6: This sounds suspicious, I have never seen such hang under Valgrind (only memory allocations at the time crash time reported as leaks), but I am not seeing what could be causing this here. 

You could also add a --source include/not_crashrep.inc here (as documentation that this is a crash-testing testcase; the include file is for Mac and mostly a no-op)

hermanlee (commented):
The rest of the changes look fine to me. Is there a root cause for the hang under valgrind? Maybe it make sense to commit these changes and then figure out the valgrind issue in a separate task/pull request?

maysamyabandeh:
I could not figure the root cause.

Yeah makes sense to me.

hermanlee (approved):

PR #672, opened 2017-07-31
Improve drop_table{,2,3} testcases

spetrunia:
- Don't truncate mysqld.1.err file, it's a bad practice
- remove --rocksdb_info_log_level=info_level from .opt file as it is
  not necessary anymore.
- Add test suppressions for harmless "Column family not found" messages.
- See https://jira.mariadb.org/browse/MDEV-13413 for the background

### DIFF
--- a/mysql-test/suite/rocksdb/t/drop_table-master.opt
+++ b/mysql-test/suite/rocksdb/t/drop_table-master.opt
@@ -1,4 +1,3 @@
 --rocksdb_max_subcompactions=1
---rocksdb_info_log_level=info_level
 --rocksdb_default_cf_options=write_buffer_size=512k;target_file_size_base=512k;level0_file_num_compaction_trigger=2;level0_slowdown_writes_trigger=-1;level0_stop_writes_trigger=1000;max_bytes_for_level_base=1m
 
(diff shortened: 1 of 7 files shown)


mysql-test/suite/rocksdb/t/drop_table-master.opt
```diff
@@ -1,4 +1,3 @@
 --rocksdb_max_subcompactions=1
---rocksdb_info_log_level=info_level
```
alxyang, line 2: @spetrunia actually the reason we have this in here right now is because we have noticed the rocksdb.drop_table test failing intermittently due tripping an assertion with manual compaction in rocksdb.   It was added in https://github.com/facebook/mysql-5.6/commit/977988340eff5cfbf651c5ccdb80c857c4c26a6f.  

This issue still isn't resolved so I don't think we should remove this just yet.

george-lorch: @alxyang Do you have an issue to track this? I have a similar issue in Percona Server that I have not yet reported to Facebook as I have not ruled out some problem as part of our port. Our tracking issue is here https://jira.percona.com/browse/MYR-127

alxyang: @georgelorchpercona  yes. this is the same issue (and therefore likely not a problem due to your port).  I think we have an issue internally to track this, but not yet on github (feel free to open and assign to me if you'd like).  For now if you need the test to pass you can also remove the manual compaction in the test, as I believe it's not necessary for the test correctness.

george-lorch: @alxyang Thanks for the confirmation, that's really all I needed to know in order to prioritize it properly on our side.

george-lorch: BTW, we see it frequently in our jenkins farm. If there is anything we can provide to help diagnose, please ask.

hermanlee (changes requested):
Requesting changes to reinclude the --rocksdb_info_log_level=info_level line.

PR #679, opened 2017-08-11
track Semisync_ack and Engine_commit time in slow query log

francoisrichard:
First pass at this.  Please note that this will modify / append to the format of the slow query log.


francoisrichard:
Created a second pull request by mistake, will use the later one.

PR #685, opened 2017-08-17
Enable TTL read filtering for tables with secondary keys

pbjc:
This extends #617 to implement read filtering for tables with secondary keys. Along with #662 this completes the basis for secondary key support. Changes:
* Modified `position_to_correct_key` and `index_next` to skip expired records.
* Fixes an SK update path bug from #662.

### DIFF
--- a/mysql-test/suite/rocksdb/r/ttl_secondary_with_partitions.result
+++ b/mysql-test/suite/rocksdb/r/ttl_secondary_with_partitions.result
@@ -1,0 +1,389 @@
+CREATE TABLE t1 (
+c1 INT,
+c2 INT,
+PRIMARY KEY (`c1`),
+KEY kc2 (`c2`)
+) ENGINE=ROCKSDB
+COMMENT="custom_p0_ttl_duration=1;"
+PARTITION BY LIST(c1) (
+PARTITION custom_p0 VALUES IN (1, 4, 7),
+PARTITION custom_p1 VALUES IN (2, 5, 8),
+PARTITION custom_p2 VALUES IN (3, 6, 9)
+);
+set global rocksdb_debug_ttl_rec_ts = -3600;
+INSERT INTO t1 values (1, 1);
+INSERT INTO t1 values (2, 2);
+INSERT INTO t1 values (3, 3);
+INSERT INTO t1 values (4, 4);
+INSERT INTO t1 values (5, 5);
+INSERT INTO t1 values (6, 6);
+INSERT INTO t1 values (7, 7);
+INSERT INTO t1 values (8, 8);
+INSERT INTO t1 values (9, 9);
+set global rocksdb_debug_ttl_rec_ts = 0;
+SELECT * FROM t1 FORCE INDEX (PRIMARY);
+c1	c2
+1	1
+2	2
+3	3
+4	4
+5	5
+6	6
+7	7
+8	8
+9	9
+SELECT * FROM t1 FORCE INDEX (kc2);
+c1	c2
+1	1
+2	2
+3	3
+4	4
+5	5
+6	6
+7	7
+8	8
+9	9
+set global rocksdb_force_flush_memtable_now=1;
+set global rocksdb_compact_cf='default';
+SELECT * FROM t1 FORCE INDEX (PRIMARY);
+c1	c2
+2	2
+3	3
+5	5
+6	6
+8	8
+9	9
+SELECT * FROM t1 FORCE INDEX (kc2);
+c1	c2
+2	2
+3	3
+5	5
+6	6
+8	8
+9	9
+DROP TABLE t1;
+CREATE TABLE t1 (
+c1 INT,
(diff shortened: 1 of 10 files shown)


mysql-test/suite/rocksdb/t/ttl_secondary_with_partitions.test
```diff
@@ -0,0 +1,298 @@
+--source include/have_debug.inc
+--source include/have_rocksdb.inc
```
laurynas-biveinis, line 2: --source include/have_partition.inc

mysql-test/suite/rocksdb/t/ttl_secondary_read_filtering.test
```diff
@@ -0,0 +1,407 @@
...
+set global rocksdb_debug_ttl_rec_ts = -300;
+INSERT INTO t1 values (1, UNIX_TIMESTAMP());
+set global rocksdb_debug_ttl_rec_ts = 300;
+INSERT INTO t1 values (2, UNIX_TIMESTAMP());
+INSERT INTO t1 values (3, UNIX_TIMESTAMP());
+set global rocksdb_debug_ttl_rec_ts = 0;
+
+set global rocksdb_force_flush_memtable_now=1;
+
+# 1 should be hidden even though compaction hasn't run.
```
hermanlee: flush_memtable was run above though, so compaction has run and key 1 would have been expired. Also, change these to --echo, so that they're in the results file and it'd be easy to see from the results file what the expected behavior is.

mysql-test/suite/rocksdb/t/ttl_secondary_with_partitions.test
```diff
@@ -0,0 +1,298 @@
...
+    KEY kc2 (`c2`)
+) ENGINE=ROCKSDB
+COMMENT="custom_p0_ttl_duration=1;custom_p1_ttl_duration=7;"
+PARTITION BY LIST(c1) (
+    PARTITION custom_p0 VALUES IN (1, 4, 7),
+    PARTITION custom_p1 VALUES IN (2, 5, 8),
+    PARTITION custom_p2 VALUES IN (3, 6, 9)
+);
+
+set global rocksdb_debug_ttl_rec_ts = -1200;
```
hermanlee, line 79: Can you have a similar test, but not for partitions? 1, 4, 7 in the same partition means they reside side-by-side in the same table. What I'd like to see is the following:

set global rocksdb_debug_ttl_rec_ts = -1200;
INSERT INTO t1 values (1,1,'a');
INSERT INTO t1 values (2,2,'b');
set global rocksdb_debug_ttl_rec_ts = 1200;
INSERT INTO t1 values (3,3,'c');
INSERT INTO t1 values (4,4,'d');
set global rocksdb_debug_ttl_rec_ts = -1200;
INSERT INTO t1 values (5,5,'e');
INSERT INTO t1 values (6,6,'f');
set global rocksdb_debug_ttl_rec_ts = 1200;
INSERT INTO t1 values (7,7,'g');
INSERT INTO t1 values (8,8,'h');

Then issue:

select * from t1 with various force index and conditionals:

select * from t1;
select * from t1 force index(primary) where c1 > 5;

etc.

hermanlee (changes requested):

mysql-test/suite/rocksdb/r/ttl_secondary_with_partitions.result
```diff
@@ -0,0 +1,389 @@
...
+INSERT INTO t1 values (4, 4);
+INSERT INTO t1 values (5, 5);
+INSERT INTO t1 values (6, 6);
+INSERT INTO t1 values (7, 7);
+INSERT INTO t1 values (8, 8);
+INSERT INTO t1 values (9, 9);
+set global rocksdb_debug_ttl_rec_ts = 0;
+SELECT * FROM t1 FORCE INDEX (PRIMARY);
+c1	c2
+1	1
```
hermanlee, line 26: This seems odd to me. Why isn't read filtering removing these keys?

hermanlee (commented):
Most of these tests change the timestamp on the keys and then flushes them. At that point, the compaction filter would have run to remove them. I would have expected the tests to use debug_ttl_read_filter_ts to adjust whether keys are returned or not.

The test should vary rocksdb_debug_ttl_rec_ts for various keys, set them to 0, 1800 and 3600. The table should have ttl_duration=7200; Then perform select on the table with debug_ttl_read_filter_ts = 0, debug_ttl_read_filter_ts = -1800, debug_ttl_read_filter_ts = -3600 and see the appropriate keys be removed.

mysql-test/suite/rocksdb/r/ttl_secondary_with_partitions.result
pbjc: Read filtering is off by default for this test case.

hermanlee (approved):

PR #693, opened 2017-08-25
crash safety for auto increment values

### LINKED ISSUE #468: Crash-safe auto_increment values
InnoDB had a long-standing bug, finally fixed in 8.0, that auto_increment values were lost on server restart: http://bugs.mysql.com/bug.php?id=199

According to a comment on #39, it seems that MyRocks might also have the same bug, which would be a regression compared to InnoDB on 8.0.

armishra:
Changes:
* Auto increment value added as type in system column family
* Added maximizing merge operator
* Merge called on write path; get called on read path
* closes #468

### DIFF
--- a/storage/rocksdb/ha_rocksdb.cc
+++ b/storage/rocksdb/ha_rocksdb.cc
@@ -1751,6 +1751,7 @@ protected:
   ulonglong m_update_count = 0;
   ulonglong m_delete_count = 0;
   ulonglong m_lock_count = 0;
+  std::unordered_map<GL_INDEX_ID, ulonglong> m_auto_incr_map;
 
   bool m_is_delayed_snapshot = false;
   bool m_is_two_phase = false;
@@ -1783,7 +1784,30 @@ protected:
   get_iterator(const rocksdb::ReadOptions &options,
                rocksdb::ColumnFamilyHandle *column_family) = 0;
 
-public:
+  /*
+    @detail
+      This function takes in the WriteBatch of the transaction to add
+      all the AUTO_INCREMENT merges. It does so by iterating through
+      m_auto_incr_map and then constructing key/value pairs to call merge upon.
+
+    @param wb
+   */
+  rocksdb::Status merge_auto_incr_map(rocksdb::WriteBatchBase *const wb) {
+    DBUG_EXECUTE_IF("myrocks_autoinc_upgrade", return rocksdb::Status::OK(););
+
+    // Iterate through the merge map merging all keys into data dictionary.
+    rocksdb::Status s;
+    for (auto &it : m_auto_incr_map) {
+      s = dict_manager.put_auto_incr_val(wb, it.first, it.second);
+      if (!s.ok()) {
+        return s;
+      }
+    }
+    m_auto_incr_map.clear();
+    return s;
+  }
+
+ public:
   const char *m_mysql_log_file_name;
   my_off_t m_mysql_log_offset;
   const char *m_mysql_gtid;
@@ -2086,6 +2110,11 @@ public:
     return false;
   }
 
+  void set_auto_incr(const GL_INDEX_ID &gl_index_id, ulonglong curr_id) {
+    m_auto_incr_map[gl_index_id] =
+        std::max(m_auto_incr_map[gl_index_id], curr_id);
+  }
+
   virtual rocksdb::Status put(rocksdb::ColumnFamilyHandle *const column_family,
                               const rocksdb::Slice &key,
                               const rocksdb::Slice &value) = 0;
@@ -2244,6 +2273,12 @@ private:
       return false;
     }
 
+    s = merge_auto_incr_map(m_rocksdb_tx->GetWriteBatch()->GetWriteBatch());
+    if (!s.ok()) {
+      rdb_handle_io_error(s, RDB_IO_ERROR_TX_COMMIT);
+      return false;
+    }
+
     s = m_rocksdb_tx->Prepare();
     if (!s.ok()) {
       rdb_handle_io_error(s, RDB_IO_ERROR_TX_COMMIT);
@@ -2254,8 +2289,16 @@ private:
 
   bool commit_no_binlog() override {
     bool res = false;
+    rocksdb::Status s;
+
+    s = merge_auto_incr_map(m_rocksdb_tx->GetWriteBatch()->GetWriteBatch());
+    if (!s.ok()) {
+      rdb_handle_io_error(s, RDB_IO_ERROR_TX_COMMIT);
+      res = true;
+    }
+
     release_snapshot();
-    const rocksdb::Status s = m_rocksdb_tx->Commit();
+    s = m_rocksdb_tx->Commit();
     if (!s.ok()) {
       rdb_handle_io_error(s, RDB_IO_ERROR_TX_COMMIT);
       res = true;
@@ -2281,6 +2324,7 @@ public:
     m_update_count = 0;
     m_delete_count = 0;
     m_lock_count = 0;
+    m_auto_incr_map.clear();
     m_ddl_transaction = false;
     if (m_rocksdb_tx) {
       release_snapshot();
@@ -2526,9 +2570,17 @@ private:
 
   bool commit_no_binlog() override {
     bool res = false;
+    rocksdb::Status s;
+
+    s = merge_auto_incr_map(m_batch->GetWriteBatch());
+    if (!s.ok()) {
+      rdb_handle_io_error(s, RDB_IO_ERROR_TX_COMMIT);
+      res = true;
+    }
+
     release_snapshot();
-    const rocksdb::Status s =
-        rdb->GetBaseDB()->Write(write_opts, m_batch->GetWriteBatch());
+
+    s = rdb->GetBaseDB()->Write(write_opts, m_batch->GetWriteBatch());
     if (!s.ok()) {
       rdb_handle_io_error(s, RDB_IO_ERROR_TX_COMMIT);
       res = true;
@@ -4412,45 +4464,6 @@ std::vector<std::string> Rdb_open_tables_map::get_table_names(void) const {
   return names;
 }
 
-void ha_rocksdb::load_auto_incr_value() {
-  const int save_active_index = active_index;
-  active_index = table->s->next_number_index;
-  const uint8 save_table_status = table->status;
-
-  /*
-    load_auto_incr_value() may be called by statements that
-    do not execute implicit commits (i.e. SHOW CREATE TABLE).
-    index_last() creates a snapshot. When a snapshot is created
-    here, it has to be released as well. (GitHub issue#189)
-  */
-  Rdb_transaction *const tx = get_or_create_tx(table->in_use);
-  const bool is_new_snapshot = !tx->has_snapshot();
-
-  // Do a lookup. We only need index column, so it should be index-only.
-  // (another reason to make it index-only is that table->read_set is
-  //  not set appropriately and non-index-only lookup will not read the value)
-  const bool save_keyread_only = m_keyread_only;
-  m_keyread_only = true;
-
-  if (!index_last(table->record[0]))
-    update_auto_incr_val();
-
-  m_keyread_only = save_keyread_only;
-  if (is_new_snapshot) {
-    tx->release_snapshot();
-  }
-
-  table->status = save_table_status;
-  active_index = save_active_index;
-
-  /*
-    Do what ha_rocksdb::index_end() does.
-    (Why don't we use index_init/index_end? class handler defines index_init
-     as private, for some reason).
-  */
-  release_scan_iterator();
-}
-
 /*
   Inspired by innobase_get_int_col_max_value from InnoDB. This returns the
   maximum value a type can take on.
@@ -4495,13 +4508,99 @@ static ulonglong rdb_get_int_col_max_value(const Field *field) {
   return max_value;
 }
 
-/* Get PK value from table->record[0]. */
-/*
-  TODO(alexyang): No existing support for auto_increment on non-pk columns, see
-  end of ha_rocksdb::create. Also see opened issue here:
-  https://github.com/facebook/mysql-5.6/issues/153
-*/
-void ha_rocksdb::update_auto_incr_val() {
+void ha_rocksdb::load_auto_incr_value() {
+  ulonglong dd_val = 0;
+  bool validate_last = false;
+  bool autoinc_upgrade_test = false;
+  DBUG_EXECUTE_IF("myrocks_autoinc_upgrade", autoinc_upgrade_test = true;);
+#ifndef NDEBUG
+  validate_last = true;
+#endif
+
+  if (!autoinc_upgrade_test &&
+      dict_manager.get_auto_incr_val(m_tbl_def->get_pk()->get_gl_index_id(),
+                                     &dd_val)) {
+    update_auto_incr_val(dd_val);
+  }
+
+  // If we find nothing in the data dictionary, or if we are in debug mode,
+  // then call index_last to get the last value.
+  //
+  // This is needed when upgrading from a server that did not support
+  // persistent auto_increment, or when doing an alter table to add
+  // auto_increment property.
+  //
+  // For debug mode, we are just verifying that the data dictionary value is
+  // greater than or equal to the maximum value in the table.
+  if (dd_val == 0 || validate_last) {
+    const int save_active_index = active_index;
+    active_index = table->s->next_number_index;
+    const uint8 save_table_status = table->status;
+
+    Rdb_transaction *const tx = get_or_create_tx(table->in_use);
+    const bool is_new_snapshot = !tx->has_snapshot();
+    if (is_new_snapshot) {
+      tx->acquire_snapshot(true);
+    }
+
+    // Do a lookup. We only need index column, so it should be index-only.
+    // (another reason to make it index-only is that table->read_set is not set
+    // appropriately and non-index-only lookup will not read the value)
+    const bool save_keyread_only = m_keyread_only;
+    m_keyread_only = true;
+
+    if (!index_last(table->record[0])) {
+      Field *field =
+          table->key_info[table->s->next_number_index].key_part[0].field;
+      ulonglong max_val = rdb_get_int_col_max_value(field);
+      my_bitmap_map *const old_map =
+          dbug_tmp_use_all_columns(table, table->read_set);
+      ulonglong last_val = field->val_int();
+      if (validate_last && last_val <= max_val) {
+        // Reload data dictionary value in case value has changed since
+        // snapshot was acquired.
+        if (dict_manager.get_auto_incr_val(
+                m_tbl_def->get_pk()->get_gl_index_id(), &dd_val)) {
+          DBUG_ASSERT(dd_val >= last_val);
+        }
+      }
+      dbug_tmp_restore_column_map(table->read_set, old_map);
+
+      // If data dictionary is empty, but rows are found, then this might be
+      // a recently upgraded table. Use the value obtained from index_last.
+      if (dd_val == 0) {
+        update_auto_incr_val_from_field();
+      }
+    }
+
+    m_keyread_only = save_keyread_only;
+    if (is_new_snapshot) {
+      tx->release_snapshot();
+    }
+
+    table->status = save_table_status;
+    active_index = save_active_index;
+
+    /*
+      Do what ha_rocksdb::index_end() does.
+      (Why don't we use index_init/index_end? class handler defines index_init
+      as private, for some reason).
+      */
+    release_scan_iterator();
+  }
+}
+
+void ha_rocksdb::update_auto_incr_val(ulonglong val) {
+  ulonglong auto_incr_val = m_tbl_def->m_auto_incr_val;
+  while (
+      auto_incr_val < val &&
+      !m_tbl_def->m_auto_incr_val.compare_exchange_weak(auto_incr_val, val)) {
+    // Do nothing - just loop until auto_incr_val is >= val or we successfully
+    // set it
+  }
+}
+
+void ha_rocksdb::update_auto_incr_val_from_field() {
   Field *field;
   ulonglong new_val, max_val;
   field = table->key_info[table->s->next_number_index].key_part[0].field;
@@ -4519,13 +4618,11 @@ void ha_rocksdb::update_auto_incr_val() {
 
   // Only update if positive value was set for auto_incr column.
   if (new_val <= max_val) {
-    ulonglong auto_incr_val = m_tbl_def->m_auto_incr_val;
-    while (auto_incr_val < new_val &&
-           !m_tbl_def->m_auto_incr_val.compare_exchange_weak(auto_incr_val,
-                                                             new_val)) {
-      // Do nothing - just loop until auto_incr_val is >= new_val or
-      // we successfully set it
-    }
+    Rdb_transaction *const tx = get_or_create_tx(table->in_use);
+    tx->set_auto_incr(m_tbl_def->get_pk()->get_gl_index_id(), new_val);
+
+    // Update the in memory auto_incr value in m_tbl_def.
+    update_auto_incr_val(new_val);
   }
 }
 
@@ -6574,6 +6671,20 @@ int ha_rocksdb::create(const char *const name, TABLE *const table_arg,
 
   m_pk_descr = m_key_descr_arr[pk_index(table_arg, m_tbl_def)];
 
+  if (create_info->auto_increment_value) {
+    bool autoinc_upgrade_test = false;
+    m_tbl_def->m_auto_incr_val = create_info->auto_increment_value;
+    DBUG_EXECUTE_IF("myrocks_autoinc_upgrade", autoinc_upgrade_test = true;);
+    if (!autoinc_upgrade_test) {
+      auto s = dict_manager.put_auto_incr_val(
+          batch, m_tbl_def->get_pk()->get_gl_index_id(),
+          m_tbl_def->m_auto_incr_val);
+      if (!s.ok()) {
+        goto error;
+      }
+    }
+  }
+
   dict_manager.lock();
   err = ddl_manager.put_and_write(m_tbl_def, batch);
   if (err != HA_EXIT_SUCCESS) {
@@ -6589,23 +6700,6 @@ int ha_rocksdb::create(const char *const name, TABLE *const table_arg,
 
   dict_manager.unlock();
 
-  if (create_info->auto_increment_value)
-    m_tbl_def->m_auto_incr_val = create_info->auto_increment_value;
-
-  /*
-    We only support auto_increment at start of the PRIMARY KEY.
-  */
-  // Field *field;
-  // if ((field= table_arg->next_number_field))
-  /* TODO mdcallag: disable this for now to let UNIQUE indexes kind of work
-  if ((field= table_arg->found_next_number_field))
-  {
-    int pk= table_arg->s->primary_key;
-    Field *pk_field= table_arg->key_info[pk].key_part[0].field;
-    if (field->field_index !=  pk_field->field_index)
-      DBUG_RETURN(HA_ERR_INTERNAL_ERROR);
-  }
-  */
   DBUG_RETURN(HA_EXIT_SUCCESS);
 
 error:
@@ -8681,7 +8775,7 @@ int ha_rocksdb::update_pk(const Rdb_key_def &kd,
   }
 
   if (table->next_number_field) {
-    update_auto_incr_val();
+    update_auto_incr_val_from_field();
   }
 
   int rc = HA_EXIT_SUCCESS;
@@ -10011,7 +10105,9 @@ int ha_rocksdb::delete_table(const char *const tablename) {
     Remove the table entry in data dictionary (this will also remove it from
     the persistent data dictionary).
   */
-  ddl_manager.remove(tbl, batch, true);
+  const GL_INDEX_ID gl_index_id = tbl->get_pk()->get_gl_index_id();
+  ddl_manager.remove(tbl, batch, gl_index_id, true);
+
   int err = dict_manager.commit(batch);
   if (err) {
     DBUG_RETURN(err);
@@ -10124,6 +10220,7 @@ int ha_rocksdb::rename_table(const char *const from, const char *const to) {
   const std::unique_ptr<rocksdb::WriteBatch> wb = dict_manager.begin();
   rocksdb::WriteBatch *const batch = wb.get();
   dict_manager.lock();
+
   if (ddl_manager.rename(from_str, to_str, batch)) {
     rc = HA_ERR_NO_SUCH_TABLE;
   } else {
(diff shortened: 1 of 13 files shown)


mysql-test/suite/rocksdb/t/autoincrement_crash_safe.test
```diff
@@ -0,0 +1,30 @@
...
+
+--exec echo "restart" > $MYSQLTEST_VARDIR/tmp/mysqld.1.expect
+SET SESSION debug="+d,crash_during_online_index_creation";
+--error 2013
+ALTER TABLE t ADD INDEX kb(b), ALGORITHM=INPLACE;
+
+--enable_reconnect
+--source include/wait_until_connected_again.inc
+
+SET SESSION debug="-d,crash_during_online_index_creation";
```
laurynas-biveinis: Server is freshly restarted here, no need to remove crash_during_online_index_creation from the session variable

storage/rocksdb/ha_rocksdb.h
```diff
@@ -598,8 +601,8 @@ class ha_rocksdb : public my_core::handler {
                             rowid_size, skip_lookup, skip_ttl_check);
   }
 
-  void update_auto_incr_val();
   void load_auto_incr_value();
+  int update_auto_increment();
```
vasilep: Like we discussed handler::update_auto_increment() is not virtual, so this is not a kosher overriding. I suggest keeping the old name update_auto_incr_val() of the method.

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -459,6 +459,7 @@ static uint32_t rocksdb_table_stats_sampling_pct;
 static my_bool rocksdb_enable_bulk_load_api = 1;
 static my_bool rocksdb_print_snapshot_conflict_queries = 0;
 static my_bool rocksdb_large_prefix = 0;
+static my_bool rocksdb_persist_auto_increment = 0;
```
vasilep: hmm, I even wonder if we should really have this sysvar since this diff is almost a bug fix rather than a feature. if this diff is tested properly, we might not want to pollute the code with extra knobs.

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -1682,6 +1690,9 @@ class Rdb_transaction {
   */
   virtual bool is_writebatch_trx() const = 0;
 
+  virtual rocksdb::Status Merge(rocksdb::ColumnFamilyHandle *cf,
```
vasilep: let's stick with the naming convention of MyRocks, rather than RocksDB, hence use `merge()` as the method name, since this class is not derived from any RocksDB class

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -4189,8 +4210,32 @@ void ha_rocksdb::load_auto_incr_value() {
...
+  dict_manager.lock();
+  auto scf = dict_manager.get_system_cf();
+  uchar key_buf[Rdb_key_def::INDEX_NUMBER_SIZE * 3] = {0};
+
+  dict_manager.get_key_from_index_id((uchar *)key_buf,
+                                     m_pk_descr->get_gl_index_id());
+
+  auto k = rocksdb::Slice((char *)key_buf, sizeof(key_buf));
+  dict_manager.unlock();
+  std::string val;
```
vasilep: add new line after unlock();

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -4189,8 +4210,32 @@ void ha_rocksdb::load_auto_incr_value() {
...
+  }
+
+  
+  longlong auto_incr_val = m_tbl_def->m_auto_incr_val;
+  while (auto_incr_val < new_val &&
+         !m_tbl_def->m_auto_incr_val.compare_exchange_weak(auto_incr_val,
+                                                           new_val)) {
+    // Do nothing - just loop until auto_incr_val is >= new_val or
+    // we successfully set it
+  }
```
vasilep: you could extract the test & set loop into a helper method that could be also called later in `ha_rocksdb::update_pk()`

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -9675,6 +9735,8 @@ int ha_rocksdb::rename_table(const char *const from, const char *const to) {
   const std::unique_ptr<rocksdb::WriteBatch> wb = dict_manager.begin();
   rocksdb::WriteBatch *const batch = wb.get();
   dict_manager.lock();
+  
+
```
vasilep: no double NLs please

storage/rocksdb/rdb_cf_options.h
```diff
@@ -64,6 +64,9 @@ class Rdb_cf_options {
   static const rocksdb::Comparator *
   get_cf_comparator(const std::string &cf_name);
 
+  std::shared_ptr<rocksdb::MergeOperator>
+  get_cf_merge_operator(const std::string & cf_name);
```
vasilep: static method?

storage/rocksdb/rdb_datadic.cc
```diff
@@ -4071,10 +4082,20 @@ bool Rdb_ddl_manager::rename(const std::string &from, const std::string &to,
...
   memcpy(new_buf + new_pos, dbname_tablename.c_str(), dbname_tablename.size());
   new_pos += dbname_tablename.size();
+  
+  std::shared_ptr<Rdb_key_def> found_key;
+  for (uint i = 0; i < new_rec -> m_key_count; i++) {
+    auto k = new_rec -> m_key_descr_arr[i];
+    if (k -> m_index_type == Rdb_key_def::INDEX_TYPE_PRIMARY) {
+      found_key = k;
+    }
+  }
```
vasilep: How about make this into a new method `Rdb_tbl_def::get_pk()`?

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -9567,7 +9616,18 @@ int ha_rocksdb::delete_table(const char *const tablename) {
     Remove the table entry in data dictionary (this will also remove it from
     the persistent data dictionary).
   */
-  ddl_manager.remove(tbl, batch, true);
+  std::shared_ptr<Rdb_key_def> found_key;
```
vasilep: How about make this into a new method `Rdb_tbl_def::get_pk()`?

storage/rocksdb/rdb_datadic.h
```diff
@@ -42,6 +42,48 @@ class Rdb_field_packing;
 class Rdb_cf_manager;
 class Rdb_ddl_manager;
 
+class Rdb_merge_operator : public rocksdb::AssociativeMergeOperator {
+ public:
+  virtual bool Merge(const rocksdb::Slice &key,
```
vasilep: you can safely drop the "virtual" keyword if you have "override" ;)

storage/rocksdb/rdb_datadic.h
```diff
@@ -1258,6 +1303,12 @@ class Rdb_dict_manager {
                     const uint &cf_flags) const;
   bool get_cf_flags(const uint &cf_id, uint *const cf_flags) const;
 
+  uint64_t get_autoinc(const GL_INDEX_ID &gl_index_id) const;
```
vasilep: I don't see where get_autoinc() is implemented

vasilep (changes requested):
Mostly cosmetic stuff from me. You'll get the real feedback from Herman & Manuel. :)

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -4331,8 +4352,32 @@ void ha_rocksdb::load_auto_incr_value() {
   const bool save_keyread_only = m_keyread_only;
   m_keyread_only = true;
 
-  if (!index_last(table->record[0]))
-    update_auto_incr_val();
+  dict_manager.lock();
```
hermanlee: Reads from the dictionary manager shouldn't require locking.

storage/rocksdb/rdb_datadic.cc
```diff
@@ -4685,6 +4706,27 @@ bool Rdb_dict_manager::get_cf_flags(const uint32_t &cf_id,
   return found;
 }
 
+void Rdb_dict_manager::get_key_from_index_id(
```
hermanlee: Is this function necessary? Wherever it's used, can dump_index_id(key_buf, Rdb_key_def::AUTO_INC, gl_index_id) be called directly?

storage/rocksdb/ha_rocksdb.cc
hermanlee: Once the code is enabled with the new autoincrement scheme, it cannot be disabled. If it's disabled, the autoincrement key will contain stale values, so if the system goes down and comes up again with this setting enabled, the server could end up using smaller values of the autoincrement count.

The code would always need to check to see if the autoincrement key is available to decide if it should ignore this variable. It would probably be better to remove this variable and support the new scheme immediately.

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -2410,6 +2421,11 @@ class Rdb_transaction_impl : public Rdb_transaction {
     }
   }
 
+  rocksdb::Status Merge(rocksdb::ColumnFamilyHandle *cf,
+                        const rocksdb::Slice &key, std::string val) override {
+    return m_rocksdb_tx->Merge(cf, key, val);
```
hermanlee: Could you also add a m_merge_count variable to track the number of times merge is called on a key?

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -4331,8 +4352,32 @@ void ha_rocksdb::load_auto_incr_value() {
...
+                                     m_pk_descr->get_gl_index_id());
+
+  auto k = rocksdb::Slice((char *)key_buf, sizeof(key_buf));
+  dict_manager.unlock();
+  std::string val;
+  tx->get(scf, k, &val);
+  longlong new_val =
+      rdb_netbuf_to_uint64(reinterpret_cast<const uchar *>(val.c_str()));
+    
+  if (new_val != std::numeric_limits<longlong>::max()) {
```
hermanlee: This code is redundant with the new code added to update_pk() for swapping in the value. In one case, the source of the new value comes from the auto increment key you've fetched, in the other case, it's coming from the field. You could modify update_auto_incr_val() to take an argument to indicate whether the source is coming from the key or from the field instead of copying the code.

storage/rocksdb/ha_rocksdb.cc
hermanlee: Yes, the code was originally in update_auto_incr_val().

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -6420,7 +6435,7 @@ int ha_rocksdb::create(const char *const name, TABLE *const table_arg,
   m_pk_descr = m_key_descr_arr[pk_index(table_arg, m_tbl_def)];
 
   dict_manager.lock();
-  err = ddl_manager.put_and_write(m_tbl_def, batch);
+  err = ddl_manager.put_and_write(m_tbl_def, batch, table_arg->found_next_number_field);
```
hermanlee: The put_and_write() call is used to construct the data dictionary keys and store them in the writebatch. Instead of passing in a side parameter, the m_tbl_def's m_auto_incr_val should be updated to the appropriate value and the put(tbl) function should be modified to generate the key.

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -6437,20 +6452,6 @@ int ha_rocksdb::create(const char *const name, TABLE *const table_arg,
   if (create_info->auto_increment_value)
     m_tbl_def->m_auto_incr_val = create_info->auto_increment_value;
 
-  /*
-    We only support auto_increment at start of the PRIMARY KEY.
-  */
-  // Field *field;
-  // if ((field= table_arg->next_number_field))
-  /* TODO mdcallag: disable this for now to let UNIQUE indexes kind of work
```
hermanlee, line 6600: What issue was this code dealing with and how has it been resolved?

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -8085,6 +8086,29 @@ int ha_rocksdb::write_row(uchar *const buf) {
   DBUG_RETURN(rv);
 }
 
+rocksdb::Status ha_rocksdb::merge_auto_increment(int curr_id) {
+  Rdb_transaction *const tx = get_or_create_tx(table->in_use);
+
+  dict_manager.lock();
```
hermanlee: lock shouldn't be needed here.

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -8085,6 +8086,29 @@ int ha_rocksdb::write_row(uchar *const buf) {
...
+  auto scf = dict_manager.get_system_cf();
+  uchar key_buf[Rdb_key_def::INDEX_NUMBER_SIZE * 3] = {0};
+
+  dict_manager.get_key_from_index_id((uchar *)key_buf,
+                                     m_pk_descr->get_gl_index_id());
+
+  auto k = rocksdb::Slice((char *)key_buf, sizeof(key_buf));
+  dict_manager.unlock();
+
+  uchar value_buf[256] = {0};
```
hermanlee: Seems like this buffer could just be 8 bytes to match the uint64. Also, this code is the same as the Deserialize(), so this should call the helper function to convert curr_id into the format to be stored in the value of the merge key.

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -8467,7 +8491,28 @@ int ha_rocksdb::update_pk(const Rdb_key_def &kd,
   }
 
   if (table->next_number_field) {
-    update_auto_incr_val();
+    longlong new_val;
```
hermanlee: This code should belong in update_auto_incr_val().

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -9779,7 +9824,18 @@ int ha_rocksdb::delete_table(const char *const tablename) {
...
+  std::shared_ptr<Rdb_key_def> found_key;
+  for (uint i = 0; i < tbl -> m_key_count; i++) {
+    auto k = tbl -> m_key_descr_arr[i];
+    if (k -> m_index_type == Rdb_key_def::INDEX_TYPE_PRIMARY) {
+      found_key = k;
+    }
+  }
+
+  DBUG_ASSERT(found_key != nullptr);
+  ddl_manager.remove(tbl, batch, found_key->get_gl_index_id(), true);
```
hermanlee: It seems extending this function to take an extra argument for the auto increment key is not the best approach. Could the check for the pk occur within the remove function, since the tbl is already being passed into the ddl_manager.remove() function?

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -10962,7 +11020,8 @@ bool ha_rocksdb::commit_inplace_alter_table(
         create_index_ids.insert(index->get_gl_index_id());
       }
 
-      if (ddl_manager.put_and_write(ctx->m_new_tdef, batch)) {
+      if (ddl_manager.put_and_write(ctx->m_new_tdef, batch,
+                                    altered_table->found_next_number_field)) {
```
hermanlee: Again, this seems like it's extraneous with the m_new_tdef being passed in?

storage/rocksdb/rdb_cf_manager.cc
```diff
@@ -94,7 +94,7 @@ Rdb_cf_manager::get_or_create_cf(rocksdb::DB *const rdb,
     /* Create a Column Family. */
     rocksdb::ColumnFamilyOptions opts;
     m_cf_options->get_cf_options(cf_name, &opts);
-
+
```
hermanlee: Remove extra space.

storage/rocksdb/rdb_cf_options.cc
```diff
@@ -323,6 +323,16 @@ Rdb_cf_options::get_cf_comparator(const std::string &cf_name) {
   }
 }
 
+std::shared_ptr<rocksdb::MergeOperator>
+Rdb_cf_options::get_cf_merge_operator(const std::string & cf_name) {
+  if (cf_name == DEFAULT_SYSTEM_CF_NAME) {
+    return std::make_shared<Rdb_merge_operator>();
+  }
```
hermanlee: } else {

or

return cf_name == DEFAULT_SYSTEM_CF_NAME ? std::make_shared<Rdb_merge_operator>() : nullptr;

storage/rocksdb/rdb_datadic.cc
```diff
@@ -4685,6 +4706,27 @@ bool Rdb_dict_manager::get_cf_flags(const uint32_t &cf_id,
...
+    uchar *key_buf, const GL_INDEX_ID &gl_index_id) const {
+  dump_index_id(key_buf, Rdb_key_def::AUTO_INC, gl_index_id);
+}
+
+void Rdb_dict_manager::add_index_autoinc_mapping(
```
hermanlee: Can we get away with just merge operators instead of adding a put key here?

storage/rocksdb/rdb_datadic.cc
```diff
@@ -4094,6 +4100,11 @@ void Rdb_ddl_manager::remove(Rdb_tbl_def *const tbl,
 
   const rocksdb::Slice tkey((char *)buf, pos);
   m_dict->delete_key(batch, tkey);
+  
+  uchar key_buf[Rdb_key_def::INDEX_NUMBER_SIZE * 3] = {0};
+  m_dict -> get_key_from_index_id((uchar *) key_buf, gl_index_id);
```
hermanlee: no space before and after ->

storage/rocksdb/rdb_datadic.h
```diff
@@ -42,6 +42,48 @@ class Rdb_field_packing;
...
+                     rocksdb::Logger *logger) const override {
+    
+    DBUG_ASSERT(new_value != nullptr);
+    std::string merged_value;
+
+    /* if no existing vaue, then persist the passed in value */
+    if (existing_value == nullptr) {
+      merged_value = Serialize(value);
+    } else {
+      merged_value = Deserialize(std::max(Serialize(value), Serialize(*existing_value)));
```
hermanlee: Serialize typically means encode for storing on media/networking. Deserialize typically means decode, so I think you have the opposite here meaning here.

Also, this code looks wrong? In one case, merged_value is the product of a Serialize() call. In the second case, it is the product of a Deserialize() call.

storage/rocksdb/rdb_datadic.h
```diff
@@ -42,6 +42,48 @@ class Rdb_field_packing;
...
+ public:
+  virtual bool Merge(const rocksdb::Slice &key,
+                     const rocksdb::Slice *existing_value,
+                     const rocksdb::Slice &value, std::string *new_value,
+                     rocksdb::Logger *logger) const override {
+    
+    DBUG_ASSERT(new_value != nullptr);
+    std::string merged_value;
+
+    /* if no existing vaue, then persist the passed in value */
```
hermanlee: Since the merge operator could be extended for other operations, it make sense to determine first which merge operation to perform based on the key. You'll want to check that the key starts with the proper index type first and ignore anything that it doesn't understand, or fail.

We also spoke about adding a versioning scheme to the value to handle extending the format in the future.

storage/rocksdb/rdb_datadic.h
```diff
@@ -42,6 +42,48 @@ class Rdb_field_packing;
...
+      merged_value = Serialize(value);
+    } else {
+      merged_value = Deserialize(std::max(Serialize(value), Serialize(*existing_value)));
+    }
+
+    *new_value = merged_value;
+    return true;
+  }
+
+  virtual const char *Name() const override { return "Rdb_merge_operator"; }
```
hermanlee: Can you add "cf_system" or something along those lines so that it's clear it's part of the system column family? Or maybe make the Rdb_merge_operator class name contain system or sys as part of its name, so it's clear this is used for the system column family?

storage/rocksdb/rdb_datadic.h
```diff
@@ -42,6 +42,48 @@ class Rdb_field_packing;
...
+
+ private:
+  std::string Deserialize(const uint64_t data) const {
+    uchar value_buf[256] = {0};
```
hermanlee: It might be simpler to allocate just 8 bytes for the uint64_t, instead of 256 bytes. Then you don't need to do the whole ptr += and ptr - value_buf calculations.

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -8467,7 +8491,28 @@ int ha_rocksdb::update_pk(const Rdb_key_def &kd,
...
   if (table->next_number_field) {
-    update_auto_incr_val();
+    longlong new_val;
+    Field *field;
+    field = table->key_info[table->s->next_number_index].key_part[0].field;
+
+    my_bitmap_map *const old_map =
+        dbug_tmp_use_all_columns(table, table->read_set);
+    new_val = field->val_int();
+    const rocksdb::Status stat = merge_auto_increment(new_val);
```
hermanlee: Small optimization: is there a way to delay calling merge_auto_increment until commit time and use the highest value for this transaction? Otherwise, we increase the number of keys that need to be added via the write batch.

Does 

```
create table t1 (a int primary key auto_increment) engine=rocksdb;
Insert into table t1 values (NULL), (NULL), (NULL), (NULL);
```

End up calling merge_auto_increment(new_val) four types?

hermanlee (changes requested):
Will there be support in information schema for dumping out the contents of these entries?

storage/rocksdb/ha_rocksdb.cc
armishra: I did exactly this earlier, but @lth suggested migrating it out. I'll reconcile this with him later.

storage/rocksdb/ha_rocksdb.cc
armishra: yes currently it would call merge_auto_increment 4 times. Will investigate into this optimization.

storage/rocksdb/rdb_datadic.cc
armishra: That would be the nicest! However dump_index_id is private

hermanlee: The places where you reference dump_index_id() should probably be inside the Rdb_dict_manager then. The keys generated for reading/writing the __system__ column family should in Rdb_dict_manager, and you can add an API to return the auto increment value, rather than return the key and require the caller to perform the read.

storage/rocksdb/rdb_datadic.h
```diff
@@ -1354,4 +1368,82 @@ struct Rdb_index_info {
...
+             rocksdb::Logger *logger) const override {
+
+    /* value corruption check and version check */
+    DBUG_ASSERT(new_value != nullptr);
+    DBUG_ASSERT(FindVersion(value) == Rdb_key_def::AUTO_INCREMENT_VERSION);
+
+    std::string merged_value;
+
+    /* if no existing vaue, then persist the passed in value */
+    if (existing_value == nullptr) {
```
hermanlee: Should check the key type matches what is expected, and the version in the value matches the expected version, before performing the merge operations here.

storage/rocksdb/rdb_datadic.h
```diff
@@ -1354,4 +1368,82 @@ struct Rdb_index_info {
...
+      merged_value =
+          Serialize(std::max(Deserialize(value), Deserialize(*existing_value)));
+    }
+
+    *new_value = merged_value;
+    return true;
+  }
+
+  virtual const char *Name() const override {
+    return "system_cf_auto_increment_merge";
```
hermanlee: I was thinking the class itself: Rdb_merge_operator, is specific to the __system__ column family. Maybe it can be renamed too? Rdb_system_merge_op. Adding "auto_increment" is probably too specific here. The merge operator can also be used to perform other types of updates with in the __system__ column family. Auto increment is just one of them.

Hence, why a key check should be performed within the Merge() routine.

storage/rocksdb/rdb_datadic.h
```diff
@@ -1354,4 +1368,82 @@ struct Rdb_index_info {
...
+  }
+
+ private:
+  std::string Serialize(const uint64_t data) const {
```
hermanlee: I think Serialize can be slightly optimized so that we minimize copies. The value is being copied from the value_buf memory to the new_value string. Can the value be assembled directly in *new_value instead of going through the value_buf array?

storage/rocksdb/rdb_datadic.h
```diff
@@ -1354,4 +1368,82 @@ struct Rdb_index_info {
...
+    @param[OUT] new_value       new value after Merge
+    @param[IN]  logger
+  */
+  bool Merge(const rocksdb::Slice &key, const rocksdb::Slice *existing_value,
+             const rocksdb::Slice &value, std::string *new_value,
+             rocksdb::Logger *logger) const override {
+
+    /* value corruption check and version check */
+    DBUG_ASSERT(new_value != nullptr);
+    DBUG_ASSERT(FindVersion(value) == Rdb_key_def::AUTO_INCREMENT_VERSION);
```
hermanlee: It seems you would also need to check the version of the existing_value too, no?

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -4331,8 +4356,26 @@ void ha_rocksdb::load_auto_incr_value() {
   const bool save_keyread_only = m_keyread_only;
   m_keyread_only = true;
 
-  if (!index_last(table->record[0]))
```
hermanlee: I think you still need this codepath to handle upgrades correctly. There should be a test case for upgrade a server containing a table that has several rows from a build without the auto_increment data dictionary key to one with the auto_increment data dictionary key.

I.e.

```
/* load old build */
create table t1 (a int primary key auto increment)) engine=rocksdb;
insert into t1 values (NULL), (NULL);
/* shutdown, load new build */
insert into t1 values (NULL), (NULL);
```

You'll want to verify when the new build starts up, the inserts will use the correct auto_incr value, instead of resetting it to 0, which is what happens below.

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -4331,8 +4356,26 @@ void ha_rocksdb::load_auto_incr_value() {
...
+                                     found_key->get_gl_index_id());
+  auto k = rocksdb::Slice((char *)key_buf, sizeof(key_buf));
+
+  std::string val;
+  auto stat = tx->get(scf, k, &val);
+  if (stat.ok()) {
+    const uchar *const value = reinterpret_cast<const uchar *>(val.c_str());
+    new_val = rdb_netbuf_to_uint64(value + RDB_SIZEOF_AUTO_INCREMENT_VERSION);
+  } else {
+    new_val = 0;
```
hermanlee: If the key is not found, the new_val should come from the loading index_last().

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -4331,8 +4356,26 @@ void ha_rocksdb::load_auto_incr_value() {
...
 
-  if (!index_last(table->record[0]))
-    update_auto_incr_val();
+  auto found_key = m_tbl_def->get_pk();
+
+  auto scf = dict_manager.get_system_cf();
+
+  // Construct the key of the data dictionary to get the auto_increment value
+  uchar key_buf[Rdb_key_def::INDEX_NUMBER_SIZE * 3] = {0};
+  dict_manager.get_key_from_index_id((uchar *)key_buf,
```
hermanlee: I think this code for fetching and writing the contents of the auto_incr key should reside within dict_manager. The dict_manager is meant to handle putting/deleting keys to the system column family. The merge operations should go through dict_manager instead of being pulled out here.

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -4350,27 +4393,24 @@ void ha_rocksdb::load_auto_incr_value() {
...
+  } else {
+    new_val = field->val_int();
+
+    // Merge the current in memory value into the auto_increment bookkeeping
+    // in the data dictionary
+    merge_auto_increment(new_val);
+  }
 
-  my_bitmap_map *const old_map =
-      dbug_tmp_use_all_columns(table, table->read_set);
```
hermanlee: Is this code not needed for !use_datadic case above? It was originally here for handling index_last(). Is it not needed anymore?

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -8085,6 +8111,29 @@ int ha_rocksdb::write_row(uchar *const buf) {
...
+rocksdb::Status ha_rocksdb::merge_auto_increment(int curr_id) {
+  Rdb_transaction *const tx = get_or_create_tx(table->in_use);
+
+  auto scf = dict_manager.get_system_cf();
+  uchar key_buf[Rdb_key_def::INDEX_NUMBER_SIZE * 3] = {0};
+  dict_manager.get_key_from_index_id((uchar *)key_buf,
+                                     m_pk_descr->get_gl_index_id());
+  auto k = rocksdb::Slice((char *)key_buf, sizeof(key_buf));
+
+  uchar value_buf[RDB_SIZEOF_AUTO_INCREMENT_VERSION +
```
hermanlee: This code looks like the same as Serialize(). Is there way to refactor the code so that they share a common function?

storage/rocksdb/rdb_datadic.cc
```diff
@@ -4095,6 +4110,11 @@ void Rdb_ddl_manager::remove(Rdb_tbl_def *const tbl,
   const rocksdb::Slice tkey((char *)buf, pos);
   m_dict->delete_key(batch, tkey);
 
+  uchar key_buf[Rdb_key_def::INDEX_NUMBER_SIZE * 3] = {0};
+  m_dict->get_key_from_index_id((uchar *)key_buf, gl_index_id);
+  auto auto_increment_key = rocksdb::Slice((char *)key_buf, sizeof(key_buf));
+  m_dict->delete_key(batch, auto_increment_key);
```
hermanlee: I thought we discussed Rdb_dict_manager::delete_index_info() would be a better place to remove the auto_increment key. There would be no need to add the gl_index_id argument to this function.

hermanlee (changes requested):

storage/rocksdb/ha_rocksdb.cc
armishra: the implementation of the Rdb_transaction class is defined in ha_rocksdb.cc and Rdb_dict_manager is unable to access the Rdb_transaction's merge() and get(), so the best solution would be to somehow extract the key from the Rdb_dict_manager and call merge() or get() in ha_rocksdb. This would either require some wrapper around dump_index_id or extension of dump_index_id to the public section.

hermanlee: Rdb_binlog_manager does something similar with the system column family for each Rdb_transaction during commit(). It pulls out the writebatch of the Rdb_transaction class and passes it as an argument to Rdb_binlog_manager. The binlog_manager then writes the key/value via writebatch->put(). The writebatch also has a merge() interface. My original thought was to record the highest auto increment value generated for each gl_index_id/table and store it local to Rdb_transaction. On commit, these merge keys can be added to the write batch.

If you decide to go with pulling out the key to ha_rocksdb.cc for reading and merging, the functions should be consolidated into common routines that the different call sites invoke. This routine reading the key should also perform the version checks, etc.

lth:
I'll open a new PR, since I can't import this due to CLA.

PR #695, opened 2017-08-29
Add stall counters to 'SHOW GLOBAL STATUS' output.

Tema:
@update-submodule: rocksdb

Summary:
The stall counters are already shown in 'SHOW ENGINE ROCKSDB STATUS
output' but this command is a bit expensive to run and you have to parse
the output to monitor these values. This branch adds these counters to
structured output of SHOW GLOBAL STATUS command. 
Also fixed the bug in INFORMATION_SCHEMA.ROCKSDB_COMPACTION_STATS  which showed only integer values but not float.

The rocksdb module was updated to bring [this](https://github.com/facebook/rocksdb/pull/2794#pullrequestreview-59667574) change  which has the following changes to GetMapProperty method:
1. Changed values from double to string
2. Change existing keys to start with "compaction." prefix for all keys returned by old version of GetMapProperty.
3. Added stall values with keys starting with "io_stall." prefix

Test plan:
* Check that compaction level metrics are still published to
INFORMATION_SCHEMA.ROCKSDB_COMPACTION_STATS properly including real double values.
BEFORE:
| __system__ | L0    | Rnp1GB         |     0 |
| __system__ | L0    | Score          |     0 |
| __system__ | L0    | SizeBytes      |  1433 |
AFTER:
| __system__ | L0    | Rnp1GB         |     0 |
| __system__ | L0    | Score          |  0.25 |
| __system__ | L0    | SizeBytes      |  1433 |
* Check the output of the command:
```
mysql> show status like 'rocksdb_stall_%';
+-----------------------------------------------------+-------+
| Variable_name                                       | Value |
+-----------------------------------------------------+-------+
| rocksdb_stall_level0_slowdown                       | 2     |
| rocksdb_stall_level0_slowdown_with_compaction       | 4     |
| rocksdb_stall_level0_numfiles                       | 10    |
| rocksdb_stall_level0_numfiles_with_compaction       | 12    |
| rocksdb_stall_stop_for_pending_compaction_bytes     | 16    |
| rocksdb_stall_slowdown_for_pending_compaction_bytes | 14    |
| rocksdb_stall_memtable_compaction                   | 6     |
| rocksdb_stall_memtable_slowdown                     | 8     |
| rocksdb_stall_total_stop                            | 32    |
| rocksdb_stall_total_slowdown                        | 24    |
| rocksdb_stall_micros                                | 0     |
+-----------------------------------------------------+-------+
11 rows in set (0.00 sec)
```
(In order to have some non-zero numbers, I've initialized internal
RocksDB stall stats with 1-8 numbers. They all doubled in output as
there are two column families in my instance: default and system.
The stall_micros has been introduced previously. total_count doesn't
include level0_slowdown_with_compaction and numfiles_with_compaction
as only their total counterparts are included in this counter.)


storage/rocksdb/ha_rocksdb.cc
```diff
@@ -11106,6 +11107,81 @@ static void show_myrocks_vars(THD *thd, SHOW_VAR *var, char *buff) {
   var->value = reinterpret_cast<char *>(&myrocks_status_variables);
 }
 
+static ulonglong io_stall_prop_value(std::map<std::string, std::string> *props,
```
laurynas-biveinis: 1st arg: const ref, 2nd arg: rvalue ref?

Tema: Sure. Not sure how to make a **const**, though, as I use iterator inside.

Tema: Ok, I think I've figured out how to do that in my latest revision. Thanks.

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -11208,6 +11209,84 @@ static void show_myrocks_vars(THD *thd, SHOW_VAR *var, char *buff) {
   var->value = reinterpret_cast<char *>(&myrocks_status_variables);
 }
 
+static ulonglong io_stall_prop_value(std::map<std::string, std::string> &props,
+                                     std::string &&key) {
```
hermanlee: This can probably be a const std::string &key I think. There's probably no need for this to be a &&.

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -11208,6 +11209,85 @@ static void show_myrocks_vars(THD *thd, SHOW_VAR *var, char *buff) {
...
+      props.find("io_stalls." + key);
+  if (iter != props.end()) {
+    return std::stoull(iter->second);
+  } else {
+    return 0;
+  }
+}
+
+static void update_rocksdb_stall_status() {
+  io_stall_stats = {};
```
hermanlee: Initializing io_stall_stats and then adding counters to it might cause strange behavior if the command is run concurrently. It might be better to not initialize io_stall_stats and set the members equal to the value returned by rocksdb, similar to the other status variables in memory_usage or export stats.

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -11208,6 +11209,85 @@ static void show_myrocks_vars(THD *thd, SHOW_VAR *var, char *buff) {
   var->value = reinterpret_cast<char *>(&myrocks_status_variables);
 }
 
+static ulonglong
+io_stall_prop_value(const std::map<std::string, std::string> &props,
+                    std::string &&key) {
```
hermanlee: I think this can just be const std::string &key.

storage/rocksdb/rdb_i_s.cc
```diff
@@ -845,17 +845,22 @@ static int rdb_i_s_compact_stats_fill_table(
       continue;
     }
 
-    std::map<std::string, double> props;
+    std::map<std::string, std::string> props;
```
hermanlee, line 848: What additional keys are returned when switching to the string version of GetMapProperty? The information schema table is still returning a double though, so the additional keys would need to have values that are doubles, which seems odd that they're not returned by the double version of GetMapProperty in the first place.

hermanlee (changes requested):

storage/rocksdb/ha_rocksdb.cc
Tema: Not sure if it possible as I pass a string directly to the method call, like 
io_stall_stats.level0_slowdown += io_stall_prop_value(props, "level0_slowdown");

LMK if I miss something.

storage/rocksdb/ha_rocksdb.cc
Tema: It is a bit different than memory_usage or export stats methods as I need to loop through multiple calls to accumulate the final result. But I see you concern, so I can create a local struct then copy it to the global one at the end of method call.

storage/rocksdb/rdb_i_s.cc
Tema: Now all keys are returned with some prefixes. The old values now have "compaction." key prefix and new ones have "io_stalls." prefix. This method now skips keys with prefixes other than "compaction."

I've tested INFORMATION_SCHEMA.ROCKSDB_COMPACTION_STATS table and see that it contains the same data it contained before.

storage/rocksdb/ha_rocksdb.cc
hermanlee: I tested this before when I saw your call with arguments like "level0_slowdown" and it looks to work with const std::string &key. Not clear to me if there is preference for one or the other.

storage/rocksdb/rdb_i_s.cc
hermanlee: What would be returned in I_S.COMPACTION_STATS without this change? Does calling GetMapProperty() using std::map<std::string, std::double> not return all of the keys with "compaction." key prefix?

storage/rocksdb/ha_rocksdb.cc
Tema: I've tested it again and it turned out my original test missed 'const'. If I use it then it works indeed. I will change it according to your comment.

Tema: I've updated summary to highlight what rocksdb changes were done relevant to this PR.

storage/rocksdb/rdb_i_s.cc
Tema: I've updated summary to highlight what rocksdb changes were done relevant to this PR.

storage/rocksdb/rdb_i_s.cc
```diff
@@ -845,17 +845,22 @@ static int rdb_i_s_compact_stats_fill_table(
...
     bool bool_ret MY_ATTRIBUTE((__unused__));
     bool_ret = rdb->GetMapProperty(cfh, "rocksdb.cfstats", &props);
     DBUG_ASSERT(bool_ret);
 
     for (auto const &prop_ent : props) {
+      std::string prefix = "compaction.";
```
hermanlee: Since this doesn't change, can probably declare this outside the for loop and make it const.

storage/rocksdb/rdb_i_s.cc
```diff
@@ -864,7 +869,7 @@ static int rdb_i_s_compact_stats_fill_table(
       field[0]->store(cf_name.c_str(), cf_name.size(), system_charset_info);
       field[1]->store(level_str.c_str(), level_str.size(), system_charset_info);
       field[2]->store(type_str.c_str(), type_str.size(), system_charset_info);
-      field[3]->store(value, true);
+      field[3]->store(std::stod(value));
```
hermanlee, line 872: Did you intend to remove the true indicating the value is unsigned?

mysql-test/suite/rocksdb/r/rocksdb.result
```diff
@@ -1595,6 +1615,16 @@ ROCKSDB_WRITE_WAL
 #  but they are shown as both session and global, like InnoDB's status vars.
 select VARIABLE_NAME from INFORMATION_SCHEMA.session_status where VARIABLE_NAME LIKE 'rocksdb%';
```
hermanlee: If there are non-deterministic results from this query, you can add an order by clause to this select.

hermanlee (approved):
Looks good overall.

storage/rocksdb/rdb_i_s.cc
Tema: Yes, the overridden method with boolean accept longlong but the value we pass is actually double. Essentially this diff fixes a bug here. Please see a Test Plan note in "Check that compaction level metrics are still published to
INFORMATION_SCHEMA.ROCKSDB_COMPACTION_STATS properly including real double values" section where previously the Score was displayed as 0 and now it is 0.25.

PR #698, opened 2017-08-30
fixed valgrind uninitialized value and segfault

armishra:
Changes:
* fixed valgrind errors by consistently typing the save value for the sysvar rocksdb_max_latest_deadlocks
* fixed segfault caused by dropped or nonexistent tables showing up in the engine transaction status query
* restructured tests to not unintentionally corrupt the deadlock buffer from or for other deadlock tests

### DIFF
--- a/mysql-test/suite/rocksdb/t/deadlock_tracking.test
+++ b/mysql-test/suite/rocksdb/t/deadlock_tracking.test
@@ -3,6 +3,9 @@ set @prior_deadlock_detect = @@rocksdb_deadlock_detect;
 set @prior_max_latest_deadlocks = @@rocksdb_max_latest_deadlocks; 
 set global rocksdb_deadlock_detect = on; 
 set global rocksdb_lock_wait_timeout = 10000;
+--echo # Clears deadlock buffer of any prior deadlocks.
+set global rocksdb_max_latest_deadlocks = 0;
+set global rocksdb_max_latest_deadlocks = @prior_max_latest_deadlocks;
 let $engine = rocksdb;
 
 --source include/count_sessions.inc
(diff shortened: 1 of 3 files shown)


storage/rocksdb/ha_rocksdb.cc
```diff
@@ -3114,8 +3114,16 @@ class Rdb_snapshot_status : public Rdb_tx_list_walker {
 
     /* extract table name and index names using the index id */
     std::string table_name = ddl_manager.safe_get_table_name(gl_index_id);
+    if (table_name == "") {
```
hermanlee: In the case where the table name could not be retrieved, maybe you can print out the gl_index_id instead?

Also, can use table_name.empty() instead of comparing it with "".

hermanlee (changes requested):

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -3114,8 +3114,17 @@ class Rdb_snapshot_status : public Rdb_tx_list_walker {
...
+          "NOT FOUND; INDEX_ID: " + std::to_string(gl_index_id.index_id);
+    }
     auto kd = ddl_manager.safe_find(gl_index_id);
-    std::string idx_name = kd->get_name();
+    std::string idx_name;
+    if (kd == nullptr) {
+      idx_name = "NOT FOUND; INDEX_ID: " + std::to_string(gl_index_id.index_id);
+    } else {
+      idx_name = kd->get_name();
+    }
```
asandryh: You can replace lines 3122-3127 with a more succinct
`std::string idx_name = kd ? kd->get_name() : "NOT FOUND; INDEX_ID: " + std::to_string(gl_index_id.index_id);`

mysql-test/suite/rocksdb/t/deadlock_tracking.test
```diff
@@ -3,6 +3,8 @@ set @prior_deadlock_detect = @@rocksdb_deadlock_detect;
 set @prior_max_latest_deadlocks = @@rocksdb_max_latest_deadlocks; 
 set global rocksdb_deadlock_detect = on; 
 set global rocksdb_lock_wait_timeout = 10000;
+set global rocksdb_max_latest_deadlocks = 0;
```
hermanlee: Can you add a comment about why this is done and also update the variable so that it's recognized what setting it to 0 does.

hermanlee (approved):

PR #707, opened 2017-09-12
Reduce memory copy for point queries

Tema:
This branch uses [PinnableSlice](http://rocksdb.org/blog/2017/08/24/pinnableslice.html) to avoid extra memory copy when values are retrieved from RocksDB. It is on the path for all SELECT clauses by primary key.

### DIFF
--- a/storage/rocksdb/ha_rocksdb.cc
+++ b/storage/rocksdb/ha_rocksdb.cc
@@ -2061,10 +2061,10 @@ public:
 
   virtual rocksdb::Status get(rocksdb::ColumnFamilyHandle *const column_family,
                               const rocksdb::Slice &key,
-                              std::string *value) const = 0;
+                              rocksdb::PinnableSlice *const value) const = 0;
   virtual rocksdb::Status
   get_for_update(rocksdb::ColumnFamilyHandle *const column_family,
-                 const rocksdb::Slice &key, std::string *const value,
+                 const rocksdb::Slice &key, rocksdb::PinnableSlice *const value,
                  bool exclusive) = 0;
 
   rocksdb::Iterator *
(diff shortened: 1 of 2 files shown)


storage/rocksdb/ha_rocksdb.cc
```diff
@@ -4895,13 +4898,16 @@ void ha_rocksdb::setup_read_decoders() {
 }
 
 #ifndef NDEBUG
-void dbug_append_garbage_at_end(std::string &on_disk_rec) {
-  on_disk_rec.append("abc");
+void dbug_append_garbage_at_end(rocksdb::PinnableSlice &on_disk_rec) {
+  on_disk_rec.Reset();
```
hermanlee: It seems this doesn't append "abc" to the end of the record, but resets the record to "abc". Could the content of on_disk_rec be copied out first and then added back to the on_disk_rec with "abc" appended to it?

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -2334,18 +2334,20 @@ class Rdb_transaction_impl : public Rdb_transaction {
 
   rocksdb::Status get(rocksdb::ColumnFamilyHandle *const column_family,
                       const rocksdb::Slice &key,
-                      std::string *value) const override {
+                      rocksdb::PinnableSlice *const value) const override {
+    value->Reset();
```
hermanlee: Once convert_from_storage_format() is called, I think m_retrieved_record needs to be Reset() to free up the underlying memory. I don't think there's a guarantee there will be a follow-up get() call.

For a point query such as "select * from table where pk = k", where does m_retrieved_record->Reset() get called?

storage/rocksdb/ha_rocksdb.cc
Tema: Yep, I can do it. Interestingly that all tests passed :)

storage/rocksdb/ha_rocksdb.cc
Tema: There is a ha_handler#reset method which calls m_retrieved_record->Reset(). [The documentation](https://github.com/Tema/mysql-5.6/blob/1f2eb46cc766543e39ae1e4e5e05eabb350d8181/sql/handler.h#L3318-L3323) says that it is called after every statement for all tables used by that statement.
So we call Reset() before each Get() to make sure it is cleared for multiple Get() per statement and rely on handler::reset() to clear the latest Get in the statement.

I looked into resetting pinnable slice after convert_from_storage_format() to make it more robust but given that we can't clear it for blobs as a result buffer [holds the pointer to the blob](https://github.com/Tema/mysql-5.6/blob/b8fee8d0de505084d3a34809c72b6952593a370e/storage/rocksdb/ha_rocksdb.cc#L5050-L5052) ([code](https://github.com/Tema/mysql-5.6/blob/b8fee8d0de505084d3a34809c72b6952593a370e/storage/rocksdb/ha_rocksdb.cc#L4978-L4983)), I think relying on handler::reset() is more robust way to go ahead.

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -8330,7 +8339,7 @@ int ha_rocksdb::check_and_lock_sk(const uint &key_id,
   /*
     psergey-todo: we just need to take lock, lookups not needed:
   */
-  std::string dummy_value;
+  rocksdb::PinnableSlice dummy_value;
```
lth: You can just pass nullptr to `get_for_update`, and it will take care of locking the key without loading a value. (And you can clean the todo above).

Tema: sure

lth (approved):

PR #708, opened 2017-09-14
fix nullifying update_cf_options sysvar

### LINKED ISSUE #696: crash when playing with rocksdb_update_cf_options variable
Noticed a crash when playing with rocksdb_update_cf_options dynamic option.

Options for starting:
`--no-defaults --default-tmp-storage-engine=MyISAM --rocksdb --skip-innodb --default-storage-engine=RocksDB`

TEST CASE:
```
create table t1(a int, primary key(a) comment 'cf1') engine=rocksdb;
SET @@global.rocksdb_update_cf_options = 'cf1={write_buffer_size=8m;target_file_size_base=2m};';
SET @@global.rocksdb_update_cf_options = NULL;
SET @@global.rocksdb_update_cf_options = NULL;
SET @@global.rocksdb_update_cf_options = NULL;
SET @@global.rocksdb_update_cf_options = 'cf1={write_buffer_size=8m;target_file_size_base=2m};';
drop table t1;
SET @@global.rocksdb_update_cf_options = 'cf1={write_buffer_size=8m;target_file_size_base=2m};';
```

GDB INFO:
```
Core was generated by `/ssd/tomislav/mysql-5.6.35-linux-x86_64/bin/mysqld --no-defaults --default-tmp-'.
Program terminated with signal SIGSEGV, Segmentation fault.
#0  __pthread_kill (threadid=<optimized out>, signo=signo@entry=11) at ../sysdeps/unix/sysv/linux/pthread_kill.c:62
62      ../sysdeps/unix/sysv/linux/pthread_kill.c: No such file or directory.
[Current thread is 1 (Thread 0x7f401f9b0700 (LWP 2434))]
(gdb) bt
#0  __pthread_kill (threadid=<optimized out>, signo=signo@entry=11) at ../sysdeps/unix/sysv/linux/pthread_kill.c:62
#1  0x0000000000b2ea17 in my_write_core (sig=sig@entry=11) at /ssd/tomislav/mysql-5.6/mysys/stacktrace.c:424
#2  0x000000000084327e in handle_fatal_signal (sig=11) at /ssd/tomislav/mysql-5.6/sql/signal_handler.cc:239
#3  <signal handler called>
#4  0x0000000000000000 in ?? ()
#5  0x0000000000ecef3b in rocksdb::(anonymous namespace)::ReadOneLine (result=0x7f401f9aa930, has_data=<synthetic pointer>, output=0x7f401f9aaa20, seq_file=0x7f4011c190c0, iss=0x7f401f9aaae0)
    at /ssd/tomislav/mysql-5.6/rocksdb/options/options_parser.cc:218
#6  rocksdb::RocksDBOptionsParser::Parse (this=this@entry=0x7f401f9abd40, file_name="./.rocksdb/OPTIONS-000028.dbtmp", env=env@entry=0x19384a0 <rocksdb::Env::Default()::default_env>,
    ignore_unknown_options=ignore_unknown_options@entry=false) at /ssd/tomislav/mysql-5.6/rocksdb/options/options_parser.cc:256
#7  0x0000000000ecf7b8 in rocksdb::RocksDBOptionsParser::VerifyRocksDBOptionsFromFile (db_opt=..., cf_names=std::vector of length 3, capacity 4 = {...},
    cf_opts=std::vector of length 3, capacity 4 = {...}, file_name="./.rocksdb/OPTIONS-000028.dbtmp", env=env@entry=0x19384a0 <rocksdb::Env::Default()::default_env>,
    sanity_check_level=rocksdb::kSanityLevelExactMatch, ignore_unknown_options=false) at /ssd/tomislav/mysql-5.6/rocksdb/options/options_parser.cc:626
#8  0x0000000000ed104d in rocksdb::PersistRocksDBOptions (db_opt=..., cf_names=std::vector of length 3, capacity 4 = {...}, cf_opts=std::vector of length 3, capacity 4 = {...},
    file_name="./.rocksdb/OPTIONS-000028.dbtmp", env=0x19384a0 <rocksdb::Env::Default()::default_env>) at /ssd/tomislav/mysql-5.6/rocksdb/options/options_parser.cc:101
#9  0x0000000000dc4ceb in rocksdb::DBImpl::WriteOptionsFile (this=this@entry=0x7f401ce57c00, need_mutex_lock=need_mutex_lock@entry=false, need_enter_write_thread=need_enter_write_thread@entry=true)
    at /ssd/tomislav/mysql-5.6/rocksdb/db/db_impl.cc:2373
#10 0x0000000000dc6446 in rocksdb::DBImpl::SetOptions (this=0x7f401ce57c00, column_family=<optimized out>, options_map=std::unordered_map with 2 elements = {...})
    at /ssd/tomislav/mysql-5.6/rocksdb/db/db_impl.cc:501
#11 0x0000000000d6c122 in rocksdb::StackableDB::SetOptions (new_options=std::unordered_map with 2 elements = {...}, column_family_handle=0x7f4011c11160, this=<optimized out>)
    at /ssd/tomislav/mysql-5.6/rocksdb/include/rocksdb/utilities/stackable_db.h:323
#12 myrocks::rocksdb_set_update_cf_options (var_ptr=0x1928cd0 <myrocks::rocksdb_update_cf_options>, save=<optimized out>) at /ssd/tomislav/mysql-5.6/storage/rocksdb/ha_rocksdb.cc:11786
#13 0x00000000008e378f in plugin_var_memalloc_global_update (value=0x7f4011c433a0 "cf1={write_buffer_size=8m;target_file_size_base=2m};", dest=<optimized out>, var=<optimized out>, thd=0x7f4019ee0000)
    at /ssd/tomislav/mysql-5.6/sql/sql_plugin.cc:3126
#14 sys_var_pluginvar::global_update (this=<optimized out>, thd=0x7f4019ee0000, var=<optimized out>) at /ssd/tomislav/mysql-5.6/sql/sql_plugin.cc:3310
#15 0x0000000000841cf6 in sys_var::update (var=0x7f4011c43300, thd=0x7f4019ee0000, this=0x7f401cf8ef80) at /ssd/tomislav/mysql-5.6/sql/set_var.cc:193
#16 set_var::update (this=0x7f4011c43300, thd=0x7f4019ee0000) at /ssd/tomislav/mysql-5.6/sql/set_var.cc:763
#17 0x00000000008429f9 in <lambda(THD*, set_var_base*)>::operator() (__closure=<optimized out>, var=<optimized out>, each_thd=<optimized out>) at /ssd/tomislav/mysql-5.6/sql/set_var.cc:667
#18 <lambda(THD*, set_var_base*)>::_FUN () at /ssd/tomislav/mysql-5.6/sql/set_var.cc:668
#19 process_set_variable_map (func=<optimized out>, map=std::unordered_map with 1 element = {...}, current_thd_id=1, thd=0x7f4019ee0000) at /ssd/tomislav/mysql-5.6/sql/set_var.cc:593
#20 sql_set_variables (thd=thd@entry=0x7f4019ee0000, var_list=var_list@entry=0x7f4019ee40a8) at /ssd/tomislav/mysql-5.6/sql/set_var.cc:668
#21 0x00000000008d7009 in mysql_execute_command (thd=thd@entry=0x7f4019ee0000, statement_start_time=statement_start_time@entry=0x7f401f9ae758, post_parse=post_parse@entry=0x7f401f9ae8a8)
    at /ssd/tomislav/mysql-5.6/sql/sql_parse.cc:4805
#22 0x00000000008db1a8 in mysql_parse (thd=thd@entry=0x7f4019ee0000, rawbuf=<optimized out>, length=<optimized out>, parser_state=parser_state@entry=0x7f401f9af180,
    last_timer=last_timer@entry=0x7f401f9ae8a8, async_commit=async_commit@entry=0x7f401f9ae8a7 "") at /ssd/tomislav/mysql-5.6/sql/sql_parse.cc:7783
#23 0x00000000008dc377 in dispatch_command (command=command@entry=COM_QUERY, thd=thd@entry=0x7f4019ee0000,
    packet=packet@entry=0x7f4019f4a001 "SET @@global.rocksdb_update_cf_options = 'cf1={write_buffer_size=8m;target_file_size_base=2m};'", packet_length=packet_length@entry=95,
    srv_session=srv_session@entry=0x0) at /ssd/tomislav/mysql-5.6/sql/sql_parse.cc:1862
#24 0x00000000008dec18 in do_command (thd=0x7f4019ee0000) at /ssd/tomislav/mysql-5.6/sql/sql_parse.cc:1207
#25 0x000000000089b28c in do_handle_one_connection (thd_arg=<optimized out>) at /ssd/tomislav/mysql-5.6/sql/sql_connect.cc:1142
#26 0x000000000089b519 in handle_one_connection (arg=<optimized out>) at /ssd/tomislav/mysql-5.6/sql/sql_connect.cc:963
#27 0x00007f401e2ac6ba in start_thread (arg=0x7f401f9b0700) at pthread_create.c:333
#28 0x00007f401d73a3dd in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:109
```

Reported here: https://jira.percona.com/browse/MYR-175 also with some more examples and MTR test created by George.
In my case here jemmalloc was LD_PRELOADED, as you may see in MYR-175 without using jemalloc it's a little bit different.

ajkr:
previously, setting this sysvar to null would result in the same memory being double-freed upon the next update to the sysvar. That's because we left `var_ptr` pointing to the string that the caller deletes.

Test Plan:

verified it fixes #696.

```
mysql> create table t1(a int, primary key(a) comment 'cf1') engine=rocksdb;
Query OK, 0 rows affected (0.01 sec)

mysql> SET @@global.rocksdb_update_cf_options = 'cf1={write_buffer_size=8m;target_file_size_base=2m};';
Query OK, 0 rows affected (0.01 sec)

mysql> SET @@global.rocksdb_update_cf_options = NULL;
ERROR 1231 (42000): Variable 'rocksdb_update_cf_options' can't be set to the value of 'NULL'
mysql> SET @@global.rocksdb_update_cf_options = NULL;
ERROR 1231 (42000): Variable 'rocksdb_update_cf_options' can't be set to the value of 'NULL'
mysql> SET @@global.rocksdb_update_cf_options = NULL;
ERROR 1231 (42000): Variable 'rocksdb_update_cf_options' can't be set to the value of 'NULL'
mysql> SET @@global.rocksdb_update_cf_options = 'cf1={write_buffer_size=8m;target_file_size_base=2m};';

mysql> drop table t1;

mysql> SET @@global.rocksdb_update_cf_options = 'cf1={write_buffer_size=8m;target_file_size_base=2m};';
Query OK, 0 rows affected (0.01 sec)
```

### DIFF
--- a/storage/rocksdb/ha_rocksdb.cc
+++ b/storage/rocksdb/ha_rocksdb.cc
@@ -408,6 +408,10 @@ static void rocksdb_set_collation_exception_list(THD *thd,
                                                  void *var_ptr,
                                                  const void *save);
 
+static int rocksdb_validate_update_cf_options(THD *thd,
+                                              struct st_mysql_sys_var *var,
+                                              void *save,
+                                              st_mysql_value *value);
 void rocksdb_set_update_cf_options(THD *thd,
                                    struct st_mysql_sys_var *var,
                                    void *var_ptr,
@@ -1164,7 +1168,8 @@ static MYSQL_SYSVAR_STR(override_cf_options, rocksdb_override_cf_options,
 static MYSQL_SYSVAR_STR(update_cf_options, rocksdb_update_cf_options,
                         PLUGIN_VAR_RQCMDARG | PLUGIN_VAR_MEMALLOC |
                             PLUGIN_VAR_ALLOCATED,
-                        "Option updates per column family for RocksDB", nullptr,
+                        "Option updates per column family for RocksDB",
+                        rocksdb_validate_update_cf_options,
                         rocksdb_set_update_cf_options, nullptr);
 
 enum rocksdb_flush_log_at_trx_commit_type : unsigned int {
@@ -11961,23 +11966,30 @@ static void rocksdb_set_max_background_jobs(THD *thd,
   RDB_MUTEX_UNLOCK_CHECK(rdb_sysvars_mutex);
 }
 
+static int rocksdb_validate_update_cf_options(THD *thd,
+                                              struct st_mysql_sys_var *var,
+                                              void *save,
+                                              st_mysql_value *value) {
+  // just disallow setting to null; otherwise same as default string checker
+  char buff[STRING_BUFFER_USUAL_SIZE];
+  const char *str;
+  int length;
+
+  length= sizeof(buff);
+  if ((str= value->val_str(value, buff, &length)))
+    str= thd->strmake(str, length);
+  *(const char**)save= str;
+  return str == nullptr;
+}
+
 void rocksdb_set_update_cf_options(THD *const /* unused */,
                                    struct st_mysql_sys_var *const /* unused */,
                                    void *const var_ptr,
                                    const void *const save) {
   const char *const val = *static_cast<const char *const *>(save);
-
-  if (!val) {
-    // NO_LINT_DEBUG
-    sql_print_warning("MyRocks: NULL is not a valid option for updates to "
-                      "column family settings.");
-    *reinterpret_cast<char**>(var_ptr) = nullptr;
-    return;
-  }
+  DBUG_ASSERT(val != nullptr);
 
   RDB_MUTEX_LOCK_CHECK(rdb_sysvars_mutex);
-
-  DBUG_ASSERT(val != nullptr);
 
   // Do the real work of applying the changes.
   Rdb_cf_options::Name_to_config_t option_map;
--- a/storage/rocksdb/ha_rocksdb.cc
+++ b/storage/rocksdb/ha_rocksdb.cc
@@ -1170,7 +1170,7 @@ static MYSQL_SYSVAR_STR(update_cf_options, rocksdb_update_cf_options,
                             PLUGIN_VAR_ALLOCATED,
                         "Option updates per column family for RocksDB",
                         rocksdb_validate_update_cf_options,
-                        rocksdb_set_update_cf_options, nullptr);
+                        rocksdb_set_update_cf_options, "");
 
 enum rocksdb_flush_log_at_trx_commit_type : unsigned int {
   FLUSH_LOG_NEVER = 0,
--- a/storage/rocksdb/ha_rocksdb.cc
+++ b/storage/rocksdb/ha_rocksdb.cc
@@ -11971,6 +11971,7 @@ void rocksdb_set_update_cf_options(THD *const /* unused */,
     // NO_LINT_DEBUG
     sql_print_warning("MyRocks: NULL is not a valid option for updates to "
                       "column family settings.");
+    *reinterpret_cast<char**>(var_ptr) = nullptr;
     return;
   }
 
@@ -12051,11 +12052,7 @@ void rocksdb_set_update_cf_options(THD *const /* unused */,
   // Reset the pointers regardless of how much success we had with updating
   // the CF options. This will results in consistent behavior and avoids
   // dealing with cases when only a subset of CF-s was successfully updated.
-  if (val) {
-    *reinterpret_cast<char**>(var_ptr) = my_strdup(val,  MYF(0));
-  } else {
-    *reinterpret_cast<char**>(var_ptr) = nullptr;
-  }
+  *reinterpret_cast<char**>(var_ptr) = my_strdup(val,  MYF(0));
 
   // Our caller (`plugin_var_memalloc_global_update`) will call `my_free` to
   // free up resources used before.


storage/rocksdb/ha_rocksdb.cc
```diff
@@ -11971,6 +11971,7 @@ void rocksdb_set_update_cf_options(THD *const /* unused */,
     // NO_LINT_DEBUG
     sql_print_warning("MyRocks: NULL is not a valid option for updates to "
```
lth: If `NULL` is not meant to be a valid option, a better solution may be to add a check callback function that checks if `val` is null and fail the set command.

ajkr: I'm not that confident it is cleaner because NULL is the default value. But I don't feel strongly about it and the code works, so will update the PR as suggested.

lth: I think we should change default value to "" then. I'm looking other similar variables like `default_cf_options` and `override_cf_options` and both default to "", not nullptr.

ajkr: Sounds good to me.

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -11961,23 +11966,31 @@ static void rocksdb_set_max_background_jobs(THD *thd,
...
+                                              void *save,
+                                              st_mysql_value *value) {
+  // just disallow setting to null; otherwise same as default string checker
+  char buff[STRING_BUFFER_USUAL_SIZE];
+  const char *str;
+  int length;
+
+  length= sizeof(buff);
+  if ((str= value->val_str(value, buff, &length)))
+    str= thd->strmake(str, length);
```
lth, line 11980: Why do we need strmake here?

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -11961,23 +11966,31 @@ static void rocksdb_set_max_background_jobs(THD *thd,
...
+  *(const char**)save= str;
```
lth, line 11981: Should we be setting save to str if str is nullptr?

lth (commented):
I think you need to rebase and modify some tests.

storage/rocksdb/ha_rocksdb.cc
ajkr: I copy/pasted the default validator and changed only the return statement.

ajkr:
At a higher level, do you know why this sysvar is configured with `PLUGIN_VAR_MEMALLOC` and `PLUGIN_VAR_ALLOCATED`? I couldn't find any use of the allocated memory outside the update handler, so don't know why we need those. The pattern of `rocksdb_create_checkpoint` looks to me like a better one to follow.

storage/rocksdb/ha_rocksdb.cc
ajkr: I'm not sure. Maybe let's discuss my high-level question first so we can potentially avoid worrying about when to save it.

storage/rocksdb/ha_rocksdb.cc
lth: What default validator were you referring to?

lth:
I think all dynamic string variables must be declared with `PLUGIN_VAR_MEMALLOC`. The relevant code is in `sys_var_pluginvar::global_update`. We basically call `plugin_var_memalloc_global_update` if this flag is set which takes care of freeing the old data when we update the variable.

I'm not sure what `PLUGIN_VAR_ALLOCATED` does. All it seems to do is move the allocation earlier into the validate function out of `plugin_var_memalloc_global_update`. The dynamic string variables that I see in this file don't have that set, so maybe we can just remove it.

The `rocksdb_create_checkpoint` is simpler but the problem is that it does both the checking/setting in the "check" phase. However, if you do `set foo = 1, bar = 2;` and the first check passes and the second check fails, the expected behaviour would be that none of the variables take effect. In this case though, the statement will be partially updated. We do it for some variables anyway, but I'd rather avoid if possible.

lth:
Ah, I misunderstood what you meant by following the `rocksdb_create_checkpoint` pattern. Yeah, we can probably just follow that. I assume you mean not setting the flags and not having the variable value persist.

storage/rocksdb/ha_rocksdb.cc
hermanlee: I think you can set save to str here?

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -11961,23 +11966,31 @@ static void rocksdb_set_max_background_jobs(THD *thd,
...
+  return str == nullptr;
```
hermanlee, line 11982: I think allowing str to be set to nullptr is fine. This function should perform the Rdb_cf_options::parse_cf_options(val, &option_map) call though and return success/failure based on the result.

hermanlee (changes requested):
A couple of points about this diff.

With the addition of the validate function, the check to validate the string should also be moved from the set function to the validate function. If the string does not contain valid options, it would set the pointer to NULL, but the existing options would still be in effect.

I think being able to set the variable to NULL is fine. We can treat it as the "" string when issuing the call to RocksDB to update the settings.

Also, did you want to continue making updates to this code Andrew? If not, we can take it over for fixes.

ajkr:
I also don't mind nullptr but was trying to address comments as I'm not that familiar with MySQL to argue my way is right. You can check if the original commit in this PR is acceptable. Otherwise I don't plan to continue working on it.

hermanlee:
Thanks Andrew! I'll go ahead and take over this diff then to put in the changes to support nullptr and validating the string.

PR #709, opened 2017-09-15
Statistics that appear in SHOW ENGINE ROCKSDB STATUS do not appear in…

george-lorch:
… SHOW STATUS LIKE 'rocksdb_%'.

This patch adds those missing values and re-records the rocksdb.rocksdb test.


george-lorch:
Corresponding patch in Percona Server can be seen testing in our jenkins CI here http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1210/

lth:
Thanks for contributing!

Issue #680, opened 2017-08-14
Re-evaluate crashing server on bulk_load error.


hermanlee:
Currently, when bulk load fails to load an sst file into RocksDB, the server triggers an abort_with_stack_trace(). The original intention for this was to prevent the user from ignore errors that happen in the bulk load path because the bulk_load files could be implicitly flushed when the user disconnects from the session.

However, this might be too draconian, and can cause problems for databases shared by multiple applications. We should re-evaluate crashing here and see if logging an error is sufficient.

george-lorch:
Returning SQL error and rolling back changes is the appropriate thing to do. The current behavior is going to lead to a large amount of user complaints.

george-lorch:
Please see https://github.com/facebook/mysql-5.6/pull/711 as a possible starting point to fix these issues.

PR #711, opened 2017-09-19
- Removed debug assertion from Rdb_sst_file_ordered::put when keys ar…

george-lorch:
…e out of order as now there is proper SQL error handling for the situation.

- Extended rocksdb.bulk_load_errors.test to induce situation that would lead to debug assertion in  Rdb_sst_file_ordered::put and recorded result.
- Fixed assertion in rocksdb_update_bulk_load by converting it to a 'check' function which can take action and return a failure.
- Fixed up rocksdb.bulk_load_errors.test to remove check for assertion on overlapping keys and just check for proper error return and validate residual state of rocksdb_bulk_load and table.
- Fixed code paths that would call my_print_error to a client where no client exists, this would cause an assertion in the server code.
- Altered rocksdb.bulk_load_errors.test to watch for new errors in log and re-recorded.
- Fixed edge case where if a handler is closed while a bulk load operation with a fault is in process ~ha_rocksdb would assert. Added another sql_print_error and test case to rocksdb.bulk_load_errors.test.

### DIFF
--- a/mysql-test/suite/rocksdb/t/bulk_load_errors.test
+++ b/mysql-test/suite/rocksdb/t/bulk_load_errors.test
@@ -1,7 +1,13 @@
 --source include/have_rocksdb.inc
+--source include/count_sessions.inc
+
+--let LOG1=$MYSQLTEST_VARDIR/tmp/rocksdb.bulk_load_errors.1.err
+--let $_mysqld_option=--log-error=$LOG1
+--replace_result $MYSQLTEST_VARDIR MYSQLTEST_VARDIR
+--source include/restart_mysqld_with_option.inc
 
 ### Bulk load ###
-CREATE TABLE t1(pk INT, PRIMARY KEY(pk));
+CREATE TABLE t1(pk INT, PRIMARY KEY(pk)) ENGINE=ROCKSDB;
 
 # Make sure we get an error with out of order keys during bulk load
 SET rocksdb_bulk_load=1;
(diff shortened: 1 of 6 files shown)


george-lorch:
Hey team, this issue https://github.com/facebook/mysql-5.6/issues/680 is I think mostly addressed in this PR. It has been a big drag on our QA team and will be a pretty big issue for users to have the server asserting rather than reporting errors.

I am pretty sure some of the changes are not really 'friendly' to the overall MyRocks design, but wanted to at least get something sent in for review, discussion, and eventual clean up. Please provide feedback on anything I may have missed or alternate approaches to solving the issues. Thanks!

george-lorch:
Something that I was considering before but didn't carry through in the initial patch is to make sure that any bulk load related sql_print_error contains the database and table names and possibly even the query that triggered the error.

mysql-test/suite/rocksdb/t/bulk_load_errors.test
```diff
@@ -21,21 +27,32 @@ INSERT INTO t1 VALUES(2);
...
 
---exec grep "RocksDB: Error 197 finalizing last SST file while setting bulk loading variable" $MYSQLTEST_VARDIR/log/mysqld.1.err | cut -d] -f2
---exec echo "" >$MYSQLTEST_VARDIR/log/mysqld.1.err
+SHOW VARIABLES LIKE 'rocksdb_bulk_load';
+
+SELECT * FROM t1;
+
+--let SEARCH_PATTERN=RocksDB: Error 198 finalizing last SST file while setting bulk loading variable
+--source include/search_pattern_in_file.inc
+--exec echo " " > $SEARCH_FILE
```
laurynas-biveinis: Restart again with a different SEARCH_FILE instead?

mysql-test/suite/rocksdb/t/bulk_load_errors.test
```diff
@@ -21,21 +27,32 @@ INSERT INTO t1 VALUES(2);
...
-# restart the crashed server
---exec echo "restart" > $MYSQLTEST_VARDIR/tmp/mysqld.1.expect
+SELECT * FROM t1;
 
-# Make sure the error exists in the .err log and then restart the server
---enable_reconnect
---source include/wait_until_connected_again.inc
+--let SEARCH_PATTERN=RocksDB: Error 198 finalizing last SST file while disconnecting
+--source include/search_pattern_in_file.inc
+--exec echo " " > $SEARCH_FILE
```
laurynas-biveinis: Likewise?

mysql-test/suite/rocksdb/t/bulk_load_errors.test
george-lorch: Slow the test down with no gain?

laurynas-biveinis: In order not to limit portability of the testcase: IIRC, on Windows the "exec echo" would fail to truncate the file because mysqld opens it without allowing shared I/O for any subsequent process trying to open it again. But, this change would be more of a favor for MariaDB MyRocks, as they are the ones with Windows build, and maybe they are in a better position to suggest a patch here.

george-lorch: Hmm, true, that is a fair point.

george-lorch: OK, changed to restarting the server with fresh logs for each search, test takes ~75x longer to run (was ~80ms locally before but is now ~6100ms) but should be safe for any 'doze users.

laurynas-biveinis: The first and the final restarts are required for error log handling correctness on all platforms, thus I hope that Windows adds only a ~2x slowdown :)

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -11973,23 +11973,30 @@ void rocksdb_set_collation_exception_list(THD *const thd,
...
-void rocksdb_set_bulk_load(THD *const thd, struct st_mysql_sys_var *const var
-                                               MY_ATTRIBUTE((__unused__)),
-                           void *const var_ptr, const void *const save) {
-  Rdb_transaction *&tx = get_tx_from_thd(thd);
+int rocksdb_check_bulk_load(THD *const thd,
+                            struct st_mysql_sys_var *var
+                                MY_ATTRIBUTE((__unused__)),
+                            void *save,
+                            struct st_mysql_value *value) {
+    Rdb_transaction *&tx = get_tx_from_thd(thd);
```
hermanlee: Nit: These look like whitespace changes. It should be two spaces per indent.

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -11973,23 +11973,30 @@ void rocksdb_set_collation_exception_list(THD *const thd,
...
+                rc);
+            THDVAR(thd, bulk_load) = 0;
+            return 1;
+        }
     }
-  }
 
-  *static_cast<bool *>(var_ptr) = *static_cast<const bool *>(save);
+    long long intbuf;
+    value->val_int(value, &intbuf);
```
hermanlee: The test rocksdb_sys_vars.bulk_load_basic is failing to use "on" as 1. I think since it's a bool, value could be a string and needs to evaluate "ON", "OFF" to 1 and 0 too.

hermanlee (changes requested):
Looks great overall. Just need a fix for the rocksdb_sys_vars.rocksdb_bulk_load_basic unit test.

storage/rocksdb/ha_rocksdb.cc
george-lorch: Oops, sorry, fat-finger on my part. Will be fixed.

storage/rocksdb/ha_rocksdb.cc
george-lorch: Really? This is weird, the test succeeds for me in a debug build. I assumed that the ON/OFF/YES/NO bool transformation was already done before coming into the check. Debugging a bit into it and yeah, you are correct, it is coming in to the check function as a string type when ON/OFF/YES/NO and not an integer type, the integer transformation just converts the string to '0' and does not recognize the ON/YES as being worth '1'. This is a good catch and in fact makes me think we have a bug over in TokuDB where it is using a similar technique to 'fail' a SET. I'll build up a fix and update when I have re-pushed.

george-lorch: OK, changes made, re-tested and pushed. Thanks!

hermanlee (approved):
Thanks!

PR #715, opened 2017-09-26
Prevent direct writes and setting rocksdb_flush_log_at_trx_commit from

hermanlee:
asserting

Summary:

Fix the interaction between rocksdb_flush_log_at_trx_commit and allowing
direct writes. When direct writes are allowed, FlushWAL() should not be
called. If these settings are conflicting, then report an error on
startup or when trying to set the variable.

Test Plan:

run mtr and add new test case

### DIFF
--- a/mysql-test/include/restart_mysqld_with_invalid_option.inc
+++ b/mysql-test/include/restart_mysqld_with_invalid_option.inc
@@ -1,0 +1,8 @@
+--source include/shutdown_mysqld.inc
+
+# Expect the server to fail to come up with these options
+--error 1
+--exec $MYSQLD_CMD $_mysqld_option
+
+# Restart the server with the default options
+--source include/start_mysqld.inc
--- a/mysql-test/suite/rocksdb/t/use_direct_reads_writes.test
+++ b/mysql-test/suite/rocksdb/t/use_direct_reads_writes.test
@@ -5,43 +5,51 @@
 # caused an assertion in RocksDB.  Now it should not be allowed and the
 # server will not start with that configuration
 
-# Write file to make mysql-test-run.pl expect the "crash", but don't restart
-# the server until it is told to
---let $_server_id= `SELECT @@server_id`
---let $_expect_file_name= $MYSQLTEST_VARDIR/tmp/mysqld.$_server_id.expect
---exec echo "wait" >$_expect_file_name
-shutdown_server 10;
+--let LOG=$MYSQLTEST_VARDIR/tmp/use_direct_reads_writes.err
+--let SEARCH_FILE=$LOG
 
-# Clear the log
---exec echo "" >$MYSQLTEST_VARDIR/log/mysqld.1.err
+--echo Checking direct reads
+--let $_mysqld_option=--log-error=$LOG --rocksdb_use_direct_reads=1 --rocksdb_allow_mmap_reads=1
+--replace_result $MYSQLTEST_VARDIR MYSQLTEST_VARDIR
+--source include/restart_mysqld_with_invalid_option.inc
 
-# Attempt to restart the server with invalid options
---exec echo "restart:--rocksdb_use_direct_reads=1 --rocksdb_allow_mmap_reads=1" >$_expect_file_name
---sleep 0.1  # Wait 100ms - that is how long the sleep is in check_expected_crash_and_restart
---exec echo "restart:" >$_expect_file_name
+--let SEARCH_PATTERN=enable both use_direct_reads
+--source include/search_pattern_in_file.inc
+--remove_file $LOG
+
+
+# Repeat with direct-writes
+--echo Checking direct writes
+--let $_mysqld_option=--log-error=$LOG --rocksdb_use_direct_io_for_flush_and_compaction=1 --rocksdb_allow_mmap_writes=1
+--replace_result $MYSQLTEST_VARDIR MYSQLTEST_VARDIR
+--source include/restart_mysqld_with_invalid_option.inc
+
+--let SEARCH_PATTERN=enable both use_direct_io_for_flush_and_compaction
+--source include/search_pattern_in_file.inc
+--remove_file $LOG
+
+
+# Verify invalid direct-writes and --rocksdb_flush_log_at_trx_commit combination at startup fails
+--echo Checking rocksdb_flush_log_at_trx_commit
+--let $_mysqld_option=--log-error=$LOG --rocksdb_flush_log_at_trx_commit=1 --rocksdb_allow_mmap_writes=1
+--replace_result $MYSQLTEST_VARDIR MYSQLTEST_VARDIR
+--source include/restart_mysqld_with_invalid_option.inc
+
+--let SEARCH_PATTERN=rocksdb_flush_log_at_trx_commit needs to be
+--source include/search_pattern_in_file.inc
+--remove_file $LOG
+
+
+# Verify rocksdb_flush_log_at_trx_commit cannot be changed if direct writes are used
+--echo Validate flush_log settings when direct writes is enabled
+--let $_mysqld_option=--rocksdb_flush_log_at_trx_commit=0 --rocksdb_allow_mmap_writes=1
+--source include/restart_mysqld_with_option.inc
+
+set global rocksdb_flush_log_at_trx_commit=0;
+--error ER_WRONG_VALUE_FOR_VAR
+set global rocksdb_flush_log_at_trx_commit=1;
+--error ER_WRONG_VALUE_FOR_VAR
+set global rocksdb_flush_log_at_trx_commit=2;
 
 # Cleanup
---enable_reconnect
---source include/wait_until_connected_again.inc
---disable_reconnect
-
-# We should now have an error message
---exec grep "enable both use_direct_reads" $MYSQLTEST_VARDIR/log/mysqld.1.err | cut -d] -f2
-
-# Repeat with --rocksdb-use-direct-writes
---let $_server_id= `SELECT @@server_id`
---let $_expect_file_name= $MYSQLTEST_VARDIR/tmp/mysqld.$_server_id.expect
---exec echo "wait" >$_expect_file_name
-shutdown_server 10;
-
---exec echo "" >$MYSQLTEST_VARDIR/log/mysqld.1.err
-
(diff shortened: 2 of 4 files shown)


mysql-test/suite/rocksdb/t/use_direct_reads_writes.test
```diff
@@ -12,8 +18,8 @@
 --exec echo "wait" >$_expect_file_name
 shutdown_server 10;
 
-# Clear the log
---exec echo "" >$MYSQLTEST_VARDIR/log/mysqld.1.err
+# Add a delimiter to the log file
+--exec echo "[Note] use_direct_reads_write.test restart direct_reads" >> $MYSQLTEST_VARDIR/log/mysqld.$_server_id.err
```
laurynas-biveinis: This will not work with MTR --repeat arg. A dedicated error log should be used instead (cheap to switch here as you are restarting anyway)

mysql-test/suite/rocksdb/t/use_direct_reads_writes.test
```diff
@@ -25,16 +31,19 @@ shutdown_server 10;
...
-# Repeat with --rocksdb-use-direct-writes
---let $_server_id= `SELECT @@server_id`
---let $_expect_file_name= $MYSQLTEST_VARDIR/tmp/mysqld.$_server_id.expect
+# Repeat with direct-writes
 --exec echo "wait" >$_expect_file_name
 shutdown_server 10;
 
---exec echo "" >$MYSQLTEST_VARDIR/log/mysqld.1.err
+# Add a delimiter to the log file
+--exec echo "[Note] use_direct_reads_write.test restart direct_writes" >> $MYSQLTEST_VARDIR/log/mysqld.$_server_id.err
```
laurynas-biveinis: Likewise

mysql-test/suite/rocksdb/t/use_direct_reads_writes.test
```diff
@@ -44,4 +53,47 @@ shutdown_server 10;
...
+--let $assert_text= Checking for use_direct_io_for_flush_and_compaction
+--let $assert_file= $MYSQLTEST_VARDIR/log/mysqld.1.err
+--let $assert_only_after= use_direct_reads_writes.test restart direct_writes
+--let $assert_select= enable both use_direct_io_for_flush_and_compaction
+--let $assert_count= 1
+--source include/assert_grep.inc
+
+# Verify invalid direct-writes and --rocksdb_flush_log_at_trx_commit combination at startup fails
+--exec echo "wait" >$_expect_file_name
+shutdown_server 10;
```
laurynas-biveinis: Here you can use a simple restart_mysqld.inc with a custom error log path due to the next comment, but in general, consider using shutdown_mysqld.inc instead of exec echo wait / shutdown_server where possible: both for shorter testcase and also this removes hardcoded shutdown timeouts (Percona Server with 10 second shutdown timeouts gets an occasional kill in our CI farm, which the testcases might not be prepared for)

mysql-test/suite/rocksdb/t/use_direct_reads_writes.test
```diff
@@ -44,4 +53,47 @@ shutdown_server 10;
...
+
+# Add a delimiter to the log file
+--exec echo "[Note] use_direct_reads_write.test restart flush_log" >> $MYSQLTEST_VARDIR/log/mysqld.$_server_id.err
```
laurynas-biveinis: Same comment re. MTR --repeat

mysql-test/suite/rocksdb/t/use_direct_reads_writes.test
```diff
@@ -44,4 +53,47 @@ shutdown_server 10;
...
+
+--exec echo "restart:--rocksdb_flush_log_at_trx_commit=1 --rocksdb_allow_mmap_writes=1" >$_expect_file_name
+--sleep 0.1
```
laurynas-biveinis: Not sure this is not racy. @georgelorchpercona, what you ended up doing for this kind of test?

mysql-test/suite/rocksdb/t/use_direct_reads_writes.test
```diff
@@ -44,4 +53,47 @@ shutdown_server 10;
...
+--exec echo "restart:" >$_expect_file_name
```
laurynas-biveinis: In general, this could be replaced by start_mysqld.inc (in this instance depends on the result of the previous comment)

mysql-test/suite/rocksdb/t/use_direct_reads_writes.test
```diff
@@ -44,4 +53,47 @@ shutdown_server 10;
...
+--let $assert_only_after= use_direct_reads_writes.test restart flush_log
+--let $assert_select= rocksdb_flush_log_at_trx_commit needs to be
+--let $assert_count= 1
+--source include/assert_grep.inc
+
+# Verify rocksdb_flush_log_at_trx_commit cannot be changed if direct writes are used
+--exec echo "wait" >$_expect_file_name
+shutdown_server 10;
+
+--exec echo "restart:--rocksdb_flush_log_at_trx_commit=0 --rocksdb_allow_mmap_writes=1" >$_expect_file_name
```
laurynas-biveinis: Same four comments as above, except that what is being tested on this line (there is no grep nor should it fail?)

mysql-test/suite/rocksdb/t/use_direct_reads_writes.test
george-lorch: You can see an example of a dedicated log in my pull request here https://github.com/facebook/mysql-5.6/pull/711/files#diff-61934317c60558ea9bedc26f897e9af1

mysql-test/suite/rocksdb/t/use_direct_reads_writes.test
george-lorch: I believe we just used restart_mysqld which waits for the wait_until_connected to timeout. Makes the test run longer but incurs no resource overhead while waiting. Here is a similar example that we fixed recently https://github.com/percona/percona-server/blob/5.7/mysql-test/suite/rocksdb/t/use_direct_reads_writes.test#L10

george-lorch: Actually, I just remembered, this is the test that we were both confused with in PS because PS would still start even if the plugin failed to initialize. So, in our case, we just start the server then grep the error log for the complaint, then stop the server and continue on. So, our (PS) is going to be different from FB-MySQL in this test behavior just like the use_direct_reads_writes.test

hermanlee:
Thanks for the great suggestions on the test case. I refactored the code for restarting mysqld into a new include file since the option settings being tested are expected to stop the server from loading up. I also followed the setup from George's test case to simplify the test.

mysql-test/include/restart_mysqld_with_invalid_option.inc
```diff
@@ -0,0 +1,35 @@
+
```
laurynas-biveinis, line 2: Based on the last @georgelorchpercona comment now I recall that the "canonical" way to test the server failure to start is
--source include/shutdown_mysqld.inc
--error 1
--exec $MYSQLD options, including a dedicated error log
grep, remove the dedicated error log (can also be done after start_mysqld.inc)
--source include/start_mysqld.inc

For example, https://github.com/mysql/mysql-server/blob/5.6/mysql-test/suite/innodb/t/innodb-log-file-size-1.test#L209

mysql-test/include/restart_mysqld_with_invalid_option.inc
```diff
@@ -0,0 +1,35 @@
...
+if ($rpl_inited)
+{
+  if (!$allow_rpl_inited)
+  {
+    --die ERROR IN TEST: This script does not support replication
+  }
+}
+
+# Write file to make mysql-test-run.pl expect the "crash", but don't start
```
laurynas-biveinis: If you keep this include file, then this snippet should be replaced by source shutdown_mysqld.inc (unless this happens to break caller's replace_result - in which case at least the explicit "10" should be removed from shutdown_server so that the default mysqltest timeout becomes effective)

mysql-test/include/restart_mysqld_with_invalid_option.inc
```diff
@@ -0,0 +1,35 @@
...
+# Send shutdown to the connected server and give
+# it 10 seconds to die before zapping it
+shutdown_server 10;
+
+# Write file to make mysql-test-run.pl start up the server again
+--exec echo "restart:$_mysqld_option" > $_expect_file_name
+
+# Expect the server to fail to come up with these options
+--sleep 0.1
+--exec echo "restart:" > $_expect_file_name
```
laurynas-biveinis: ...and this one by start_mysqld.inc

mysql-test/suite/rocksdb/t/use_direct_reads_writes.test
```diff
@@ -1,47 +1,60 @@
...
+--echo Validate flush_log settings when direct writes is enabled
+--let $_mysqld_option=--rocksdb_flush_log_at_trx_commit=0 --rocksdb_allow_mmap_writes=1
+--source include/restart_mysqld_with_option.inc
 
---exec grep "enable both use_direct_io_for_flush_and_compaction" $MYSQLTEST_VARDIR/log/mysqld.1.err | cut -d] -f2
+set global rocksdb_flush_log_at_trx_commit=0;
+--error ER_WRONG_VALUE_FOR_VAR
+set global rocksdb_flush_log_at_trx_commit=1;
+--error ER_WRONG_VALUE_FOR_VAR
+set global rocksdb_flush_log_at_trx_commit=2;
```
laurynas-biveinis, line 52: The server should be restarted with the default options at the end

hermanlee:
I updated with the suggested changes to use the canonical method of restarting with error and simplified the scripts.

mysql-test/suite/rocksdb/t/use_direct_reads_writes.test
```diff
@@ -1,47 +1,61 @@
 --source include/have_rocksdb.inc
 
+call mtr.add_suppression("RocksDB: Can't enable both");
```
laurynas-biveinis: The suppressions should be redundant now as they are not making it to the main server error log

laurynas-biveinis (commented):
One last comment and I don't think I'll have any further ones - thanks for making the changes

lth (approved):

PR #720, opened 2017-10-02
Fix auto_increment bugs with large values

lth:
Fix auto_increment bugs with large values

There are currently a few related bugs in our autoincrement implementation relating to large values.

1. When inserting values larger than LLONG_MAX, we do not bump the autoincrement counter on the table because we are using signed 64bit integers to store autoincrement. Fix by converting to use ulonglong.

2. When `auto_increment_increment` is set, and the next number in the sequence is larger than ULLONG_MAX, we are not currently do not error out, and instead insert smaller values, due to underflow. Fix by calculating next value more carefully, and return ULLONG_MAX when limit is reached, so that the server can handle appropriately.

3. `get_auto_increment` can return values that are greater than the maximum allowed value for the column type. Although the SQL layer truncates the value, the value is still stored inside the SE, which means that SHOW CREATE TABLE can show invalid values from auto_increment columns.


storage/rocksdb/ha_rocksdb.cc
```diff
@@ -10402,32 +10450,62 @@ void ha_rocksdb::get_auto_increment(ulonglong off, ulonglong inc,
...
+    //
+    // If we return values greater than the max, the SQL layer will "truncate"
+    // the value anyway, but it means that we store invalid values into
+    // auto_incr that will be visible in SHOW CREATE TABLE.
     new_val = auto_incr;
-    while (new_val != std::numeric_limits<longlong>::max()) {
-      if (auto_incr.compare_exchange_weak(new_val, new_val + 1)) {
+    while (new_val != std::numeric_limits<ulonglong>::max()) {
+      if (auto_incr.compare_exchange_weak(new_val,
+                                          std::min(new_val + 1, max_val))) {
```
hermanlee, line 10490: It looks like it's intended we return new_value+1 even when it's larger than max_val, and rely on the SQL layer to truncate the value as needed? For signed values, we return new_val + 1, which could be negative. It seems like this would still work though.

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -10439,31 +10517,23 @@ void ha_rocksdb::get_auto_increment(ulonglong off, ulonglong inc,
         //      may not be in our sequence, but it is guaranteed to be equal
         //      to or larger than any other value already inserted.
         //
-        //  For now I'm going to take option @2.
-        new_val = std::numeric_limits<longlong>::max();
+        //  For now I'm going to take option 2.
+        new_val = std::numeric_limits<ulonglong>::max();
         auto_incr = new_val;  // Store the largest value into auto_incr
```
hermanlee: Shouldn't this be using max_val? Otherwise, auto_incr could be larger than the field type's max value, no?

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -10402,32 +10450,62 @@ void ha_rocksdb::get_auto_increment(ulonglong off, ulonglong inc,
...
+      // offset back in.
+      //
+      // The additions are rearranged to avoid overflow.
+      //
+      DBUG_ASSERT(last_val > 0);
+      // The following is equivalent to (last_val - 1 + inc - off) / inc.
+      ulonglong n =
+          (last_val - 1) / inc + ((last_val - 1) % inc + inc - off) / inc;
+
+      if (n > (std::numeric_limits<ulonglong>::max() - off) / inc) {
```
hermanlee: Shouldn't this be using max_val to determine if n is larger than the max?

mysql-test/suite/rocksdb/t/issue255.test
```diff
@@ -8,6 +8,38 @@ SHOW TABLE STATUS LIKE 't1';
...
+
+--error ER_DUP_ENTRY
+INSERT INTO t1 VALUES ();
+SELECT * FROM t1;
+--replace_column 3 # 6 # 7 # 8 # 9 # 10 #
+SHOW TABLE STATUS LIKE 't1';
+
+DROP TABLE t1;
+
+CREATE TABLE t1 (pk TINYINT NOT NULL PRIMARY KEY AUTO_INCREMENT);
```
hermanlee, line 28: Can you add some show create table statements to show the auto_increment value of t1 after trying to insert a few times when auto_incr is at max.

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -10402,32 +10450,62 @@ void ha_rocksdb::get_auto_increment(ulonglong off, ulonglong inc,
...
+      // To get the next number in the sequence above you subtract out the
+      // offset, calculate the next sequence (N * increment) and then add the
+      // offset back in.
+      //
+      // The additions are rearranged to avoid overflow.
+      //
+      DBUG_ASSERT(last_val > 0);
+      // The following is equivalent to (last_val - 1 + inc - off) / inc.
+      ulonglong n =
+          (last_val - 1) / inc + ((last_val - 1) % inc + inc - off) / inc;
```
hermanlee: I'm having some trouble reasoning through the correctness of this calculation. For example, suppose

last_val = 75
max_val = 128
inc = 3
offset = 120

I'd expect the next value to be 120, followed by 123, 126 and then we hit the max. Is this correct? The statement indicates we generate sequences of offset + N * inc, where N in [0, 1, 2, 3...], so 120, 123, 126, 129...

However: (75 - 1) / 3 + ((75 - 1) % 3 + 3 - 120) / 3 = -14

Later, we have new_val * inc + offset: -14 * 3 + 120 = 78, which seems different from the original code.

hermanlee (changes requested):

storage/rocksdb/ha_rocksdb.cc
lth: We could rely on the SQL layer to truncate but:
1. You get "out of range" warning for a value the user didn't specify and
2. You get odd output from show create table.

I think this is better behaviour than what we had before.

storage/rocksdb/ha_rocksdb.cc
lth: There's an assumption that off <= inc, which would make the (inc - off) calculation always return a non-negative number. I can add a DBUG_ASSERT if needed, but the assumption is enforced a few lines above in the same function.

Here's the reasoning. I'm using the fact that `(a+b)/c = a/c + b/c + (a%c + b%c) / c`. Here's a short proof:

```
(a+b)/c
= (a - a%c + a%c + b - b%c + b%c) / c
= (a - a%c) / c + (b - b%c) / c + (a%c + b%c) / c
= a/c + b/c + (a%c + b%c) / c
```

Now let `a = last_val -1` and `b = inc - off` and `c = inc` to get my statement.

storage/rocksdb/ha_rocksdb.cc
lth: This is just for checking for overflow, so that the next line `new_val = n * inc + off` doesn't overflow.

We compare with max_val in the `std::min` call below.

storage/rocksdb/ha_rocksdb.cc
lth: If you have `n * inc + off > ULLONG_MAX`, then you definitely have a UNSIGNED BIGINT here, so it's the same as max_val. I can add an assert if needed.

As a side note, a return value of ULLONG_MAX for `get_auto_increment` has special meaning. This will return `ER_AUTOINC_READ_FAILED` to the user. So we have a quirk due to the SE API that we have different autoinc behaviour between UNSIGNED BIGINT and any of the other integer type. For UNSIGNED BIGINT, you get `ER_AUTOINC_READ_FAILED` once you hit max, whereas for everyone else, you just get duplicate key error (or no error if autoinc is on SK).

mysql-test/suite/rocksdb/t/issue255.test
lth: I'm not sure what you mean, isn't that what I added to this test?

If you look at the results file, you can see for the bigint case, we're at `9223372036854775807` for a few failed inserts, and for the tiny int case, we're at `127` for a few failed inserts.

storage/rocksdb/ha_rocksdb.cc
hermanlee: Ah, okay. The check will skip the degenerate case, so the rest of the logic follows then.

storage/rocksdb/ha_rocksdb.cc
hermanlee: Yes, that makes sense. Can you update the diff with these comments on the logic of the math and what happens in the UNSIGNED BIGINT vs. non-UNSIGNED BIGINT cases?

mysql-test/suite/rocksdb/t/issue255.test
hermanlee: Yes, I see a mixture of SHOW CREATE TABLE vs. SHOW TABLE STATUS, but just realized the auto_increment column name is also in the show table status output, just off-screen. Each of the two files is consistent in its use of the command to show the value, but show create table is probably clearer. Just a slight preference for show create table is all. Your changes above is sufficient.

hermanlee (approved):
Changes look good then. Can you update your diff to include your explanations? It will be easier to reason through the logic if we need to revisit this code.

storage/rocksdb/ha_rocksdb.cc
hermanlee: There's some interesting interaction with modify column. Could it lead to auto_incr being larger than max_val after modify is performed?

create table t1 (pk bigint(8) auto_increment, primary key(pk)) engine=rocksdb;
insert into t1 values (1), (2), (10), (130), (140);
delete from t1 where pk > 127;
alter table t1 modify column pk tinyint auto_increment;
show create table t1;

The current implementation seems to show auto_increment = 141.

lth: I just tested on this diff, it seems to also return 141. We'd probably need to hook into the alter code paths to update this info for the table.

lth: It gets corrected on the next insert though.

hermanlee: Yeah, the modify column should work correctly with this patch. It does means we can't assert() though on the condition that the value is larger than max_val since it's possible.

PR #722, opened 2017-10-07
Implement persistent auto_increment values

lth:
This implements persistent auto_increment values for MyRocks by adding a new data dictionary type for auto increment, and using the merge operator to calculate the latest auto increment.

Closes https://github.com/facebook/mysql-5.6/issues/468


mysql-test/suite/rocksdb/t/autoinc_debug.test
```diff
@@ -0,0 +1,110 @@
...
+--error 2013
+insert into t values ();
+--source include/wait_until_disconnected.inc
+--enable_reconnect
+--exec echo "restart" > $MYSQLTEST_VARDIR/tmp/mysqld.1.expect
+--source include/wait_until_connected_again.inc
+--disable_reconnect
+select table_schema, table_name, auto_increment from information_schema.tables where table_name = 't';
+select max(i) from t;
+set debug="-d,crash_commit_before";
```
laurynas-biveinis: Redundant on a newly started server

mysql-test/suite/rocksdb/t/autoinc_debug.test
```diff
@@ -0,0 +1,110 @@
...
+--error 2013
+insert into t values ();
+--source include/wait_until_disconnected.inc
+--enable_reconnect
+--exec echo "restart" > $MYSQLTEST_VARDIR/tmp/mysqld.1.expect
+--source include/wait_until_connected_again.inc
+--disable_reconnect
+select table_schema, table_name, auto_increment from information_schema.tables where table_name = 't';
+select max(i) from t;
+set debug="-d,crash_commit_after_prepare";
```
laurynas-biveinis: Likewise

mysql-test/suite/rocksdb/t/autoinc_debug.test
```diff
@@ -0,0 +1,110 @@
...
+--error 2013
+insert into t values ();
+--source include/wait_until_disconnected.inc
+--enable_reconnect
+--exec echo "restart" > $MYSQLTEST_VARDIR/tmp/mysqld.1.expect
+--source include/wait_until_connected_again.inc
+--disable_reconnect
+select table_schema, table_name, auto_increment from information_schema.tables where table_name = 't';
+select max(i) from t;
+set debug="-d,crash_commit_after_log";
```
laurynas-biveinis: Likewise

mysql-test/suite/rocksdb/t/autoinc_debug.test
```diff
@@ -0,0 +1,110 @@
...
+--error 2013
+insert into t values ();
+--source include/wait_until_disconnected.inc
+--enable_reconnect
+--exec echo "restart" > $MYSQLTEST_VARDIR/tmp/mysqld.1.expect
+--source include/wait_until_connected_again.inc
+--disable_reconnect
+select table_schema, table_name, auto_increment from information_schema.tables where table_name = 't';
+select max(i) from t;
+set debug="-d,crash_commit_after";
```
laurynas-biveinis: Likewise

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -1783,7 +1784,30 @@ class Rdb_transaction {
...
+    @param wb
+   */
+  rocksdb::Status merge_auto_incr_map(rocksdb::WriteBatchBase *const wb) {
+    DBUG_EXECUTE_IF("myrocks_autoinc_upgrade", return rocksdb::Status::OK(););
+
+    // Iterate through the merge map merging all keys into data dictionary.
+    rocksdb::Status s;
+    for (auto &it : m_auto_incr_map) {
+      s = dict_manager.put_auto_incr_val(wb, it.first, it.second);
+      if (!s.ok()) {
```
hermanlee, line 1811: In the error case, does m_auto_incr_map.clear() get called through rollback()?

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -2526,9 +2579,17 @@ class Rdb_writebatch_impl : public Rdb_transaction {
...
+
+    s = merge_auto_incr_map(m_batch->GetWriteBatch());
+    if (!s.ok()) {
+      rdb_handle_io_error(s, RDB_IO_ERROR_TX_COMMIT);
+      res = true;
+    }
+
     release_snapshot();
-    const rocksdb::Status s =
-        rdb->GetBaseDB()->Write(write_opts, m_batch->GetWriteBatch());
```
hermanlee, line 2540: If merge_auto_incr_map() returned an error, shouldn't we need to skip this commit step?

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -2254,8 +2298,16 @@ class Rdb_transaction_impl : public Rdb_transaction {
...
+
+    s = merge_auto_incr_map(m_rocksdb_tx->GetWriteBatch()->GetWriteBatch());
+    if (!s.ok()) {
+      rdb_handle_io_error(s, RDB_IO_ERROR_TX_COMMIT);
+      res = true;
+    }
+
     release_snapshot();
-    const rocksdb::Status s = m_rocksdb_tx->Commit();
+    s = m_rocksdb_tx->Commit();
```
hermanlee, line 2320: If merge_auto_incr_map() returns an error, shouldn't we skip this Commit() call?

storage/rocksdb/rdb_datadic.h
```diff
@@ -1355,4 +1372,107 @@ struct Rdb_index_info {
...
+  bool Merge(const rocksdb::Slice &key, const rocksdb::Slice *existing_value,
+             const rocksdb::Slice &value, std::string *new_value,
+             rocksdb::Logger *logger) const override {
+    DBUG_ASSERT(new_value != nullptr);
+
+    if (key.size() != Rdb_key_def::INDEX_NUMBER_SIZE * 3 ||
+        GetKeyType(key) != Rdb_key_def::AUTO_INC ||
+        value.size() !=
+            RDB_SIZEOF_AUTO_INCREMENT_VERSION + ROCKSDB_SIZEOF_AUTOINC_VALUE ||
+        GetVersion(value) != Rdb_key_def::AUTO_INCREMENT_VERSION) {
```
hermanlee: This might be too strict. Currently, it's fine since there's only one version, but in future versions, if we update the version, the compaction filter might still run with older versions of existing keys in the system column family. This code will probably need to be able to merge the old versions together to generate the new version. I'd think GetVersion(value) > Rdb_key_def::AUTO_INCREMENT_VERSION would be a better check.

storage/rocksdb/rdb_datadic.h
```diff
@@ -1355,4 +1372,107 @@ struct Rdb_index_info {
...
+
+  virtual const char *Name() const override { return "Rdb_system_merge_op"; }
+
+ private:
+  /*
+    Serializes the integer data to the new_value buffer or the target buffer
+    the merge operator will update to
+   */
+  void Serialize(const uint64_t data, std::string *new_value) const {
+    uchar value_buf[RDB_SIZEOF_AUTO_INCREMENT_VERSION +
```
hermanlee, line 1436: Small optimization to allocate the buffer directly in new_value, rather than write to the stack and copy over? But not really an issue.

storage/rocksdb/rdb_datadic.cc
```diff
@@ -3459,6 +3460,20 @@ void Rdb_tbl_def::set_name(const std::string &name) {
   check_if_is_mysql_system_table();
 }
 
+std::shared_ptr<Rdb_key_def> Rdb_tbl_def::get_pk() {
```
hermanlee: We issue get_pk() mainly to pull out the gl_index_id() that is used for the auto_increment key. Can we be more specific here for this function? Call it get_autoincr_gl_index_id() perhaps, so it's clear it's generating the index used for storing the auto_increment key. This also allows us to skip copying the Rdb_key_def share ptr.

mysql-test/suite/rocksdb/t/autoinc_debug.test
```diff
@@ -0,0 +1,106 @@
...
+create table t (i int primary key auto_increment);
+insert into t values ();
+insert into t values ();
+insert into t values ();
+
+--echo # Before anything
+set debug="+d,crash_commit_before";
+--exec echo "restart" > $MYSQLTEST_VARDIR/tmp/mysqld.1.expect
+--error 2013
+insert into t values ();
```
hermanlee: Could you turn these into multi-statement transactions to help show that even if some statements complete before the transaction could be committed, the allocated ids are not committed.

mysql-test/suite/rocksdb/t/autoinc_crash_safe.test
```diff
@@ -0,0 +1,130 @@
...
+--source include/rpl_restart_server.inc
+select table_schema, table_name, auto_increment from information_schema.tables where table_name = 't';
+
+--echo # Remove auto_increment property.
+alter table t modify i int;
+--let $rpl_server_number = 1
+--source include/rpl_restart_server.inc
+select table_schema, table_name, auto_increment from information_schema.tables where table_name = 't';
+
+--echo # Add auto_increment property.
```
hermanlee, line 101: Can you also add a test that performs

`alter table t auto_increment=1000;`

This will just recreate the table in our current implementation, but probably a good test case for future features.

hermanlee (changes requested):
Just minor requests. Feel free to disregard if they're not useful. The diff looks very solid overall and the new validation code is great.

storage/rocksdb/ha_rocksdb.cc
lth: The point is kind of moot just because `rdb_handle_io_error` will abort the server anyway, but I'll try to rework the code in case this ever changes.

storage/rocksdb/rdb_datadic.h
lth: I'll leave it as, just because I don't know if std::string actually supports semantics like this. I don't think there is stipulation that its memory be contiguous.

mysql-test/suite/rocksdb/t/autoinc_crash_safe.test
lth: I was planning to add this when I implement in-place alter for auto_increment.

hermanlee (approved):

mdcallag:
There is a perf regression in full index scan perf and I will guess that this diff caused it - https://github.com/facebook/mysql-5.6/issues/741

Issue #724, opened 2017-10-10
Curious sleeps in write_sync.test


george-lorch:
While importing https://github.com/facebook/mysql-5.6/commit/8b9734948e0023f6adeadb536375f22016d8e521  "use new FlushWAL API in MyRocks" It was noticed on these lines https://github.com/facebook/mysql-5.6/commit/8b9734948e0023f6adeadb536375f22016d8e521#diff-4d4f80f6019c0d48cefc62d836532afaR10 that there are curious sleeps after setting rocksdb_flush_log_at_trx_commit. We are guessing at the purpose of these sleeps and wondering if there is a better, more deterministic way to implement this test without the use of sleeps.

hermanlee:
Yes, the sleep calls are used to wait for the background thread to flush the wal. Definitely having a deterministic solution is better. The rocksdb_wal_synced counter is there, but I'd think we'd need a flushwal call that is exposed to the test to do this deterministically.

laurynas-biveinis:
The testcase already uses wait_for_status_var.inc with rocksdb_wal_synced in other places. Would doing that everywhere remove the need for sleeps?

hermanlee:
I think I can rework the test to remove the sleep calls by leveraging the fact setting trx_commit=1 should result in the wal being flushed and the counter being incremented.

PR #725, opened 2017-10-11
Rework rocksdb.write_sync test without using sleep calls

hermanlee:
Summary:

Remove sleep calls from rocksdb.write_sync and make the test more
deterministic.

Test Plan:

mtr rocksdb.write_sync


hermanlee:
https://github.com/facebook/mysql-5.6/issues/724

lth (approved):

PR #729, opened 2017-10-12
Created new information_schema.rocksdb_deadlock table.

george-lorch:
This table tracks historic deadlocks based on the functionality introduced in
"Display Deadlock Paths".
Copied the deadlock_tracking test and touched it up to test the new table as
opposed to the SHOW ENGINE ROCKSDB TRANSACTION STATUS.


lth:
The tests `main.mysqlshow`, `main.mysqld--help-notwin-profiling`, `main.mysqld--help-notwin` need updating.

george-lorch:
Ahh yeah, sorry about that. I forgot that in fb-mysql I_S, sysvars and others show in these tests whereas they do not in PS. I'll get them updated and re-push

lth:
It would also be great if the existing SHOW ENGINE code could reuse this code, but it's okay if you don't want to do it.

george-lorch:
That is not a problem at all. I can remove the 'parallel' implementation bits in the Rdb_snapshot_status that fills the string data buffer and change the populate_deadlock_buffer to behave more like the loop in the I_S fill or even remove it entirely and put the loop right in populate_deadlock_buffer. I considered that initially but figured I would keep things apart and see what kind of review feedback emerged.

If you have any suggestions on the schema to make it more clear of the relationships between rows for an individual deadlock or any other info that might be useful, please let me know.

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -3367,8 +3347,52 @@ class Rdb_snapshot_status : public Rdb_tx_list_walker {
     m_data += "----------LATEST DETECTED DEADLOCKS----------\n";
 
     for (auto path_entry : dlock_buffer) {
-      m_data += get_dlock_path_info(path_entry);
+      std::string path_data;
+      const auto dl_info = get_dl_path_trx_info(path_entry);
```
lth: I think `get_dl_path_trx_info` does not need to be called if limit is exceeded.

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -3367,8 +3347,52 @@ class Rdb_snapshot_status : public Rdb_tx_list_walker {
...
+      if (path_entry.limit_exceeded) {
+        path_data += "\n-------DEADLOCK EXCEEDED MAX DEPTH-------\n";
+      } else {
+        path_data += "\n*** DEADLOCK PATH\n"
+                     "=========================================\n";
+        for (auto it = dl_info.path.begin(); it != dl_info.path.end(); it++) {
+          const auto trx_info = *it;
+          path_data += format_string(
+              "TRANSACTIONID: %u\n"
```
lth: This isn't from your diff, but I think "TRANSACTION ID" would be more consistent.

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -3367,8 +3347,52 @@ class Rdb_snapshot_status : public Rdb_tx_list_walker {
...
+          if (it != dl_info.path.end() - 1) {
+            path_data += "---------------WAITING FOR---------------\n";
+          }
+        }
+
+        if (!path_entry.path.empty() && !path_entry.limit_exceeded) {
+          /* prints the txn id of the transaction that caused the deadlock */
+          auto deadlocking_txn = *(path_entry.path.end() - 1);
+          path_data += format_string(
+              "\n--------TRANSACTIONID: %u GOT DEADLOCK---------\n",
```
lth: Same thing, "TRANSACTION ID".

storage/rocksdb/rdb_i_s.cc
```diff
@@ -1476,6 +1476,117 @@ static int rdb_i_s_trx_info_init(void *const p) {
...
+                       0),
+    ROCKSDB_FIELD_INFO("CAUSER_ID", sizeof(ulonglong), MYSQL_TYPE_LONGLONG,
+                       0),
+    ROCKSDB_FIELD_INFO("TRANSACTION_ID", sizeof(ulonglong), MYSQL_TYPE_LONGLONG,
+                       0),
+    ROCKSDB_FIELD_INFO("CF_NAME", NAME_LEN + 1, MYSQL_TYPE_STRING, 0),
+    ROCKSDB_FIELD_INFO("WAITING_KEY", FN_REFLEN + 1, MYSQL_TYPE_STRING, 0),
+    ROCKSDB_FIELD_INFO("LOCK_TYPE", NAME_LEN + 1, MYSQL_TYPE_STRING, 0),
+    ROCKSDB_FIELD_INFO("INDEX NAME", NAME_LEN + 1, MYSQL_TYPE_STRING, 0),
+    ROCKSDB_FIELD_INFO("TABLE NAME", NAME_LEN + 1, MYSQL_TYPE_STRING, 0),
```
lth: Can we standardize on underscores vs spaces? I think we tend to use underscore.

storage/rocksdb/ha_rocksdb.h
```diff
@@ -91,6 +91,25 @@ struct Rdb_trx_info {
...
+  struct Rdb_dl_trx_info {
+    ulonglong trx_id;
+    std::string cf_name;
+    std::string waiting_key;
+    bool exclusive_lock;
+    std::string index_name;
+    std::string table_name;
+  };
+  std::vector <Rdb_dl_trx_info> path;
+  ulonglong causer_trx_id;
```
lth: I think the intention of the message `TXN_ID GOT DEADLOCK` was to denote which transaction got rolled back. It's not so much the causer as it is the victim (although in rocksdb, the causer is *always* the victim).

How about a bool column called "Error" or "Rolled back" or something to denote that transaction got deadlock error? It would avoid repeating the id for every row.

lth (commented):
I think the schema looks pretty good overall!

Can you also rebase?

storage/rocksdb/ha_rocksdb.h
george-lorch: That actually makes a lot more sense than re-listing the transaction. Perhaps something similar can be added to the txn block output in the SHOW STATUS.

george-lorch:
All great comments. I am travelling the next two days so may not get a push with the changes until Friday or next week, just didn't want you to think I abandoned it.

george-lorch:
OK @lth, I rebased and cleaned up all things as discussed and re-worked the output structure of both the I_S table and the SERTS output. I eliminated the redundant reporting of the victim/rolled back transaction and added the simple ROLLED BACK field which should indicate the same idea. Cleaned up/normalized the field names, etc... Please take another look when you have a few minutes and let me know if I missed or misunderstood anything or if you have any more ideas on how to improve the feature. Thanks!

mysql-test/suite/rocksdb/r/deadlock_tracking.result
```diff
@@ -59,22 +59,15 @@ KEY
...
 TABLE NAME: test.t
----------------WAITING FOR---------------
-TXN_ID
-COLUMN FAMILY NAME: default
-KEY
-LOCK TYPE: EXCLUSIVE
-INDEX NAME: PRIMARY
-TABLE NAME: test.t
-
---------TXN_ID GOT DEADLOCK---------
```
lth: Actually, what I was going for was that the SHOW ENGINES code can keep using `TRANSACTION ID GOT DEADLOCK`, since I think that's a more concise way of expressing that info.

I think the "rolled back" column makes more sense in the I_S table, than in the show engines command.

lth (commented):
One more test to update: `main.information_schema`.

mysql-test/suite/rocksdb/r/deadlock_tracking.result
george-lorch: Ahh, OK, no problem, I can bump that back, but what about the original also re-printing the deadlocked txn info ? Maybe just list the transaction path and finish with the TRANSACTION ID <n> GOT DEADLOCK ?

lth: Yeah, we can take out the reprinted transaction.

george-lorch:
OK, I added back the 'GOT DEADLOCK' and removed the 'ROLLED BACK' on the status output as discussed, re-recorded the test and ran another mtr cycle.

lth:
The diff looks good to me. The only thing is to update `main.information_schema`.

george-lorch:
Thanks Manuel, I fixed up the main.information_schema result and just re-pushed.

PR #738, opened 2017-10-17
Implement inplace alter for auto_increment

lth:
This implements inplace alter for changing auto_increment values. It works by reading the last value of the auto increment index and compares that with the request auto_increment value in the alter statement. The larger of the two values is set as the new autoincrement value for the table.

To handle partitions, every partition is handled as a separate table. Since we call `prepare_inplace_alter_table` for every partition but `commit_inplace_alter_table` only once, we load the last value in the index during prepare and write into the data dictionary on commit.

### DIFF
--- a/mysql-test/suite/rocksdb/t/autoinc_crash_safe_partition.test
+++ b/mysql-test/suite/rocksdb/t/autoinc_crash_safe_partition.test
@@ -1,8 +1,9 @@
 --source include/have_rocksdb.inc
 --source include/have_binlog_format_row.inc
 --source include/master-slave.inc
+--source include/have_partition.inc
 
-create table t (i int primary key auto_increment) engine=rocksdb;
+create table t (i int primary key auto_increment) engine=rocksdb partition by key (i) partitions 3;
 
 --source suite/rocksdb/include/autoinc_crash_safe.inc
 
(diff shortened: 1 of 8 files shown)


mysql-test/suite/rocksdb/t/autoinc_crash_safe_partition.test
```diff
@@ -0,0 +1,9 @@
+--source include/have_rocksdb.inc
```
laurynas-biveinis, line 1: Could also add have_partition.inc

mysql-test/suite/rocksdb/t/autoinc_crash_safe.test
```diff
@@ -2,129 +2,8 @@
...
---let $rpl_server_number = 1
---source include/rpl_restart_server.inc
-select table_schema, table_name, auto_increment from information_schema.tables where table_name = 't';
-
---echo # Drop table.
-drop table t;
-
---let $rpl_server_number = 1
---source include/rpl_restart_server.inc
+--source suite/rocksdb/include/autoinc_crash_safe.inc
```
laurynas-biveinis, line 7: There is a rocksdb_rpl suite, which would seem as a better fit for replication bits, and ideally the tests could be split into non-rpl and rpl autoinc_crash_safe. But it looks like the autoinc tests are quite tightly coupled to replication, thus it would take some effort to split the two.

hermanlee (commented):
Mostly looks good to me, though github made reading this diff difficult.

Do you have a test case that does the following?

create table t (a int, b int, primary key (a)) engine = rocksdb;
insert into t values (1,1), (10,10);
alter table t add column (c int auto_increment), add key(c), auto_increment=15;

lth:
No, I'll add it.

hermanlee (approved):

PR #745, opened 2017-10-25
[MyRocks] Setting iterate_upper_bounds as rocksdb.ReadOptions

yoshinorim:
Summary: This diff explicitly sets iterate_upper_bounds
and passes to rocksdb read options. This will avoid
scanning tombstones outside of the given key ranges.
This does not change behaviors when prefix bloom filter is used,
since prefix_same_as_start can be set in that case.

This diff will be updated when RocksDB supports
iterate_lower_bounds.

### DIFF
--- a/storage/rocksdb/ha_rocksdb.cc
+++ b/storage/rocksdb/ha_rocksdb.cc
@@ -126,10 +126,6 @@ static handler *rocksdb_create_handler(my_core::handlerton *hton,
                                        my_core::TABLE_SHARE *table_arg,
                                        my_core::MEM_ROOT *mem_root);
 
-bool can_use_bloom_filter(THD *thd, const Rdb_key_def &kd,
-                          const rocksdb::Slice &eq_cond,
-                          const bool use_all_keys);
-
 static rocksdb::CompactRangeOptions getCompactRangeOptions() {
   rocksdb::CompactRangeOptions compact_range_options;
   compact_range_options.bottommost_level_compaction =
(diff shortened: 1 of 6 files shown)


storage/rocksdb/ha_rocksdb.cc
```diff
@@ -8659,9 +8676,14 @@ int ha_rocksdb::check_and_lock_sk(const uint &key_id,
     return row_info.tx->set_status_error(table->in_use, s, kd, m_tbl_def,
                                          m_table_handler);
   }
+  if (total_order_seek) {
+    setup_iterator_bounds(kd, &new_slice);
```
hermanlee: I think the setup_iterator_bounds code should go into can_use_bloom_filter() and the latter should be renamed so that it's clear there's a side effect. This makes it clear the iterator bounds needs to be setup if bloom filters are not being used. So when can_use_bloom_filter() is going to return false, then within the function, immediately call setup_iterator_bounds() to setup the upper/lower bounds.

hermanlee (commented):
Functionally, the code looks fine. The code around dropping indexes, remove_rows() can run with total_seek_order = true and can potentially hit this problem. In those cases, the boundary should be index+1 (or index in the case of reverse_cf). Do you want to add fixes for those calls too?

storage/rocksdb/rdb_datadic.cc
```diff
@@ -806,6 +806,25 @@ int Rdb_key_def::successor(uchar *const packed_tuple, const uint &len) {
...
+
+  int changed = 0;
+  uchar *p = packed_tuple + len - 1;
+  for (; p > packed_tuple; p--) {
+    changed++;
+    if (*p != uchar(0x00)) {
+      *p = *p - 1;
+      break;
+    }
+    *p = '\0';
```
lth: Is this correct? Shouldn't it wrap around to 0xff?

hermanlee (commented):
Changes look good to me. I'll let Manuel take a look at them before accepting.

yoshinorim:
rebased

lth (approved):
Looks good to me!

PR #746, opened 2017-10-25
Cleaned up bulk load tests.

george-lorch:
- Moved .inc files into include dir
- Added count_sessions and wait_for_count_sessions includes around multiple session use
- Closed session that was being left open for consistent snapshot
- Removed deprecated idioms (DROP TABLE IF EXISTS)
- Removed spurious re-setting of session variable to default value on test cleanup
- Expicitly specify ROCKSDB storage engine on CREATE TABLE
- Re-recorded results


mysql-test/suite/rocksdb/include/bulk_load_unsorted.inc
```diff
@@ -1,14 +1,11 @@
---disable_warnings
-DROP TABLE IF EXISTS t1;
---enable_warnings
-
```
asandryh: Missing `--source include/count_sessions.inc` here

george-lorch: Now how the heck did I do that, this is the file that started me in on this work :-\

lth:
@georgelorchpercona can you rebase past https://github.com/facebook/mysql-5.6/commit/5969a3ef953cbd7838581c4553edb11c341ec174 It's failing to build internally.

Also, just a heads up that that commit changes how dependencies are given to MyRocks, so you might want to update your setups.

george-lorch:
Ack, thanks for the heads up on that Manuel, that should come with our October merge in a few weeks. It doesn't look like it will be terribly disruptive due to the way we build and package.

Rebased the change on this mornings HEAD of fb-mysql-5.6.35 and also touched another test that I missed the first time around. Seems to build and clear mtr.

Issue #741, opened 2017-10-21
table scans are slower in MyRocks build from October 16


mdcallag:
There is a regression between Aug 15 and Oct 16 that:
1) makes full index scans slower
2) make some explain full index scans slower

My workload is the insert benchmark with a cached database. The repro is 1) create table with 1 PK and 3 secondary indexes and 2) insert 10M rows and then 3) do a full scan of the PK, then each secondary index, then the PK and 4) do an explain for each of the queries from the previous step

The problems I see are that the full scans in step 3 are about 3X slower and  then the explain of the full scan for the PK takes 10+ seconds. Explain has always taken less than 1 second.

I will guess that the problem comes from https://github.com/facebook/mysql-5.6/pull/722

PMP for the slow explain shows:
```
      1 std::swap<rocksdb::Slice>,std::iter_swap<__gnu_cxx::__normal_iterator<rocksdb::Slice*,,std::__reverse<__gnu_cxx::__normal_iterator<rocksdb::Slice*,,std::reverse<__gnu_cxx::__normal_iterator<rocksdb::Slice*,,rocksdb::MergeContext::SetDirectionBackward,rocksdb::MergeContext::PushOperand,rocksdb::SaveValue,rocksdb::(anonymous,rocksdb::MemTable::Get,rocksdb::MemTable::Get,rocksdb::DBImpl::GetImpl,rocksdb::DBImpl::Get,rocksdb::DB::Get,myrocks::Rdb_dict_manager::get_value,myrocks::Rdb_dict_manager::get_auto_incr_val,myrocks::ha_rocksdb::load_auto_incr_value,myrocks::ha_rocksdb::open,handler::ha_open,open_table_from_share,open_table,open_and_process_table,open_tables,open_normal_and_derived_tables,execute_sqlcom_select,mysql_execute_command,mysql_parse,dispatch_command,do_command,do_handle_one_connection,handle_one_connection,start_thread,clone
```

Linux perf for the slow explain is:
```
    99.69%  my-oneconnectio  mysqld              [.] _ZN7rocksdbL9SaveValueEPvPKc
            |
            ---_ZN7rocksdbL9SaveValueEPvPKc
               _ZN7rocksdb12_GLOBAL__N_111SkipListRep3GetERKNS_9LookupKeyEPvPFbS5_PKcE
[...]

PR #747, opened 2017-10-26
Load autoincrement and hidden pk only once

lth:
Currently, we load hidden pk and auto increment ids unconditionally on every `ha_rocksdb::open`. This is actually not necessary, since we can rely on the in-memory value after the server initially starts up. It's only on restarts that we need to reload the value, in which case we just load from disk on the first few opens.

Closes https://github.com/facebook/mysql-5.6/issues/741


hermanlee (approved):
Awesome fix!

PR #755, opened 2017-10-30
Fixed rocksdb_bulk_load_allow_unsorted option callback.

george-lorch:
The option uses a 'set' callback to set/return an error to the client which is
an invalid thing to do. This is proven by the debug assertion that is hit on a
debug build when executing the sequence in the rocksdb.bulk_load_unsorted_errors
test. The proper way to return an error from a SET is to use a 'check' callback
instead. This patch changes the 'set' to a 'check' and removes the no-debug
constraint on the bulk_load_unsorted_errors test.

### DIFF
--- a/storage/rocksdb/ha_rocksdb.cc
+++ b/storage/rocksdb/ha_rocksdb.cc
@@ -415,9 +415,9 @@ static int rocksdb_check_bulk_load(THD *const thd,
                                    void *save,
                                    struct st_mysql_value *value);
 
-static void rocksdb_set_bulk_load_allow_unsorted(
-    THD *thd, struct st_mysql_sys_var *var MY_ATTRIBUTE((__unused__)),
-    void *var_ptr, const void *save);
+static int rocksdb_check_bulk_load_allow_unsorted(
+    THD *const thd, struct st_mysql_sys_var *var MY_ATTRIBUTE((__unused__)),
+    void *save, struct st_mysql_value *value);
 
 static void rocksdb_set_max_background_jobs(THD *thd,
                                             struct st_mysql_sys_var *const var,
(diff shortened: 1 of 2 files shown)


asandryh (approved):
Thanks for fixing the test for the debug mode!

george-lorch:
Not a problem Aliaksei. Thanks for considering.

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -12518,16 +12519,43 @@ int rocksdb_check_bulk_load(THD *const thd, struct st_mysql_sys_var *var
...
-void rocksdb_set_bulk_load_allow_unsorted(
-    THD *const thd,
-    struct st_mysql_sys_var *const var MY_ATTRIBUTE((__unused__)),
-    void *const var_ptr, const void *const save) {
+int rocksdb_check_bulk_load_allow_unsorted(
+    THD *const thd, struct st_mysql_sys_var *var MY_ATTRIBUTE((__unused__)),
+    void *save, struct st_mysql_value *value) {
+  my_bool new_value;
+  int new_value_type = value->value_type(value);
+  if (new_value_type == MYSQL_VALUE_TYPE_STRING) {
```
laurynas-biveinis, line 12527: This is duplicated code - make a helper function to extract a my_bool from a st_mysql_value?

asandryh: Good point, I'll refactor this shortly.

george-lorch: When I did the first fix for rocksdb_bulk_load I tried to find some place in sys_vars where this transformation of a bool-string value to a real bool but could not find anything that made sense, thus duplicating the server behavior that transforms "true", "false", "on", "off" to a true bool. It would be nice if we could somehow come up with a change to sys_var functionality that did this in some kind of a 'natural' way. I do not know if this is the first time this has come up but I am pretty sure it will not be the last.

PR #756, opened 2017-10-30
Fix rocksdb.skip_validate_tmp_table

george-lorch:
- Updated test to remove invalid/inaccurate comment and converted to using
  modern mtr semantics via include/restart_mysqld.inc.
- Added some extra validation via 'show tables' to confirm tables that
  server believes should exist.


asandryh (approved):
Thank you!

george-lorch:
My pleasure Aliaksei.

PR #759, opened 2017-11-06
Set rocksdb_max_open_files to reasonable value at startup from open_f…

george-lorch:
…iles_limit

- Leave the default of max_open_files to -1 to trigger auto-tune-ish.
- On start, after server sets open_files_limit, during plugin init, check to see
  if max_open_files is > open_files_limit or < 0, if it is, set it to
  open_files_limit / 2.
- Added new test that restarts the server with different known good and bad
  values for max_open_files and validates that it is either left alone or set
  to open_files_limit / 2.
- Touched rocksdb.rocksdb test to account for 'floating' effective default value
  of rocksdb_max_open_files across installs, distros, and local configs.

### DIFF
--- a/mysql-test/suite/rocksdb_sys_vars/t/rocksdb_max_open_files_basic.test
+++ b/mysql-test/suite/rocksdb_sys_vars/t/rocksdb_max_open_files_basic.test
@@ -1,6 +1,8 @@
 --source include/have_rocksdb.inc
 
---let $sys_var=ROCKSDB_MAX_OPEN_FILES
---let $read_only=1
---let $session=0
---source ../include/rocksdb_sys_var.inc
+# We can not use rocksdb_sys_var.inc here as this is a global, read only option
+# whose value is dependent on the servers open_files_limit. It is more fully
+# tested in the rocksdb.max_open_files test.
+
+--replace_column 2 #
+show variables like 'rocksdb_max_open_files';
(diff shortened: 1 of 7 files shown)


mysql-test/suite/rocksdb_sys_vars/t/rocksdb_max_open_files_basic.test
```diff
@@ -1,6 +1,8 @@
 --source include/have_rocksdb.inc
 
---let $sys_var=ROCKSDB_MAX_OPEN_FILES
---let $read_only=1
---let $session=0
---source ../include/rocksdb_sys_var.inc
+# We can not use rocksdb_sys_var.inc here as this is a global, read only option
+# whose value is dependent on the servers open_files_limit. It is more fully
+# tested in the tocksdb.max_open_files test.
```
lth: `tocksdb` -> `rocksdb`

hermanlee: @georgelorchpercona These changes look good. Could you update the pull request to fix the typo? We can commit it after.

hermanlee (changes requested):
Fix typo.

mysql-test/suite/rocksdb_sys_vars/t/rocksdb_max_open_files_basic.test
george-lorch: Rebased it to current fb-mysql-5.6.35 HEAD, fixed the typo and pushed. Thanks Herman!

hermanlee (approved):

PR #764, opened 2017-11-13
Rename concurrent_prepare to two_write_queues

maysamyabandeh:
Summary: concurrent_prepare in rocksdb is renamed to two_write_queues.
This patch updates the rocksdb submodule and make similar change in
myrocks.
@update-submodule: rocksdb

### DIFF
--- a/mysql-test/r/mysqld--help-notwin-profiling.result
+++ b/mysql-test/r/mysqld--help-notwin-profiling.result
@@ -1161,9 +1161,6 @@ The following options may be given as the first argument:
  Enable or disable ROCKSDB_COMPACTION_STATS plugin.
  Possible values are ON, OFF, FORCE (don't start if the
  plugin fails to load).
- --rocksdb-concurrent-prepare 
- DBOptions::concurrent_prepare for RocksDB
- (Defaults to on; use --skip-rocksdb-concurrent-prepare to disable.)
  --rocksdb-create-checkpoint=name 
  Checkpoint directory
  --rocksdb-create-if-missing 
@@ -1438,6 +1435,9 @@ The following options may be given as the first argument:
  --rocksdb-trx[=name] 
  Enable or disable ROCKSDB_TRX plugin. Possible values are
  ON, OFF, FORCE (don't start if the plugin fails to load).
+ --rocksdb-two-write-queues 
+ DBOptions::two_write_queues for RocksDB
+ (Defaults to on; use --skip-rocksdb-two-write-queues to disable.)
  --rocksdb-unsafe-for-binlog 
  Allowing statement based binary logging which may break
  consistency
@@ -2126,7 +2126,6 @@ rocksdb-compaction-sequential-deletes-count-sd FALSE
 rocksdb-compaction-sequential-deletes-file-size 0
 rocksdb-compaction-sequential-deletes-window 0
 rocksdb-compaction-stats ON
-rocksdb-concurrent-prepare TRUE
 rocksdb-create-checkpoint 
 rocksdb-create-if-missing TRUE
 rocksdb-create-missing-column-families FALSE
@@ -2219,6 +2218,7 @@ rocksdb-table-stats-sampling-pct 10
 rocksdb-tmpdir (No default value)
 rocksdb-trace-sst-api FALSE
 rocksdb-trx ON
+rocksdb-two-write-queues TRUE
 rocksdb-unsafe-for-binlog FALSE
 rocksdb-update-cf-options (No default value)
 rocksdb-use-adaptive-mutex FALSE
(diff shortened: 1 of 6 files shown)


lth:
`rocksdb_sys_vars.all_vars` is failing because you need to rename `mysql-test/suite/rocksdb_sys_vars/t/rocksdb_concurrent_prepare_basic.test`.

mysql-test/r/mysqld--help-notwin-profiling.result
```diff
@@ -1162,7 +1162,7 @@ The following options may be given as the first argument:
  Possible values are ON, OFF, FORCE (don't start if the
  plugin fails to load).
  --rocksdb-concurrent-prepare
```
laurynas-biveinis: This is suspicious - the variable help text is changed but not the variable name itself. Likewise re. --skip-rocksdb-concurrent-prepare two lines down. Does the testcase pass?

@georgelorchpercona, does the var rename make it to our GA freeze?

mysql-test/r/mysqld--help-notwin.result
```diff
@@ -1160,7 +1160,7 @@ The following options may be given as the first argument:
  Possible values are ON, OFF, FORCE (don't start if the
  plugin fails to load).
  --rocksdb-concurrent-prepare
```
laurynas-biveinis: Likewise

mysql-test/suite/rocksdb/r/rocksdb.result
```diff
@@ -879,7 +879,7 @@ rocksdb_compaction_sequential_deletes	0
 rocksdb_compaction_sequential_deletes_count_sd	OFF
 rocksdb_compaction_sequential_deletes_file_size	0
 rocksdb_compaction_sequential_deletes_window	0
-rocksdb_concurrent_prepare	ON
+rocksdb_two_write_queues	ON
```
laurynas-biveinis: The list looks sorted. Does the testcase pass?

maysamyabandeh: Thanks. I just noticed the test failure. I rushed into submitting the PR. I will submit a fix soon.

mysql-test/r/mysqld--help-notwin-profiling.result
maysamyabandeh: I am going to keep the myrocks var name. We only need to update the rocksdb var.

george-lorch: @laurynas-biveinis, no, this will not likely make the PS 5.7 MyRocks GA. This PR hasn't even landed yet. If it did tomorrow, it would still be be within the November merge group which is not planned to be included in the GA. We are currently planning on releasing GA merged up through September or October, with October being a bit of a stretch. So we are effectively already frozen. The change, as originally authored, would eventually be reviewed and discarded or merged in a fixed up manner that would not violate our compatibility policy. As stated many times over the last year, once we ship GA, we will no longer merge into our 5.7 tree any changes that introduce renaming of variables or alterations of default behaviors of variables in a mid-series release. Users demand that mid-series upgrades just work without needing to alter configurations or their applications. Any such changes will have to wait until we port to 8.0 or we would need very explicit direction from Vadim and/or PZ to override that policy.

george-lorch: These kinds of things are also problematic in rocksdb, not just myrocks, if an option name is changed with no aliasing, rocksdb will fail to start until the user edits the rocksdb OPTIONS file and corrects it or uses the recently added "fb-mysql-5.6.35/13d2e6f2da9808ad7b4cfe86394aa56734572e7d : Enable unknown options being passed to rocksDB". This is terrible practice to alter option names and default values and expect users to know and understand that they need to perform manual steps on upgrades.

Here is a case already that illustrates what an end user potentially could hit if options are changed and released into the wild with no backward compatibility or automatic upgrade. I do concede though that his particular instance looks like the user is attempting to shuffle variants without fully cleaning their datadir in between installs, but you get the idea of what we will run into. https://groups.google.com/forum/#!topic/percona-discussion/JEF4njF1094

maysamyabandeh: at rocksdb the old var name is still accepted but simply ignored. the "enable unknown option" is only for forward compatibility: when you downgrade to an earlier version where the old binary does not understand the options added by the new binary.

lth: We added `ignore_unknown_options` for the purpose of downgrading, with the understanding that it's okay if miss setting some of these options on an older version. Upgrades should work just fine.

Since this is going to cause problems anyway with the OPTIONS file, I think we should make a clean break and rename the MyRocks variable as well, so it's clear to the user that a variable rename has occurred.

lth (approved):

PR #765, opened 2017-11-29
Fixed incorrect use of pointer to internal data of a temporary instance.

george-lorch:
- std::string instance is being returned as a temporary. C++ standard defines
  that its lifetime is the end of the complete lexical expession. The expression
  is maintaining a pointer to the internal char* (c_str) pointer that is getting
  destroyed at the completion of the expression, this rendering the pointer
  as pointing to invalid data.
- Changed to a const reference to the returned temporary, avoiding a little
  extra overhead of an additional copy and forcing the temporary to remain in
  scope beyond the expression where it is created.
- Slightly optimized rdb_corruption_file_name to reduce number of copy
  constructions and temporary heap allocations.

### DIFF
--- a/storage/rocksdb/ha_rocksdb.cc
+++ b/storage/rocksdb/ha_rocksdb.cc
@@ -12805,8 +12805,9 @@ double ha_rocksdb::read_time(uint index, uint ranges, ha_rows rows) {
 }
 
 std::string rdb_corruption_marker_file_name() {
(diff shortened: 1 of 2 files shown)


hermanlee (commented):
These changes look fine to me. I'll let Artem take a look too.

hermanlee (approved):

PR #771, opened 2017-12-14
Alter the rocksdb_max_open_files limit checking to allow pass-through…

george-lorch:
… of -1.

New behavior for rocksdb_max_open_files is as such :
* -1 : infinite, pass -1 to rocksdb
* 0 : 'auto-tune', silently set rocksdb_max_open_files to 1/2 of mysql
  open_files_limit
* > 0 <= open_files_limit : take value as-is
* > open_files_lmit :  set rocksdb_max_open_files to 1/2 of mysql open_files_limit
  and emit a warning to the mysqld error log.

### DIFF
--- a/storage/rocksdb/ha_rocksdb.cc
+++ b/storage/rocksdb/ha_rocksdb.cc
@@ -500,6 +500,7 @@ static std::unique_ptr<rocksdb::DBOptions> rdb_init_rocksdb_db_options(void) {
   o->listeners.push_back(std::make_shared<Rdb_event_listener>(&ddl_manager));
   o->info_log_level = rocksdb::InfoLogLevel::INFO_LEVEL;
   o->max_subcompactions = DEFAULT_SUBCOMPACTIONS;
+  o->max_open_files = -2; // auto-tune to 50% open_files_limit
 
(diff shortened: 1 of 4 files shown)


storage/rocksdb/ha_rocksdb.cc
```diff
@@ -500,6 +500,7 @@ static std::unique_ptr<rocksdb::DBOptions> rdb_init_rocksdb_db_options(void) {
   o->listeners.push_back(std::make_shared<Rdb_event_listener>(&ddl_manager));
   o->info_log_level = rocksdb::InfoLogLevel::INFO_LEVEL;
   o->max_subcompactions = DEFAULT_SUBCOMPACTIONS;
+  o->max_open_files = -2; // auto-tune to 50% open_files_limit
```
hermanlee, line 503: Could define the -2 in a #define or const static int variable to be reused later in the code.

hermanlee (commented):
Looks good to me.

hermanlee (changes requested):
Can you also re-record the test main.mysqld--help-notwin and change the values from -1 to -2?

hermanlee (approved):

PR #782, opened 2018-01-09
Support building with libressl

jyaworski:
Libressl reports a major version number of '2', which causes the ssl check to fail.


jyaworski:
Any update? @Tema?

PR #788, opened 2018-01-25
[MyRocks] Cheking optimize_filters_for_hits cf opt for bulk loading

yoshinorim:
Summary: SstFileWriter added a feature whether to create a bloom filter.
By default, bloom filter is always created. This diff checks
optimize_filters_for_hits cf option and skips bloom filter if it is
turned on.

Test Plan: mtr
@update-submodule: rocksdb


hermanlee (approved):

mysql-test/suite/rocksdb/t/bloomfilter_bulk_load-master.opt
```diff
@@ -0,0 +1,2 @@
+--rocksdb_default_cf_options=write_buffer_size=16k;block_based_table_factory={filter_policy=bloomfilter:10:false;whole_key_filtering=0;};prefix_extractor=capped:12
```
george-lorch, line 1: suggest using --loose-rocksdb_*

PR #791, opened 2018-02-08
Optimize favicon

zpao:
This was a 16MB 2000x2000 pixel image. It may be long cached but this is
never rendered at more than maybe 64x64. It's now 99.97% smaller.

I also fixed the mime type, it's not an x-icon.


JoelMarcey (approved):
This is good. Thanks!

Issue #779, opened 2018-01-03
Alter table with string indexed column to non-binary collation works


tplavcic:
MyRocks doesn't seem to support indexes on char fields without using binary collations and it prevents that on create table, but I can still create table with binary collation and alter it to non-binary collation.

Here's two examples:
1. alter rocksdb table with binary collation to non-binary collation
```
5.6.35>create table t1 (a int primary key, b char(3), key(b)) engine=rocksdb default charset=latin1 collate=latin1_general_ci;
ERROR 12049 (HY000): Unsupported collation on string indexed column test.t1.b Use binary collation (binary, latin1_bin, utf8_bin).

5.6.35>create table t1 (a int primary key, b char(3), key(b)) engine=rocksdb default charset=latin1 collate=latin1_bin;
Query OK, 0 rows affected (0.03 sec)

5.6.35>insert into t1 values (1, "  a");
Query OK, 1 row affected (0.00 sec)

5.6.35>alter table t1 collate=latin1_general_ci;
Query OK, 1 row affected (0.05 sec)
Records: 1  Duplicates: 0  Warnings: 0

5.6.35>insert into t1 values (2, "  a");
Query OK, 1 row affected (0.00 sec)

5.6.35>show create table t1;
+-------+------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| Table | Create Table                                                                                                                                                                                                           |
+-------+------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| t1    | CREATE TABLE `t1` (
  `a` int(11) NOT NULL,
  `b` char(3) CHARACTER SET latin1 COLLATE latin1_bin DEFAULT NULL,
  PRIMARY KEY (`a`),
  KEY `b` (`b`)
) ENGINE=ROCKSDB DEFAULT CHARSET=latin1 COLLATE=latin1_general_ci |
+-------+------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
1 row in set (0.00 sec)

5.6.35>show table status like 't1';
+------+---------+---------+------------+------+----------------+-------------+-----------------+--------------+-----------+----------------+-------------+-------------+------------+-------------------+----------+----------------+---------+
| Name | Engine  | Version | Row_format | Rows | Avg_row_length | Data_length | Max_data_length | Index_length | Data_free | Auto_increment | Create_time | Update_time | Check_time | Collation         | Checksum | Create_options | Comment |
[...]

PR #796, opened 2018-02-26
Introduce new variable 'rocksdb_error_on_suboptimal_collation'.

george-lorch:
New variable specifies whether to raise an error or a warning if an index is
created on a char field where the table has a sub-optimal collation (case
insensitive).

Default is TRUE. Variable is global read only.

### DIFF
--- a/mysql-test/suite/rocksdb/t/collation.test
+++ b/mysql-test/suite/rocksdb/t/collation.test
@@ -2,14 +2,12 @@
 --source include/have_fullregex.inc
 
 call mtr.add_suppression("Invalid pattern");
-SET @start_global_value = @@global.ROCKSDB_STRICT_COLLATION_EXCEPTIONS;
-
---disable_warnings
-DROP TABLE IF EXISTS t1;
---enable_warnings
 
 # ci non-indexed column is allowed
 CREATE TABLE t1 (id INT primary key, value varchar(50), value2 varbinary(50), value3 text) engine=rocksdb charset utf8;
+# ci indexed column is not allowed
+--error ER_UNSUPPORTED_COLLATION
+ALTER TABLE t1 ADD INDEX (value);
 DROP TABLE t1;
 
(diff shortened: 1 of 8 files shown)


mysql-test/suite/rocksdb/t/collation.test
```diff
@@ -29,6 +27,8 @@ DROP TABLE t1;
 
 # cs latin1_bin is allowed
 CREATE TABLE t1 (id varchar(20), value varchar(50), value2 varchar(50), value3 text, primary key (id), index(value, value2)) engine=rocksdb charset latin1 collate latin1_bin;
+# THIS SHOULD FAIL BUT IT DOES NOT
```
laurynas-biveinis, line 30: ?

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -492,6 +492,7 @@ static my_bool rocksdb_large_prefix = 0;
 static my_bool rocksdb_allow_to_start_after_corruption = 0;
 static uint32_t rocksdb_write_policy =
     rocksdb::TxnDBWritePolicy::WRITE_COMMITTED;
+static my_bool rocksdb_error_on_suboptimal_collation = 1;
```
laurynas-biveinis, line 495: s/1/TRUE

mysql-test/suite/rocksdb/t/collation.test
george-lorch: Added simple test and notation to cover https://github.com/facebook/mysql-5.6/issues/779 which is related/overlaps this test. See also https://jira.percona.com/browse/MYR-223

storage/rocksdb/ha_rocksdb.cc
george-lorch: Not in upstream fb, 0/1 is the standard. We 'fix' it on merge downstream.

Issue #802, opened 2018-03-29
MyRocks: Statement rollback doesnt work correctly for nested statements


spetrunia:
Consider this example:

Prepare the testcase:
```
create table t1 (a varchar(100)) engine=rocksdb;
create table two_rows(a int) engine=rocksdb;
insert into two_rows values (1), (2);

create table t2(a varchar(100)) engine=rocksdb;

delimiter //
create function func() returns varchar(100) deterministic 
begin
  insert into t2 values ('func-called');
  set @a= (select a from two_rows);
  return 'func-returned';
end;//
delimiter ;
```

Testcase itself:
```
MySQL [tmyrocks]> begin;
Query OK, 0 rows affected (0.00 sec)

MySQL [tmyrocks]>   insert into t1 values (func());
ERROR 1242 (21000): Subquery returns more than 1 row
MySQL [tmyrocks]>   select * from t1;
Empty set (0.00 sec)

MySQL [tmyrocks]>   select * from t2;
+-------------+
| a           |
+-------------+
| func-called |
+-------------+
1 row in set (0.00 sec)
```

This is a wrong result.  The statement `insert into t1` failed with an error, and so all of its effects should have been rolled back.

Trying the same with InnoDB:
```
create table t1 (a varchar(100)) engine=innodb;
create table two_rows(a int) engine=innodb;
insert into two_rows values (1), (2);

create table t2(a varchar(100)) engine=innodb;

delimiter //
create function func() returns varchar(100) deterministic 
begin
  insert into t2 values ('func-called');
  set @a= (select a from two_rows);
  return 'func-returned';
end;//

delimiter ;
```

```
MySQL [test]> begin;
Query OK, 0 rows affected (0.00 sec)

MySQL [test]>   insert into t1 values (func());
ERROR 1242 (21000): Subquery returns more than 1 row
MySQL [test]>   select * from t1;
Empty set (0.00 sec)

MySQL [test]>   select * from t2;
Empty set (0.00 sec)
```

spetrunia:
I've studied how InnoDB does it.

`trx_mark_sql_stat_end()` has this line to update the information about to where statement rollback should rollback:

```
	trx->last_sql_stat_start.least_undo_no = trx->undo_no;
```

InnoDB calls it from `innobase_xa_prepare` and from `innobase_commit`, when either of these functions is called for the statement inside the transaction (the parameter `commit_trx=false`).

[...]

PR #804, opened 2018-04-03
Issue #802: MyRocks: Statement rollback doesnt work correctly for nes…

spetrunia:
…ted statements

Variant #1: When the statement fails, we should roll back to the latest
savepoint taken at the top level.

### DIFF
--- a/storage/rocksdb/ha_rocksdb.cc
+++ b/storage/rocksdb/ha_rocksdb.cc
@@ -1854,6 +1854,17 @@ protected:
   bool m_is_delayed_snapshot = false;
   bool m_is_two_phase = false;
 
+private:
+  /* Number of RockDB savepoints taken */
+  int m_n_savepoints;
+  /*
+    Number of write operations this transaction had when we took the last
+    savepoint (the idea is not to take another savepoint if we haven't made
+    any changes)
+  */
+  ulonglong m_writes_at_last_savepoint;
+
+protected:
   THD *m_thd = nullptr;
 
   static std::multiset<Rdb_transaction *> s_tx_list;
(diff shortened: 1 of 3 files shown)


storage/rocksdb/ha_rocksdb.cc
```diff
@@ -2449,6 +2453,10 @@ class Rdb_transaction_impl : public Rdb_transaction {
   rocksdb::Transaction *m_rocksdb_tx = nullptr;
   rocksdb::Transaction *m_rocksdb_reuse_tx = nullptr;
 
+  // Number of savepoints to roll back to get to the start of the
+  // "top-level" statement.
+  int m_n_savepoints;
```
hermanlee: It would probably make more sense to have these variables and collapse_savepoint() call in Rdb_transaction instead?

hermanlee (changes requested):

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -1854,6 +1854,17 @@ class Rdb_transaction {
   bool m_is_delayed_snapshot = false;
   bool m_is_two_phase = false;
 
+private:
+  /* Number of RockDB savepoints taken */
+  int m_n_savepoints;
```
hermanlee, line 1859: m_n_savepoints and m_writes_at_last_savepoint need to be initialized at similar locations to m_write_count, no?

hermanlee (changes requested):

storage/rocksdb/ha_rocksdb.cc
spetrunia: Yes. Lost that initialization when re-working the patch. Will now add it back

hermanlee (approved):

PR #818, opened 2018-04-25
fix gcc-8 compilation warnings

inikep:
This PR is based on our patches for Percona Server 5.6. It fixes the following gcc-8 warnings:
    - type qualifiers ignored on cast result type [-Werror=ignored-qualifiers]
    - this statement may fall through [-Werror=implicit-fallthrough=]
    - clearing an object with no trivial copy-assignment [-Werror=class-memaccess]
    - this ‘if’ clause does not guard... [-Werror=misleading-indentation]
    - ‘%s’ directive writing up to 16384 bytes into a region of size 232 [-Werror=format-overflow=]
    - cast between incompatible function types [-Werror=cast-function-type]
    - directive output may be truncated [-Werror=format-truncation=]

The server will compile with `-DMYSQL_MAINTAINER_MODE=ON -DWITH_ZLIB=system` without any errors after the following changes: 
- Advance RocksDB submodule commit pointer to `dee95a1afc6c63515e7d94dec33acdb79638b6d7` or later
- Add `ADD_DEFINITIONS("-Wno-implicit-fallthrough -Wno-shift-negative-value")` to `storage/innobase/CMakeLists.txt` as files in `storage/innobase/zlib_embedded/` are not fixed


inikep:
Are you interested in the submitted changes? 
I can remove changing of the RocksDB submodule commit pointer (as it's no longer needed).

laurynas-biveinis (commented):
If I had to guess, the subset of fixes which applies to storage/rocksdb and to the top of any Facebook-added patches elsewhere is more likely to be accepted

hermanlee:
Laurynas is correct. Most of these changes affect upstream code and we are looking to move to MySQL 8.0, where this patch would likely cause significant conflict and probably be dropped. We also don't have a strong need for these fixes in our current 5.6 environment, so we're not as interested in the changes.

Changes that fix issues in storage/rocksdb are definitely of greater interest.

PR #821, opened 2018-04-30
Turn PIC on for libservices to enable LTO build

taewookoh:
Turning on link-time optimization (LTO) fails the build of .so's linked agains libmysqlservices.a with
"requires dynamic R_X86_64_PC32 reloc against 'my_snprintf_service' which may overflow at runtime; recompile with -fPIC". 

This patch addresses the issue and enables LTO build.


lth:
This has been addressed already in our build tools.

Issue #809, opened 2018-04-09
Wrong query result with bloom filters


spetrunia:
I am using debug build of the current tree (tip cset 6f8a6f19b46023c0f08d87a8abd5225b4c3ed154)

When I run the attached test file without the `-master.opt` file, I get a result with a correct number of rows in the table t2:
```
...
insert t1
select (seq+9) div 10, (seq+4) div 5, (seq+4) div 5, seq, seq, 1000, "aaabbbccc"
  from ten_k;
insert t2 select * from t1;
select count(*) from t1;
count(*)
10000
select sin(0);
sin(0)
0
select count(*) from t2;
count(*)
10000
```

When I put the `b-master.opt` in place, I get a wrong query result:
```
...
insert t1
select (seq+9) div 10, (seq+4) div 5, (seq+4) div 5, seq, seq, 1000, "aaabbbccc"
  from ten_k;
insert t2 select * from t1;
select count(*) from t1;
count(*)
10000
select sin(0);
sin(0)
0
select count(*) from t2;
count(*)
0
```

spetrunia:
It's not possible attach arbitrary files here... The files:
b.test: https://gist.github.com/spetrunia/9630b2aad51738eadff54e96f31d2312
b-master.opt :  https://gist.github.com/spetrunia/63b400f25b525a137d69c10144a338b3

They are excerpts from MariaDB's variant of `bloomfilter.test`  (it is different from FB/MySQL variant of bloomfilter.test, the difference is faster loading of data into the tables.  "select sin(0)" is just for debugging, easy place to put a breakpoint at)

hermanlee:
Yes, this looks very strange:

```
mysql> select * from t1;
Empty set (0.00 sec)

mysql> explain select * from t1;
+----+-------------+-------+------+---------------+------+---------+------+-------+-------+
| id | select_type | table | type | possible_keys | key  | key_len | ref  | rows  | Extra |
+----+-------------+-------+------+---------------+------+---------+------+-------+-------+
|  1 | SIMPLE      | t1    | ALL  | NULL          | NULL | NULL    | NULL | 10000 | NULL  |
+----+-------------+-------+------+---------------+------+---------+------+-------+-------+
1 row in set (0.00 sec)

mysql> explain select count(*) from t1;
+----+-------------+-------+-------+---------------+------+---------+------+-------+-------------+
| id | select_type | table | type  | possible_keys | key  | key_len | ref  | rows  | Extra       |
[...]

PR #825, opened 2018-05-07
Issue #809: Wrong query result with bloom filters

spetrunia:
In reverse-ordered column families, if one wants to start reading at the
logical end of the index, they should Seek() to a key value that is not
covered by the index. This may (and typically does) prevent use of a bloom
filter.

The calls to setup_scan_iterator() that are made for index and table scan
didn't take this into account and passed eq_cond_len=INDEX_NUMBER_SIZE.
Fixed them to compute and pass correct eq_cond_len.

Also, removed an incorrect assert in ha_rocksdb::setup_iterator_bounds.


hermanlee (approved):
The changes look good overall.

Issue #834, opened 2018-05-18
ALTER TABLE table_with_hidden_pk causes Can't write; duplicate key in table error and/or crash


spetrunia:
The issue is originally from  https://jira.mariadb.org/browse/MDEV-15304 , but is reproducible on the current ee761fc7a7d95af943a97aada5d453d441741b3c as well.

The test uses two client connections.

Perpare the dataset (I do it in client1):

```
CREATE TABLE t1 (a INT, KEY(a)) ENGINE=RocksDB;
INSERT INTO t1 VALUES (1),(1+1);
create table t2 (a int);
insert into t2 values (10),(20),(30);
```

Client1 starts a transaction and gets a snapshot:
```
MySQL [j55]> BEGIN;                                                                                                               
Query OK, 0 rows affected (0.00 sec)

MySQL [j55]> select * from t2;
+------+
| a    |
+------+
|   10 |
|   20 |
|   30 |
+------+
3 rows in set (0.00 sec)
```

Client2: perform an ALTER TABLE:
```
MariaDB [j25]> alter table t1 force;
Query OK, 2 rows affected (0.32 sec)
Records: 2  Duplicates: 0  Warnings: 0
```

Client1: read something from t1 (this causes incorrect hidden PK value to be loaded):

```
MySQL [j55]> select * from t1;
Empty set (0.00 sec)
```

Client2:  Try inserting a row into t1, and get a "duplicate key" error even if the table doesn't have a unique key as far as the user is concerned:

```
MariaDB [j25]> insert into t1 values (100);
ERROR 1022 (23000): Can't write; duplicate key in table 't1'
```

The last query may also cause the server to crash with a stack trace like this:

```
Program received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7ffff420a700 (LWP 18798)]
0x0000000000f473ab in key_unpack (to=0x7ffff42059a0, table=0x7fffd401ea80, key=0x7fffd401bd88) at /home/psergey/dev-git/mysql-5.6-rocksdb-look800/sql/key.cc:457
457         if (key_part->null_bit)
(gdb) wher
#0  0x0000000000f473ab in key_unpack (to=0x7ffff42059a0, table=0x7fffd401ea80, key=0x7fffd401bd88) at /home/psergey/dev-git/mysql-5.6-rocksdb-look800/sql/key.cc:457
[...]

PR #835, opened 2018-05-18
Issue #834: Server crash in print_keydup_error / key_unpack or unexpe…

spetrunia:
…cted ER_DUP_KEY

Fix two issues:
1. Rdb_ddl_manager::rename() loses the value of m_hidden_pk_val. new
object used to get 0, which means "not loaded from the db yet".

2. ha_rocksdb::load_hidden_pk_value() uses current transaction (and its
snapshot) when loading hidden PK value from disk. This may cause it to
load an out-of-date value.

### DIFF
--- a/mysql-test/suite/rocksdb/r/allow_no_primary_key.result
+++ b/mysql-test/suite/rocksdb/r/allow_no_primary_key.result
@@ -262,3 +262,28 @@ SELECT * FROM t1;
 a	b
 36	foo
 DROP TABLE t1;
+#
+# Issue #834/MDEV-15304 ALTER TABLE table_with_hidden_pk causes Can't
+#   write; duplicate key in table error and/or crash
+#
+CREATE TABLE t1 (a INT, KEY(a)) ENGINE=RocksDB;
+INSERT INTO t1 VALUES (1),(1+1);
+create table t2 (a int) ENGINE=rocksdb;
+insert into t2 values (10),(20),(30);
+BEGIN;
+select * from t2;
+a
+10
+20
+30
+alter table t1 force;
+select * from t1;
+a
+insert into t1 values (100);
+select * from t1;
+a
+1
+2
+100
+rollback;
+drop table t1,t2;
--- a/mysql-test/suite/rocksdb/t/allow_no_primary_key.test
+++ b/mysql-test/suite/rocksdb/t/allow_no_primary_key.test
@@ -96,3 +96,31 @@ DELETE FROM t1 WHERE a = 35 AND b = 'foo';
 --sorted_result
 SELECT * FROM t1;
 DROP TABLE t1;
+
+--echo #
+--echo # Issue #834/MDEV-15304 ALTER TABLE table_with_hidden_pk causes Can't
+--echo #   write; duplicate key in table error and/or crash
+--echo #
+CREATE TABLE t1 (a INT, KEY(a)) ENGINE=RocksDB;
+INSERT INTO t1 VALUES (1),(1+1);
+create table t2 (a int) ENGINE=rocksdb;
+insert into t2 values (10),(20),(30);
+
+BEGIN;
+select * from t2;
+
+connect (con1,localhost,root,,);
+connection con1;
+alter table t1 force;
+
+connection default;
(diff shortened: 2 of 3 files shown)


mysql-test/suite/rocksdb/t/allow_no_primary_key.test
```diff
@@ -96,3 +96,31 @@ DELETE FROM t1 WHERE a = 35 AND b = 'foo';
...
 SELECT * FROM t1;
 DROP TABLE t1;
+
+--echo #
+--echo # Issue #834/MDEV-15304 ALTER TABLE table_with_hidden_pk causes Can't
+--echo #   write; duplicate key in table error and/or crash
+--echo #
+CREATE TABLE t1 (a INT, KEY(a)) ENGINE=RocksDB;
+INSERT INTO t1 VALUES (1),(1+1);
+create table t2 (a int);
```
laurynas-biveinis: Do I understand correctly that t2 engine is not important as long as it's transactional?

mysql-test/suite/rocksdb/r/allow_no_primary_key.result
```diff
@@ -262,3 +262,34 @@ SELECT * FROM t1;
...
+INSERT INTO t1 VALUES (1),(1+1);
+create table t2 (a int);
+insert into t2 values (10),(20),(30);
+BEGIN;
+select * from t2;
+a
+10
+20
+30
+connect  con1,localhost,root,,;
```
hermanlee: Could you re-record the test? I do not see these lines in the test output and the test itself fails.

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -5216,8 +5216,17 @@ int ha_rocksdb::load_hidden_pk_value() {
   active_index = m_tbl_def->m_key_count - 1;
   const uint8 save_table_status = table->status;
 
-  Rdb_transaction *const tx = get_or_create_tx(table->in_use);
-  const bool is_new_snapshot = !tx->has_snapshot();
+  /*
+    We should read the latest committed value in the database.
```
hermanlee: Is this needed, or is this being extra defensive, but not strictly required? It seems the fix needed is transferring m_hidden_pk_val from the old table_def to the new table_def during alter, and that is already in this diff. This code only populates the m_hidden_pk_val of the table if it hasn't been loaded yet. If there is an open transaction with the table, then we shouldn't be able to rename the table. If the current snapshot does not read the latest data, then another transaction must have written the new m_hidden_pk_val after this snapshot was created. In that case, the table definition should have the latest data, and whether we read with the latest snapshot or the existing snapshot shouldn't matter, no?

hermanlee (changes requested):

mysql-test/suite/rocksdb/t/allow_no_primary_key.test
spetrunia: No, I think it needs to be MyRocks.
(t2 is present in the test to cause MyRocks to start a transaction. It is a MyRocks table in my setup. I didn't try other engines for it but I don't think using another engine will cause MyRocks transaction to be created)

I'll add ENGINE=RocksDB.

mysql-test/suite/rocksdb/r/allow_no_primary_key.result
spetrunia: Yes, will do.

storage/rocksdb/ha_rocksdb.cc
spetrunia: Also discovered this with valgrind: if one calls `tx->start_tx`, they must call set_params beforehand, like it's done at other places:
```
    tx->set_params(THDVAR(thd, lock_wait_timeout), THDVAR(thd, max_row_locks));
    tx->start_tx();
```

spetrunia: ... as discussed on the call - the patch is extra defensive, and I will remove this part.  Will leave only the one in `Rdb_ddl_manager::rename`.

hermanlee (approved):

PR #842, opened 2018-05-25
Add cahce_high_pri_pool_ratio option and cache_index_and_filter_blocks_with_high_priority

yiwu-arbug:
Summary:
Adding rocksdb options cahce_high_pri_pool_ratio and cache_index_and_filter_blocks_with_high_priority. I'm adding LRU cache midpoint insertion with rocksdb PR https://github.com/facebook/rocksdb/pull/3877. My preliminary db_bench shows some latency and throughput drop for a read only workload when background full table scan presents (See the PR for more details). To enable the midpoint insertion strategy, high_pri_pool_ratio needs to be set to set the location for the midpoint. cache_index_and_filter_blocks_with_high_priority can be enabled so that index and filter blocks don't insert to the midpoint, but the tail of the full list.

Test Plan:
What are the test I need to run with myrocks?

### DIFF
--- a/storage/rocksdb/ha_rocksdb.cc
+++ b/storage/rocksdb/ha_rocksdb.cc
@@ -1227,7 +1227,7 @@ static MYSQL_SYSVAR_BOOL(
     *reinterpret_cast<my_bool *>(
         &rocksdb_tbl_options->cache_index_and_filter_blocks_with_high_priority),
     PLUGIN_VAR_RQCMDARG | PLUGIN_VAR_READONLY,
-    "BlockBasedTableOptions::cache_index_and_filter_blocks_with_high_priority for RocksDB",
+    "Cache index and filter blocks as high-pri cache entry",
     nullptr, nullptr, true);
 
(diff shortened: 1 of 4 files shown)


storage/rocksdb/ha_rocksdb.cc
```diff
@@ -1206,6 +1207,13 @@ static MYSQL_SYSVAR_BOOL(
     "Use ClockCache instead of default LRUCache for RocksDB",
     nullptr, nullptr, false);
 
+static MYSQL_SYSVAR_DOUBLE(
+    cache_high_pri_pool_ratio,
+    rocksdb_cache_high_pri_pool_ratio,
+    PLUGIN_VAR_RQCMDARG | PLUGIN_VAR_READONLY,
+    "Use ClockCache instead of default LRUCache for RocksDB",
```
laurynas-biveinis: Description string of a different variable?

lth (commented):
You need to add a test under the rocksdb_sys_vars suite. This is why the rocksdb_sys_vars.all_vars test is failing.

yiwu-arbug:
Closing. Will send an internal diff instead.

PR #843, opened 2018-05-25
Fixing slow_log_extra_big.test failure

yoshinorim:
Summary: slow_log_extra_big.test was incorrect
and caused test failures. This diff fixes it.

Test Plan: slow_log_extra_big.test


hermanlee:
Volodymyr should have a diff to fix this. The test requires a tag to be added to enable tracing for it, since https://github.com/facebook/mysql-5.6/commit/5832232 caused the issue.

PR #850, opened 2018-07-03
Integrate deadlock timestamp to MySQL

chouxi:
- Summary
Integrate deadlock timestamp info to MySQL based on RocksDB changes.

- Test Plan

`~/mysql/tools/mysqltest.sh rocksdb.i_s_deadlock`


george-lorch:
Why does PR have multiple commits and code changes completely unrelated to the summary? In general, pull requests should consist of a single commit based on the HEAD of the fb-mysql-5.6.35 branch that contain only changes directly related to the task/summary.

PR #851, opened 2018-07-10
[Feature] cluster key

justfortaste:
Support secondary clustering key in MyRocks

```   
 Support new syntax
      create table t1(c1 int priamry key, c2 int clustering, c3 int) engine=rocksdb;
      alter table t1 add clustering key idx1(c3);

 New clustering key row layout as follows, PK and SK has not changed
    1) CK (secondary cluster key)
      key: index_id + ck_cols + pk_cols
      value: [TTL] +  NULL-bitmap + unpack_info + other_cols

    2) PK (primary key)
      key: index_id + pk_cols
      value: [TTL] +  NULL-bitmap + unpack_info + other_cols

    3) SK (ordinary sencordary key)
      key: index_id + sk_cols + pk_cols
      value: [unpack_info (TTL is in unpack_info, TTL_FLAG)
```


laurynas-biveinis (commented):
The syntax and server parts of the patch appear to be based on the Percona Server TokuDB clustering key support, which appears to be a good idea from future compatibility point of view

PR #852, opened 2018-07-10
Adding resizable block cache support in MyRocks

yoshinorim:
Summary: This diff makes rocksdb_block_cache_size dynamically
configurable -- both increasing and decreasing. This relies on
RocksDB's Cache::SetCapacity API. Calling Cache::SetCapacity()
is done outside of rdb_sysvars_mutex lock, since reducing
large block cache may take a long time. Instead, this diff
introduces a dedicated mutex to protect block cache resize.

Test Plan: mtr

Tasks:

Tags:


yoshinorim:
https://github.com/facebook/mysql-5.6/commit/8a1c5c11ae874c4253c5ee2596bed24951e58a52

PR #855, opened 2018-07-10
Integrate deadlock timestamp to MySQL

chouxi:
Integrate deadlock timestamp info to MySQL based on RocksDB changes.


storage/rocksdb/rdb_i_s.cc
```diff
@@ -1504,6 +1505,7 @@ static ST_FIELD_INFO rdb_i_s_deadlock_info_fields_info[] = {
     ROCKSDB_FIELD_INFO("TABLE_NAME", NAME_LEN + 1, MYSQL_TYPE_STRING, 0),
     ROCKSDB_FIELD_INFO("ROLLED_BACK", sizeof(ulonglong), MYSQL_TYPE_LONGLONG,
                        0),
+    ROCKSDB_FIELD_INFO("TIMESTAMP", sizeof(ulonglong), MYSQL_TYPE_LONGLONG, 0),
```
lth: I don't think we have anything relying on the order of columns here.

I think I'd slightly prefer having the timestamp begin the first column.

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -3678,10 +3681,12 @@ class Rdb_snapshot_status : public Rdb_tx_list_walker {
               "LOCK TYPE: %s\n"
               "INDEX NAME: %s\n"
               "TABLE NAME: %s\n",
-              trx_info.trx_id, trx_info.cf_name.c_str(),
+              "Timestamp: %" PRId64 "\n", trx_info.trx_id,
```
lth: Same thing here with timestamp being first. Also for consistency, it looks like we should capitalize TIMESTAMP.

lth (approved):
Looks good to me.

PR #858, opened 2018-07-13
Issue#857: MyRocks: Incorrect behavior when multiple statements fail …

spetrunia:
…inside a transaction

rollback_to_stmt_savepoint() calls do_rollback_to_savepoint(). This
removes the changes made by this statement, and also removes the last
set savepoint.

Before we start processing the next statement, we need to set the new
savepoint, so we will have something to rollback to if the next statement
fails.

Since rollback_to_stmt_savepoint always sets a new savepoint now,
m_n_savepoints is now redundant and is removed.

### DIFF
--- a/mysql-test/suite/rocksdb/r/transaction.result
+++ b/mysql-test/suite/rocksdb/r/transaction.result
@@ -958,3 +958,20 @@ a
 rollback;
 drop function func;
 drop table t1,t2,t3;
+#
+# MDEV-16710: Slave SQL: Could not execute Update_rows_v1 event with RocksDB and triggers
+# Issue#857: MyRocks: Incorrect behavior when multiple statements fail inside a transaction
+#
+CREATE TABLE t1 (a INT PRIMARY KEY) ENGINE=RocksDB;
+INSERT INTO t1 VALUES (1);
+CREATE TABLE t2 (b INT PRIMARY KEY) ENGINE=RocksDB;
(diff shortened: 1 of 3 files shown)


mysql-test/suite/rocksdb/r/transaction.result
```diff
@@ -958,3 +958,20 @@ a
 rollback;
 drop function func;
 drop table t1,t2,t3;
+#
+# MDEV-16710: Slave SQL: Could not execute Update_rows_v1 event with RocksDB and triggers
+# Issue#857: MyRocks: Incorrect behavior when miltiple statements fail inside a transaction
```
hermanlee: nit: typo: multiple

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -2436,7 +2436,6 @@ class Rdb_transaction {
       entire transaction.
     */
     do_set_savepoint();
-    m_n_savepoints= 1;
```
hermanlee, line 2439: Since this is not longer used, the definition in Rdb_transaction should also be removed.

hermanlee (changes requested):
Can you also update the description with:

Squash with D7509380

On a new rebase, we'll merge this fix with the original change.

storage/rocksdb/ha_rocksdb.cc
spetrunia: Done

spetrunia:
Addressed all review input.

hermanlee (approved):

PR #868, opened 2018-08-09
Update AUTO_INCREMENT in the UPDATE statement

minggr:
Currently RocksDB engine doesn't update AUTO_INCREMENT in the UPDATE statement.
For example,

CREATE TABLE t1 (pk INT AUTO_INCREMENT, a INT, PRIMARY KEY(pk)) ENGINE=RocksDB;
INSERT INTO t1 (a) VALUES (1);
UPDATE t1 SET pk = 3; ==> AUTO_INCREMENT should be updated to 4.

Without this fix, it hits the Assertion `dd_val >= last_val' failed in
myrocks::ha_rocksdb::load_auto_incr_value_from_index.

Please see detail at:
https://jira.mariadb.org/browse/MDEV-16703

### DIFF
--- a/storage/rocksdb/ha_rocksdb.cc
+++ b/storage/rocksdb/ha_rocksdb.cc
@@ -9410,7 +9410,7 @@ int ha_rocksdb::update_pk(const Rdb_key_def &kd,
     }
   }
 
-  if (table->next_number_field) {
+  if (table->found_next_number_field) {
     update_auto_incr_val_from_field();
   }
 
--- a/mysql-test/suite/rocksdb/r/autoinc_vars.result
+++ b/mysql-test/suite/rocksdb/r/autoinc_vars.result
@@ -151,3 +151,11 @@ CREATE TABLE t0(c0 BLOB) ENGINE=ROCKSDB;
 INSERT INTO t0 VALUES(0);
 ALTER TABLE t0 AUTO_INCREMENT=0;
 DROP TABLE t0;
+#----------------------------------
+# Issue #869 Crash in autoincrement
+#----------------------------------
+CREATE TABLE t1 (pk INT AUTO_INCREMENT, a INT, PRIMARY KEY(pk)) ENGINE=RocksDB;
+INSERT INTO t1 (a) VALUES (1);
+UPDATE t1 SET pk = 3;
+ALTER TABLE t1 AUTO_INCREMENT 2;
+DROP TABLE t1;
--- a/mysql-test/suite/rocksdb/t/autoinc_vars.test
+++ b/mysql-test/suite/rocksdb/t/autoinc_vars.test
@@ -114,3 +114,13 @@ CREATE TABLE t0(c0 BLOB) ENGINE=ROCKSDB;
 INSERT INTO t0 VALUES(0);
 ALTER TABLE t0 AUTO_INCREMENT=0;
 DROP TABLE t0;
+
+--echo #----------------------------------
+--echo # Issue #869 Crash in autoincrement
+--echo #----------------------------------
+
+CREATE TABLE t1 (pk INT AUTO_INCREMENT, a INT, PRIMARY KEY(pk)) ENGINE=RocksDB;
+INSERT INTO t1 (a) VALUES (1);
+UPDATE t1 SET pk = 3;
+ALTER TABLE t1 AUTO_INCREMENT 2;
+DROP TABLE t1;


storage/rocksdb/ha_rocksdb.cc
```diff
@@ -9583,6 +9583,36 @@ int ha_rocksdb::update_indexes(const struct update_row_info &row_info,
...
+  const char* n_ptr;
+  uint i;
+
+  auto_inc = 0;
+
+  if (!old_row) {
+    return;
+  }
+
+  for (i = 0; i < table->s->fields; i++) {
```
lth: Do we have to do the for loop? Doesn't `table->s->next_number_index` contain the field we want?

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -9646,6 +9676,16 @@ int ha_rocksdb::update_write_row(const uchar *const old_data,
...
   }
 
+  ulonglong autoinc;
+  calc_autoinc_difference(old_data, new_data, table, autoinc);
+  if (autoinc) {
+    /*
+     A value for an AUTO_INCREMENT column
+     was specified in the UPDATE statement.
+    */
+    update_auto_incr_val_from_field();
```
lth: We already call `update_auto_incr_val_from_field` from `update_pk`, but it's not working as expected because it's using the 'old' row instead of the new one. I think a better fix is to unify these paths.

lth (changes requested):
Can we add an mtr test as well?

storage/rocksdb/ha_rocksdb.cc
minggr: Not always, for example,

CREATE TABLE t1 (a INT, pk INT AUTO_INCREMENT, PRIMARY KEY(pk)) ENGINE=RocksDB;
INSERT INTO t1 (a) VALUES (20);

(gdb) p table->s->next_number_index
$18 = 0
(gdb) p table->field[0]->val_uint()
$19 = 20

lth: Actually, I got confused on what `table->s->next_number_index` meant. It is actually telling us which index the autoincrement is one (not the "index" of the field array). In this case, it's 0 because it is on primary key.

Looking at `ha_rocksdb::load_auto_incr_value_from_index`, we just assume that the first column of the index is the autoincrement field. Looks like something similar is assumed in InnoDB (row_search_autoinc_read_column is always called with col_no == 0).

minggr: How about the updated patch?

storage/rocksdb/ha_rocksdb.cc
minggr: CREATE TABLE t1 (pk INT AUTO_INCREMENT, a INT, PRIMARY KEY(pk)) ENGINE=RocksDB;
INSERT INTO t1 (a) VALUES (1);
UPDATE t1 SET pk = 3;

table->next_number_field is null so update_auto_incr_val_from_field() is not called in update_pk()
<pre>
(gdb) list
9408	    } else {
9409	      bytes_written = row_info.old_pk_slice.size();
9410	    }
9411	  }
9412	
9413	  if (table->next_number_field) {
9414	    update_auto_incr_val_from_field();
9415	  }
9416	
9417	  int rc = HA_EXIT_SUCCESS;
(gdb) p table->next_number_field
$2 = (Field *) 0x0
</pre>

lth: Can we merge these two paths, since I think they're trying to do the same thing?

lth:
Can we add an mtr test as well?

storage/rocksdb/ha_rocksdb.cc
lth: Also, `table->next_number_field` isn't nullptr for me, but it is using the old row, which is why this bug reproduces for me.

```
(gdb) list
9395          bytes_written = row_info.old_pk_slice.size();
9396        }
9397      }
9398
9399      if (table->next_number_field) {
9400        update_auto_incr_val_from_field();
9401      }
9402
9403      int rc = HA_EXIT_SUCCESS;
9404      rocksdb::Slice value_slice;
(gdb) p table->next_number_field
$2 = (Field *) 0x7f2a9c825150
```

minggr:
Yes, I'll add a mtr test.

storage/rocksdb/ha_rocksdb.cc
minggr: That's weird. Can you set a memory watchpoint to show the call stack that sets table->next_number_field?
<pre>
Breakpoint 13, mysql_update (thd=0x2c8f850, table_list=0x7fffdc005710, fields=..., values=..., conds=0x0, order_num=0, order=0x0, limit=18446744073709551615, handle_duplicates=DUP_ERROR, 
    ignore=false, found_return=0x7ffff7e8d4a0, updated_return=0x7ffff7e8d498) at /workspace/mysql/sql/sql_update.cc:266
266	  if (!table_list->updatable)
(gdb) list
261	  }
262	
263	  THD_STAGE_INFO(thd, stage_init);
264	  table= table_list->table;
265	
266	  if (!table_list->updatable)
267	  {
268	    my_error(ER_NON_UPDATABLE_TABLE, MYF(0), table_list->alias, "UPDATE");
269	    DBUG_RETURN(1);
270	  }
(gdb) p table
$53 = (TABLE *) 0x7fffdc021180
(gdb) watch ((TABLE *) 0x7fffdc021180)->next_number_field
Hardware watchpoint 14: ((TABLE *) 0x7fffdc021180)->next_number_field
</pre>

It's not clear to me how to merge these two paths. Can we simply do below?
<pre>
@@ -9410,7 +9410,7 @@ int ha_rocksdb::update_pk(const Rdb_key_def &kd,
     }
   }
 
-  if (table->next_number_field) {
+  if (table->found_next_number_field) {
     update_auto_incr_val_from_field();
   }
</pre>

minggr:
What are the correct parameters to run mtr test?
I tried as below,
<pre>
./mtr --mysqld="--default-tmp-storage-engine=MyISAM --rocksdb --skip-innodb --default-storage-engine=RocksDB"  --suite=rocksdb 
</pre>

But error log shows:
<pre>
2018-08-15 15:02:32 76594 [Note] Plugin 'ROCKSDB' is disabled.
2018-08-15 15:02:32 76594 [Note] Plugin 'FEDERATED' is disabled.
2018-08-15 15:02:32 76594 [Note] InnoDB: Using atomics to ref count buffer pool pages
2018-08-15 15:02:32 7f34b2cd2880 InnoDB: !!!!!!!! UNIV_DEBUG switched on !!!!!!!!!
2018-08-15 15:02:32 7f34b2cd2880 InnoDB: !!!!!!!! UNIV_SYNC_DEBUG switched on !!!!!!!!!
2018-08-15 15:02:32 76594 [Note] InnoDB: The InnoDB memory heap is disabled
2018-08-15 15:02:32 76594 [Note] InnoDB: Mutexes and rw_locks use GCC atomic builtins
2018-08-15 15:02:32 76594 [Note] InnoDB: Memory barrier is not used
2018-08-15 15:02:32 76594 [Note] InnoDB: Compressed tables use zlib 1.2.3
2018-08-15 15:02:32 76594 [Note] InnoDB: Using Linux native AIO
</pre>

spetrunia:
For me running tests like this works:
 ```
./mysql-test-run --mysqld=--default-storage-engine=rocksdb \
  --mysqld=--skip-innodb --mysqld=--default-tmp-storage-engine=MyISAM \
  --mysqld=--rocksdb  --suite=rocksdb
```
that is, mysqld arguments go one by one

minggr:
Yes, that works. Thanks.

storage/rocksdb/ha_rocksdb.cc
lth: Checking `found_next_number_field` will probably be good enough for the fix. You might need to change `update_auto_incr_val_from_field` to look at new row.

For some reason, I'm seeing nullptr now for `next_number_field`, so I'm not sure how I got the previous results in the first place...

minggr: Nice. I'll update the patch and also add a mtr test.

minggr: Patch updated.
In update_auto_incr_val_from_field(),
field = table->key_info[table->s->next_number_index].key_part[0].field;
field->val_int() is already new AUTO_INCREMENT value.

lth (approved):
Thanks for fixing this!

PR #873, opened 2018-08-30
Don't require readline + zstd if not building client+server

fredemmott:
e.g. HHVM is just building the client library, so:
- doesn't need either
- **must not** link against readline (GPL vs the PHP licenses)

HHVM has previously been patching the CMake file to remove the readline
dependency, but lets' end that.


fredemmott:
FYI, this is still an issue - T23034557

HHVM's applying this patch in our OSS builds.

PR #879, opened 2018-09-27
Fix issue #878: Descending scans from reverse cf return no results

### LINKED ISSUE #878: Descending scans from reverse column families return no results due to iterator bounds
mtr test case:
```
create table t (i int primary key, j int, key(j) comment 'rev:bf5_2');

let $i=253;
while ($i)
{
  --disable_query_log
  truncate table t;
  --enable_query_log
  dec $i;
}

select *, hex(index_number) from information_schema.rocksdb_ddl where table_name = 't';

create table u (i int primary key, j int, key(j) comment 'rev:bf5_2');

insert into t values (1, 1);
insert into u values (1, 1);

select j from t order by j desc;

drop table t;
drop table u;
```

Using these options:
```
--rocksdb_default_cf_options=write_buffer_size=256k;block_based_table_factory={filter_policy=bloomfilter:10:false;whole_key_filtering=0;}
--rocksdb_override_cf_options=rev:bf5_1={prefix_extractor=capped:12};
```

The issue here is that in `ha_rocksdb::setup_iterator_bounds`, the predecessor function is underflowing because the eq_cond buffer is just 0x00 0x00:
```
(gdb) x/2x upper_bound
0x7fffe8e15100: 0x00    0x01
(gdb) x/2x lower_bound
0x7fffe8e150f0: 0x00    0xff
```

This causes the bounds to be non-overlapping leading to empty result sets.

I think the confusion that caused this bug is the fact that `check_bloom_and_set_bounds` is doing both checking bloom filters and setting bounds based on eq_cond, but we can actually set tighter bounds if we knew more than just eq_cond (eg. if we knew index number)

I think there are a few solutions:
1. Change the predecessor/successor functions to detect underflow/overflow, and do not set ReadOptions.iterate_lower_bound (and upper), if underflow/overflow is detected. The drawback here is that we greatly increase the coarseness of the bounds.
2. Pass index id down to `check_bloom_and_set_bounds` so that if eq_cond has length less than 4, it can be smarter about setting bounds. The calculate will call predecessor on index id rather than on eq_cond.

I slightly prefer the 2nd solution.

lth:
Currently, `Rdb_key_def::predecessor` can underflow during `ha_rocksdb::setup_iterator_bounds` if eq_cond being passed in is too short. The fix is to use index id directly in that case, instead of relying on eq_cond.

Fixes https://github.com/facebook/mysql-5.6/issues/878

### DIFF
--- a/mysql-test/suite/rocksdb/r/iterator_bounds.result
+++ b/mysql-test/suite/rocksdb/r/iterator_bounds.result
@@ -1,13 +1,15 @@
+create table t (i int primary key) engine=rocksdb;
+drop table t;
 create table t (i int primary key, j int, key(j) comment 'rev:bf5_2') engine=rocksdb;
-select *, hex(index_number) from information_schema.rocksdb_ddl where table_name = 't';
-TABLE_SCHEMA	TABLE_NAME	PARTITION_NAME	INDEX_NAME	COLUMN_FAMILY	INDEX_NUMBER	INDEX_TYPE	KV_FORMAT_VERSION	TTL_DURATION	INDEX_FLAGS	CF	AUTO_INCREMENT	hex(index_number)
-test	t	NULL	PRIMARY	0	766	1	13	0	0	default	NULL	2FE
-test	t	NULL	j	2	767	2	13	0	0	rev:bf5_2	NULL	2FF
-create table u (i int primary key, j int, key(j) comment 'rev:bf5_2') engine=rocksdb;
+select RIGHT(HEX(index_number), 2) from information_schema.rocksdb_ddl where table_name = 't';
+RIGHT(HEX(index_number), 2)
+FE
+FF
 insert into t values (1, 1);
-insert into u values (1, 1);
+select j from t order by j asc;
+j
+1
 select j from t order by j desc;
 j
 1
 drop table t;
-drop table u;
--- a/mysql-test/suite/rocksdb/r/iterator_bounds.result
+++ b/mysql-test/suite/rocksdb/r/iterator_bounds.result
@@ -1,9 +1,9 @@
-create table t (i int primary key, j int, key(j) comment 'rev:bf5_2');
+create table t (i int primary key, j int, key(j) comment 'rev:bf5_2') engine=rocksdb;
 select *, hex(index_number) from information_schema.rocksdb_ddl where table_name = 't';
 TABLE_SCHEMA	TABLE_NAME	PARTITION_NAME	INDEX_NAME	COLUMN_FAMILY	INDEX_NUMBER	INDEX_TYPE	KV_FORMAT_VERSION	TTL_DURATION	INDEX_FLAGS	CF	AUTO_INCREMENT	hex(index_number)
 test	t	NULL	PRIMARY	0	766	1	13	0	0	default	NULL	2FE
 test	t	NULL	j	2	767	2	13	0	0	rev:bf5_2	NULL	2FF
-create table u (i int primary key, j int, key(j) comment 'rev:bf5_2');
+create table u (i int primary key, j int, key(j) comment 'rev:bf5_2') engine=rocksdb;
 insert into t values (1, 1);
 insert into u values (1, 1);
 select j from t order by j desc;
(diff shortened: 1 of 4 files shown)


mysql-test/suite/rocksdb/t/iterator_bounds.test
```diff
@@ -0,0 +1,27 @@
+#
+# Issue #878: Descending scans from reverse column families return no results
+# due to iterator bounds
+#
+
+create table t (i int primary key, j int, key(j) comment 'rev:bf5_2');
```
laurynas-biveinis: I assume your tree is correct in this regard (although I was not able to find --default-storage-engine=rocksdb for this MTR suite in all of 10 seconds of searching), but at least our downstream would appreciate explicit engine=rocksdb clauses.

lth: We add the option when we run mtr. See Sergey's comment:
https://github.com/facebook/mysql-5.6/pull/868#issuecomment-413220539

mysql-test/suite/rocksdb/r/iterator_bounds.result
```diff
@@ -0,0 +1,13 @@
+create table t (i int primary key, j int, key(j) comment 'rev:bf5_2') engine=rocksdb;
+select *, hex(index_number) from information_schema.rocksdb_ddl where table_name = 't';
+TABLE_SCHEMA	TABLE_NAME	PARTITION_NAME	INDEX_NAME	COLUMN_FAMILY	INDEX_NUMBER	INDEX_TYPE	KV_FORMAT_VERSION	TTL_DURATION	INDEX_FLAGS	CF	AUTO_INCREMENT	hex(index_number)
```
spetrunia: I'm not sure if it is safe to have INDEX_NUMBER in the .result file?  (I would expect the numbers to be different on every test re-run.  An experiment shows this is not happening, but is this something we can depend on, or just luck?)

I tried running `rocksdb-mtr --repeat=2 rocksdb.iterator_bounds` and the second run failed:
https://gist.github.com/spetrunia/54fcab13b5de5a8bf5e2218c0da808a7

The problem is that if this is not safe, I'm not aware of any solution - there's no way to cause MTR to run a test on a clean data dir (we have discussed this in MariaDB on some other occasion)

george-lorch: It is possible, but hack-ish, to have a server instance start on forced clean data-dir by shutting down the current server instance, making the mtr test perform a manual bootstrap on a new temporary data-dir, then launch/restart the server instance against that temporary data-dir, perform whatever test is necessary, shutting down the server again and restarting it with defaults, then removing the temporary data-dir.  Otherwise, if it is not worth it here for the purpose of this test, probably just masking the one column out of the result file is the next best thing.

lth: The reason this "works" is because I have an .opt file, and mtr always restarts the server with the provided options. You're right that repeating this would fail though.

The purpose of the test is actually to confirm that the index id ends in 0xff, and I think it actually fails (due to other index ids being allocated), if we do repeat like that.

Let me look into the hack-ish solution @georgelorchpercona suggested...

spetrunia: ... Maybe, just make a script that does `TRUNCATE TABLE t1` until `SELECT RIGHT(HEX(index_number), 2) FROM I_S...` returns 0xFF  ? 

This will be simpler, I think

spetrunia:
Other than the above observation about the testcase, the patch looks good to me.

(I side with an observation that in the longer term, it would be better to pass upper/lower scan boundaries into setup_iterator_bounds, not eq_cond.)

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -9705,18 +9705,36 @@ void ha_rocksdb::setup_iterator_bounds(
...
-  uint min_len = std::min(eq_cond.size(), bound_len);
-  memcpy(upper_bound, eq_cond.data(), min_len);
-  kd.successor(upper_bound, min_len);
-  memcpy(lower_bound, eq_cond.data(), min_len);
-  kd.predecessor(lower_bound, min_len);
+  // bound_len represents the lower_bound and upper_bound buffer size. It
+  // should always be greater than Rdb_key_def::INDEX_NUMBER_SIZE because
+  // callers either with a local buffer of size Rdb_key_def::INDEX_NUMBER_SIZE
+  // or it is called with m_scan_it_lower_bound.
+  DBUG_ASSERT(bound_len >= Rdb_key_def::INDEX_NUMBER_SIZE);
```
hermanlee: We can probably move this DBUG_ASSERT into the first conditional and assert it's Rdb_key_def::INDEX_NUMBER_SIZE.

hermanlee (approved):

PR #893, opened 2018-10-19
issue-884 : Prefixed PK and index_merge causes "ERROR 1032 (HY000): C…

george-lorch:
…an't find record in ..."

PS-4593 : MyRocks 8.0 inserts/returns incorrect data for tables with no explicit primary key

- Fix for PS-4593 backported from Percona Server 5.7 to Facebook-MySQL 5.6.35
  and issue-884 mtr test integrated into patch.

- This appears as a real issue in 8.0 as it is passing HA_EXTRA_KEYREAD more
  aggressively. The core issue is that the logic in MyRocks to perform a
  secondary key read directly does not check to see if the packed key can be
  properly unpacked to satisfy the read. When this happens, the previous value
  for Field is left and no error returned, leading to incorrect query results.
  This is difficult to reproduce in 5.7 to for a test case, some DBUG trickery
  is needed to force the same situation that happens naturally in 8.0.

- The fix is quite simple and that is to validate that all key parts are marked
  as 'covering', meaning that they are known to be able to be unpacked, with the
  predicate for HA_EXTRA_KEYREAD.

### DIFF
--- a/mysql-test/suite/rocksdb/r/issue884.result
+++ b/mysql-test/suite/rocksdb/r/issue884.result
@@ -1,0 +1,79 @@
+create table test (
+a bigint(20) not null,
+b bigint(20) not null,
+c varchar(500) not null,
+d bigint(20) not null,
+e bigint(20) not null,
+f varchar(500) not null,
+g varchar(500) not null,
+h varchar(500) not null,
+i varchar(1000) not null,
+j varchar(16384) not null,
+k varchar(200) not null,
+l varchar(500) not null,
+m varchar(100) not null,
+n bigint(20) not null,
+primary key (a, b, m, c(100), l(100), d, e, f(100), g(100), h(100), n),
+key n (n),
+key d (d, a)
+) engine = rocksdb default charset = latin1;
+Table	Op	Msg_type	Msg_text
+test.test	analyze	status	OK
+explain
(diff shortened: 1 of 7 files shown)


mysql-test/suite/rocksdb/r/issue884.result
```diff
@@ -0,0 +1,79 @@
...
+primary key (a, b, m, c(100), l(100), d, e, f(100), g(100), h(100), n),
+key n (n),
+key d (d, a)
+) engine = rocksdb default charset = latin1;
+Table	Op	Msg_type	Msg_text
+test.test	analyze	status	OK
+explain
+select * from test where d = 10 and a = 10 and b = 2;
+id	select_type	table	type	possible_keys	key	key_len	ref	rows	Extra
+1	SIMPLE	test	index_merge	PRIMARY,d	d,PRIMARY	24,16	NULL	14	Using intersect(d,PRIMARY); Using where
```
lth: Can you replace the "rows" column with a placeholder? This is causing test flakiness.

lth (commented):
Code looks good to me.

mysql-test/suite/rocksdb/r/issue884.result
george-lorch: Done and pushed

PR #897, opened 2018-11-01
Issue-895 : rocksdb/env/io_posix.cc:449: rocksdb::PosixMmapReadableFi…

george-lorch:
…le::PosixMmapReadableFile(int, const string&, void*, size_t, const rocksdb::EnvOptions&): Assertion `!options.use_direct_reads' failed.

- Extended previous fix fo Issue 778 to also check if
  rocksdb_use_direct_io_for_flush_and_compaction=1 is being used on tmpfs.
  Duplicated simple test that skips when on tmpfs.

### DIFF
--- a/mysql-test/suite/rocksdb/t/use_direct_io_for_flush_and_compaction.test
+++ b/mysql-test/suite/rocksdb/t/use_direct_io_for_flush_and_compaction.test
@@ -1,5 +1,5 @@
 --source include/have_rocksdb.inc
 
---let $io_option=--rocksdb_use_direct_reads=1
+--let $io_option=--rocksdb_use_direct_io_for_flush_and_compaction=1
 
 --source ../include/use_direct_io_option.inc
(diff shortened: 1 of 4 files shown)


mysql-test/suite/rocksdb/t/use_direct_io_for_flush_and_compaction.test
```diff
@@ -0,0 +1,37 @@
+--source include/have_rocksdb.inc
+
+--perl
```
laurynas-biveinis: This is copy-paste with use_direct_reads.test - begs for an include file

hermanlee (changes requested):
Refactor tests

laurynas-biveinis (commented):
What I had in mind to was to factor out some skip_if_data_in_dev_shm.inc, but this looks good to me too :)

hermanlee:
I found another test was broken, so I refactored headers into have_direct_io.inc. A better name would be have_not_in_shm.inc though, but this should be fine.

PR #898, opened 2018-11-12
issue-896 : Segmentation fault in myrocks::Rdb_string_reader::read

george-lorch:
- Code for skipping ttl bytes for secondary keys is slightly incorrect.

### DIFF
--- a/mysql-test/suite/rocksdb/r/issue896.result
+++ b/mysql-test/suite/rocksdb/r/issue896.result
@@ -1,0 +1,17 @@
+CREATE TABLE `t1` (
+`a` bigint(20) NOT NULL,
+`b` varchar(10) CHARACTER SET utf8 COLLATE utf8_bin NOT NULL,
+`u` bigint(20) unsigned NOT NULL,
+`d` bigint(20) DEFAULT NULL,
+PRIMARY KEY (`a`,`b`),
+KEY `d` (`d`)
+) ENGINE=ROCKSDB DEFAULT CHARSET=utf8 COLLATE=utf8_bin COMMENT='ttl_duration=1000;ttl_col=u';
+INSERT INTO t1 VALUES (100, 'aaabbb', UNIX_TIMESTAMP(), 200);
+EXPLAIN SELECT COUNT(*) FROM t1;
+id	select_type	table	type	possible_keys	key	key_len	ref	rows	Extra
+1	SIMPLE	t1	index	NULL	d	9	NULL	#	Using index
+# segfault here without the fix
+SELECT COUNT(*) FROM t1;
+COUNT(*)
+1
+DROP TABLE t1;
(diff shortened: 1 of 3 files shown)


storage/rocksdb/rdb_datadic.cc
```diff
@@ -1416,14 +1416,19 @@ int Rdb_key_def::unpack_record(TABLE *const table, uchar *const buf,
   // For secondary keys, we expect the value field to contain unpack data and
```
lth: We should probably update the comment here to say, we expect for SKs to have index flags in the beginning as well.(which is just TTL only at the moment)

george-lorch: Ack,makes sense.  I'll update and re-push later.

mysql-test/suite/rocksdb/r/issue896.result
```diff
@@ -0,0 +1,17 @@
...
+`b` varchar(10) CHARACTER SET utf8 COLLATE utf8_bin NOT NULL,
+`u` bigint(20) unsigned NOT NULL,
+`d` bigint(20) DEFAULT NULL,
+PRIMARY KEY (`a`,`b`),
+KEY `d` (`d`)
+) ENGINE=ROCKSDB DEFAULT CHARSET=utf8 COLLATE=utf8_bin COMMENT='ttl_duration=1000;ttl_col=u';
+INSERT INTO t1 VALUES (100, 'aaabbb', UNIX_TIMESTAMP(), 200);
+EXPLAIN SELECT COUNT(*) FROM t1;
+id	select_type	table	type	possible_keys	key	key_len	ref	rows	Extra
+1	SIMPLE	t1	index	NULL	d	9	NULL	1	Using index
```
lth: Also, we need to mask out the 'rows' value because of test flakiness.

storage/rocksdb/rdb_datadic.cc
```diff
@@ -1413,17 +1413,22 @@ int Rdb_key_def::unpack_record(TABLE *const table, uchar *const buf,
     return HA_ERR_ROCKSDB_CORRUPT_DATA;
   }
 
-  // For secondary keys, we expect the value field to contain unpack data and
-  // checksum data in that order. One or both can be missing, but they cannot
-  // be reordered.
+  // For secondary keys, we expect the value field to contain unpack data,
```
lth: @georgelorchpercona What I meant was:

For secondary keys, we expect the value field to contain index flags, unpack data,
and checksum data in that order. One or all can be missing,
but they cannot be reordered.

george-lorch: Ack, thanks, I glanced at the code and it looked like the flags were being packed in last.  I'll go back and re-review for my own sanity and re-push.

Issue #904, opened 2018-11-19
Crash in rocksdb::IOStatsContext::Reset, this=NULL


spetrunia:
The following tests crash for me: rocksdb.bytes_written rocksdb.rocksdb_icp rocksdb.perf_context rocksdb.slow_query_log.  The with top-level frame of the crash is rocksdb::IOStatsContext::Reset which is called with this=NULL.

The crash happens on some systems/builds, but not others.

A stack trace from bytes_written.test:

```
#4  0x000055926e07e9ae in rocksdb::IOStatsContext::Reset (this=0x0) at /home/psergey/dev-git/mysql-5.6-try4/rocksdb/monitoring/iostats_context.cc:25
#5  0x000055926de8fbb8 in myrocks::Rdb_io_perf::start (this=0x7f99a404df70, perf_context_level=3) at /home/psergey/dev-git/mysql-5.6-try4/storage/rocksdb/rdb_perf_context.cc:202
#6  0x000055926ddbd1e0 in myrocks::(anonymous namespace)::Rdb_perf_context_guard::Rdb_perf_context_guard (this=0x7f99c034fc10, io_perf=0x7f99a404df70, level=3) at /home/psergey/dev-git/mysql-5.6-try4/storage/rocksdb/ha_rocksdb.cc:3286
#7  0x000055926ddc9b7d in myrocks::ha_rocksdb::open (this=0x7f99a404d950, name=0x7f99a40a8de0 "./test/stats_test_table", mode=2, test_if_locked=2) at /home/psergey/dev-git/mysql-5.6-try4/storage/rocksdb/ha_rocksdb.cc:6587
#8  0x000055926d4dc24a in handler::ha_open (this=0x7f99a404d950, table_arg=0x7f99a404cff0, name=0x7f99a40a8de0 "./test/stats_test_table", mode=2, test_if_locked=2) at /home/psergey/dev-git/mysql-5.6-try4/sql/handler.cc:2843
#9  0x000055926d7e229c in open_table_from_share (thd=0x55926ffc14e0, share=0x7f99a40a8a30, alias=0x7f99a40058c0 "stats_test_table", db_stat=39, prgflag=44, ha_open_flags=0, outparam=0x7f99a404cff0, is_create_table=false) at /home/psergey/dev-git/mysql-5.6-try4/sql/table.cc:2675
[...]

PR #905, opened 2018-11-19
Issue #904: Crash in rocksdb::IOStatsContext::Reset, this=NULL

spetrunia:
Fix both code paths:
- Change the test source code so it doesn't cause the "Unused variable"
  warning (which -Werror converted into error and caused CMake not to set
  HAVE_THREAD_LOCAL)

- If the system doesn't seem to support HAVE_THREAD_LOCAL, refuse to
  compile (rather than producing a binary that crashes for some tests)

### DIFF
--- a/storage/rocksdb/CMakeLists.txt
+++ b/storage/rocksdb/CMakeLists.txt
@@ -47,10 +47,15 @@ CHECK_CXX_SOURCE_COMPILES("
 #endif
 int main() {
   static __thread int tls;
+  tls=0;
+  return tls;
 }
 " HAVE_THREAD_LOCAL)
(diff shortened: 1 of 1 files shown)


laurynas-biveinis:
C++11 thread_local seems to be fully supported starting with GCC 4.8, clang 3.3, MSVC 2015 - maybe that's good enough to switch to thread_local and drop CMake checks altogether?

hermanlee (approved):

hermanlee:
Removing the check should be okay too. The check looks to be inherited from RocksDB cmake files. @spetrunia, I assume none of your builds actually fail with your updated cmake check and thread local is supported across all builds?

PR #913, opened 2018-12-13
Optimize MyRocks Replace Into Statement

justfortaste:
Summary:
When rocksdb tables only have primary key and without secondary indexes,
Replace Into can use Insert Into instead and this insert into no
need to check unique constraint.

### DIFF
--- a/mysql-test/suite/rocksdb/t/optimize_myrocks_replace_into_base.test
+++ b/mysql-test/suite/rocksdb/t/optimize_myrocks_replace_into_base.test
@@ -1,0 +1,79 @@
+--source include/have_rocksdb.inc
+SET @prior_rocksdb_perf_context_level = @@rocksdb_perf_context_level;
+SET GLOBAL rocksdb_perf_context_level=3;
+
+#
+# cast 1: table only with primary key, support replace blind write
+#
+create table t1(c1 int,c2 int, primary key (c1)) engine=rocksdb;
+insert into t1 values(1,1),(2,2),(3,3);
+select * from t1;
+
+select value into @c from information_schema.rocksdb_perf_context where table_schema='test' and table_name='t1' and stat_type='GET_FROM_MEMTABLE_COUNT';
+replace into t1 values(1,11);
+select case when value-@c > 0 then 'false' else 'true' end as read_free from information_schema.rocksdb_perf_context where table_schema='test' and table_name='t1' and stat_type='GET_FROM_MEMTABLE_COUNT';
+drop table t1;
+
+#
+# cast 2: table only with primary key but with trigger, not support replace blind write
+#
+create table t1(c1 int,c2 int, primary key (c1)) engine=rocksdb;
+create trigger trg before insert on t1 for each row set @a:=1;
+insert into t1 values(1,1),(2,2),(3,3);
+select * from t1;
+
+select value into @c from information_schema.rocksdb_perf_context where table_schema='test' and table_name='t1' and stat_type='GET_FROM_MEMTABLE_COUNT';
+replace into t1 values(1,11);
+select case when value-@c > 0 then 'false' else 'true' end as read_free from information_schema.rocksdb_perf_context where table_schema='test' and table_name='t1' and stat_type='GET_FROM_MEMTABLE_COUNT';
+drop table t1;
+
+
+#
+# cast 3: table without primary key, not support replace blind write
+#
+
+create table t1(c1 int,c2 int) engine=rocksdb;
+insert into t1 values(1,1),(2,2),(3,3);
+select * from t1;
+
+select value into @c from information_schema.rocksdb_perf_context where table_schema='test' and table_name='t1' and stat_type='GET_FROM_MEMTABLE_COUNT';
+replace into t1 values(1,11);
+select case when value-@c > 0 then 'false' else 'true' end as read_free from information_schema.rocksdb_perf_context where table_schema='test' and table_name='t1' and stat_type='GET_FROM_MEMTABLE_COUNT';
+drop table t1;
+
+
+create table t1(c1 int,c2 int unique) engine=rocksdb;
+insert into t1 values(1,1),(2,2),(3,3);
+select * from t1;
+
+select value into @c from information_schema.rocksdb_perf_context where table_schema='test' and table_name='t1' and stat_type='GET_FROM_MEMTABLE_COUNT';
+replace into t1 values(1,11);
+select case when value-@c > 0 then 'false' else 'true' end as read_free from information_schema.rocksdb_perf_context where table_schema='test' and table_name='t1' and stat_type='GET_FROM_MEMTABLE_COUNT';
+drop table t1;
+
+
+
+#
+# cast 4: table with primary key and secondary key, not support replace blind write
+#
+create table t1(c1 int primary key,c2 int unique) engine=rocksdb;
+insert into t1 values(1,1),(2,2),(3,3);
+select * from t1;
+
+select value into @c from information_schema.rocksdb_perf_context where table_schema='test' and table_name='t1' and stat_type='GET_FROM_MEMTABLE_COUNT';
+replace into t1 values(1,11);
+select case when value-@c > 0 then 'false' else 'true' end as read_free from information_schema.rocksdb_perf_context where table_schema='test' and table_name='t1' and stat_type='GET_FROM_MEMTABLE_COUNT';
+drop table t1;
+
+
+create table t1(c1 int primary key,c2 int, key idx1(c2)) engine=rocksdb;
+insert into t1 values(1,1),(2,2),(3,3);
+select * from t1;
+
+select value into @c from information_schema.rocksdb_perf_context where table_schema='test' and table_name='t1' and stat_type='GET_FROM_MEMTABLE_COUNT';
+replace into t1 values(1,11);
+select case when value-@c > 0 then 'false' else 'true' end as read_free from information_schema.rocksdb_perf_context where table_schema='test' and table_name='t1' and stat_type='GET_FROM_MEMTABLE_COUNT';
+drop table t1;
+
+SET GLOBAL rocksdb_perf_context_level = @prior_rocksdb_perf_context_level;
+
--- a/mysql-test/suite/rocksdb/t/optimize_myrocks_replace_into_lock.test
+++ b/mysql-test/suite/rocksdb/t/optimize_myrocks_replace_into_lock.test
@@ -1,0 +1,75 @@
+connect (con1,localhost,root,,);
+connect (con2,localhost,root,,);
+
+
+
+create table t1(c1 int,c2 int, primary key (c1)) engine=rocksdb;
+insert into t1 values(1,1),(2,2),(3,3);
+select * from t1;
+
+#
+# cast 1: update is blocked by replace into 
+#
+connection con1;
+begin;
+replace into t1 values(1,11);
+
+
+connection con2;
+begin;
+send update t1 set c2=22 where c1=1;
+
+
+connection default;
+# Check that the above update is blocked
+let $wait_condition=
+  select count(*) = 1 from information_schema.processlist
+    where state = 'Waiting for row lock' and
+          info = 'update t1 set c2=22 where c1=1';
+--source include/wait_condition.inc
+
+
+connection con1;
+commit;
+
+connection con2;
+--echo # Reap update.
+--reap
+commit;
+select * from t1;
+
+
+#
+# cast 2: replace into is blocked by update
+#
+
+connection con1;
+begin;
+update t1 set c2=55 where c1=1;
+
+connection con2;
+begin;
+send replace into t1 values(1,66);
+
+
+connection default;
+# Check that the above replace into is blocked
+let $wait_condition=
+  select count(*) = 1 from information_schema.processlist
+    where state = 'Waiting for row lock' and
+          info = 'replace into t1 values(1,66)';
+--source include/wait_condition.inc
+
+
+connection con1;
+commit;
+
+connection con2;
+--echo # Reap replace into.
+--reap
+commit;
+select * from t1;
+
+connection default;
+drop table t1;
+
--- a/mysql-test/suite/rocksdb/t/optimize_myrocks_replace_into_lock.test
+++ b/mysql-test/suite/rocksdb/t/optimize_myrocks_replace_into_lock.test
@@ -1,7 +1,6 @@
+--source include/have_rocksdb.inc
 connect (con1,localhost,root,,);
 connect (con2,localhost,root,,);
-
-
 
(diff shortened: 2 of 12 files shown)


mysql-test/suite/rocksdb_rpl/r/optimize_myrocks_replace_info.result
```diff
@@ -0,0 +1,52 @@
...
+replace into t1 values(1,11);
+replace into t1 values(2,22);
+replace into t1 values(3,33);
+select * from t1;
+c1	c2
+1	11
+2	22
+3	33
+replace into t1 values(2,44),(3,55);
+show binlog events;
```
yoshinorim: Outputs of "show binlog events" may change between versions or build types, so they make tests unstable. Instead, how about checking Get count from perf context for each replace, in both master and slave? Expected behavior is Get count does not increase in both master and slave.

```
select value into @g from information_schema.rocksdb_perf_context_global where stat_type='GET_FROM_MEMTABLE_COUNT';

REPLACE...

select case when value-@c > 0 then 'false' else 'true' end as read_free from information_schema.rocksdb_perf_context_global where stat_type='GET_FROM_MEMTABLE_COUNT';
```

rocksdb_perf_context_level should be non zero, to count Gets (3 is fine).

mysql-test/suite/rocksdb_rpl/t/optimize_myrocks_replace_info.test
```diff
@@ -0,0 +1,29 @@
...
+
+connection master;
+
+create table t1(c1 int,c2 int, primary key (c1)) engine=rocksdb;
+insert into t1 values(1,1),(2,2),(3,3);
+select * from t1;
+
+
+
+replace into t1 values(1,11);
```
yoshinorim: Could you add another test case about locking --  1. "replace into t1 values(1,11);" blocks updates/deletes to the same key (1) until a transaction is committed, and 2. replace is blocked if there is already a running transaction locking the same key?

yoshinorim (commented):
Thanks a lot for the awesome feature! Overall code looks good to me. I am asking a few more people to review. I have a few requests about your test case.

mysql-test/suite/rocksdb_rpl/r/optimize_myrocks_replace_info.result
justfortaste: That is a good idead, 
use  -- source include/show_binlog_events.inc is also ok.

mysql-test/suite/rocksdb_rpl/t/optimize_myrocks_replace_info.test
justfortaste: OK

mysql-test/suite/rocksdb_rpl/r/optimize_myrocks_replace_info.result
yoshinorim: You may combine both rocksdb_perf_context_global and source include/show_binlog_events.inc. Raw "show binlog events" caused test failures on some of our builds, since version strings were different.

```
@@ -21,7 +21,7 @@
 replace into t1 values(2,44),(3,55);
 show binlog events;
 Log_name	Pos	Event_type	Server_id	End_log_pos	Info
-master-bin.000001	4	Format_desc	1	120	Server ver: 5.6.35-log, Binlog ver: 4
+master-bin.000001	4	Format_desc	1	120	Server ver: 5.6.35-fb-log, Binlog ver: 4
```

mysql-test/suite/rocksdb_rpl/t/optimize_myrocks_replace_into.test
```diff
@@ -0,0 +1,38 @@
...
+--source include/have_binlog_format_row.inc
+
+connection master;
+SET GLOBAL rocksdb_perf_context_level=3;
+
+create table t1(c1 int,c2 int, primary key (c1)) engine=rocksdb;
+insert into t1 values(1,1),(2,2),(3,3);
+select * from t1;
+
+
```
yoshinorim, line 16: Could you check GET_FROM_MEMTABLE_COUNT on slave as well? Something like this:

insert ...
```
connection slave;
select value into @c from information_schema.rocksdb_perf_context where table_schema='test' and table_name='t1' and stat_type='GET_FROM_MEMTABLE_COUNT';
connection master;
```
replace...
sync_slave_with_master;
```
select case when value-@c > 0 then 'false' else 'true' end as read_free from information_schema.rocksdb_perf_context where table_schema='test' and table_name='t1' and stat_type='GET_FROM_MEMTABLE_COUNT';
```

mysql-test/suite/rocksdb_rpl/r/optimize_myrocks_replace_into.result
```diff
@@ -0,0 +1,70 @@
+include/master-slave.inc
+Warnings:
+Note	####	Sending passwords in plain text without SSL/TLS is extremely insecure.
+Note	####	Storing MySQL user name or password information in the master info repository is not secure and is therefore not recommended. Please consider using the USER and PASSWORD connection options for START SLAVE; see the 'START SLAVE Syntax' in the MySQL Manual for more information.
+[connection master]
+SET GLOBAL rocksdb_perf_context_level=3;
+create table t1(c1 int,c2 int, primary key (c1)) engine=rocksdb;
```
yoshinorim, line 8: How about testing a table with secondary keys, like this? Then it can verify if secondary keys are correctly updated.
```
create table t1 (c1 int primary key, c2 int, index (c2)) engine=rocksdb;
```

mysql-test/suite/rocksdb_rpl/t/optimize_myrocks_replace_into.test
```diff
@@ -0,0 +1,38 @@
...
+update t1 set c2=66 where c1=3;
+select case when value-@c > 0 then 'false' else 'true' end as read_free from information_schema.rocksdb_perf_context where table_schema='test' and table_name='t1' and stat_type='GET_FROM_MEMTABLE_COUNT';
+--source include/show_binlog_events.inc
+
+sync_slave_with_master;
+
+select * from t1;
+
+connection master;
+drop table t1;
```
yoshinorim, line 63: You may add tests about
- Table without primary key
- Table with a trigger

and confirm blind replace is disabled.

mysql-test/suite/rocksdb/t/optimize_myrocks_replace_into_lock.test
```diff
@@ -0,0 +1,75 @@
...
+commit;
+
+connection con2;
+--echo # Reap replace into.
+--reap
+commit;
+select * from t1;
+
+connection default;
+drop table t1;
```
laurynas-biveinis, line 73: There are no disconnects for con1 and con2. Since disconnect processing is async in the server, it is also a good idea to surround the testcase with include/count_sessions.inc & wait_until_count_sessions.inc

mysql-test/suite/rocksdb/t/optimize_myrocks_replace_into_lock.test
```diff
@@ -0,0 +1,75 @@
+connect (con1,localhost,root,,);
```
laurynas-biveinis, line 2: --source include/have_rocksdb.inc

mysql-test/suite/rocksdb/t/optimize_myrocks_replace_into_lock.test
```diff
@@ -0,0 +1,75 @@
...
+connect (con2,localhost,root,,);
+
+
+
+create table t1(c1 int,c2 int, primary key (c1)) engine=rocksdb;
+insert into t1 values(1,1),(2,2),(3,3);
+select * from t1;
```
laurynas-biveinis, line 7: IMHO redundant

mysql-test/suite/rocksdb_rpl/t/optimize_myrocks_replace_into.test
```diff
@@ -0,0 +1,38 @@
+--source include/have_rocksdb.inc
+--source include/master-slave.inc
+--source include/have_binlog_format_row.inc
```
laurynas-biveinis: Please swap these two: the master-slave.inc is heavy, it sets up replication, and have_binlog_format_row.inc will throw away all that work for the other two replication formats

mysql-test/suite/rocksdb_rpl/t/optimize_myrocks_replace_into.test
```diff
@@ -0,0 +1,38 @@
...
+
+connection master;
```
laurynas-biveinis, line 4: Redundant

mysql-test/suite/rocksdb_rpl/t/optimize_myrocks_replace_into.test
```diff
@@ -0,0 +1,38 @@
...
+SET GLOBAL rocksdb_perf_context_level=3;
+
+create table t1(c1 int,c2 int, primary key (c1)) engine=rocksdb;
+insert into t1 values(1,1),(2,2),(3,3);
+select * from t1;
```
laurynas-biveinis, line 10: IMHO redundant

mysql-test/suite/rocksdb_rpl/t/optimize_myrocks_replace_into.test
```diff
@@ -0,0 +1,38 @@
...
+select value into @c from information_schema.rocksdb_perf_context where table_schema='test' and table_name='t1' and stat_type='GET_FROM_MEMTABLE_COUNT';
+replace into t1 values(2,44),(3,55);
+select case when value-@g > 0 then 'false' else 'true' end as read_free from information_schema.rocksdb_perf_context where table_schema='test' and table_name='t1' and stat_type='GET_FROM_MEMTABLE_COUNT';
+
+select value into @c from information_schema.rocksdb_perf_context where table_schema='test' and table_name='t1' and stat_type='GET_FROM_MEMTABLE_COUNT';
+update t1 set c2=66 where c1=3;
+select case when value-@c > 0 then 'false' else 'true' end as read_free from information_schema.rocksdb_perf_context where table_schema='test' and table_name='t1' and stat_type='GET_FROM_MEMTABLE_COUNT';
+--source include/show_binlog_events.inc
+
+sync_slave_with_master;
```
laurynas-biveinis, line 57: Consider --source include/sync_slave_sql_with_master.inc instead

mysql-test/suite/rocksdb_rpl/t/optimize_myrocks_replace_into.test
```diff
@@ -0,0 +1,38 @@
+--source include/have_rocksdb.inc
+--source include/master-slave.inc
+--source include/have_binlog_format_row.inc
+
+connection master;
+SET GLOBAL rocksdb_perf_context_level=3;
```
laurynas-biveinis, line 6: The usual idiom for global vars in MTR is not to reset to default in the end, but to save the original value in a user SQL variable first, and restore from it in the end. This way it becomes more compatible with MTR --mysqld=--some-option=non-default invocations

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -8968,6 +8982,8 @@ int ha_rocksdb::check_and_lock_unique_pk(const uint &key_id,
   DBUG_ASSERT(found != nullptr);
   DBUG_ASSERT(pk_changed != nullptr);
 
+  bool ignore_get = false;
```
laurynas-biveinis, line 8985: Declare at initialization instead?

mysql-test/suite/rocksdb_rpl/t/optimize_myrocks_replace_into.test
justfortaste: good suggestion

mysql-test/suite/rocksdb_rpl/t/optimize_myrocks_replace_into.test
justfortaste: Replace into bline write only effect row base replication, Maybe I need to make a new testcase to test all replication formats

sql/log_event.h
```diff
@@ -650,6 +650,10 @@ struct sql_ex_info
 #define LOG_EVENT_MTS_ISOLATE_F 0x200
 
 
+/**
+  flag for myrocks replace into
+ */
+#define LOG_EVENT_DUP_REPLACE_F 0x400
```
hermanlee, line 656: It seems like this could easily collide with changes made by Oracle, even though it looks like 8.0 hasn't added anything new here. It also seems this might be better off in the row_log_event structure rather than use one up one of flag bits that apply to all binlog events. There are also 16 bits there and only 4 are used so far. Using one of the upper bits there would have a small chance of collision.

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -8986,6 +9002,9 @@ int ha_rocksdb::check_and_lock_unique_pk(const uint &key_id,
     *pk_changed = true;
   }
 
+  ignore_get = can_optimize_replace_into();
+  if (ignore_get)
+    ha_thd()->lex->can_optimize_replace_into = true;
```
hermanlee, line 9007: This seems to be the wrong dependency ordering to me. Here, MyRocks determines if optimized replace_into can be used, if so, then can_optimize_replace_into is set, which then tells the binlog to write out the LOG_EVENT_DUP_REPLACE_F bit. So binlog's decision to use DUP_REPLACE depends on MyRocks, which depends on MySQL's processing of the command. But binlog's decision to indicate DUP_REPLACE should be independent of the storage engine.

Would a MyRocks slave reading binlog entries from an InnoDB master which processed the REPLACE query be allowed to use the optimization if it satisfies the other conditions (no hidden primary key, no triggers).

Alternatively, if a MyRocks master had triggers set on its table, but a MyRocks slave did not have triggers, should the MyRocks slave be able to perform the REPLACE optimization too?

mysql-test/suite/rocksdb/t/optimize_myrocks_replace_into_base.test
```diff
@@ -0,0 +1,79 @@
+--source include/have_rocksdb.inc
+SET @prior_rocksdb_perf_context_level = @@rocksdb_perf_context_level;
+SET GLOBAL rocksdb_perf_context_level=3;
+
+#
+# cast 1: table only with primary key, support replace blind write
```
hermanlee, line 6: These test cases are great. Can they be extended to master/slave setups too to verify the slaves? But we'd also want to test combinations where the master has triggers/secondary keys, the slave doesn't, or where the slave has triggers/secondary keys, but the master does not.

The reason these cases are important is because schema changes and triggers are applied to each member of a replicaset at different times, so there are times when schemas between master and slave will different.

hermanlee (commented):
Can a variable setting controlling the use of the replace into optimization and binlog modifications be added?

PR #916, opened 2018-12-24
Bug Fix: MyRocks locking scan with RC may not return full result sets

yoshinorim:
Summary:
With Read Committed isolation level, MyRocks locking full/range scan
(updates/deletes with Seek) may end up GetForUpdate returning kNotFound,
if other transactions delete the same key after Seek and before
GetForUpdate. MyRocks was expected to skip to next row in that case.
However, under certain conditions (locking range scan with equal predicates,
and locking range/full scan with secondary index),
MyRocks stops scanning records further and end up not updating after
that, even if further rows may match.
From caller's perspective, the update succeeds with updating
zero/fewer rows.
This diff catches the cases and preventing from stopping scanning.

Test Plan:
Merging delete_before_lock into rocksdb_concurrent_delete. It now covers
RR/RC, and PK/SK range/full scan.


mysql-test/suite/rocksdb/t/delete_before_lock.inc
```diff
@@ -0,0 +1,37 @@
...
+--source include/have_debug_sync.inc
+
+# This is a test case to reproduce https://github.com/facebook/mysql-5.6/issues/162
+# Expected output of the last select for update was (1,2,100) and (1,3,100), but
+# currently it returns (1,2,1) and (1,3,1), which must be fixed.
+
+connect (con, localhost, root,,);
+connection default;
+
+--disable_warnings
```
laurynas-biveinis: Obsolete idiom, please consider removing

mysql-test/suite/rocksdb/t/delete_before_lock.inc
```diff
@@ -0,0 +1,37 @@
+--source include/have_rocksdb.inc
+--source include/have_debug_sync.inc
+
+# This is a test case to reproduce https://github.com/facebook/mysql-5.6/issues/162
+# Expected output of the last select for update was (1,2,100) and (1,3,100), but
+# currently it returns (1,2,1) and (1,3,1), which must be fixed.
+
+connect (con, localhost, root,,);
```
laurynas-biveinis: Please consider wrapping with count_sessions.inc / wait_until_count_sessions.inc

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -12979,6 +12987,23 @@ bool Rdb_manual_compaction_thread::is_manual_compaction_finished(int mc_id) {
...
+ * someone else in the meantime.
+ * If so, we re-scan an iterator and try again.
+ */
+bool ha_rocksdb::should_recreate_snapshot(const bool lock_row, const int rc,
+                                          const bool is_new_snapshot) {
+  if (!(lock_row && rc == HA_ERR_KEY_NOT_FOUND &&
+        my_core::thd_tx_isolation(ha_thd()) == ISO_READ_COMMITTED) &&
+      (rc != HA_ERR_ROCKSDB_STATUS_BUSY || !is_new_snapshot)) {
+    return false;
+  }
```
lth, line 13013: I think rewriting this as a positive expression makes it more readable:
```
  if ((lock_row && rc == HA_ERR_KEY_NOT_FOUND &&
        my_core::thd_tx_isolation(ha_thd()) == ISO_READ_COMMITTED) ||
      (rc == HA_ERR_ROCKSDB_STATUS_BUSY && is_new_snapshot)) {
    return true;
  }
```

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -12979,6 +12987,23 @@ bool Rdb_manual_compaction_thread::is_manual_compaction_finished(int mc_id) {
...
+/**
+ * Locking read + Not Found + Read Committed occurs if we accessed
+ * a row by Seek, tried to lock it, failed, released and reacquired the
+ * snapshot (because of READ COMMITTED mode) and the row was deleted by
+ * someone else in the meantime.
+ * If so, we re-scan an iterator and try again.
+ */
+bool ha_rocksdb::should_recreate_snapshot(const bool lock_row, const int rc,
+                                          const bool is_new_snapshot) {
+  if (!(lock_row && rc == HA_ERR_KEY_NOT_FOUND &&
```
lth: You can just read m_lock_rows.

lth (commented):
The case where the non-first key is removed needs to be handled and tested as well.

From discussion offline:

```
create table (i int, j int, primary key (i), key (j));
insert into t values (1, 1), (2, 2), (3, 3), (4, 4), (5, 5), (6, 6), (7, 7), (8, 8), (9, 9);
mysql> select * from t force index (j) where 0=sleep(1) for update;
...
(in a different session)
mysql> delete from t where i = 8;
..
mysql> select * from t force index (j) where 0=sleep(1) for update;

+---+------+
| i | j    |
+---+------+
| 1 |    1 |
| 2 |    2 |
| 3 |    3 |
| 4 |    4 |
| 5 |    5 |
| 6 |    6 |
| 7 |    7 |
+---+------+

```

mysql-test/suite/rocksdb/include/rocksdb_concurrent_delete_range.inc
```diff
@@ -0,0 +1,85 @@
...
+eval SET SESSION TRANSACTION ISOLATION LEVEL $isolation_level;
+create table t1 (id1 int, id2 int, value int, primary key (id1, id2), index sk (id1, value)) engine=rocksdb;
+insert into t1 values (1, 1, 1),(1, 2, 1),(1, 3, 1),(1, 4, 1),(1, 5, 1),(2, 2, 2);
+
+# deleting a first row
+--echo --First row delete with $index
+connection con;
+eval SET SESSION TRANSACTION ISOLATION LEVEL $isolation_level;
+set debug_sync='rocksdb.get_row_by_rowid SIGNAL parked WAIT_FOR go';
+send update t1 $index set value=100 where id1=1;
```
lth: Don't you need send_eval here?

I'm really surprised that `update t1 $index`  is valid syntax!

lth (commented):
One thing that confused me was the existence of both
mysql-test/suite/rocksdb/include/rocksdb_concurrent_delete.inc
and
mysql-test/suite/rocksdb/t/rocksdb_concurrent_delete.inc

I would prefer that we named them differently.

Because we're not refreshing snapshot during iteration, this means that we might still miss some (committed) rows. This is probably okay, since you'll still miss rows even with refresh, but might be worth commenting about this somewhere.

mysql-test/suite/rocksdb/include/rocksdb_concurrent_delete_range.inc
lth: Looks like it's just table alias.

yoshinorim:
Fixed tests, added more source comments for better understanding.

lth (approved):
Looks good to me.

PR #920, opened 2019-01-08
Skip snapshot validation on Read Committed

yoshinorim:
Summary:
This diff adopts RocksDB's new API (Extend Transaction::GetForUpdate
with do_validate (#4680)). If transaction isolation level is read
committed, snapshot validation is skipped on GetForUpdate. This slightly
improves performance.

Snapshot conflict still happens, even if skipping snapshot validation.
Typical example is T1 Seek (finding key1) -> T2 Delete(key 1) -> T1
GetForUpdate (key 1) . GetForUpdate now returns kNotFound with this
diff. Prior to this diff, it was encapsulated in
ha_rocksdb::get_for_update and it eventually returned kNotFound.
Overall behavior unchanges -- in both
cases, MyRocks either implicitly re-creates a snapshot or proceeds to
next row.

@update-submodule: rocksdb

Test Plan: rocksdb_concurrent_delete has already enough test coverage

Tasks:

Tags:


lth:
I'm okay if you don't want to include in this diff, but can't you also skip snapshot validation (ie. set assume_tracked = true) for RR? The only difference between RR/RC should be be just the get_for_update path. In RR, get_for_update can return snapshot conflict, whereas it should always succeed in RC (thus removing retry logic).

I think this will simplify the diff a bit.

yoshinorim:
Skipping snapshot validation with RR changes behavior from PostgreSQL style (may get snapshot conflict errors) to InnoDB style (locking reads do read committed behavior) so I think it should be in a separated diff. The diff will simplify retry logic, and will also update several test files.

lth:
I don't think it actually changes RR behaviour, because get_for_update should still do snapshot reads (not locking reads), and return snapshot conflict.

We should never do locking puts/deletes/SDs without first locking with get_for_update, which is why we can assume that keys are already locked.

lth:
Actually, have you tried running the blind deletes test with RC? Does that work (since we can't assume locked)?

yoshinorim:
- Setting assume_tracked=true for both RR and RC, except under blind replication/deletions
- Added blind_delete_rc.test

lth (approved):
Looks good to me (assuming tests pass).

Issue #923, opened 2019-01-10
(FB8-211) Building with -DWITH_ZSTD=system fails on both CentOS and Ubuntu


percona-ysorokin:
https://github.com/facebook/mysql-5.6/blob/71a0757b8792210c3a102e6c07a3d38ca9cf0277/cmake/zstd.cmake#L2

https://jira.percona.com/browse/FB8-211

Executing
```
cmake  <path>/facebook-mysql \
-DCMAKE_BUILD_TYPE=Debug \
-DBUILD_CONFIG=mysql_release \
-DFEATURE_SET=community \
-DMYSQL_MAINTAINER_MODE=ON \
-DWITH_SSL=system \
-DWITH_ZSTD=system \
-DENABLE_DOWNLOADS=ON \
-DDOWNLOAD_BOOST=ON \
-DWITH_BOOST=../deps
```
generates the following cmake error
```
...
-- OpenSSL include: /usr/include
-- OpenSSL lib:
-- LZMA_INCLUDE_DIR /mnt/hgfs/repos/facebook-mysql/extra/lzma
-- LZMA_LIBRARY lzma_lib
CMake Error at cmake/zstd.cmake:28 (MESSAGE):
  Cannot find system zstd libraries.
Call Stack (most recent call first):
  CMakeLists.txt:922 (MYSQL_CHECK_ZSTD)


-- Configuring incomplete, errors occurred!
See also "/home/yura/addon/mysql-build-8.0-debug_gcc8_broken/CMakeFiles/CMakeOutput.log".
See also "/home/yura/addon/mysql-build-8.0-debug_gcc8_broken/CMakeFiles/CMakeError.log".
```

The problem is in the `cmake/zstd.cmake` file where `FIND_SYSTEM_ZSTD` macro tries to find `zstd/zstd.h` instead of just `zstd.h`.

On both Debian-style (including Ubuntu) and RHEL-style (including CentOS)  platforms this file, when installed from the system packages (`libzstd1-dev` / `libzstd-devel`), is located in `/usr/include` and therefore does not match `zstd/zstd.h` pattern.

hermanlee:
@lth , it looks like https://github.com/facebook/mysql-5.6/commit/aff4780f45b and  https://github.com/facebook/mysql-5.6/commit/9579bd99290 probably both need to be ported to 8.0 too.

PR #924, opened 2019-01-10
Fix FB8-211 (Building with -DWITH_ZSTD=system fails on both CentOS and Ubuntu)

percona-ysorokin:
https://jira.percona.com/browse/FB8-211

Fixed a problem in the 'cmake/zstd.cmake' file where 'FIND_SYSTEM_ZSTD'
macro tries to find 'zstd/zstd.h' instead of just 'zstd.h'.
On both Debian-style (including Ubuntu) and RHEL-style (including CentOS)
platforms this file, when installed from the system packages
('libzstd1-dev' / 'libzstd-devel'), is located in '/usr/include' and
therefore does not match 'zstd/zstd.h' pattern.

This PR fixes Issue #923 

Squash with D7142958


hermanlee (commented):
Can you add

`Squash with D7142958`

to the commit description? Thanks!

percona-ysorokin:
Updated commit message with a reference to D7142958.

hermanlee (approved):

PR #933, opened 2019-01-21
FB8-64: Fix time granularity in dump thread

dutow:
Summary:
Dump thread used second granularity to decide whether
heartbeat needs to be sent to slaves. With milli-second heartbeat
configuration on slaves, this caused dump thread to send heartbeat
events for every skipped event. This significantly increases slave
start time. Fix this by increasing the granularity.

JIRA: https://jira.percona.com/browse/FB8-64

Fixes upstream bug: https://bugs.mysql.com/bug.php?id=81164

Reference patch: https://github.com/facebook/mysql-5.6/commit/1ac73cf

Test Plan:
The MTR changes done by the original patch doesn't result in measurable slave start time increase.
The change was verified manually with traces for 8.0:
with modified (millisecond) heartbeat_period in some rpl MTR testcases, the number of heartbeats is significatly lower with the patch applied.

### DIFF
--- a/sql/rpl_binlog_sender.cc
+++ b/sql/rpl_binlog_sender.cc
@@ -22,12 +22,12 @@
 
 #include "sql/rpl_binlog_sender.h"
 
+#include <errno.h>
 #include <stdio.h>
+#include <sys/socket.h>
 #include <algorithm>
 #include <atomic>
-#include <errno.h>
 #include <memory>
-#include <sys/socket.h>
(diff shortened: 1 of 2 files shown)


percona-ysorokin:
LGTM

percona-ysorokin (approved):
LGTM

hermanlee (approved):

hermanlee (approved):

PR #934, opened 2019-01-22
FB8-54, FB8-55, FB8-70, FB8-101: Expose more information to audit plugin

dutow:
Summary:
This commit adds the following fields to the generic event in audit log:
* query_id
* database
* affected_rows
* connection_certificate

Reference Patch: https://github.com/facebook/mysql-5.6/commit/1def6b7
Reference Patch: https://github.com/facebook/mysql-5.6/commit/ce95a09
Reference Patch: https://github.com/facebook/mysql-5.6/commit/588be34
Reference Patch: https://github.com/facebook/mysql-5.6/commit/ba03c70
Reference Patch: https://github.com/facebook/mysql-5.6/commit/be8c587
Reference Patch: https://github.com/facebook/mysql-5.6/commit/22b2508

We need some extra info for the shadowing and security logging. This is a
simple first step of info that MariaDB actually also exposes.
Now we would have the `query_id` and the database name for general events.
Making as few changes as possible to accomplish it, so I'm just taking the
information from the TDH and exposing it through `mysql_event_general`
struct and as a argument to disconnect.

Forward the connection certificate to the audit plugin. The connection certificate can then be parsed by the audit plugin and handled appropriately. It made more sense for the certificate to live in the connection events, since they generally don't change between every general event, so the move was done.

This is done by caching a BUF_MEM struct on the THD object. Since it's not possible to change certificates on the same connection, this caching should be correct. The BUF_MEM is released on THD::release_resources.

If upstream bumps the MYSQL_AUDIT_INTERFACE_VERSION, we should bump ours to be greater or equal to it.

Expose the port current mysqld is running on for the audit plugin. If no port, 0 is used.

Test Plan:
This commit also introduces changes to the audit_null plugin, which makes
the fields, and further similar changes testable.
The audit_null plugin now has the "extended_log" variable, which can be
turned on. When it's ON, the plugin logs the last generic event log
into the Audit_null_generic_event_response system variable.
This can be easily verified in MTR tests, which is done in the new
audit_null.event_params testcase.

### DIFF
--- a/mysql-test/suite/audit_null/r/event_params.result
+++ b/mysql-test/suite/audit_null/r/event_params.result
@@ -1,0 +1,33 @@
+INSTALL PLUGIN null_audit SONAME 'adt_null.so';
+SET @@null_audit_extended_log = 1;
+CREATE TABLE foo (v INT);
+DROP TABLE foo;
+SHOW STATUS LIKE "Audit_null_generic_event_response";
+Variable_name	Value
+Audit_null_generic_event_response	database:test
+CREATE TABLE foo (v INT);
+SHOW STATUS LIKE "Audit_null_generic_event_response";
+Variable_name	Value
+Audit_null_generic_event_response	affected_rows:0
+INSERT INTO foo VALUES (1), (2);
+SHOW STATUS LIKE "Audit_null_generic_event_response";
+Variable_name	Value
+Audit_null_generic_event_response	affected_rows:2
+SELECT * FROM foo;
+v
+1
+2
+SHOW STATUS LIKE "Audit_null_generic_event_response";
+Variable_name	Value
+Audit_null_generic_event_response	affected_rows:-1
+DELETE FROM foo;
+SHOW STATUS LIKE "Audit_null_generic_event_response";
+Variable_name	Value
+Audit_null_generic_event_response	affected_rows:2
+DROP TABLE foo;
+SHOW STATUS LIKE "Audit_null_generic_event_response";
+Variable_name	Value
+Audit_null_generic_event_response	port:MYSQLD_PORT
+UNINSTALL PLUGIN null_audit;
+Warnings:
+Warning	1620	Plugin is busy and will be uninstalled on shutdown
--- a/mysql-test/suite/audit_null/t/event_params.test
+++ b/mysql-test/suite/audit_null/t/event_params.test
@@ -1,0 +1,44 @@
+--source include/have_null_audit_plugin.inc
+--source include/count_sessions.inc
+--source include/have_debug.inc
+
+eval INSTALL PLUGIN null_audit SONAME '$AUDIT_NULL';
+
+SET @@null_audit_extended_log = 1;
+
+## database name in generic event
+CREATE TABLE foo (v INT);
+DROP TABLE foo;
+
+--replace_regex /.*(database:[^;]*).*/\1/
+SHOW STATUS LIKE "Audit_null_generic_event_response";
+
+## affected_rows in generic event
+CREATE TABLE foo (v INT);
+--replace_regex /.*(affected_rows:[^;]*).*/\1/
+SHOW STATUS LIKE "Audit_null_generic_event_response";
+
+INSERT INTO foo VALUES (1), (2);
+--replace_regex /.*(affected_rows:[^;]*).*/\1/
+SHOW STATUS LIKE "Audit_null_generic_event_response";
+
+SELECT * FROM foo;
+--replace_regex /.*(affected_rows:[^;]*).*/\1/
+SHOW STATUS LIKE "Audit_null_generic_event_response";
+
+DELETE FROM foo;
+--replace_regex /.*(affected_rows:[^;]*).*/\1/
+SHOW STATUS LIKE "Audit_null_generic_event_response";
+
+DROP TABLE foo;
+
+## port
+let $MYSQLD_PORT= `SELECT @@port`;
+--replace_result $MYSQLD_PORT MYSQLD_PORT
+--replace_regex /.*(port:[^;]*).*/\1/
+SHOW STATUS LIKE "Audit_null_generic_event_response";
+
+
+UNINSTALL PLUGIN null_audit;
+
+--source include/wait_until_count_sessions.inc
--- a/mysql-test/suite/audit_null/t/event_params_cert.test
+++ b/mysql-test/suite/audit_null/t/event_params_cert.test
@@ -1,0 +1,27 @@
+--source include/have_ssl.inc
+--source include/count_sessions.inc
+--source include/have_debug.inc
+
+eval INSTALL PLUGIN null_audit SONAME '$AUDIT_NULL';
+
+CREATE USER cert_auth@localhost REQUIRE X509;
+GRANT SELECT ON test.* TO cert_auth@localhost;
+CREATE TABLE foo (i INT);
+FLUSH PRIVILEGES;
+connect(con1,localhost,cert_auth,,,,,SSL);
+
+SET @@null_audit_extended_log = 1;
+
+SELECT * FROM foo;
+
+--replace_regex /.*(connection_certificate:[^;]*).*/\1/
+SHOW STATUS LIKE "Audit_null_generic_event_response";
+
+disconnect con1;
+connection default;
+DROP USER cert_auth@localhost;
+DROP TABLE foo;
+
+UNINSTALL PLUGIN null_audit;
+
+--source include/wait_until_count_sessions.inc
--- a/plugin/audit_null/audit_null.cc
+++ b/plugin/audit_null/audit_null.cc
@@ -25,6 +25,8 @@
 #include <mysqld_error.h>
 #include <stdio.h>
 #include <sys/types.h>
+#include <sstream>
+#include <boost/algorithm/string/replace.hpp>
 
 #include "lex_string.h"
 #include "m_ctype.h"
@@ -130,6 +132,13 @@ static char *g_record_buffer;
 
 #undef AUDIT_NULL_VAR
 
+#ifndef DBUG_OFF
+static const constexpr size_t event_response_buffer_len = 1000;
+static char generic_event_response[event_response_buffer_len + 1] = {
+    0,
+};
+#endif
+
 /*
   Plugin status variables for SHOW STATUS
 */
@@ -143,6 +152,11 @@ static SHOW_VAR simple_status[] = {
 #include "plugin/audit_null/audit_null_variables.h"
 
 #undef AUDIT_NULL_VAR
+
+#ifndef DBUG_OFF
+    {"Audit_null_generic_event_response", (char *)generic_event_response,
+     SHOW_CHAR, SHOW_SCOPE_GLOBAL},
+#endif
 
     {0, 0, SHOW_UNDEF, SHOW_SCOPE_GLOBAL}};
 
@@ -174,6 +188,12 @@ static MYSQL_THDVAR_INT(event_order_started, PLUGIN_VAR_RQCMDARG,
 static MYSQL_THDVAR_INT(event_order_check_exact, PLUGIN_VAR_RQCMDARG,
                         "Plugin checks exact event order.", NULL, NULL, 1, 0, 1,
                         0);
+
+#ifndef DBUG_OFF
+static MYSQL_THDVAR_INT(extended_log, PLUGIN_VAR_RQCMDARG,
+                        "Provide extended debug information with audit_null.",
+                        NULL, NULL, 1, 0, 1, 0);
+#endif
 
 static MYSQL_THDVAR_STR(event_record_def,
                         PLUGIN_VAR_RQCMDARG | PLUGIN_VAR_MEMALLOC,
@@ -411,6 +431,45 @@ static int process_command(MYSQL_THD thd, LEX_CSTRING event_command,
   return 0;
 }
 
+#ifndef DBUG_OFF
+/*
+ * Exposes a generic audit log event in a status variable
+ */
+static void log_event(const mysql_event_general *event) {
+#define EVENT_PARAM(name) event_str << #name ":" << event->name << ";";
+#define EVENT_PARAM_STR(name)                             \
+  {                                                       \
+    std::string tmp(event->name.str, event->name.length); \
+    boost::replace_all(tmp, "\n", "\\n");                 \
+    event_str << #name ":" << tmp << ";";                 \
+  }
+  std::stringstream event_str;
+  EVENT_PARAM(event_subclass);
+  EVENT_PARAM(general_error_code);
+  // skipping general_thread_id
+  EVENT_PARAM_STR(general_user);
+  EVENT_PARAM_STR(general_command);
+  EVENT_PARAM_STR(general_query);
+  // EVENT_PARAM_STR(general_charset);
+  EVENT_PARAM(general_time);
+  EVENT_PARAM(general_rows);
+  EVENT_PARAM_STR(general_host);
+  EVENT_PARAM_STR(general_sql_command);
+  EVENT_PARAM_STR(general_external_user);
+  EVENT_PARAM_STR(general_ip);
+  EVENT_PARAM(query_id);
+  EVENT_PARAM_STR(database);
+  EVENT_PARAM(affected_rows);
+  EVENT_PARAM(port);
+  EVENT_PARAM_STR(connection_certificate);
+#undef EVENT_PARAM
+#undef EVENT_PARAM_STR
+
+  const std::string str = event_str.str();
+  strncpy(generic_event_response, str.c_str(), event_response_buffer_len);
+}
+#endif
+
 /**
   @brief Plugin function handler.
 
@@ -454,9 +513,16 @@ static int audit_null_notify(MYSQL_THD thd, mysql_event_class_t event_class,
       case MYSQL_AUDIT_GENERAL_RESULT:
         number_of_calls_general_result++;
         break;
-      case MYSQL_AUDIT_GENERAL_STATUS:
+      case MYSQL_AUDIT_GENERAL_STATUS: {
+#ifndef DBUG_OFF
+        const int extended_info = static_cast<int>(THDVAR(thd, extended_log));
+        if (extended_info != 0) {
+          log_event(static_cast<const mysql_event_general *>(event));
+        }
+#endif
         number_of_calls_general_status++;
         break;
+      }
       default:
         break;
     }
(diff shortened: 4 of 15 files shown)


plugin/audit_null/audit_null.cc
```diff
@@ -455,6 +501,9 @@ static int audit_null_notify(MYSQL_THD thd, mysql_event_class_t event_class,
         number_of_calls_general_result++;
         break;
       case MYSQL_AUDIT_GENERAL_STATUS:
+        if (extended_info) {
```
percona-ysorokin: move `extended_info` declaration here (inside the `case` block).

plugin/audit_null/audit_null.cc
```diff
@@ -175,6 +183,10 @@ static MYSQL_THDVAR_INT(event_order_check_exact, PLUGIN_VAR_RQCMDARG,
                         "Plugin checks exact event order.", NULL, NULL, 1, 0, 1,
                         0);
 
+static MYSQL_THDVAR_INT(extended_log, PLUGIN_VAR_RQCMDARG,
```
percona-ysorokin, line 193: It's better to declare this var as `GLOBAL`-only (`MYSQL_SYSVAR_INT`) since we are using a single global buffer for the output anyway.

plugin/audit_null/audit_null.cc
```diff
@@ -431,6 +476,7 @@ static int audit_null_notify(MYSQL_THD thd, mysql_event_class_t event_class,
   const char *order_str = (const char *)THDVAR(thd, event_order_check);
   int event_order_started = (int)THDVAR(thd, event_order_started);
   int exact_check = (int)THDVAR(thd, event_order_check_exact);
+  int extended_info = (int)THDVAR(thd, extended_log);
```
percona-ysorokin: `const auto extended_info = static_cast<int>(THDVAR(thd, extended_log));`

plugin/audit_null/audit_null.cc
```diff
@@ -411,6 +423,39 @@ static int process_command(MYSQL_THD thd, LEX_CSTRING event_command,
...
+  EVENT_PARAM_STR(general_external_user);
+  EVENT_PARAM_STR(general_ip);
+  EVENT_PARAM(query_id);
+  EVENT_PARAM_STR(database);
+  EVENT_PARAM(affected_rows);
+  EVENT_PARAM(port);
+#undef EVENT_PARAM
+#undef EVENT_PARAM_STR
+
+  std::string str = event_str.str();
```
percona-ysorokin: `const auto str = ...`

plugin/audit_null/audit_null.cc
percona-ysorokin: `if (extended_info !=0)`

mysql-test/suite/audit_null/t/event_params.test
```diff
@@ -0,0 +1,40 @@
...
+eval INSTALL PLUGIN null_audit SONAME '$AUDIT_NULL';
+
+SET @@null_audit_extended_log = 1;
+
+## database name in generic event
+CREATE TABLE foo (v INT);
+DROP TABLE foo;
+
+--replace_regex /.*(database:[^;]*).*/\1/
+SHOW STATUS LIKE "Audit_null_generic_event_response";
```
percona-ysorokin, line 14: Try to use something like
```sql
SET @param_name = 'database:';
SELECT SUBSTRING(REGEX_SUBSTR(variable_value, CONCAT('.*(', @param_name, '[^;]*).*' )), LENGTH(@param_name)) INTO @extracted_param
FROM performance_schema.global_status
WHERE variable_name='Audit_null_generic_event_response';
```
Then use `--source include/assert.inc` to check `DATABASE() == @extracted_param`

mysql-test/suite/audit_null/t/event_params.test
```diff
@@ -0,0 +1,40 @@
...
+
+## affected_rows in generic event
+CREATE TABLE foo (v INT);
+--replace_regex /.*(affected_rows:[^;]*).*/\1/
+SHOW STATUS LIKE "Audit_null_generic_event_response";
```
percona-ysorokin, line 19: Likewise

mysql-test/suite/audit_null/t/event_params.test
```diff
@@ -0,0 +1,40 @@
...
+
+INSERT INTO foo VALUES (1), (2);
+--replace_regex /.*(affected_rows:[^;]*).*/\1/
+SHOW STATUS LIKE "Audit_null_generic_event_response";
```
percona-ysorokin, line 23: Likewise

mysql-test/suite/audit_null/t/event_params.test
```diff
@@ -0,0 +1,40 @@
...
+
+SELECT * FROM foo;
+--replace_regex /.*(affected_rows:[^;]*).*/\1/
+SHOW STATUS LIKE "Audit_null_generic_event_response";
```
percona-ysorokin, line 27: Likewise

mysql-test/suite/audit_null/t/event_params.test
```diff
@@ -0,0 +1,40 @@
...
+
+DELETE FROM foo;
+--replace_regex /.*(affected_rows:[^;]*).*/\1/
+SHOW STATUS LIKE "Audit_null_generic_event_response";
```
percona-ysorokin, line 31: Likewise

mysql-test/suite/audit_null/t/event_params.test
```diff
@@ -0,0 +1,40 @@
...
+
+DROP TABLE foo;
+
+## port
+let $MYSQLD_PORT= `SELECT @@port`;
+--replace_result $MYSQLD_PORT MYSQLD_PORT
+--replace_regex /.*(port:[^;]*).*/\1/
+SHOW STATUS LIKE "Audit_null_generic_event_response";
```
percona-ysorokin, line 39: Likewise

mysql-test/suite/audit_null/r/event_params.result
```diff
@@ -0,0 +1,33 @@
...
+SHOW STATUS LIKE "Audit_null_generic_event_response";
+Variable_name	Value
+Audit_null_generic_event_response	affected_rows:2
+DROP TABLE foo;
+SHOW STATUS LIKE "Audit_null_generic_event_response";
+Variable_name	Value
+Audit_null_generic_event_response	port:MYSQLD_PORT
+UNINSTALL PLUGIN null_audit;
+Warnings:
+Warning	1620	Plugin is busy and will be uninstalled on shutdown
```
percona-ysorokin, line 33: Make sure the test works with `--repeat=2`

percona-ysorokin (changes requested):

percona-ysorokin:
Also, please, add Jira ticket references

mysql-test/suite/audit_null/t/event_params.test
dutow: I'm not sure if that's better - with this approach, the query is easier to read, and the value is there in the result file. The test also won't stop at the first error, and in the end, it will display all issues, not only the first.

plugin/audit_null/audit_null.cc
dutow: But doesn't that make thread level control useful? If it's a threadvar, you can turn it on for only one thread, and other threads won't accidentally modify the status variables, breaking tests.

laurynas-biveinis: Such change is too visible for user to be discussed only here (unless FB chimes in)

mysql-test/suite/audit_null/t/event_params.test
percona-ysorokin: The only reason for this complexity is to have everything related to the assertion in one place (in the .test file) and not split between .test and .result. From my experience, .result files are never reviewed properly and developers very often do './mtr --record' without checking if the output is still expected.
This is just a suggestion - leaving this up to you.

mysql-test/suite/audit_null/r/event_params.result
dutow: I checked, it works.

dutow:
Updated: alsoadded FB8-55.

mysql-test/suite/audit_null/t/event_params_cert.test
```diff
@@ -0,0 +1,24 @@
+--source include/have_ssl.inc
+
```
percona-ysorokin, line 4: Add `--source include/count_sessions.inc` as you are establishing/closing a new connection.

mysql-test/suite/audit_null/t/event_params_cert.test
```diff
@@ -0,0 +1,24 @@
...
+--replace_regex /.*(connection_certificate:[^;]*).*/\1/
+SHOW STATUS LIKE "Audit_null_generic_event_response";
+
+disconnect con1;
+connection default;
+DROP USER cert_auth@localhost;
+DROP TABLE foo;
+
+UNINSTALL PLUGIN null_audit;
+
```
percona-ysorokin, line 26: `--source include/wait_until_count_sessions.inc`

plugin/audit_null/audit_null.cc
```diff
@@ -411,6 +424,44 @@ static int process_command(MYSQL_THD thd, LEX_CSTRING event_command,
...
+/*
+ * Exposes a generic audit log event in a status variable
+ */
+static void log_event(const mysql_event_general *event) {
+#define EVENT_PARAM(name) event_str << #name ":" << event->name << ";";
+#define EVENT_PARAM_STR(name)                                                  \
+  {                                                                            \
+    std::string tmp(event->name.str, event->name.length);                      \
+    boost::replace_all(tmp, "\n", "\\n");                                      \
+    event_str << #name ":" << std::string(event->name.str, event->name.length) \
```
percona-ysorokin: Should this be
```c++
event_str << #name ":" << tmp
```
?

sql/sql_class.cc
```diff
@@ -893,6 +895,15 @@ void THD::cleanup_connection(void) {
 #endif
 }
 
+void THD::set_connection_certificate(std::string const &cert) {
+  DBUG_ASSERT(m_connection_certificate == "");
```
percona-ysorokin:
```c++
DBUG_ASSERT(m_connection_certificate.empty());
```

sql/sql_class.h
```diff
@@ -3323,6 +3323,13 @@ class THD : public MDL_context_owner,
   Gtid_set owned_gtid_set;
 #endif
 
+#ifdef HAVE_OPENSSL
+  std::string m_connection_certificate;
+#endif
+
+  std::string const &connection_certificate() const;
```
percona-ysorokin: `noexcept`

sql/sql_class.h
```diff
@@ -3323,6 +3323,13 @@ class THD : public MDL_context_owner,
   Gtid_set owned_gtid_set;
 #endif
 
+#ifdef HAVE_OPENSSL
```
percona-ysorokin: Should `connection_certificate()` / `set_connection_certificate()` also be wrapped with `#ifdef HAVE_OPENSSL`?
Because otherwise, this does not seem to compile without `HAVE_OPENSSL` defined

sql/sql_class.h
```diff
@@ -4032,6 +4039,7 @@ class THD : public MDL_context_owner,
     m_persist_variables_init = is_init;
   }
   bool is_persist_variables_init() { return m_persist_variables_init; }
+
```
percona-ysorokin, line 4040: unnecessary change

percona-ysorokin (changes requested):
Please also add full Jira tickets URLs to the commit message
```
https://jira.percona.com/browse/FB8-54
https://jira.percona.com/browse/FB8-55
https://jira.percona.com/browse/FB8-70
https://jira.percona.com/browse/FB8-101

File: plugin/audit_null/audit_null.cc [hunk 19]
```diff
@@ -130,6 +131,11 @@ static char *g_record_buffer;
 
 #undef AUDIT_NULL_VAR
 
+static const constexpr size_t event_response_buffer_len = 1000;
```
percona-ysorokin, line 136: Is 1000 enough now when we have certificates also included in GENERAL STATUS event?

dutow: For now yes, as all other fields are short.

sql/sql_class.h
dutow: I think this was clang-format, but I'll recheck what happens if I remove it.

sql/sql_class.h
dutow: That's a trivial setter / getter, works without SSL.

sql/sql_class.h
dutow: Technically, the std::string copy constructor could throw a bad_alloc.

sql/sql_class.h
dutow: Yes, it was indeed clang-format.

sql/sql_class.h
percona-ysorokin: No, no copy constructor here, the result is returned by const reference
`std::string const &`

sql/sql_class.h
percona-ysorokin: Actually, my main concern here was that when `HAVE_OPENSSL` is not defined, the class won't have `m_connection_certificate` member and therefore `connection_certificate() / set_connection_certificate()` will fail to compile as `m_connection_certificate = cert;` / `return m_connection_certificate;` will be an invalid code.

mysql-test/suite/audit_null/t/event_params_cert.test
percona-ysorokin: Still no `--source include/wait_until_count_sessions.inc`

sql/sql_class.h
dutow: Uh, right, I completely missed the & aligned to the function name, fixed.

sql/sql_class.h
dutow: Fixed.

mysql-test/suite/audit_null/t/event_params_cert.test
dutow: Fixed.

percona-ysorokin (approved):
LGTM

plugin/audit_null/audit_null.cc
hermanlee: Can this be debug only since it's mainly used for testing? Are there security issues with non-privileged users getting access to see the last logged query.

sql/sql_class.cc
```diff
@@ -863,6 +863,8 @@ void THD::cleanup_connection(void) {
   sp_cache_clear(&sp_proc_cache);
   sp_cache_clear(&sp_func_cache);
 
+  m_connection_certificate = "";
```
hermanlee, line 866: Does this need to be wrapped in #ifdef HAVE_OPENSSL?

sql/sql_class.cc
```diff
@@ -1058,6 +1071,8 @@ void THD::release_resources() {
 
   if (current_thd == this) restore_globals();
 
+  m_connection_certificate = "";
```
hermanlee, line 1072: Same, does this need to be wrapped in #ifdef OPENSSL?

hermanlee (changes requested):

plugin/audit_null/audit_null.cc
```diff
@@ -411,6 +424,43 @@ static int process_command(MYSQL_THD thd, LEX_CSTRING event_command,
...
+ */
+static void log_event(const mysql_event_general *event) {
+#define EVENT_PARAM(name) event_str << #name ":" << event->name << ";";
+#define EVENT_PARAM_STR(name)                             \
+  {                                                       \
+    std::string tmp(event->name.str, event->name.length); \
+    boost::replace_all(tmp, "\n", "\\n");                 \
+    event_str << #name ":" << tmp << ";";                 \
+  }
+  std::stringstream event_str;
```
hermanlee, line 446: I'm also seeing an internal error compiling this:

../../../plugin/audit_null/audit_null.cc: In function ‘void log_event(const mysql_event_general*)’:
../../../plugin/audit_null/audit_null.cc:438:21: error: aggregate ‘std::stringstream event_str’ has incomplete type and cannot be defined
   std::stringstream event_str;

sql/sql_class.cc
laurynas-biveinis: Probably the opposite: any existing `HAVE_OPENSSL` uses should be removed from patches. E.g. Now CMakeLists.txt contains
`ADD_DEFINITIONS(-DHAVE_OPENSSL)            # TODO: remove #ifdef from C++ code`

I guess it is assumed that WolfSSL is sufficiently similar to OpenSSL, and any code specific for it can be guarded with HAVE_WOLFSSL.

plugin/audit_null/audit_null.cc
dutow: And is audit_null used outside MTR testing, in production? This is only present in that plugin, that's why I didn't see it as an issue. I can make it debug only, but that will also make all related tests debug only.

hermanlee: It's not used in production, but if it's used as a template for other audit plugins, it might run the risk of it getting out. Let's make it debug-only and having the tests be debug-only would be fine.

plugin/audit_null/audit_null.cc
```diff
@@ -130,6 +132,11 @@ static char *g_record_buffer;
 
 #undef AUDIT_NULL_VAR
 
+static const constexpr size_t event_response_buffer_len = 1000;
```
percona-ysorokin, line 136: Should `event_response_buffer_len` and `generic_event_response` definitions be also wrapped in `#ifndef DBUG_OFF`?

plugin/audit_null/audit_null.cc
```diff
@@ -144,6 +151,9 @@ static SHOW_VAR simple_status[] = {
 
 #undef AUDIT_NULL_VAR
 
+    {"Audit_null_generic_event_response", (char *)generic_event_response,
```
percona-ysorokin, line 157: Should this also be wrapped in `#ifndef DBUG_OFF`?

plugin/audit_null/audit_null.cc
```diff
@@ -411,6 +427,43 @@ static int process_command(MYSQL_THD thd, LEX_CSTRING event_command,
   return 0;
 }
 
+/*
+ * Exposes a generic audit log event in a status variable
+ */
+static void log_event(const mysql_event_general *event) {
```
percona-ysorokin, line 438: Should this function also be wrapped in `#ifndef DBUG_OFF`?

percona-ysorokin (changes requested):

hermanlee (approved):

PR #935, opened 2019-01-22
FB8-69: Don't force flush relay log info

inikep:
Summary:

With gtid enabled, flushing relay log info is not necessary. The force flushing is required to ensure recovery of MTS in non-gtid mode. The coordinator thread relies on positions in the relay log info repository during recovery. With GTID, this is not required and sql_thread doesn't even need to do any recovery. The slave (io_thread) uses auto-positioning directly.

JIRA: https://jira.percona.com/browse/FB8-69

Reference Patch: https://github.com/facebook/mysql-5.6/commit/ed39d49

### DIFF
--- a/sql/rpl_slave.cc
+++ b/sql/rpl_slave.cc
@@ -6151,7 +6151,8 @@ bool mts_checkpoint_routine(Relay_log_info *rli, bool force) {
          waiter: set wait_flag; waits....; drops wait_flag;
   */
 
-  error = rli->flush_info(true);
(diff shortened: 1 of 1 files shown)


percona-ysorokin (approved):
LGTM

Consider creating a separate task which would add an MTR test case for this scenario.

inikep:
The ticket for an MTR test creation is at https://jira.percona.com/browse/FB8-215

hermanlee (approved):

PR #938, opened 2019-01-24
FB8-100: Handle out-of-space on binlog cache file

inikep:
JIRA: https://jira.percona.com/browse/FB8-100

Reference Patch: https://github.com/facebook/mysql-5.6/commit/e29ff14

Summary:
The binlog cache file code is designed to handle an error that would cause binlog corruption.  It does this by checking the error for a set of values that it expects and when it finds one it marks the cache to indicate an incident has occurred and this mark keeps it from writing corrupt entries to the binlog.  Unfortunately, during an out-of-space event, there is a lower level error that gets set which is not being overwritten when we set the expected error.  This means that when the code checks for the expected error it is not found - the original lower level error is found.  This diff modifies the checking code to scan through all associated conditions (which is a list of all encountered errors) to see if the expected error is present anywhere in the list.

This diff can be reverted if/when https://bugs.mysql.com/bug.php?id=72457 is resolved upstream.

### DIFF
--- a/sql/binlog.cc
+++ b/sql/binlog.cc
@@ -5013,14 +5013,20 @@ bool MYSQL_BIN_LOG::check_write_error(THD *thd) {
 
   if (!thd->is_error()) DBUG_RETURN(checked);
 
-  switch (thd->get_stmt_da()->mysql_errno()) {
-    case ER_TRANS_CACHE_FULL:
-    case ER_STMT_CACHE_FULL:
-    case ER_ERROR_ON_WRITE:
-    case ER_BINLOG_LOGGING_IMPOSSIBLE:
-      checked = true;
-      break;
+  // Check all conditions for one that matches the expected error
+  const Sql_condition *err;
+  auto it = thd->get_stmt_da()->sql_conditions();
+  while ((err = it++) != nullptr && !checked) {
+    switch (err->mysql_errno()) {
+      case ER_TRANS_CACHE_FULL:
+      case ER_STMT_CACHE_FULL:
+      case ER_ERROR_ON_WRITE:
+      case ER_BINLOG_LOGGING_IMPOSSIBLE:
+        checked = true;
+        break;
+    }
   }
+
   DBUG_PRINT("return", ("checked: %s", YESNO(checked)));
   DBUG_RETURN(checked);
 }


sql/binlog.cc
```diff
@@ -5013,14 +5013,20 @@ bool MYSQL_BIN_LOG::check_write_error(THD *thd) {
...
-  switch (thd->get_stmt_da()->mysql_errno()) {
-    case ER_TRANS_CACHE_FULL:
-    case ER_STMT_CACHE_FULL:
-    case ER_ERROR_ON_WRITE:
-    case ER_BINLOG_LOGGING_IMPOSSIBLE:
-      checked = true;
-      break;
+  // Check all conditions for one that matches the expected error
+  const Sql_condition *err;
+  auto it= thd->get_stmt_da()->sql_conditions();
```
laurynas-biveinis: clang-format

laurynas-biveinis (changes requested):

sql/binlog.cc
inikep: My changes should be now clang-formatted but the rest of file is not.
There is a task for this issue at https://jira.percona.com/browse/FB8-214

sql/binlog.cc
```diff
@@ -5013,14 +5013,20 @@ bool MYSQL_BIN_LOG::check_write_error(THD *thd) {
...
   if (!thd->is_error()) DBUG_RETURN(checked);
 
-  switch (thd->get_stmt_da()->mysql_errno()) {
-    case ER_TRANS_CACHE_FULL:
-    case ER_STMT_CACHE_FULL:
-    case ER_ERROR_ON_WRITE:
-    case ER_BINLOG_LOGGING_IMPOSSIBLE:
-      checked = true;
-      break;
+  // Check all conditions for one that matches the expected error
```
percona-ysorokin, line 5016: This is a known upstream bug
https://bugs.mysql.com/bug.php?id=88223 / https://bugs.mysql.com/bug.php?id=72457
that we fixed in Percona Server
https://jira.percona.com/browse/PS-1107
https://github.com/percona/percona-server/pull/2574 (Pull Request for 8.0)
Please, consider taking MTR test cases from there.

inikep: I created a JIRA ticket for importing MTR test cases at https://jira.percona.com/browse/FB8-217

sql/binlog.cc
```diff
@@ -5013,14 +5013,20 @@ bool MYSQL_BIN_LOG::check_write_error(THD *thd) {
...
+  const Sql_condition *err;
+  auto it = thd->get_stmt_da()->sql_conditions();
+  while ((err = it++) != nullptr && !checked) {
```
percona-ysorokin, line 5019: Percona's fix also checks `thd->get_stmt_da()->mysql_errno()` in addition to iterating through `thd->get_stmt_da()->sql_conditions()`. Could you confirm that checking only `thd->get_stmt_da()->sql_conditions()` is enough.

inikep: It looks that checking only `thd->get_stmt_da()->sql_conditions()` should be enough:
1. `m_mysql_errno` is changed only in `set_error_status()` and `reset_diagnostics_area()`.
In both cases `m_status` is also changed (to `DA_ERROR` and `DA_EMPTY` respectively)
2. `check_write_error()` checks if 
`if (!thd->is_error()) DBUG_RETURN(checked);`
and we have
```
bool is_error() const { return m_status == DA_ERROR; }
```

percona-ysorokin (approved):
LGTM

hermanlee (approved):

PR #939, opened 2019-01-24
FB8-86: Delete crash safe index file if index file exists

inikep:
JIRA: https://jira.percona.com/browse/FB8-86

Reference Patch: https://github.com/facebook/mysql-5.6/commit/78fa270

Summary:
Crash safe index file is created when any change is made to the index
file. The index file is first copied over to the crash safe one and then stuff
is added/removed from the crash safe index file. Finally, the old index file is
deleted and crash safe index file is renamed to the index file. There might be a
scenario where mysql crashes after stuff is changed in the crash safe file
but before the old index file is deleted. In this case we should delete the
crash safe index file at restart. If we don't delete it subsequent changes to
the index file can corrupt it because it can be copied to the old crash safe
file which already has some data in it.

### DIFF
--- a/sql/binlog.cc
+++ b/sql/binlog.cc
@@ -3599,18 +3599,30 @@ bool MYSQL_BIN_LOG::open_index_file(const char *index_file_name_arg,
     goto end;
   }
 
-  /*
-    We need move crash_safe_index_file to index_file if the index_file
-    does not exist and crash_safe_index_file exists when mysqld server
-    restarts.
-  */
-  if (my_access(index_file_name, F_OK) &&
-      !my_access(crash_safe_index_file_name, F_OK) &&
-      my_rename(crash_safe_index_file_name, index_file_name, MYF(MY_WME))) {
-    LogErr(ERROR_LEVEL, ER_BINLOG_CANT_MOVE_TMP_TO_INDEX,
-           "MYSQL_BIN_LOG::open_index_file");
-    error = true;
-    goto end;
+  // case: crash_safe_index_file exists
+  if (!my_access(crash_safe_index_file_name, F_OK)) {
+    /*
+      We need to move crash_safe_index_file to index_file if the index_file
+      does not exist or delete it if the index_file exists when mysqld server
+      restarts.
+    */
+
+    // case: index_file does not exist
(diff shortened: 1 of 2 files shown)


sql/binlog.cc
```diff
@@ -3599,18 +3599,31 @@ bool MYSQL_BIN_LOG::open_index_file(const char *index_file_name_arg,
...
-      !my_access(crash_safe_index_file_name, F_OK) &&
-      my_rename(crash_safe_index_file_name, index_file_name, MYF(MY_WME))) {
-    LogErr(ERROR_LEVEL, ER_BINLOG_CANT_MOVE_TMP_TO_INDEX,
-           "MYSQL_BIN_LOG::open_index_file");
-    error = true;
-    goto end;
+  // case: crash_safe_index_file exists
+  if (!my_access(crash_safe_index_file_name, F_OK)) {
+    /*
+      We need move crash_safe_index_file to index_file if the index_file
```
percona-ysorokin: we need **to** move

sql/binlog.cc
```diff
@@ -3599,18 +3599,31 @@ bool MYSQL_BIN_LOG::open_index_file(const char *index_file_name_arg,
...
+      does not exist or delete it if the index_file exists when mysqld server
+      restarts.
+    */
+
+    // case: index_file does not exist
+    if (my_access(index_file_name, F_OK)) {
+      if (my_rename(crash_safe_index_file_name, index_file_name, MYF(MY_WME))) {
+        LogErr(ERROR_LEVEL, ER_BINLOG_CANT_MOVE_TMP_TO_INDEX,
```
percona-ysorokin, line 3613: Why different log functions (`LogErr` vs `sql_print_error`) for different branches?

inikep: There is no proper error for delete crash_safe_index_file in `share/errmsg-utf8.txt`. Do you want me to add it?

percona-ysorokin: Looks like the right thing to do.

inikep:
PR was updated with requested changes

percona-ysorokin (approved):
LGTM

hermanlee (approved):

PR #944, opened 2019-01-29
FB8-97: Add mtr suppression rule for rpl.rpl_gtid_skip_writing_previous_gtid_set test

inikep:
Summary:

JIRA: https://jira.percona.com/browse/FB8-97

Reference Patch: https://github.com/facebook/mysql-5.6/commit/eaabc64
Reference Patch: https://github.com/facebook/mysql-5.6/commit/62c3073

Add mtr suppression rule for rpl.rpl_gtid_skip_writing_previous_gtid_set test

Adding slave start/stop calls to the
rpl.rpl_gtid_skip_writing_previous_gtid_set test can sometimes trigger a
warning message in the log. Add a new suppression rule for this warning
message since it does not cause the test to fail.

A flaky warning `failed registering on master` may happen in this rpl test,
when the master restarts twice in a row. The slave may still try to connect to
the master while the master restarts again. Adding stop/start slaves when
master restarts.

Test Plan:
mtr.

Also stressed with `--repeat=32 --parallel=32` and passed.

### DIFF
--- a/mysql-test/suite/rpl/r/rpl_gtid_skip_writing_previous_gtid_set.result
+++ b/mysql-test/suite/rpl/r/rpl_gtid_skip_writing_previous_gtid_set.result
@@ -10,8 +10,10 @@ insert into t1 values(1);
 insert into t1 values(2);
 FLUSH LOGS;
 include/assert.inc [gtid set on master.]
+include/stop_slave.inc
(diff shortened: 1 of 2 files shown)


percona-ysorokin (approved):
LGTM

hermanlee (approved):

PR #947, opened 2019-02-04
FB8-49: Don't remove padding chars for binary column in RBR

dutow:
Jira issue: https://jira.percona.com/browse/FB8-49

Reference patch: https://github.com/facebook/mysql-5.6/commit/915b131

MySQL client doesn't remove padding chars for binary columns. Do the same for RBR. Tools parsing the replication stream vs. the client output can then use memcmp to compare keys. Otherwise, the tools need to handle padding characters correctly.


percona-ysorokin (approved):
LGTM

hermanlee (changes requested):
I think there's a number of tests failing, but it's not clear to me if all of them just need to be re-recorded, or if something else is going on.

main.mysqlbinlog_rbr_column_names
binlog_nogtid.binlog_mysqlbinlog_row
binlog_nogtid.binlog_mysqlbinlog_row_myisam
binlog_nogtid.binlog_mysqlbinlog_row_innodb

sql/field.cc
```diff
@@ -6324,8 +6324,10 @@ uchar *Field_string::pack(uchar *to, const uchar *from, uint max_length,
            (this is for not packing padding adding bytes in BINARY
            fields).
   */
-  if (field_charset->mbmaxlen == 1) {
-    while (length && from[length - 1] == field_charset->pad_char) length--;
+  if (key_type() != HA_KEYTYPE_BINARY) {
```
lth, line 6327: I don't think the body of this if statement is correct. It should both the `mbmaxlen == 1` case and the else statement.

You can see that the re-recorded results are actually incorrect, because we're spacing padding varchars. This is because we're now keeping padding characters for charsets with mbmaxlen  > 1 (eg. utf8mb4)

lth:
Can you update 
```
binlog_gtid.binlog_gtid_mysqlbinlog_row_myisam
binlog_gtid.binlog_gtid_mysqlbinlog_row
binlog_gtid.binlog_gtid_mysqlbinlog_row_innodb
```

lth:
https://github.com/facebook/mysql-5.6/commit/939e5261ddc89824e77a7acba9666983b8b93724

PR #951, opened 2019-02-05
FB8-103: Heartbeat event should carry the latest master timestamp

dutow:
Jira issue: https://jira.percona.com/browse/FB8-103

Reference patch: https://github.com/facebook/mysql-5.6/commit/eb9a3db
Reference patch: https://github.com/facebook/mysql-5.6/commit/5803e4f
Reference patch: https://github.com/facebook/mysql-5.6/commit/746c217

Compared to the 5.6 patch:
* fixed the rpl_heartbeat_timestamp testcase, which resulted in the test executor killing all servers because of a connection drop
* split the rpl_heartbeat_zero_timestamp test into two

Heartbeat events should carry the now() timestamp from the master and
the last_master_timstamp from the slave. In other words, HB should always carry
the lastet master timestamp. This will help services connected to slaves know if
they are lagging even when no real events are flowing in the system.

Also, for this to work last_master_timestamp should be updated whenever a HB
event is received.

Let master only send HB timestamp while waiting for new trx, this will fix the SBM being inaccurate when the slave is stopped and trying to catchup from the master.

### DIFF
--- a/mysql-test/suite/rpl/include/rpl_heartbeat_zero_timestamp_input.inc
+++ b/mysql-test/suite/rpl/include/rpl_heartbeat_zero_timestamp_input.inc
@@ -1,0 +1,36 @@
+################################################################################
+# The workload generator on the master for rpl_heartbeat_zero_timestamp
+################################################################################
+
+# Create the databases and tables
+
+--disable_query_log
+--disable_result_log
+
+let $i = $databases;
+
+while ($i)
+{
+  eval create database test$i;
+  eval use test$i;
+  eval create table t$i (a int) engine=InnoDB;
+  dec $i;
+}
+
+# Run the queries
+
+while ($iter)
+{
+  let $i=$databases;
+
+  while ($i)
+  {
+    eval use test$i;
+    eval insert into t$i values ($iter);
+    dec $i;
+  }
+  dec $iter;
+}
+
+--enable_result_log
+--enable_query_log
--- a/mysql-test/suite/rpl/t/rpl_heartbeat_timestamp.cnf
+++ b/mysql-test/suite/rpl/t/rpl_heartbeat_timestamp.cnf
@@ -1,0 +1,7 @@
+!include ../my.cnf
+
+
+[mysqld.3]
+
+[ENV]
+SERVER_MYPORT_3= @mysqld.3.port
--- a/mysql-test/suite/rpl_gtid/t/rpl_heartbeat_zero_timestamp_mts.test
+++ b/mysql-test/suite/rpl_gtid/t/rpl_heartbeat_zero_timestamp_mts.test
@@ -1,0 +1,7 @@
+################################################################################
+# Testing the second behind master is correct with timestamp in HB event
+################################################################################
+
+# Start a new master-slave
+let $parallel_workers=4;
+--source suite/rpl/include/rpl_heartbeat_zero_timestamp.inc
--- a/mysql-test/suite/rpl_gtid/t/rpl_heartbeat_zero_timestamp_nomts.test
+++ b/mysql-test/suite/rpl_gtid/t/rpl_heartbeat_zero_timestamp_nomts.test
@@ -1,0 +1,7 @@
+################################################################################
+# Testing the second behind master is correct with timestamp in HB event
+################################################################################
+
+# Start a new master-slave
+let $parallel_workers=0;
+--source suite/rpl/include/rpl_heartbeat_zero_timestamp.inc
--- a/sql/rpl_binlog_sender.cc
+++ b/sql/rpl_binlog_sender.cc
@@ -528,12 +528,17 @@ int Binlog_sender::send_events(File_reader *reader, my_off_t end_pos) {
       set_timespec_nsec(&cur_clock, 0);
       DBUG_ASSERT(cmp_timespec(&cur_clock, &m_last_event_sent_ts) >= 0);
       // Both diff_timespec() and heartbeat_period are in nano seconds.
-      const bool time_for_hb_event =
-          (diff_timespec(&cur_clock, &m_last_event_sent_ts) >=
-           m_heartbeat_period);
+#ifdef DBUG_OFF
+      const
+#endif
+          bool time_for_hb_event =
+              (diff_timespec(&cur_clock, &m_last_event_sent_ts) >=
+               m_heartbeat_period);
+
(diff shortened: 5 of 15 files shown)


mysql-test/suite/rpl/include/rpl_heartbeat_zero_timestamp_input.inc
```diff
@@ -0,0 +1,38 @@
...
+# Create the databases and tables
+
+--disable_query_log
+--disable_result_log
+
+let $i = $databases;
+
+while ($i)
+{
+  eval drop database if exists test$i;
```
laurynas-biveinis: Redundant

mysql-test/suite/rpl/include/rpl_heartbeat_zero_timestamp_input.inc
```diff
@@ -0,0 +1,38 @@
...
+  eval create database test$i;
+  eval use test$i;
+  eval drop table if exists t$i;
```
laurynas-biveinis: Doubly redundant

mysql-test/suite/rpl/t/rpl_create_drop_temp_table.cnf
```diff
@@ -0,0 +1,11 @@
+!include ../my.cnf
+
+[mysqld.1]
+log-slave-updates
```
laurynas-biveinis: Default in 8.0. The cnf file should only set SERVER_MYPORT_3

mysql-test/suite/rpl/t/rpl_heartbeat_timestamp.cnf
```diff
@@ -0,0 +1,11 @@
+!include ../my.cnf
+
+[mysqld.1]
```
laurynas-biveinis: Default in 8.0. The cnf file should only set SERVER_MYPORT_3

mysql-test/suite/rpl_gtid/t/rpl_heartbeat_zero_timestamp_mts.test
```diff
@@ -0,0 +1,121 @@
+################################################################################
+# Testing the second behind master is correct with timestamp in HB event
+################################################################################
+
+source include/not_valgrind.inc;
+source include/not_parallel.inc;
+
+# Start a new master-slave
+--disable_warnings
```
laurynas-biveinis: Why?

mysql-test/suite/rpl_gtid/t/rpl_heartbeat_zero_timestamp_mts.test
```diff
@@ -0,0 +1,121 @@
...
+--connection slave
+--source include/stop_slave.inc
+set @save.slave_parallel_workers= @@global.slave_parallel_workers;
+SET @@global.slave_parallel_workers= 4;
+CHANGE MASTER TO MASTER_AUTO_POSITION=1;
+--source include/start_slave.inc
+
+# Run a few queries on the master to create skip scneario
+--connection master
+SET GLOBAL DEBUG="+d, send_zero_hb_event";
```
laurynas-biveinis: --source include/have_debug.inc at the top?

mysql-test/suite/rpl_gtid/t/rpl_heartbeat_zero_timestamp_nomts.test
```diff
@@ -0,0 +1,85 @@
+################################################################################
```
laurynas-biveinis, line 1: Same comments as for MTS one.
The testcase looks copy-pasted, consider making an include file

laurynas-biveinis (changes requested):
I don't like timing-dependant MTR style in this patch but not sure how to rewrite it without spending much effort neither (perhaps sync by waiting until the desired number of heartbeats have happened).

sql/rpl_binlog_sender.cc
```diff
@@ -528,12 +528,14 @@ int Binlog_sender::send_events(File_reader *reader, my_off_t end_pos) {
       set_timespec_nsec(&cur_clock, 0);
       DBUG_ASSERT(cmp_timespec(&cur_clock, &m_last_event_sent_ts) >= 0);
       // Both diff_timespec() and heartbeat_period are in nano seconds.
-      const bool time_for_hb_event =
+      bool time_for_hb_event =
```
percona-ysorokin:
```suggestion
#ifdef DBUG_OFF
const
#endif
bool time_for_hb_event = ...
```

sql/rpl_binlog_sender.cc
```diff
@@ -1164,7 +1166,14 @@ int Binlog_sender::send_heartbeat_event(my_off_t log_pos) {
   uchar *header = (uchar *)m_packet.ptr() + event_offset;
 
   /* Timestamp field */
-  int4store(header, 0);
+  // NOTE: if @last_master_timestamp is provided we're a slave and we use this
+  // value in the HB event otherwise we use now()
+  time_t ts = 0;
+  if (send_timestamp) {
+    ts = mysql_bin_log.last_master_timestamp.load();
+    if (!ts) ts = time(0);
```
percona-ysorokin:
```suggestion
    if (ts == 0) ts = time(nullptr);
```

mysql-test/suite/rpl/include/rpl_heartbeat_zero_timestamp_input.inc
dutow: This part isn't redundant: the testcase first creates a few small databases, then drops them, and creates a few large.

laurynas-biveinis: I see, in which case I'd rewrite this with drops after use

dutow:
Updated.

percona-ysorokin (approved):
LGTM

hermanlee (changes requested):
I'm seeing an error running the rpl.rpl_heartbeat_timestamp test. It results in:

```
There is no group named 'mysqld.3' that can be used to resolve 'port' at /home/herman/rocks-mysql/8.0/mysql-test/mysql-test-run.pl line 4359. 
```

I've added the missing lines to mysql-test/suite/rpl/t/rpl_heartbeat_timestamp.cnf:

```
[mysqld.1]
log-slave-updates
[mysqld.2]
log-slave-updates
[mysqld.3]
log-slave-updates
```

and the test looks to start running properly again.

hermanlee (approved):

PR #943, opened 2019-01-29
FB8-77: Changed the behavior of purging logs. Now deleting logs first and then updating the index

inikep:
Summary: 
JIRA: https://jira.percona.com/browse/FB8-77

Reference Patch: https://github.com/facebook/mysql-5.6/commit/832be30
Reference Patch: https://github.com/facebook/mysql-5.6/commit/0d85250
Reference Patch: https://github.com/facebook/mysql-5.6/commit/e8be5be

Changed the behavior of purging logs. Now deleting logs first and then updating the index.

Previously when purging logs, the index was updated before deleting the logs. This creates problems the disk is full. We should first delete logs to free up disk and then update index.

Test Plan: Modified existing test cases to test the change in order of operations.

PR #952, opened 2019-02-06
FB8-85: Copy and move index file only when there are changes

inikep:
Jira ticket: https://jira.percona.com/browse/FB8-85

Reference Patch: https://github.com/facebook/mysql-5.6/commit/4a1823a

Can be squashed with FB8-77 `Changed the behavior of purging logs. Now deleting logs first and then updating the index.`:
https://jira.percona.com/browse/FB8-77
https://github.com/facebook/mysql-5.6/commit/97ae20bd3aa24aee6d3ad87ec7833931c451d887

When index file is updated, it is first copied to a crash safe index
file, updates are made to the crash safe file and finally the crash safe file is
renamed. During purge, all this should happen only when at least one log is
being removed.

Squash with D13930811

### DIFF
--- a/mysql-test/suite/binlog/r/binlog_index.result
+++ b/mysql-test/suite/binlog/r/binlog_index.result
@@ -350,7 +350,6 @@ call mtr.add_suppression("Failed to delete the existing index file");
 call mtr.add_suppression("failed to move crash safe index file to index file");
 call mtr.add_suppression("failed to update the index file");
 PURGE BINARY LOGS TO 'binlog.000014';;
-ERROR HY000: I/O error reading log index file
 # Test if both the index file and binlogs are in sync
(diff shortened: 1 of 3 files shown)


percona-ysorokin (approved):
LGTM

I think this should have been squashed with 
FB8-77 `Changed the behavior of purging logs. Now deleting logs first and then updating the index.`
https://jira.percona.com/browse/FB8-77
https://github.com/facebook/mysql-5.6/pull/943
https://github.com/facebook/mysql-5.6/commit/97ae20bd3aa24aee6d3ad87ec7833931c451d887
which is already merged to the trunk.

May be just leave a comment in the commit message.

hermanlee (approved):

PR #955, opened 2019-02-08
FB8-102 : Add posix_fadvise(FADV_DONTNEED) to flush buffered pages

satya-bodapati:
JIRA : https://jira.percona.com/browse/FB8-102

Referenced patch : https://github.com/facebook/mysql-5.6/commit/ab9d60b

When InnoDB opens a .ibd or log file in O_DIRECT mode, request the os to
flush the buffer cache. Otherwise, direct IO is serialized when buffered
pages for the file existing within the filesystem cache.

Originally Reviewed By: alxyang

### DIFF
--- a/storage/innobase/os/os0file.cc
+++ b/storage/innobase/os/os0file.cc
@@ -5428,6 +5428,13 @@ void os_file_set_nocache(int fd MY_ATTRIBUTE((unused)),
                               << strerror(errno_save) << ", continuing anyway.";
     }
   }
+
+#ifdef POSIX_FADV_DONTNEED
+  /* Request the filesystem to flush cached pages. Otherwise,
+  DIRECTIO requests are serialized. */
+  posix_fadvise(fd, 0, 0, POSIX_FADV_DONTNEED);
+#endif /* POSIX_FADV_DONTNEED  */
+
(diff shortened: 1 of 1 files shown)


storage/innobase/os/os0file.cc
```diff
@@ -97,6 +97,11 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301  USA
 #include <data0type.h>
 #endif /* UNIV_HOTBACKUP */
 
+/* Ignore posix_fadvise() on those platforms where it does not exist */
```
percona-ysorokin: This define is probably redundant. `POSIX_FADV_DONTNEED` will not be defined on Windows even if we cared about this platform.
Moreover, there are two more occurrences of `posix_fadvise()` in the code tree: in `row0log.cc` and `row0merge.cc`, where we do not redefine `posix_fadvise` as a macro.

percona-ysorokin (approved):
LGTM with `#ifdef _WIN32` block removed

storage/innobase/os/os0file.cc
satya-bodapati: Right, we can safely remove this.

percona-ysorokin (approved):
LGTM

hermanlee (approved):

PR #957, opened 2019-02-11
FB8-134: Fix crash during shutdown in audit plugin code path

dutow:
Jira ticket: https://jira.percona.com/browse/FB8-134

Reference patch: https://github.com/facebook/mysql-5.6/commit/0fc059a

------  0fc059a ------

On server shutdown, there exists a race condition that can trigger a segfault if an audit plugin is installed. The root cause because the `THD::audit_class_plugins` array is accessed concurrently without locks.

This is the sequence of actions that can trigger a crash:
1. The shutdown thread calls close_connections on all connected clients.
2. The connection thread will call into the audit plugin because it will receive a network error from `close_connections`, and will log the error into the audit plugin. The thread then adds the plugin to `THD::audit_class_plugins`, but it has not updated `audit_class_mask` yet.
3. The shutdown thread sees that `audit_class_mask` is not set, and also adds the plugin into `THD::audit_class_plugins`, but it has only extended the array, without setting the value of the 2nd element of the array.
4. The connection thread then goes ahead and updates `audit_class_mask` and finishes `acquire_plugins`. It then calls `mysql_audit_notify` and sees the array of size 2, and when it loops over the 2nd element, it segfaults because the value is uninitalized.

The fix is to add a lock per THD that protects these data structures. They should be relatively uncontended since the shutdown thread is the only thread that calls `acquire_plugins` on another THD.

### DIFF
--- a/sql/sql_audit.cc
+++ b/sql/sql_audit.cc
@@ -26,6 +26,7 @@
 
 #include "lex_string.h"
 #include "m_ctype.h"
+#include "mutex_lock.h"  // MUTEX_LOCK
 #include "my_compiler.h"
 #include "my_dbug.h"
 #include "my_inttypes.h"
@@ -971,9 +972,11 @@ int mysql_audit_notify(THD *thd, mysql_event_authentication_subclass_t subclass,
 
   @return false is always returned.
 */
-static bool acquire_lookup_mask(THD *, plugin_ref plugin, void *arg) {
+static bool acquire_lookup_mask(THD *thd, plugin_ref plugin, void *arg) {
   st_mysql_subscribe_event *evt = static_cast<st_mysql_subscribe_event *>(arg);
   st_mysql_audit *audit = plugin_data<st_mysql_audit *>(plugin);
+
+  MUTEX_LOCK(lock, &thd->LOCK_thd_audit_data);
 
   /* Check if this plugin is interested in the event */
   if (!check_audit_mask(audit->class_mask[evt->event_class],
@@ -996,6 +999,8 @@ static bool acquire_lookup_mask(THD *, plugin_ref plugin, void *arg) {
 static bool acquire_plugins(THD *thd, plugin_ref plugin, void *arg) {
   st_mysql_subscribe_event *evt = static_cast<st_mysql_subscribe_event *>(arg);
   st_mysql_audit *data = plugin_data<st_mysql_audit *>(plugin);
+
+  MUTEX_LOCK(lock, &thd->LOCK_thd_audit_data);
 
   /* Check if this plugin is interested in the event */
   if (check_audit_mask(data->class_mask, evt->lookup_mask)) {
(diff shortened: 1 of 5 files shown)


sql/sql_audit.cc
```diff
@@ -1311,6 +1316,13 @@ static int event_class_dispatch(THD *thd, mysql_event_class_t event_class,
                ? 1
                : 0;
   } else {
+    // mutexes are non-reentrant, and audit_null dispatches events from this
+    // function
+    thread_local bool mutex_locked = false;
```
percona-ysorokin: Although, this might work, it looks kinda hackish.
Did you manage to figure out the original source of recursive `event_class_dispatch()` calls?

dutow: It only happens in audit null and mtr: it has a feature that creates (user specified) events on an event dispatch. You assign the evets you want to a variable, and then select the same variable - and on that select event, audit null triggers all the events you specified.

percona-ysorokin: Please consider marking events generated from inside `audit_null` plugin with a special flag (say, `GENERATED_FROM_DISPATCH`) and lock `thd->LOCK_thd_audit_data` depending on the value of this flag.

If this is too much trouble and / or requires significant audit interface changes, then `thread_local` solution is fine.

dutow: I tried a few approaches, but I think this would end up more complex/hackish than this. While this bool isn't the best, it only adds 6 lines of code to the patch, is a localized change, and quite easy to understand.

percona-ysorokin: OK, then it's good as is.

percona-ysorokin (approved):
LGTM

sql/sql_audit.cc
lth: Actually, any audit plugin that calls my_message (or generates errors) will end up recursively calling the dispatch function. I don't think this will interact well if we have different THDs on different physical threads.

Since the `audit_class_plugins` list is append-only, and plugins are never free'd until the THD is done, I think a possible solution is just to copy the list of plugins under the lock, but iterate through the list while unlocked.

lth (changes requested):
Are we missing a mutex lock in the `acquire_plugins` function?

sql/sql_audit.cc
```diff
@@ -1311,11 +1318,16 @@ static int event_class_dispatch(THD *thd, mysql_event_class_t event_class,
                ? 1
                : 0;
   } else {
+	  decltype(thd->audit_class_plugins) plugins_copy{PSI_NOT_INSTRUMENTED};
```
lth: The indenting here looks wrong.

lth (commented):
Do you mind rebasing?

dutow:
Rebased and fixed the formatting.

sql/sql_audit.cc
dutow: I wanted to avoid possible additional allocations for every notification - but I guess if you are using less audit plugins than the preallocated size for the container, that won't be an issue there.

lth (approved):

PR #958, opened 2019-02-11
FB8-76: Add current master host:port to server read only error when available

inikep:
Summary:

Jira ticket: https://jira.percona.com/browse/FB8-76

Reference Patch: https://github.com/facebook/mysql-5.6/commit/270854f
Reference Patch: https://github.com/facebook/mysql-5.6/commit/9f00827
Reference Patch: https://github.com/facebook/mysql-5.6/commit/3e50e46
Reference Patch: https://github.com/facebook/mysql-5.6/commit/9f18725

When a server is in read_only, we append the current master_host and
master_port to the error message, if the server is a slave. The master_host
could be either hostname or ip address, depending on how a slave is connected
to the master.

A new generic error code `ER_OPTION_PREVENTS_STATEMENT_EXTRA_INFO` is added, so
that we can append any additional info to the fixed error message.

Proposing a way to set a custom message in the `--read-only (super)` error.
Note: the variable can only be set when server is in read-only (super) state.

p.s. I am hesitant to add a mutex for the variable, so we will get a snapshot
of the current pointer of the variable (const char*) when printing the error.

Per request, we allow setting read_only_error_msg_extra without being in
read_only state. The custom message can be set/reset at any time.

D61383 added a new error code to allow appending extra info to read_only error
message. The new error code breaks existing dependencies which expects to
receive code 1290. Reverting back to use the old error, and extending the 1290
error message instead of adding a new one. This involves updating many files
(which was the motivation I added a new one in D61383).

Ideally we should have a more elegant solution to send extra info back to
clients, which will be worked on separately. This is to fix the broken
dependencies.

Test Plan:
Updated rpl.rpl_read_only. Master_port is non-deterministic in the test, so I
masked in the test result.

Will update any test that comes up in the sandcastle if needed.

Originally Reviewed By: santoshb

### DIFF
--- a/share/errmsg-utf8.txt
+++ b/share/errmsg-utf8.txt
@@ -5080,12 +5080,12 @@ ER_FEATURE_DISABLED
         spa "El recurso '%s' fue deshabilitado; usted necesita construir MySQL con '%s' para tener eso funcionando"
         swe "'%s' är inte aktiverad; För att aktivera detta måste du bygga om MySQL med '%s' definierad"
 ER_OPTION_PREVENTS_STATEMENT  
-        eng "The MySQL server is running with the %s option so it cannot execute this statement"
-        ger "Der MySQL-Server läuft mit der Option %s und kann diese Anweisung deswegen nicht ausführen"
-        jpn "MySQLサーバーが %s オプションで実行されているので、このステートメントは実行できません。"
-        por "O servidor MySQL está rodando com a opção %s razão pela qual não pode executar esse commando"
-        spa "El servidor MySQL está rodando con la opción %s tal que no puede ejecutar este comando"
-        swe "MySQL är startad med %s. Pga av detta kan du inte använda detta kommando"
+        eng "The MySQL server is running with the %s option so it cannot execute this statement%s"
+        ger "Der MySQL-Server läuft mit der Option %s und kann diese Anweisung deswegen nicht ausführen%s"
+        jpn "MySQLサーバーが %s オプションで実行されているので、このステートメントは実行できません。%s"
+        por "O servidor MySQL está rodando com a opção %s razão pela qual não pode executar esse commando%s"
+        spa "El servidor MySQL está rodando con la opción %s tal que no puede ejecutar este comando%s"
+        swe "MySQL är startad med %s. Pga av detta kan du inte använda detta kommando%s"
 ER_DUPLICATED_VALUE_IN_TYPE  
         eng "Column '%-.100s' has duplicated value '%-.64s' in %s"
         ger "Feld '%-.100s' hat doppelten Wert '%-.64s' in %s"
--- a/sql/rpl_slave.cc
+++ b/sql/rpl_slave.cc
@@ -9749,6 +9749,29 @@ static int check_slave_sql_config_conflict(const Relay_log_info *rli) {
   return 0;
 }
 
+// This function will generate the string containing current master host and
+// port info if available.
+std::string get_active_master_info() {
+  std::string str_ptr;
+
+  channel_map.rdlock();
+  Master_info *mi = channel_map.get_default_channel_mi();
+
+  if (Master_info::is_configured(mi)) {
+    str_ptr = ". Current master_host: ";
+    str_ptr += mi->host;
+    str_ptr += ", master_port: ";
+    str_ptr += std::to_string(mi->port);
+    if (opt_read_only_error_msg_extra && opt_read_only_error_msg_extra[0]) {
+      str_ptr += ". ";
+      str_ptr += opt_read_only_error_msg_extra;
+    }
+  }
+
+  channel_map.unlock();
+  return str_ptr;
+}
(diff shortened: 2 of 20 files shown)


sql/rpl_slave.cc
```diff
@@ -9736,6 +9736,27 @@ static int check_slave_sql_config_conflict(const Relay_log_info *rli) {
   return 0;
 }
 
+// This function will generate the string containing current master host and
+// port info if available.
+void get_active_master_info(std::string &str_ptr) noexcept {
```
laurynas-biveinis: Return std::string

laurynas-biveinis (changes requested):

sql/rpl_slave.cc
inikep: I kept the original solution because `std::string extra_info = get_active_master_info();` would require an additional string copy.

laurynas-biveinis: It should not: RVO

inikep:
PR was updated

sql/rpl_slave.cc
percona-ysorokin: Technically, this should not be `noexcept` as string concatenations may throw.

sql/rpl_slave.cc
```diff
@@ -9749,6 +9749,30 @@ static int check_slave_sql_config_conflict(const Relay_log_info *rli) {
...
+
+  channel_map.rdlock();
+  Master_info *mi = channel_map.get_default_channel_mi();
+
+  if (Master_info::is_configured(mi)) {
+    str_ptr = ". Current master_host: ";
+    str_ptr += mi->host;
+    str_ptr += ", master_port: ";
+    str_ptr += std::to_string(mi->port);
+    const char *extra_str = opt_read_only_error_msg_extra;
```
percona-ysorokin: Seems like `extra_str` was declared here to give a shorter name to `opt_read_only_error_msg_extra`. Seems redundant in my opinion.

sql/rpl_slave.cc
```diff
@@ -9749,6 +9749,30 @@ static int check_slave_sql_config_conflict(const Relay_log_info *rli) {
   return 0;
 }
 
+// This function will generate the string containing current master host and
+// port info if available.
+std::string get_active_master_info() noexcept {
```
percona-ysorokin: Technically speaking, this should not be `noexcept` as string concatenations may throw.

percona-ysorokin: It should also be `const` but I believe it is not so because of the `channel_map` locking.

sql/rpl_slave.h
```diff
@@ -466,6 +466,7 @@ bool start_slave_thread(
     std::atomic<uint> *slave_running, std::atomic<ulong> *slave_run_id,
     Master_info *mi);
 
+std::string get_active_master_info() noexcept;
```
percona-ysorokin:
```suggestion
std::string get_active_master_info();
```

inikep:
All comments were addressed

percona-ysorokin (approved):
LGTM

share/errmsg-utf8.txt
```diff
@@ -5080,12 +5080,12 @@ ER_FEATURE_DISABLED
         spa "El recurso '%s' fue deshabilitado; usted necesita construir MySQL con '%s' para tener eso funcionando"
         swe "'%s' är inte aktiverad; För att aktivera detta måste du bygga om MySQL med '%s' definierad"
 ER_OPTION_PREVENTS_STATEMENT
```
hermanlee, line 5082: The original code added a new ER_OPTION_PREVENTS_STATEMENT_EXTRA_INFO code. It would be better to keep to the original code instead. The value of the error code returned would match the existing 5.6 version. The return codes would be different between servers that have a master vs. those that do not, without having to parse the return string to determine if host:port were appended to it.

hermanlee (changes requested):

share/errmsg-utf8.txt
inikep: This change matches existing 5.6 version:
`ER_OPTION_PREVENTS_STATEMENT_EXTRA_INFO` was removed with https://github.com/facebook/mysql-5.6/commit/9f18725 in favor of `ER_OPTION_PREVENTS_STATEMENT` with additional string param.

inikep: I only changed output to match upstream when additional parameter is an empty string e.g.
```
ERROR HY000: The MySQL server is running with the --secure-file-priv option so it cannot execute this statement
```
instead of
```
ERROR HY000: The MySQL server is running with the --secure-file-priv option so it cannot execute this statement. 
```
in 5.6 with additional `. ` at the end.

hermanlee: Ah, thanks. I missed checking the last diff that reverted this behavior.

hermanlee (approved):

PR #959, opened 2019-02-11
FB8-53: Expose user certificate details to command line

percona-ysorokin:
Jira ticket: https://jira.percona.com/browse/FB8-53

Reference Patch: https://github.com/facebook/mysql-5.6/commit/f29eb03
Reference Patch: https://github.com/facebook/mysql-5.6/commit/c631017

'THD' class extended with the following convenience methods:
- 'has_net_vio()'
- 'get_net_vio()'
- 'has_net_vio_ssl_arg()'
- 'get_net_vio_ssl_arg()'
- 'get_peer_certificate_info()'

Implemented new 'Fill_authinfo_list' class derived from 'Do_THD_Impl'
which populates 'INFORMATION_SCHEMA.AUTHINFO' table.

This is a change to help the implementation of SSL based authentication
efforts. Currently the certificate details are unavailable to the sys admin.
The v3 extensions usually carry important information for identifying the
user.

This change adds a new table to the information schema to expose auth
details. It is organized by process ID like processlist table.

Test Plan: Run newly added test case

Originally Reviewed By: santoshb


sql/sql_class.h
```diff
@@ -1510,6 +1510,24 @@ class THD : public MDL_context_owner,
   NET net;        // client connection descriptor
   String packet;  // dynamic buffer for network I/O
  public:
+  bool has_net_vio() const noexcept { return net.vio != nullptr; }
+  Vio *get_net_vio() const noexcept { return net.vio; }
+  bool has_net_vio_ssl_arg() const noexcept {
+#ifdef HAVE_OPENSSL
```
laurynas-biveinis: IMHO in 8.0 HAVE_OPENSSL should be assumed to be always defined

sql/sql_show.cc
```diff
@@ -1914,6 +1914,100 @@ static const char *thread_state_info(THD *tmp) {
   }
 }
 
+/**
+  This class implements callback function used by fill_schema_authinfo()
+  to populate all the client SSL auth information into I_S table.
+*/
+class Fill_authinfo_list : public Do_THD_Impl {
+ private:
+  THD *m_thd;
```
laurynas-biveinis: const all the fields that are set in the constructor

laurynas-biveinis (changes requested):

sql/sql_class.cc
```diff
@@ -2721,6 +2722,45 @@ bool THD::is_current_stmt_binlog_row_enabled_with_write_set_extraction() const {
           !is_current_stmt_binlog_disabled());
 }
 
+std::string THD::get_peer_certificate_info() const {
```
dutow: This isn't the same as THD::connection_certificate() I introduced in #934 ?

sql/sql_class.h
percona-ysorokin: Actually, this comes from `include/violite.h`
```c++
#ifdef HAVE_OPENSSL
  void *ssl_arg = {nullptr};
#endif
```
Not sure if these are just refactoring leftovers, but I just made it match.

percona-ysorokin: On the other hand here is a line from top-level `CMakeLists.txt`
```
ADD_DEFINITIONS(-DHAVE_OPENSSL)            # TODO: remove #ifdef from C++ code
```

percona-ysorokin:
Rebased on a newer trunk with audit extensions, addressed review comments. Ready for another round of review.

sql/sql_class.cc
```diff
@@ -895,13 +896,44 @@ void THD::cleanup_connection(void) {
...
+
+  // Print the certificate to the buffer
+  int print_result = 0;
+  if (printable)
+    print_result = X509_print(bio.get(), cert.get());
+  else
+    print_result = PEM_write_bio_X509(bio.get(), cert.get());
+  if (print_result != 1) return {};
+
+  // decouple buffer and close bio object
```
dutow: Are you sure this comment is true?
The bio object is closed when the unique_ptr is destructed, at the end of the function. And if it decouples it, and it doesn't free the memory, we are leaking memory - but based on the documentation, I don't think BIO_get_mem_ptr does that.

sql/sql_class.cc
```diff
@@ -895,13 +896,44 @@ void THD::cleanup_connection(void) {
...
 }
 
-void THD::set_connection_certificate(std::string const &cert) {
-  DBUG_ASSERT(m_connection_certificate.empty());
-  m_connection_certificate = cert;
-}
+std::string THD::extract_peer_certificate_info(const THD *thd, bool printable) {
+  if (!thd->has_net_vio_ssl_arg()) return {};
+
+  auto ssl = static_cast<SSL *>(thd->get_net_vio_ssl_arg());
```
dutow: This could be const SSL*, and the get_net_vio getters could also return const pointers.

dutow (commented):
Looking at the changes/tests, the commit message is also a bit strange - it isn't a new command line option or anything like that, but an information schema extension.

sql/sql_class.cc
percona-ysorokin: Oh, this is just an outdated comment. There is no decoupling here at all. Removing.

sql/sql_class.cc
percona-ysorokin: Good point. Reworked.

dutow (approved):

sql/auth/sql_authentication.cc
```diff
@@ -3375,6 +3370,8 @@ int acl_authenticate(THD *thd, enum_server_command command) {
       DBUG_RETURN(1);
     }
 
+    thd->update_connection_certificate();
```
hermanlee, line 3373: Should this check for failure? If acl_check_ssl() returned 0, but then update_connection_certificate() hits a problem allocating memory during Bio_new, the certificate stored for the connection would be "". Would it be safer to fail the authentication?

hermanlee (changes requested):

hermanlee (approved):
Chances of the failure in update_connection_certificate is low, and returning "" is probably just fine.

sql/auth/sql_authentication.cc
percona-ysorokin: @hermanlee My reasoning is that this certificate string stored in THD is used only for information purposes and in my opinion in case of OOM it is better to provide inaccurate information but let users log in rather than completely deny their access. In any case, leaving this up to you to decide. It's not a big deal to rework this.

PR #967, opened 2019-02-21
FB8-167: Add variable to force range plans if force index is used

inikep:
Because of conflicts I removed changes in mysql-test/t/all_persisted_variables.test. This test has to be modified at let $total_persistent_vars=XXX; (+1 increase) and it should be re-recorded.
```
Jira ticket: https://jira.percona.com/browse/FB8-167

Reference Patch: https://github.com/facebook/mysql-5.6/commit/b01ff6a

Summary:
Currently, when force index is used, full table scans are highly penalized, but it is still possible to do a full index scan which is expensive. In many cases, we only use force index when we know there is better way to traverse the index, and not necessarily because we want a full index scan.

To address these cases, add a session variable called optimizer_force_index_for_range, which when turned on, will try to search for a 'range' plan before falling back to an index plan.

fbshipit-source-id: 4a4641b
```

### DIFF
--- a/mysql-test/t/optimizer_force_index_for_range.test
+++ b/mysql-test/t/optimizer_force_index_for_range.test
@@ -1,0 +1,52 @@
+CREATE TABLE t (i INT, j INT, PRIMARY KEY (i, j));
+INSERT INTO t VALUES (1, 1);
+INSERT INTO t VALUES (1, 2);
+INSERT INTO t VALUES (2, 1);
+INSERT INTO t VALUES (2, 2);
+INSERT INTO t VALUES (3, 1);
+INSERT INTO t VALUES (3, 2);
+
+--echo # Test range plans
+SET optimizer_force_index_for_range = ON;
+--replace_column 10 #
+EXPLAIN SELECT i FROM t WHERE i IN (1, 2, 3) AND j IN (1, 2);
+--replace_column 10 #
+EXPLAIN SELECT i FROM t FORCE INDEX (PRIMARY) WHERE i IN (1, 2, 3) AND j IN (1, 2);
+SET optimizer_force_index_for_range = OFF;
+--replace_column 10 #
+EXPLAIN SELECT i FROM t WHERE i IN (1, 2, 3) AND j IN (1, 2);
+--replace_column 10 #
+EXPLAIN SELECT i FROM t FORCE INDEX (PRIMARY) WHERE i IN (1, 2, 3) AND j IN (1, 2);
+
+--echo # Test group-by plans.
+SET optimizer_force_index_for_range = ON;
+--replace_column 10 #
+EXPLAIN SELECT DISTINCT i FROM t;
+--replace_column 10 #
+EXPLAIN SELECT DISTINCT i FROM t FORCE INDEX (PRIMARY);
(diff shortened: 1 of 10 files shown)


mysql-test/t/optimizer_force_index_for_range.test
```diff
@@ -0,0 +1,55 @@
+set @orig_optimizer_force_index_for_range = @@optimizer_force_index_for_range;
```
percona-ysorokin: no need to save/restore a session variable

mysql-test/t/optimizer_force_index_for_range.test
```diff
@@ -0,0 +1,55 @@
...
+explain select i from t force index (primary) where j > 1;
+set optimizer_switch = 'skip_scan=off';
+
+--echo # Test that in absence of range plan, index is used.
+set optimizer_force_index_for_range = on;
+--replace_column 10 #
+explain select i from t force index (primary) where j > 1;
+
+drop table t;
+set optimizer_force_index_for_range = @orig_optimizer_force_index_for_range;
```
percona-ysorokin: no need to save/restore a session variable

sql/sys_vars.cc
```diff
@@ -2746,6 +2746,12 @@ static Sys_var_ulong Sys_open_files_limit(
     /* open_files_limit is used as a sizing hint by the performance schema. */
     sys_var::PARSE_EARLY);
 
+static Sys_var_bool Sys_optimizer_force_index_for_range(
+    "optimizer_force_index_for_range",
+    "If enabled, FORCE INDEX will also try to force a range plan.",
+    SESSION_VAR(optimizer_force_index_for_range), CMD_LINE(OPT_ARG),
```
percona-ysorokin: Consider `HINT_UPDATEABLE` here.

percona-ysorokin (changes requested):

inikep:
PR was updated according to comments.

mysql-test/t/optimizer_force_index_for_range.test
```diff
@@ -0,0 +1,61 @@
+SET @session_optimizer_force_index_for_range = @@session.optimizer_force_index_for_range;
```
percona-ysorokin: not fixed
no need to save/restore session variables

mysql-test/t/optimizer_force_index_for_range.test
```diff
@@ -0,0 +1,61 @@
...
+SET optimizer_force_index_for_range = ON;
+--replace_column 10 #
+EXPLAIN SELECT i FROM t FORCE INDEX (PRIMARY) WHERE j > 1;
+
+DROP TABLE t;
+
+
+# Restore initial value
+
+SET @@session.optimizer_force_index_for_range = @session_optimizer_force_index_for_range;
```
percona-ysorokin: not fixed
no need to save/restore session variables

percona-ysorokin (approved):
LGTM with saving/restoring session var issue addressed

inikep:
Because of conflicts I removed changes in mysql-test/t/all_persisted_variables.test. This test has to be modified at let $total_persistent_vars=XXX; (+1 increase) and it should be re-recorded.

PR #968, opened 2019-02-22
FB8-222:  Add --gterm option to mtr

satya-bodapati:
https://jira.percona.com/browse/FB8-222

Problem:
--------
When using --gdb option with mtr, it spawns xterm as terminal window
by default.

xterm is really outdated.
1. Can't easily copy text from it.
2. Text sometimes appear small by deafult
3. Window buffer isn't resized properly on maximize

Fix:
----
On linux systems, where gnome-termnial (/usr/bin/gnome-terminal) is
available, try to use it.

Introduce --gterm option. This is to be used along with --gdb option.

Ex: ./mtr --mem main.1st --gdb --gterm

Note: Not all gnome-terminal versions are supported. Only those
gnome-terminals which has "wait" option are supported.

gnome-terminal --help-all

should show:
--wait                          Wait until the child exits

Squash with /dev/null

### DIFF
--- a/mysql-test/mysql-test-run.pl
+++ b/mysql-test/mysql-test-run.pl
@@ -185,6 +185,8 @@ our $opt_client_ddd;
 our $opt_client_debugger;
 our $opt_client_gdb;
 our $opt_client_lldb;
+# option for gnome-terminal with gdb
+our $opt_gterm;
 our $opt_ctest_path;
 our $opt_ctest_report;
 our $opt_dbx;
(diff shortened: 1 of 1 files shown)


inikep:
LGTM

hermanlee:
We can import this for 8.0.13 and drop it later.

satya-bodapati:
@hermanlee Thanks for importing it. Ideally I would like it to be there post 8.0.13 too but if there are issues, please feel free to drop the commit.

hermanlee (approved):

PR #969, opened 2019-02-22
FB8-59: Database admission control

satya-bodapati:
JIRA: https://jira.percona.com/browse/FB8-59

Reference patch: https://github.com/facebook/mysql-5.6/commit/5cbfab0
Reference patch: https://github.com/facebook/mysql-5.6/commit/0ebaf75
Reference patch: https://github.com/facebook/mysql-5.6/commit/cf7ce48
Reference patch: https://github.com/facebook/mysql-5.6/commit/3842a8f
Reference patch: https://github.com/facebook/mysql-5.6/commit/bcbd77a
Reference patch: https://github.com/facebook/mysql-5.6/commit/e6c2b87
Reference patch: https://github.com/facebook/mysql-5.6/commit/928c2a5
Reference patch: https://github.com/facebook/mysql-5.6/commit/86587af
Reference patch: https://github.com/facebook/mysql-5.6/commit/7870828
Reference patch: https://github.com/facebook/mysql-5.6/commit/0151b45
Reference patch: https://github.com/facebook/mysql-5.6/commit/75de821
Reference patch: https://github.com/facebook/mysql-5.6/commit/08d4174
Reference patch: https://github.com/facebook/mysql-5.6/commit/c1e5b4f
Reference patch: https://github.com/facebook/mysql-5.6/commit/08368d4
Reference patch: https://github.com/facebook/mysql-5.6/commit/a2198e0
Reference patch: https://github.com/facebook/mysql-5.6/commit/0730342
Reference patch: https://github.com/facebook/mysql-5.6/commit/d868c1b
Reference patch: https://github.com/facebook/mysql-5.6/commit/645fc5e

<Comment by Percona developers>

This patch ports Database admission control. The multi-tenancy plugin
changes are not ported. This code will be eventually moved to new plugin
of audit type.

There are possible improvements with code. Below are the deferred items
(makes it easier for review. Can compare with 5.6 code)

1. Remove MT_RESOURCE_ATTRS. we only need database name (also removes the typedefs)
2. Remove enums enum_multi_tenancy_resource_type & enum_multi_tenancy_return_type
3. Remove the usage of 'multi tenancy' word

thd->is_real_trans is not ported from 5.6. Instead an equivalent is used. This is
verified in 5.6 and it works well. (5.6 equivalent is:
 thd->transaction.all.ha_list == NULL;)

"!thd->get_transaction()->is_active(Transaction_ctx::SESSION)" is used instead of
thd->is_real_trans.

The above check seems sufficient but after porting thd->is_real_trans, it should be
possible to replace the above check with thd->is_real_trans
</end>

-------  https://github.com/facebook/mysql-5.6/commit/5cbfab0 --------

Add two global sys vars max_running_queries and max_waiting_queries.
They control maximum number of running queries on a database and maximum
waiting queries when max_running_queries limit is crossed on that database.
A value of 0 implies no limits are applied for the queries. If max_waiting_queries
limit is crossed, then new queries will simply fail.

For waiting threads SHOW PROCESSLIST will show the state as 'waiting for admission'.
Only dagtabase set at the session level is considered in these checks. Admission checks
are by-passed in the following cases
1. Query is run by super user.
2. Query is run by replication threads.
3. No database is set for the session.
4. max_running_queries is 0.

Performance testing

1. start MySQL using mtr
> mtr --start --mysqld=--default_storage_engine=InnoDB

2. Connect to mysql using
> mysql --socket=mysql-test/var/tmp/mysqld.1.sock --user=root

3. Set innodb_flush_log_at_trx_commit=2.
mysql> set @@global.innodb_flush_log_at_trx_commit=2

4. Check rows inserted using
> mysqladmin extended-status --user=root -i 1 -r --socket=mysql-test/var/tmp/mysqld.1.sock status | grep "Innodb_rows_inserted"

5. Run mysqlslap using
> mysqlslap --auto-generate-sql --number-of-queries=5000000 --concurrency=$concurrency --auto-generate-sql-load-type=write --auto-generate-sql-add-autoincrement --csv=a --user=root --socket=mysql-test/var/tmp/mysqld.1.sock

== Innodb_rows_inserted with concurrency=100 ==
38106 (max_running_queries=0, max_waiting_queries=0)
38733 (max_running_queries=1000, max_waiting_queries=0)
38365 (max_running_queries=100, max_waiting_queries=0)
40308 (max_running_queries=70, max_waiting_queries=0)
43790 (max_running_queries=50, max_waiting_queries=0)
47321 (max_running_queries=20, max_waiting_queries=0)
49473 (max_running_queries=10, max_waiting_queries=0)
54851 (max_running_queries=5, max_waiting_queries=0)

== Innodb_rows_inserted with concurrency=100 ==
26516 (max_running_queries=0, max_waiting_queries=0)
44494 (max_running_queries=10, max_waiting_queries=0)
51193 (max_running_queries=5, max_waiting_queries=0)

The above numbers show the improvement in quality of service.

== Innodb_rows_inserted with concurrency=20 ==
47228 (max_running_queries=0, max_waiting_queries=0)
47107 (max_running_queries=100, max_waiting_queries=0)
47082 (max_running_queries=20, max_waiting_queries=0)
47515 (max_running_queries=10, max_waiting_queries=0)

* There isn't much difference under low concurrency workloads.

* There isn't much difference with non zero max_waiting_queries values greater than max_running_queries. With values lower than max_running_queries, lots of queries simply fail, so the qps in this scenario is of no value.

Original Reviewers: tianx, kradhakrishnan, jtolmer, jkedgar

-----  https://github.com/facebook/mysql-5.6/commit/0ebaf75 -----

Added the following status variables:

1. Database_admission_control_aborted_queries gives the total number of queries aborted because of queue overfill.
2. Database_admission_control_running_queries gives the total number of running queries across all the databases.
3. Database_admission_control_waiting_queries gives the total number of waiting queries across all the databases.

-----  https://github.com/facebook/mysql-5.6/commit/cf7ce48 ------

In order to avoid possible mysqld hangs, we need an option to blacklist
specific commands in admission control. Example usage is

  set @@global.admission_control_filter='BEGIN,COMMIT,SET';

An empty value for admission_control_filter implies no commands are blacklisted
in the admission control.

---- https://github.com/facebook/mysql-5.6/commit/bcbd77a ------

Initial version of admission control applied limits on each query statement in multi
query packet. The behavior is modified to the following:

A thread enters admission control when executing the first non-filtered command
in the multi query packet and exits after executing all the query statements in
the multi query packet.

This diff also fixes a bug where admission control limits are not initialized during
mysqld startup.

----- https://github.com/facebook/mysql-5.6/commit/928c2a5 --------
Get accurate status variables

Hold global admission control locks to get accurate value of running/waiting queries. Initial stats were
not accurate causing test failures

------- https://github.com/facebook/mysql-5.6/commit/86587af -----------

Added SHOW command filter in admission control.

-------  https://github.com/facebook/mysql-5.6/commit/7870828 ------

The original ac_map was a map of unique_ptr<Ac_info>.  When accessing the value we had to use a reference to the value.  This diff changes that so that we use a shared_ptr<Ac_info> instead so we can just pass it around have have reference counting automatically.

---- https://github.com/facebook/mysql-5.6/commit/75de821 ------
Fixing the following two issues in admission control when changing database is
in a multi-statement query:
- If the session hasn't set a database, admission control will allow arbitrary
  number of multi-query packet to run, even the packet starts with `use
  [db_name]`.
- If the session already has a default database, the admission control will use
  the limit of the default database, even if we change database in the
  multi-query packet.

This diff fixes the above issues in the following way:
- The admission control flag (is_in_ac) is reset when a `use` command is
  executed. So in the multi-statement query, the subsequent sub-queries are not
  bypassed.
- Admission control uses the `thd->db` directly instead of attribute map, so
  the attribute map doesn't need to store the session database.
- `USE [db_name]` statement is added to `admission_control_filter`.

In addition, I also changed the multitenancy plugin interface to use a wrapper
object to pass down the connection_attrs_map, query_attrs_map, and session
database info. Attribute maps will not be modified.

--------- https://github.com/facebook/mysql-5.6/commit/c1e5b4f  --------

Fix resource leaks by admission_control_by_trx

- Rollback will need to mark is_real_trans
- Query resource needs to be released when THD is terminated.

### DIFF
--- a/mysql-test/t/admission_control.test
+++ b/mysql-test/t/admission_control.test
@@ -1,0 +1,379 @@
+--source include/count_sessions.inc
+create database test_db;
+create user 'test_user'@'localhost';
+grant all on test_db.* to 'test_user'@'localhost';
+grant all on test.* to 'test_user'@'localhost';
+
+let $start_aborted_queries = query_get_value(show global status like "Database_admission_control_aborted_queries", Value, 1);
+let $start_running_queries = query_get_value(show global status like "Database_admission_control_running_queries", Value, 1);
+let $start_waiting_queries = query_get_value(show global status like "Database_admission_control_waiting_queries", Value, 1);
+
+connection default;
+use test_db;
+set @start_max_running_queries= @@global.max_running_queries;
+set @start_max_waiting_queries= @@global.max_waiting_queries;
+set @@global.max_running_queries=10;
+set @@global.max_waiting_queries=5;
+create table t1(a int) engine=InnoDB;
+
+# Write lock here ensures the insert queries will block on row lock and be
+# in running queue until unlock tables is executed.
+lock table t1 write;
+
+disable_query_log;
+disable_result_log;
+let $i= 15;
+while ($i)
+{
+  connect (con$i, localhost, test_user,,test_db);
+  dec $i;
+}
+let $i= 10;
+while($i)
+{
+  connection con$i;
+  # These queries will be in running queue
+  send insert into t1 values(1);
+  dec $i;
+}
+let $i= 15;
+while ($i != 10)
+{
+  # These queries will be in waiting queue
+  connection con$i;
+  send insert into t1 values(1);
+  dec $i;
+}
+enable_query_log;
+enable_result_log;
+
+connection default;
+--echo Threads waiting for admission will have appropriate state set in processlist.
+let $wait_condition=
+  select count(*)=5 from information_schema.processlist where state='waiting for admission';
+source include/wait_condition.inc;
+--echo Super user is exempted from admission control checks.
+select * from t1;
+
+# set filter to bypass USE statement
+--let $saved_admission_control_filter = `SELECT @@global.admission_control_filter`
+set @@global.admission_control_filter = 'USE';
+select @@global.admission_control_filter;
+
+# Test failure case
+# Default db is intentionally not set when creating the connection
+connect (con_max_wait, localhost, test_user,,);
+connection con_max_wait;
+
+# Test multi-statement packet with use statement.
+delimiter ||||;
+--echo Maximum waiting queries reached. So this would hit an error.
+--error ER_DB_ADMISSION_CONTROL
+# switch the session's default db to test_db, which (the use statement) should
+# be bypassed in admission control. The select statement will fail.
+use test_db;
+select * from t1||||
+
+# Query will only fail on test_db, but not on test
+--echo Maximum waiting queries reached. So this would hit an error.
+--error ER_DB_ADMISSION_CONTROL
+use test;
+create table t1_test(aaa int);
+insert into t1_test values (1);
+select aaa from t1_test;
+drop table t1_test;
+use test_db;
+select * from t1||||
+
+delimiter ;||||
+
+# USE is ok
+use test_db;
+
+# Single statement
+--error ER_DB_ADMISSION_CONTROL
+select * from t1;
+
+disconnect con_max_wait;
+
+connection default;
+# reset admission_control_filter
+--eval set @@global.admission_control_filter = '$saved_admission_control_filter'
+select @@global.admission_control_filter;
+
+--echo Check status variables
+let $aborted_queries = query_get_value(show global status like "Database_admission_control_aborted_queries", Value, 1);
+let $running_queries = query_get_value(show global status like "Database_admission_control_running_queries", Value, 1);
+let $waiting_queries = query_get_value(show global status like "Database_admission_control_waiting_queries", Value, 1);
+let $aborted_queries = `select $aborted_queries - $start_aborted_queries`;
+let $running_queries = `select $running_queries - $start_running_queries`;
+let $waiting_queries = `select $waiting_queries - $start_waiting_queries`;
+--echo aborted_queries = $aborted_queries
+--echo running_queries = $running_queries
+--echo waiting_queries = $waiting_queries
+
+--echo Filled up queues on one db doesn't affect queries on other db.
+connect (con_test, localhost, test_user,,test);
+connection con_test;
+use test;
+disconnect con_test;
+
+connection default;
+set @@global.max_waiting_queries=6;
+
+--echo Kill a thread that is waiting for admission.
+connect (killed_connection, localhost, test_user,,test_db);
+connection killed_connection;
+send select count(*) from t1;
+connection default;
+# Need to first wait for the state in processlist, otherwise we run the risk
+# of a race condition and test will fail because $id would be empty.
+let $wait_timeout= 300;
+let $wait_condition=
+  select count(*) = 1 from information_schema.processlist where state='waiting for admission' and info='select count(*) from t1';
+source include/wait_condition.inc;
+let $id=
+  `select id from information_schema.processlist where state='waiting for admission' and info='select count(*) from t1'`;
+--replace_result $id ID
+eval kill $id;
+disconnect killed_connection;
+
+# Insert queries sent above will be unblocked by this.
+use test_db;
+unlock tables;
+
+disable_query_log;
+disable_result_log;
+let $i= 15;
+while ($i)
+{
+  connection con$i;
+  reap;
+  dec $i;
+}
+enable_query_log;
+enable_result_log;
+connection default;
+--echo Verify the waiting queries received wakeup signal.
+select count(*) from t1;
+
+set @save_admission_control_by_trx = @@global.admission_control_by_trx;
+select @save_admission_control_by_trx;
+set @@global.max_running_queries=5;
+set @@global.max_waiting_queries=10;
+--echo # By default, open transaction has no effect on running queries
+disable_query_log;
+disable_result_log;
+let $i= 5; # max_running_queries is 5
+while($i)
+{
+  connection con$i;
+  # 5 Open transactions
+  begin;
+  select count(*) from t1;
+  dec $i;
+}
+enable_query_log;
+enable_result_log;
+# query will succeed
+connection con11;
+select count(*) from t1;
+
+--echo # Test: open transactions will take slots in running queries,
+--echo #       and will not be blocked
+connection default;
+--let $saved_admission_control_filter = `SELECT @@global.admission_control_filter`
+set @@global.admission_control_filter = 'BEGIN,COMMIT,ROLLBACK';
+select @@global.admission_control_filter;
+set @@global.admission_control_by_trx = true;
+SELECT @@global.admission_control_by_trx;
+disable_query_log;
+disable_result_log;
+let $i= 5; # max_running_queries is 5
+while($i)
+{
+  connection con$i;
+  # Open transactions will take up slots in running queue
+  begin;
+  select count(*) from t1;
+  dec $i;
+}
+let $i= 15; # max_waiting_queries is 10
+while ($i != 5)
+{
+  # These queries will be in waiting queue
+  connection con$i;
+  send select count(*) from t1;
+  dec $i;
+}
+enable_query_log;
+enable_result_log;
+
+connection default;
+let $wait_timeout= 300;
+let $wait_condition=
+  select count(*)=10 from information_schema.processlist where state='waiting for admission';
+source include/wait_condition.inc;
+
+--echo Open transaction is able to continue running queries
+disable_query_log;
+disable_result_log;
+let $i= 5;
+while($i)
+{
+  connection con$i;
+  select count(*) from t1;
+  dec $i;
+}
+enable_query_log;
+enable_result_log;
+
+connect (con_max_wait, localhost, test_user,,test_db);
+--echo connection con_max_wait;
+connection con_max_wait;
+--echo New queries will be rejected (waiting queue is full)
+error ER_DB_ADMISSION_CONTROL;
+select * from t1;
+--echo New transactions will be rejected (waiting queue is full)
+begin;
+error ER_DB_ADMISSION_CONTROL;
+select * from t1;
+disconnect con_max_wait;
+--echo aborted_queries will increase by 2
+
+--echo Committing a transaction will free up the running query slots
+disable_query_log;
+let $i= 5;
+# commit con4 and con5
+while($i > 3)
+{
+  connection con$i;
+  commit;
+  dec $i;
+}
+# rollback con2 explicitly
+connection con2;
+rollback;
+# close con1 to rollback
+connection con1;
+disconnect con1;
+connection con3;
+# get connection id of con3
+let $con3_id = `select connection_id()`;
+# kill con3 to rollback
+connection default;
+eval kill $con3_id;
+enable_query_log;
+
+--echo The waiting queries will be unblocked
+disable_query_log;
+disable_result_log;
+let $i= 15;
+while ($i != 5)
+{
+  connection con$i;
+  reap;
+  dec $i;
+}
+enable_query_log;
+enable_result_log;
+
+connection default;
+let $wait_timeout= 300;
+
+# Total 15-2 = 13 test_user connections (con1 & con3 were closed)
+# wait for con1 and con3 to be killed or disconnected
+let $wait_condition=
+  select count(*)=13 from information_schema.processlist where user='test_user';
+source include/wait_condition.inc;
+
+# wait for all test_user connections becoming idle.
+let $wait_condition=
+  select count(*)=13 from information_schema.processlist where command='Sleep' and state='' and user='test_user';
+source include/wait_condition.inc;
+
+--echo Check status variables
+let $aborted_queries = query_get_value(show global status like "Database_admission_control_aborted_queries", Value, 1);
+let $running_queries = query_get_value(show global status like "Database_admission_control_running_queries", Value, 1);
+let $waiting_queries = query_get_value(show global status like "Database_admission_control_waiting_queries", Value, 1);
+let $aborted_queries = `select $aborted_queries - $start_aborted_queries`;
+let $running_queries = `select $running_queries - $start_running_queries`;
+let $waiting_queries = `select $waiting_queries - $start_waiting_queries`;
+
+--let $assert_text= DB Admission control waiting queries should be zero
+--let $assert_cond= $waiting_queries = 0
+--source include/assert.inc
+
+--let $assert_text= DB Admission control running queries should be zero
+--let $assert_cond= $running_queries = 0
+--source include/assert.inc
+
+--let $assert_text= DB Admission control aborted queries should be five
+--let $assert_cond= $aborted_queries = 5
+--source include/assert.inc
+
+
+# restore admission_control_by_trx
+connection default;
+set @@global.admission_control_by_trx = @save_admission_control_by_trx;
+select @@global.admission_control_by_trx;
+# reset admission_control_filter
+--eval set @@global.admission_control_filter = '$saved_admission_control_filter'
+select @@global.admission_control_filter;
+# recreate con1 and con3
+disconnect con3;
+connect (con1, localhost, test_user,,test_db);
+connect (con3, localhost, test_user,,test_db);
+--echo # End of open transaction test
+
+connection default;
+--echo reset global.max_running_queries and global.max_waiting_queries
+set @@global.max_running_queries=10;
+set @@global.max_waiting_queries=5;
+--echo Run parallel load and drop the database.
+# Verify the waiting queries will receive the signal from DROP DATABASE
+# and exit with appropriate error ER_NO_SUCH_TABLE. max_waiting_queries=0
+# is used to not have a limit on number of waiting queries.
+set @@global.max_waiting_queries=0; # No limit on waiting queue.
+disable_query_log;
+disable_result_log;
+let $i= 15;
+while($i)
+{
+  connection con$i;
+  # 10 queries (max_running_queries is 10) will be in running queue.
+  # Using sleep(10) in the insert ensures some queries end up in waiting queue.
+  send insert into t1(select sleep(10));
+  dec $i;
+}
+connection default;
+drop database test_db;
+let $i= 15;
+while($i)
+{
+  connection con$i;
+  error 0,ER_BAD_DB_ERROR;
+  reap;
+  dec $i;
+}
+enable_query_log;
+enable_result_log;
+
+--echo Cleanup.
+connection default;
+--echo Verify there are no waiting threads.
+select count(*) from information_schema.processlist where state='waiting for admission';
+set @@global.max_running_queries=@start_max_running_queries;
+set @@global.max_waiting_queries=@start_max_waiting_queries;
+drop user test_user@localhost;
+disable_query_log;
+disable_result_log;
+let $i= 15;
+while ($i)
+{
+  disconnect con$i;
+  dec $i;
+}
+enable_query_log;
+enable_result_log;
+--source include/wait_until_count_sessions.inc
--- a/mysql-test/t/admission_control_hang.test
+++ b/mysql-test/t/admission_control_hang.test
@@ -1,0 +1,126 @@
+--source include/count_sessions.inc
+create database test_db;
+create user test_user@localhost;
+grant all on test_db.* to test_user@localhost;
+
+set @start_max_running_queries = @@global.max_running_queries;
+set @@global.max_running_queries = 4;
+set @start_innodb_lock_wait_timeout = @@global.innodb_lock_wait_timeout;
+set @@global.innodb_lock_wait_timeout = 10000;
+set @start_admission_control_filter = @@global.admission_control_filter;
+set @@global.admission_control_filter = 'COMMIT';
+
+let $i= 6;
+while ($i)
+{
+  connect (con$i, localhost, test_user,,test_db);
+  dec $i;
+}
+
+connection con5;
+create table t1 (a int) engine=innodb;
+insert into t1 values(1);
+begin;
+update t1 set a=2 where a=1;
+
+connection default;
+let $i= 4;
+# Fill up the admission control running limit.
+while($i)
+{
+  connection con$i;
+  # Blocked on innodb row lock
+  send update t1 set a=2 where a=1;
+  dec $i;
+}
+
+# Test filter USE statement
+connection default;
+set @@global.admission_control_filter = 'USE';
+select @@global.admission_control_filter;
+
+connection con6;
+use test;
+use test_db;
+
+connection default;
+set @@global.admission_control_filter = 'ALTER,BEGIN,COMMIT,CREATE,DELETE,DROP,INSERT,LOAD,SELECT,SET,REPLACE,TRUNCATE,UPDATE,SHOW,ROLLBACK';
+select @@global.admission_control_filter;
+
+# Verify the commands filtered above run fine.
+connection con6;
+create table t2(a int) engine=innodb;
+begin;
+insert into t2 values(1);
+update t2 set a=2 where a=1;
+commit;
+SHOW TABLES LIKE 't2';
+begin;
+alter table t2 rename t3;
+select * from t3;
+delete from t3;
+set @val = 1;
+truncate table t3;
+rollback;
+drop table t3;
+
+# Verify that admission_control_by_trx will honor the filters
+connection default;
+set @save_admission_control_by_trx = @@global.admission_control_by_trx;
+select @save_admission_control_by_trx;
+--echo # Turn on admission_control_by_trx
+set @@global.admission_control_by_trx = true;
+SELECT @@global.admission_control_by_trx;
+
+connection con6;
+create table t2(a int) engine=innodb;
+begin;
+insert into t2 values(1);
+update t2 set a=2 where a=1;
+commit;
+SHOW TABLES LIKE 't2';
+begin;
+alter table t2 rename t3;
+select * from t3;
+delete from t3;
+set @val = 1;
+truncate table t3;
+rollback;
+drop table t3;
+
+connection default;
+set @@global.admission_control_filter = default;
+select @@global.admission_control_filter;
+
+# open transaction will be able to go through
+connection con5;
+select count(*) from t1;
+
+# restore admission_control_by_trx
+connection default;
+set @@global.admission_control_by_trx = @save_admission_control_by_trx;
+select @@global.admission_control_by_trx;
+
+connection default;
+set @@global.admission_control_filter = 'COMMIT';
+select @@global.admission_control_filter;
+
+connection con5;
+# Without the COMMIT filter set above, this query gets blocked until conflicting
+# queries hit timeout.
+commit;
+
+let $i= 6;
+while ($i)
+{
+  disconnect con$i;
+  dec $i;
+}
+
+connection default;
+set @@global.max_running_queries = @start_max_running_queries;
+set @@global.innodb_lock_wait_timeout = @start_innodb_lock_wait_timeout;
+set @@global.admission_control_filter = @start_admission_control_filter;
+drop database test_db;
+drop user test_user@localhost;
+--source include/wait_until_count_sessions.inc
--- a/mysql-test/t/admission_control_multi_query.test
+++ b/mysql-test/t/admission_control_multi_query.test
@@ -1,0 +1,15 @@
+source include/big_test.inc;
+source include/not_valgrind.inc;
+create database test_db;
+create user test_user@localhost;
+grant all on test_db.* to test_user@localhost;
+grant all on test.* to test_user@localhost;
+
+use test_db;
+create table t1 (a int primary key, b int) engine=InnoDB;
+
+let $exec = python $MYSQL_TEST_DIR/t/admission_control_multi_query.py --user='test_user' --host=127.0.0.1 --port=$MASTER_MYPORT --database='test_db';
+exec $exec;
+
+drop database test_db;
+drop user test_user@localhost;
--- a/mysql-test/t/admission_control_stress.test
+++ b/mysql-test/t/admission_control_stress.test
@@ -1,0 +1,95 @@
+source include/big_test.inc;
+# valgrind is pretty slow with lot of threads.
+source include/not_valgrind.inc;
+source include/count_sessions.inc;
+
+
+--echo Test setup.
+
+let $num_databases = 2;
+let $num_connections = 100;
+let $num_connection_per_db = 50;
+let $num_iterations = 10000;
+
+--disable_query_log
+--disable_result_log
+
+set @start_max_connections = @@global.max_connections;
+eval set @@global.max_connections = $num_connections + 10;
+set @start_max_running_queries = @@global.max_running_queries;
+set global max_running_queries = 2;
+
+# Create databases and tables.
+connection default;
+let $i = $num_databases;
+while ($i) {
+  eval create database test_db$i;
+  eval use test_db$i;
+  eval create table test_table (a int auto_increment primary key, b text) engine=InnoDB;
+  dec $i;
+}
+
+# Create test user and connection pool.
+create user test_user@localhost;
+grant all on *.* to test_user@localhost;
+let $i = $num_connections;
+while ($i) {
+  let $db = `select ceiling($i/$num_connection_per_db)`;
+  connect (con$i, localhost, test_user,,test_db$db);
+  dec $i;
+}
+
+--echo Generate load. Toggle max_running_queries and randomly kill a query.
+let $j = $num_iterations;
+while ($j) {
+  let $i = $num_connections;
+  while ($i) {
+    connection con$i;
+    send insert into test_table values (NULL, uuid());
+    dec $i;
+  }
+
+  connection default;
+  let $i = 20;
+  while ($i) {
+    let $val = `select 5*floor(10*rand())`;
+    eval set global max_running_queries = $val;
+    let $id = `select id from information_schema.processlist where user='test_user' order by rand() limit 1;`;
+    eval kill query $id;
+    dec $i;
+  }
+
+  let $i = $num_connections;
+  while ($i) {
+    connection con$i;
+    error 0, ER_QUERY_INTERRUPTED; # Error may be caused by kill query.
+    reap;
+    dec $i;
+  }
+
+  dec $j;
+}
+
+--echo Cleanup
+connection default;
+let $i = $num_databases;
+while ($i) {
+  eval select count(*) from test_db$i.test_table;
+  eval drop database if exists test_db$i;
+  dec $i;
+}
+
+drop user test_user@localhost;
+
+let $i = $num_connections;
+while ($i) {
+  disconnect con$i;
+  dec $i;
+}
+
+--enable_result_log
+--enable_query_log
+
+set global max_connections = @start_max_connections;
+set global max_running_queries = @start_max_running_queries;
+--source include/wait_until_count_sessions.inc
--- a/share/errmsg-utf8.txt
+++ b/share/errmsg-utf8.txt
@@ -18728,6 +18728,9 @@ ER_RPL_FAILED_IN_RLI_INIT_INFO
 ER_SEMISYNC_FORCE_SHUTDOWN
    eng "Force shutdown: Semi-sync master is being switched off while there are active un-acked transactions"
 
+ER_DB_ADMISSION_CONTROL
+  eng "Maximum waiting queries %lu reached for database `%s`"
+
 #
 # End of 8.0 FB MySQL error messages.
 # (Please read comments from the header of this section before adding error
--- a/sql/mysqld.cc
+++ b/sql/mysqld.cc
@@ -715,6 +715,7 @@ The documentation is based on the source files such as:
 #include "sql/dd/dictionary.h"               // dd::get_dictionary
 #include "sql/dd/performance_schema/init.h"  // performance_schema::init
 #include "sql/dd/upgrade/upgrade.h"          // dd::upgrade::in_progress
+#include "sql/sql_admission_control.h"
 #include "sql/srv_session.h"
 
 #ifdef HAVE_JEMALLOC
@@ -1127,6 +1128,7 @@ ulong binlog_cache_use = 0, binlog_cache_disk_use = 0;
 ulong binlog_stmt_cache_use = 0, binlog_stmt_cache_disk_use = 0;
 ulong max_connections, max_connect_errors;
 ulong max_nonsuper_connections = 0, nonsuper_connections = 0;
+ulong opt_max_running_queries, opt_max_waiting_queries;
 ulong rpl_stop_slave_timeout = LONG_TIMEOUT;
 bool log_bin_use_v1_row_events = 0;
 bool thread_cache_size_specified = false;
@@ -2128,6 +2130,7 @@ static void clean_up(bool print_message) {
   ha_pre_dd_shutdown();
   dd::shutdown();
 
+  delete db_ac;
   Events::deinit();
   stop_handle_manager();
 
@@ -6361,6 +6364,10 @@ int mysqld_main(int argc, char **argv)
   }
 #endif
 
+  db_ac = new AC();
+  db_ac->update_max_running_queries(opt_max_running_queries);
+  db_ac->update_max_waiting_queries(opt_max_waiting_queries);
+
   if (init_server_components()) unireg_abort(MYSQLD_ABORT_EXIT);
 
   /*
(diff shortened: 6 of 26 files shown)


sql/sql_admission_control.h
```diff
@@ -0,0 +1,356 @@
...
+extern PSI_stage_info stage_waiting_for_admission;
+#endif
+
+// TODO: remove this block eventually when this becomes part of audit plugin?
+// TBD later
+// Resource isolation types that a multi-tenancy plugin will handle.
+// Currently connection and query limits are two resource types. More will be
+// supported in the future.
+enum class enum_multi_tenancy_resource_type : int32_t {
+  MULTI_TENANCY_RESOURCE_CONNECTION,
```
laurynas-biveinis, line 62: enum classes always require type name, thus the value name prefixes can be dropped: CONNECTION, QUERY, NUM_RESOURCE_TYPES

sql/sql_admission_control.h
```diff
@@ -0,0 +1,356 @@
...
+  MULTI_TENANCY_NUM_RESOURCE_TYPES
+};
+
+// Callback function return types.
+// - ACCEPT: resource can be granted
+// - WAIT: may need to wait for resource to be freed up
+// - REJECT: resource cannot be granted
+// - FALLBACK: plugin is disabled
+enum class enum_multi_tenancy_return_type : int32_t {
+  MULTI_TENANCY_RET_ACCEPT = 0,
```
laurynas-biveinis, line 75: Likewise: ACCEPT, WAIT, ...

sql/sql_admission_control.h
```diff
@@ -0,0 +1,356 @@
...
+  ADMISSION_CONTROL_ROLLBACK,
+  ADMISSION_CONTROL_TRUNCATE,
+  ADMISSION_CONTROL_UPDATE,
+  ADMISSION_CONTROL_SHOW,
+  ADMISSION_CONTROL_USE,
+  ADMISSION_CONTROL_END = 64
+};
+
+/*
+ * sql_admission_control.h/cc
```
laurynas-biveinis: Was this supposed to be a file header comment?

sql/sql_admission_control.h
```diff
@@ -0,0 +1,356 @@
...
+  std::atomic<ulonglong> total_aborted_queries;
+
+ public:
+  AC() {
+#ifdef HAVE_PSI_INTERFACE
+    mysql_rwlock_register("sql", key_rwlock_LOCK_ac_info,
+                          array_elements(key_rwlock_LOCK_ac_info));
+#endif
+    mysql_rwlock_init(key_rwlock_LOCK_ac, &LOCK_ac);
+    max_running_queries = 0;
```
laurynas-biveinis: Field initializers or constructor initializer list please

sql/sys_vars.cc
```diff
@@ -2466,6 +2467,63 @@ static Sys_var_ulong Sys_max_connections(
...
+  return false;
+}
+
+static Sys_var_ulong Sys_max_running_queries(
+    "max_running_queries",
+    "The maximum number of running queries allowed for a database. "
+    "If this value is 0, no such limits are applied.",
+    GLOBAL_VAR(opt_max_running_queries), CMD_LINE(REQUIRED_ARG),
+    VALID_RANGE(0, 100000), DEFAULT(0), BLOCK_SIZE(1), NO_MUTEX_GUARD,
+    NOT_IN_BINLOG, ON_CHECK(0), ON_UPDATE(update_max_running_queries));
```
```suggestion
    NOT_IN_BINLOG, ON_CHECK(nullptr), ON_UPDATE(update_max_running_queries));
```

sql/sys_vars.cc
```diff
@@ -2466,6 +2467,63 @@ static Sys_var_ulong Sys_max_connections(
...
+
+static Sys_var_ulong Sys_max_waiting_queries(
+    "max_waiting_queries",
+    "The maximum number of waiting queries allowed for a database."
+    "If this value is 0, no such limits are applied.",
+    GLOBAL_VAR(opt_max_waiting_queries), CMD_LINE(REQUIRED_ARG),
+    VALID_RANGE(0, 100000), DEFAULT(0), BLOCK_SIZE(1), NO_MUTEX_GUARD,
+    NOT_IN_BINLOG, ON_CHECK(0), ON_UPDATE(update_max_waiting_queries));
```
```suggestion
    NOT_IN_BINLOG, ON_CHECK(nullptr), ON_UPDATE(update_max_waiting_queries));
```

laurynas-biveinis (changes requested):

mysql-test/t/admission_control.test
```diff
@@ -0,0 +1,366 @@
...
+connection default;
+--echo Threads waiting for admission will have appropriate state set in processlist.
+let $wait_condition=
+  select count(*)=5 from information_schema.processlist where state='waiting for admission';
+source include/wait_condition.inc;
+--echo Super user is exempted from admission control checks.
+select * from t1;
+
+# set filter to bypass USE statement
+set @@global.admission_control_filter = 'USE';
```
percona-ysorokin, line 60: It's better to save the value of `admission_control_filter` global var here before update rather than restoring it to default.

mysql-test/t/admission_control.test
```diff
@@ -0,0 +1,366 @@
...
+select @@global.admission_control_filter;
+
+--echo Check status variables
+let $aborted_queries = query_get_value(show global status like "Database_admission_control_aborted_queries", Value, 1);
+let $running_queries = query_get_value(show global status like "Database_admission_control_running_queries", Value, 1);
+let $waiting_queries = query_get_value(show global status like "Database_admission_control_waiting_queries", Value, 1);
+let $aborted_queries = `select $aborted_queries - $start_aborted_queries`;
+let $running_queries = `select $running_queries - $start_running_queries`;
+let $waiting_queries = `select $waiting_queries - $start_waiting_queries`;
+--echo aborted_queries = $aborted_queries
```
percona-ysorokin, line 111: Substitute with `asserts`

mysql-test/t/admission_control.test
```diff
@@ -0,0 +1,366 @@
+create database test_db;
```
percona-ysorokin, line 2: add `--source/count_sessions.inc`

mysql-test/t/admission_control.test
```diff
@@ -0,0 +1,366 @@
...
+disable_query_log;
+disable_result_log;
+let $i= 15;
+while ($i)
+{
+  disconnect con$i;
+  dec $i;
+}
+enable_query_log;
+enable_result_log;
```
percona-ysorokin, line 378: add `--source/wait_until_count_sessions.inc`

mysql-test/t/admission_control_hang.test
```diff
@@ -0,0 +1,124 @@
...
+  disconnect con$i;
+  dec $i;
+}
+
+connection default;
+set @@global.max_running_queries = @start_max_running_queries;
+set @@global.innodb_lock_wait_timeout = @start_innodb_lock_wait_timeout;
+set @@global.admission_control_filter = @start_admission_control_filter;
+drop database test_db;
+drop user test_user@localhost;
```
percona-ysorokin, line 125: add `--source include/wait_until_count_sessions.inc`

mysql-test/t/admission_control_hang.test
```diff
@@ -0,0 +1,124 @@
+create database test_db;
```
percona-ysorokin, line 2: add `--source include/count_sessions.inc`

mysql-test/t/admission_control_multi_query.test
```diff
@@ -0,0 +1,16 @@
+source include/big_test.inc;
+source include/not_valgrind.inc;
+create database test_db;
+create user test_user@localhost;
+grant all on test_db.* to test_user@localhost;
+grant all on test.* to test_user@localhost;
+
+connection default;
```
percona-ysorokin: redundant. Already on the `default` connection

mysql-test/t/admission_control_stress.test
```diff
@@ -0,0 +1,93 @@
+source include/big_test.inc;
+# valgrind is pretty slow with lot of threads.
+source include/not_valgrind.inc;
+
```
percona-ysorokin, line 6: `--source include/count_sessions.inc`

mysql-test/t/admission_control_stress.test
```diff
@@ -0,0 +1,93 @@
...
+while ($i) {
+  disconnect con$i;
+  dec $i;
+}
+
+--enable_result_log
+--enable_query_log
+
+set global max_connections = @start_max_connections;
+set global max_running_queries = @start_max_running_queries;
```
percona-ysorokin, line 94: `--source include/wait_until_count_sessions.inc`

mysql-test/t/admission_control_stress.test
```diff
@@ -0,0 +1,93 @@
...
+    eval set global max_running_queries = $val;
+    let $id = `select id from information_schema.processlist where user='test_user' order by rand() limit 1;`;
+    eval kill query $id;
+    dec $i;
+  }
+
+  let $i = $num_connections;
+  while ($i) {
+    connection con$i;
+    error 0, 1317; # Error may be caused by kill query.
```
percona-ysorokin: Mnemonic instead of `1317`

mysql-test/t/admission_control_stress.test
```diff
@@ -0,0 +1,93 @@
...
+set @start_max_connections = @@global.max_connections;
+eval set @@global.max_connections = $num_connections + 10;
+set @start_max_running_queries = @@global.max_running_queries;
+set global max_running_queries = 2;
+
+# Create databases and tables.
+connection default;
+let $i = $num_databases;
+while ($i) {
+  eval drop database if exists test_db$i;
```
percona-ysorokin: redundant

sql/mysqld.cc
```diff
@@ -7881,6 +7887,33 @@ static int show_connection_errors_tcpwrap(THD *, SHOW_VAR *var, char *buff) {
   return 0;
 }
 
+static int get_db_ac_total_aborted_queries(THD *thd MY_ATTRIBUTE((unused)),
+                                           SHOW_VAR *var, char *buff) {
+  var->type = SHOW_LONGLONG;
```
percona-ysorokin, line 7969: The type is set as `SHOW_LONGLONG` whereas the value is cast to `long *`

sql/sql_admission_control.cc
```diff
@@ -0,0 +1,396 @@
...
+static bool filter_command(enum_sql_command sql_command) {
+  switch (sql_command) {
+    case SQLCOM_ALTER_TABLE:
+    case SQLCOM_ALTER_DB:
+    case SQLCOM_ALTER_PROCEDURE:
+    case SQLCOM_ALTER_FUNCTION:
+    case SQLCOM_ALTER_TABLESPACE:
+    case SQLCOM_ALTER_SERVER:
+    case SQLCOM_ALTER_EVENT:
+      //    case SQLCOM_ALTER_DB_UPGRADE:
```
percona-ysorokin, line 48: Commented out code

sql/sql_admission_control.cc
```diff
@@ -0,0 +1,396 @@
...
+      return IS_BIT_SET(admission_control_filter, ADMISSION_CONTROL_BEGIN);
+
+    case SQLCOM_COMMIT:
+      return IS_BIT_SET(admission_control_filter, ADMISSION_CONTROL_COMMIT);
+
+    case SQLCOM_CREATE_TABLE:
+    case SQLCOM_CREATE_INDEX:
+    case SQLCOM_CREATE_DB:
+    case SQLCOM_CREATE_FUNCTION:
+      //    case SQLCOM_CREATE_NPROCEDURE:
```
percona-ysorokin, line 67: Commented out code

sql/sql_admission_control.cc
```diff
@@ -0,0 +1,396 @@
...
+
+    case SQLCOM_DELETE:
+    case SQLCOM_DELETE_MULTI:
+      return IS_BIT_SET(admission_control_filter, ADMISSION_CONTROL_DELETE);
+
+    case SQLCOM_DROP_TABLE:
+    case SQLCOM_DROP_INDEX:
+    case SQLCOM_DROP_DB:
+    case SQLCOM_DROP_FUNCTION:
+      //   case SQLCOM_DROP_NPROCEDURE:
```
percona-ysorokin, line 88: Commented out code

sql/sql_admission_control.cc
```diff
@@ -0,0 +1,396 @@
...
+    case SQLCOM_SHOW_TABLES:
+    case SQLCOM_SHOW_FIELDS:
+    case SQLCOM_SHOW_KEYS:
+    case SQLCOM_SHOW_VARIABLES:
+    case SQLCOM_SHOW_STATUS:
+    case SQLCOM_SHOW_ENGINE_LOGS:
+    case SQLCOM_SHOW_ENGINE_STATUS:
+    case SQLCOM_SHOW_ENGINE_MUTEX:
+    case SQLCOM_SHOW_PROCESSLIST:
+      //    case SQLCOM_SHOW_TRANSACTION_LIST:
```
percona-ysorokin, line 136: Commented out code

sql/sql_admission_control.cc
```diff
@@ -0,0 +1,396 @@
...
+    case SQLCOM_SHOW_CHARSETS:
+    case SQLCOM_SHOW_COLLATIONS:
+    case SQLCOM_SHOW_CREATE_DB:
+    case SQLCOM_SHOW_TABLE_STATUS:
+    case SQLCOM_SHOW_TRIGGERS:
+    case SQLCOM_SHOW_BINLOGS:
+    case SQLCOM_SHOW_OPEN_TABLES:
+    case SQLCOM_SHOW_SLAVE_HOSTS:
+    case SQLCOM_SHOW_BINLOG_EVENTS:
+      //   case SQLCOM_SHOW_BINLOG_CACHE:
```
percona-ysorokin, line 150: Commented out code

sql/sql_admission_control.cc
```diff
@@ -0,0 +1,396 @@
...
+    case SQLCOM_SHOW_PLUGINS:
+    case SQLCOM_SHOW_CREATE_EVENT:
+    case SQLCOM_SHOW_EVENTS:
+    case SQLCOM_SHOW_CREATE_TRIGGER:
+    case SQLCOM_SHOW_PROFILE:
+    case SQLCOM_SHOW_PROFILES:
+    case SQLCOM_SHOW_RELAYLOG_EVENTS:
+    case SQLCOM_SHOW_ENGINE_TRX:
+    case SQLCOM_SHOW_MEMORY_STATUS:
+      //    case SQLCOM_SHOW_CONNECTION_ATTRIBUTES:
```
percona-ysorokin, line 170: Commented out code

sql/sql_admission_control.cc
```diff
@@ -0,0 +1,396 @@
...
+  }
+  if (release_lock_ac) {
+    mysql_rwlock_unlock(&LOCK_ac);
+  }
+  thd->proc_info = prev_proc_info;
+  return error;
+}
+
+void AC::wait_for_signal(THD *thd, std::shared_ptr<st_ac_node> &ac_node,
+                         std::shared_ptr<Ac_info> ac_info) {
```
percona-ysorokin:
```suggestion
                         const Ac_info_ptr& ac_info) {
```

sql/sql_admission_control.h
```diff
@@ -0,0 +1,356 @@
...
+  inline ulong get_max_waiting_queries() {
+    mysql_rwlock_rdlock(&LOCK_ac);
+    ulong res = max_waiting_queries;
+    mysql_rwlock_unlock(&LOCK_ac);
+    return res;
+  }
+
+  bool admission_control_enter(THD *, const MT_RESOURCE_ATTRS *);
+  void admission_control_exit(THD *, const MT_RESOURCE_ATTRS *);
+  void wait_for_signal(THD *, std::shared_ptr<st_ac_node> &,
```
percona-ysorokin:
```suggestion
  using st_ac_node_ptr = std::shared_ptr<st_ac_node>;
  using Ac_info_ptr = std::shared_ptr<Ac_info>
  void wait_for_signal(THD *, st_ac_node_pt &,
```

sql/sql_admission_control.h
```diff
@@ -0,0 +1,356 @@
...
+                       std::shared_ptr<Ac_info> ac_info);
```
percona-ysorokin:
```suggestion
                       const Ac_info_ptr& ac_info);
```

sql/sql_admission_control.h
```diff
@@ -0,0 +1,356 @@
...
+enum class enum_multi_tenancy_return_type : int32_t {
+  MULTI_TENANCY_RET_ACCEPT = 0,
+  MULTI_TENANCY_RET_WAIT,
+  MULTI_TENANCY_RET_REJECT,
+
+  MULTI_TENANCY_RET_FALLBACK,
+  MULTI_TENANCY_NUM_RETURN_TYPES
+};
+
+typedef enum enum_multi_tenancy_resource_type MT_RESOURCE_TYPE;
```
percona-ysorokin, line 83: redundant
```
enum class xxx {...};
typedef enum xxx yyy;
```
can just be 
```
enum class yyy {...};
```

sql/sql_admission_control.h
```diff
@@ -0,0 +1,356 @@
...
+typedef enum enum_multi_tenancy_return_type MT_RETURN_TYPE;
```
percona-ysorokin, line 84: redundant

sql/sql_admission_control.h
```diff
@@ -0,0 +1,356 @@
...
+ * Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301 USA
+ */
+
+#ifndef _sql_admission_control_h
+#define _sql_admission_control_h
+
+#include "mysql/psi/mysql_mutex.h"
+#include "mysql/psi/psi_base.h"
+
+#include <deque>
```
percona-ysorokin, line 40:
```suggestion
#include <deque>
#include <unordered_map>
```

sql/mysqld.cc
```diff
@@ -6341,6 +6343,10 @@ int mysqld_main(int argc, char **argv)
   }
 #endif
 
+  db_ac = new AC();
```
percona-ysorokin, line 6367: `db_ac` is never deleted - consider declaring it as a `unique_ptr`

sql/sql_admission_control.h
```diff
@@ -0,0 +1,356 @@
...
+#include "sql/sql_class.h"
+
+#define IS_BIT_SET(val, n) ((val) & (1 << (n)))
+
+extern bool opt_admission_control_by_trx;
+extern ulonglong admission_control_filter;
+class AC;
+extern AC *db_ac;
```
percona-ysorokin, line 48: declare `db_ac` as a `scoped_ptr`

sql/sql_admission_control.h
```diff
@@ -0,0 +1,356 @@
...
+  MULTI_TENANCY_RET_WAIT,
+  MULTI_TENANCY_RET_REJECT,
+
+  MULTI_TENANCY_RET_FALLBACK,
+  MULTI_TENANCY_NUM_RETURN_TYPES
+};
+
+typedef enum enum_multi_tenancy_resource_type MT_RESOURCE_TYPE;
+typedef enum enum_multi_tenancy_return_type MT_RETURN_TYPE;
+typedef std::unordered_map<std::string, std::string> ATTRS_MAP_T;
```
percona-ysorokin, line 85: using instead of typedef

sql/sql_admission_control.h
```diff
@@ -0,0 +1,356 @@
...
+typedef std::vector<std::pair<std::string, std::string>> ATTRS_VEC_T;
```
percona-ysorokin, line 86: using instead of typedef

sql/sql_admission_control.h
```diff
@@ -0,0 +1,356 @@
...
+
+struct multi_tenancy_resource_attributes {
+  const ATTRS_MAP_T *connection_attrs_map;
+  const ATTRS_VEC_T *query_attrs_vec;
+  const char *database;
+};
+
+typedef struct multi_tenancy_resource_attributes MT_RESOURCE_ATTRS;
```
percona-ysorokin, line 94: redundant

sql/sql_admission_control.h
```diff
@@ -0,0 +1,356 @@
...
+
+// The contents here must match entries in admission_control_filter_names array
+enum enum_admission_control_filter {
```
percona-ysorokin, line 97: change to `enum class`?

sql/sql_admission_control.h
```diff
@@ -0,0 +1,356 @@
...
+
+  count = static_cast<int>(array_elements(all_ac_node_mutexes));
+  mysql_mutex_register(category, all_ac_node_mutexes, count);
+
+  count = static_cast<int>(array_elements(all_ac_node_conds));
+  mysql_cond_register(category, all_ac_node_conds, count);
+}
+#endif
+
+struct st_ac_node {
```
percona-ysorokin, line 149: make `st_ac_node` non-copyable
```suggestion
struct st_ac_node {
  st_ac_node(const st_ac_node&) = delete;
  st_ac_node& operator = (const st_ac_node&) = delete;
```

sql/sql_admission_control.h
```diff
@@ -0,0 +1,356 @@
...
+
+*/
+class Ac_info {
+  friend class AC;
+#ifdef HAVE_PSI_INTERFACE
+  PSI_mutex_key key_lock;
+  PSI_mutex_info key_lock_info[1] = {{&key_lock, "Ac_info::lock", 0}};
+#endif
+  // Queue to track the running and waiting threads.
+  std::deque<std::shared_ptr<st_ac_node>> queue;
```
percona-ysorokin: usings/typedef for custom containers
```suggestion
  using ac_node_ptr_container = std::deque<st_ac_node_ptr>;
  ac_node_ptr_container queue;
```

sql/sql_admission_control.h
```diff
@@ -0,0 +1,356 @@
...
+  std::atomic<unsigned long> waiting_queries;
```
percona-ysorokin:
```suggestion
  std::atomic_ulong waiting_queries;
```

sql/sql_admission_control.h
```diff
@@ -0,0 +1,356 @@
...
+  Ac_info(const Ac_info &) = delete;
+  Ac_info &operator=(const Ac_info &) = delete;
+};
+
+/**
+  Global class used to enforce per admission control limits.
+*/
+class AC {
+  // This map is protected by the rwlock LOCK_ac.
+  std::unordered_map<std::string, std::shared_ptr<Ac_info>> ac_map;
```
percona-ysorokin: usings/typedefs for custom containers
```suggestion
  using Ac_info_ptr_container = std::unordered_map<std::string, Ac_info_ptr;
  Ac_info_ptr_container ac_map;
```

sql/sql_admission_control.h
```diff
@@ -0,0 +1,356 @@
...
+    Locking order followed is LOCK_ac, Ac_info::lock, st_ac_node::lock.
+  */
+  mysql_rwlock_t LOCK_ac;
+#ifdef HAVE_PSI_INTERFACE
+  PSI_rwlock_key key_rwlock_LOCK_ac;
+  PSI_rwlock_info key_rwlock_LOCK_ac_info[1] = {
+      {&key_rwlock_LOCK_ac, "AC::rwlock", 0}};
+#endif
+
+  std::atomic<ulonglong> total_aborted_queries;
```
percona-ysorokin:
```suggestion
  std::atomic_ullong total_aborted_queries;
```

sql/sql_admission_control.h
```diff
@@ -0,0 +1,356 @@
...
+    max_waiting_queries = 0;
+    total_aborted_queries = 0;
+  }
+
+  ~AC() { mysql_rwlock_destroy(&LOCK_ac); }
+  // Disable copy constructor.
+  AC(const AC &) = delete;
+  AC &operator=(const AC &) = delete;
+
+  inline void signal(std::shared_ptr<st_ac_node> &ac_node) {
```
percona-ysorokin: this is already inline
```suggestion
  void signal(st_ac_node_ptr &ac_node) {
```

sql/sql_admission_control.h
```diff
@@ -0,0 +1,356 @@
...
+    if (it != ac_map.end()) {
+      ac_map.erase(it);
+    }
+    mysql_rwlock_unlock(&LOCK_ac);
+  }
+
+  void insert(const std::string &entity) {
+    mysql_rwlock_wrlock(&LOCK_ac);
+    if (ac_map.find(entity) == ac_map.end()) {
+      ac_map[entity] = std::make_shared<Ac_info>();
```
percona-ysorokin:
```suggestion
      ac_map.emplace(entity, std::make_shared<Ac_info>())
```

sql/sql_admission_control.h
```diff
@@ -0,0 +1,356 @@
...
+      if (ac_info->queue.size() > max_running_queries)
+        res += ac_info->queue.size() - max_running_queries;
+      mysql_mutex_unlock(&ac_info->lock);
+    }
+    mysql_rwlock_unlock(&LOCK_ac);
+    return res;
+  }
+};
+
+extern AC *db_ac;
```
percona-ysorokin: redundant

sql/sql_class.h
```diff
@@ -4037,6 +4038,10 @@ class THD : public MDL_context_owner,
   }
   bool is_persist_variables_init() { return m_persist_variables_init; }
 
+  /* whether the session is already in admission control for queries */
+  bool is_in_ac = false;
+  std::shared_ptr<st_ac_node> ac_node;
```
percona-ysorokin:
```suggestion
  st_ac_node_ptr ac_node;
```

sql/sql_admission_control.cc
```diff
@@ -0,0 +1,396 @@
...
+    switch (ret) {
+      case MT_RETURN_TYPE::MULTI_TENANCY_RET_REJECT:
+        ++total_aborted_queries;
+        // We reached max waiting limit. Error out
+        mysql_mutex_unlock(&ac_info->lock);
+        error = true;
+        break;
+
+      case MT_RETURN_TYPE::MULTI_TENANCY_RET_WAIT:
+        ac_info->queue.push_back(thd->ac_node);
```
percona-ysorokin:
```suggestion
        ac_info->queue.emplace_back(thd->ac_node);
```

sql/sql_admission_control.cc
```diff
@@ -0,0 +1,396 @@
...
+        */
+        mysql_rwlock_unlock(&LOCK_ac);
+        release_lock_ac = false;
+        wait_for_signal(thd, thd->ac_node, ac_info);
+        --ac_info->waiting_queries;
+        break;
+
+      case MT_RETURN_TYPE::MULTI_TENANCY_RET_ACCEPT:
+        // We are below the max running limit.
+        ac_info->queue.push_back(thd->ac_node);
```
percona-ysorokin:
```suggestion
        ac_info->queue.emplace_back(thd->ac_node);
```

sql/sql_admission_control.h
```diff
@@ -0,0 +1,356 @@
...
+  // Protects the queue.
+  mysql_mutex_t lock;
+
+ public:
+  Ac_info() {
+#ifdef HAVE_PSI_INTERFACE
+    mysql_mutex_register("sql", key_lock_info, array_elements(key_lock_info));
+#endif
+    mysql_mutex_init(key_lock, &lock, MY_MUTEX_INIT_FAST);
+    waiting_queries = 0;
```
percona-ysorokin: move to the constructor's initialization list

sql/sql_admission_control.h
percona-ysorokin: this method has nothing to do with `AC` class, should probably be moved to `st_ac_node`

sql/sql_admission_control.h
```diff
@@ -0,0 +1,356 @@
...
+  // This map is protected by the rwlock LOCK_ac.
+  std::unordered_map<std::string, std::shared_ptr<Ac_info>> ac_map;
+  // Variables to track global limits
+  ulong max_running_queries, max_waiting_queries;
+  /**
+    Protects ac_map and max_running_queries/max_waiting_queries.
+
+    Locking order followed is LOCK_ac, Ac_info::lock, st_ac_node::lock.
+  */
+  mysql_rwlock_t LOCK_ac;
```
percona-ysorokin:
```suggestion
  mutable mysql_rwlock_t LOCK_ac;
```

sql/sql_admission_control.h
```diff
@@ -0,0 +1,356 @@
...
+    mysql_rwlock_unlock(&LOCK_ac);
+  }
+
+  void update_max_waiting_queries(ulong val) {
+    mysql_rwlock_wrlock(&LOCK_ac);
+    max_waiting_queries = val;
+    mysql_rwlock_unlock(&LOCK_ac);
+  }
+
+  inline ulong get_max_running_queries() {
```
percona-ysorokin: already inline
```suggestion
  ulong get_max_running_queries() const {
```

sql/sql_admission_control.h
```diff
@@ -0,0 +1,356 @@
...
+    mysql_rwlock_rdlock(&LOCK_ac);
+    ulong res = max_running_queries;
+    mysql_rwlock_unlock(&LOCK_ac);
+    return res;
+  }
+
+  inline ulong get_max_waiting_queries() {
```
percona-ysorokin: already inline
```suggestion
  ulong get_max_waiting_queries() const {
```

percona-ysorokin (changes requested):

sql/sql_admission_control.h
satya-bodapati: I would like to keep this as is for now. Any improvements should be done when moved to plugin.
Neverthless I will add TODO here

sql/sql_admission_control.h
satya-bodapati: Mutable is suggested because of the const function
get_max_running_queries() ?

sql/sql_admission_control.h
satya-bodapati: This enum will be removed when the code moves to separate plugin. It is just there to keep the changes to minimal(when compared with 5.6 patch).

Will add TODO here

sql/sql_admission_control.h
satya-bodapati: ok, just read about "enum classes". Interesting addition in C++11. But..

This enum will be removed when the code moves to separate plugin. It is just there to keep the changes to minimal(when compared with 5.6 patch).

Will add TODO here

sql/sql_admission_control.h
satya-bodapati: No idea. It was there in original FB source. You want me to change this?

sql/sql_admission_control.cc
satya-bodapati: This is because I wasn't sure if FB will add them back. 

@hermanlee Hello! any plans to do this SQL COMMAND to 8.0

sql/sql_admission_control.cc
satya-bodapati: @hermanlee Hello! any plans to add the below SQL COMMANDS to 8.0

SQLCOM_ALTER_DB_UPGRADE
SQLCOM_CREATE_NPROCEDURE
SQLCOM_SHOW_TRANSACTION_LIST
SQLCOM_DROP_NPROCEDURE:
SQLCOM_SHOW_BINLOG_CACHE:
SQLCOM_SHOW_CONNECTION_ATTRIBUTES

sql/sql_admission_control.cc
satya-bodapati: Are you sure about this change? In one of the commits, it was changed to std::shared_ptr

satya-bodapati: Ignore, it is typedef/using variant..

sql/sql_admission_control.h
satya-bodapati: Enums to be removed

sql/sql_admission_control.h
satya-bodapati: Enums to be removed

sql/mysqld.cc
satya-bodapati: isn't this defined as smart pointer? Let me check

sql/sql_admission_control.h
satya-bodapati: Why this unordered_map is added? Don't see any failure without it

sql/sql_admission_control.h
satya-bodapati: You mean this https://www.boost.org/doc/libs/1_61_0/libs/smart_ptr/scoped_ptr.htm ?

sql/sql_admission_control.h
satya-bodapati: "There are possible improvements with code. Below are the deferred items
(makes it easier for review. Can compare with 5.6 code)" -> from commit messsage

sql/sql_admission_control.h
satya-bodapati: "There are possible improvements with code. Below are the deferred items
(makes it easier for review. Can compare with 5.6 code)" -> from commit messsage

sql/sql_admission_control.h
satya-bodapati: This will be removed and there is TODO on the struct.

sql/sql_admission_control.h
satya-bodapati: yes that would be nice but it needs more thought. We have to implement operatior & | for this enum  class. So deferring this for now.

sql/mysqld.cc
satya-bodapati: Good catch. missed to add delete. Done

sql/sql_admission_control.h
percona-ysorokin: Actually, what I meant was `std::unique_ptr` - `scoped_ptr` is a flashback from non-c++11 times  :-)

satya-bodapati:
Ready for internal review again @percona-ysorokin @laurynas-biveinis.

Testcase improvements will be done later.

sql/sql_admission_control.h
satya-bodapati: Ok, got it. Currently not using unique_ptr as the delete was added.

mysql-test/t/admission_control.test
satya-bodapati: Am not big fan of this "save" and restore. Infact I prefer to use the "Set global var = default'
I know that there is a reason to use save and restore (i.e when mtr tested with non default values) but is barely ever used. You also need to add disable_query_log around the set statements.

Nevertheless, I have done this time ;)

satya-bodapati:
Testcase comments addressed too. Ready for internal review @laurynas-biveinis @percona-ysorokin

percona-ysorokin (approved):
LGTM

satya-bodapati:
Thanks @percona-ysorokin Ready for FB review

share/errmsg-utf8.txt
```diff
@@ -18724,6 +18724,9 @@ ER_BINLOG_CANT_READ_INDEX
 ER_RPL_FAILED_IN_RLI_INIT_INFO
   eng "Failed in %s called from Relay_log_info::rli_init_info()."
 
+ER_DB_ADMISSION_CONTROL
```
hermanlee, line 18731: This should replace ER_PLACEHOLDER_50028. In the original 5.6 code, it came after ER_DB_READ_ONLY.

sql/sql_admission_control.cc
hermanlee: No, we don't plan to add this back in, so it can be removed.

sql/sql_admission_control.cc
hermanlee: The only one we'd likely add back is SQLCOM_SHOW_CONNECTION_ATTRIBUTES. But the rest will be dropped.

sql/sql_admission_control.h
```diff
@@ -0,0 +1,365 @@
...
+extern PSI_stage_info stage_waiting_for_admission;
+#endif
+
+// TODO: remove this block eventually when this becomes part of audit plugin?
+// TBD later
+// Resource isolation types that a multi-tenancy plugin will handle.
+// Currently connection and query limits are two resource types. More will be
+// supported in the future.
+enum class enum_multi_tenancy_resource_type : int32_t {
+  MULTI_TENANCY_RESOURCE_CONNECTION,
```
hermanlee, line 62: Would it be too onerous to remove the references to the multi_tenancy_* enums and use most of the original code which did not depend on these enums? Or do the subsequent fixes added after the refactoring into sql_multi_tenancy.cc make this difficult?

hermanlee (changes requested):

sql/sql_admission_control.h
satya-bodapati: I delayed the removals, renaming (and refactoring) because I want this patch to be tested for correctness first. Once it is good, all these things can be handled with new proposed JIRA that moves this entire code to a audit plugin type.

How did the internal testing go? Any issues found?

sql/sql_admission_control.h
```diff
@@ -0,0 +1,365 @@
...
+
+  Every entity (database or table or user name) will have this
+  object created and stored in the global map AC::ac_map.
+
+*/
+class Ac_info {
+  friend class AC;
+#ifdef HAVE_PSI_INTERFACE
+  PSI_mutex_key key_lock;
+  PSI_mutex_info key_lock_info[1] = {{&key_lock, "Ac_info::lock", 0}};
```
hermanlee, line 181: compiler error: need to add PSI_FLAG_SINGLETON and PSI_DOCUMENT_ME.

sql/sql_admission_control.h
```diff
@@ -0,0 +1,365 @@
...
+  /**
+    Protects ac_map and max_running_queries/max_waiting_queries.
+
+    Locking order followed is LOCK_ac, Ac_info::lock, st_ac_node::lock.
+  */
+  mutable mysql_rwlock_t LOCK_ac;
+#ifdef HAVE_PSI_INTERFACE
+  PSI_rwlock_key key_rwlock_LOCK_ac;
+  PSI_rwlock_info key_rwlock_LOCK_ac_info[1] = {
+      {&key_rwlock_LOCK_ac, "AC::rwlock", 0}};
```
hermanlee, line 224: compiler error: need to add PSI_FLAG_SINGLETON and PSI_DOCUMENT_ME.

sql/sql_admission_control.cc
```diff
@@ -0,0 +1,396 @@
...
+    unlock ac_node
+    The locks are interleaved to avoid possible races which makes
+    this waiting thread miss the signal from admission_control_exit().
+  */
+  mysql_mutex_unlock(&ac_info->lock);
+  thd->ENTER_COND(&ac_node->cond, &ac_node->lock, &stage_waiting_for_admission,
+                  &old_stage);
+  // Spurious wake-ups are rare and fine in this design.
+  mysql_cond_wait(&ac_node->cond, &ac_node->lock);
+  mysql_mutex_unlock(&ac_node->lock);
```
hermanlee, line 361: Hm.. the original code doesn't have this unlock call. Seems like the feature would be broken without it! I'm not sure where the unlock() for the waiting case.

hermanlee (changes requested):
Looks like the following tests need to be fixed or re-recorded:

sys_vars.all_vars main.mysqld--help-notwin main.admission_control_multi_query perfschema.dml_setup_instruments main.information_schema_cs

Should mysql_rm_db() call db_ac->remove(db) somewhere, or has it been intentionally removed?

The unit tests, other than the ones listed, look to be passing.

I'm going to setup some more comprehensive testing with this patch.

sql/sql_admission_control.cc
hermanlee: Ah, okay, 5.6 has exit_cond unlock the mutex, but this is different in 8.0.

satya-bodapati: Right! spot on. In 8.0, it is different.

sql/sql_admission_control.h
satya-bodapati: How did you get this error? I cannot see this. 

Some part I actually converted to new interface but old was working fine too. Let me know the steps to get compiler error

satya-bodapati:
> Looks like the following tests need to be fixed or re-recorded:
> 
> sys_vars.all_vars main.mysqld--help-notwin main.admission_control_multi_query perfschema.dml_setup_instruments main.information_schema_cs

what is the failure with main.admission_control_multi_query  test? Locally it passes for me.
Other tests I have fixed now.

> 
> Should mysql_rm_db() call db_ac->remove(db) somewhere, or has it been intentionally removed?

Good catch. Don't know how it got missed.

> 
> The unit tests, other than the ones listed, look to be passing.

Don't know the "list" you are referring here. Hopefully nothing related to admission control are failing.

> 
> I'm going to setup some more comprehensive testing with this patch.

Please do and let me know the results.

hermanlee:
The main.admission_control_multi_query test problem was due to an internal configuration difference. I'll update the test locally.

The list of tests was referrring to: "sys_vars.all_vars main.mysqld--help-notwin main.admission_control_multi_query perfschema.dml_setup_instruments main.information_schema_cs"

The testing is still on my TODO. I'm hoping to get to it soon.

sql/sql_admission_control.h
hermanlee: Here's the error:

```
In file included from /home/herman/rocks-mysql/8.0/sql/sql_admission_control.cc:17:
/home/herman/rocks-mysql/8.0/sql/sql_admission_control.h:182:37: error: missing field 'm_volatility' initializer [-Werror,-Wmissing-field-initializers]
      {&key_lock, "Ac_info::lock", 0}};
```

So missing-field-initializers?

hermanlee (approved):
Internal testing is going well.

PR #970, opened 2019-02-22
FB8-172: Add variable to disable full table/index scans

inikep:
Jira issue: https://jira.percona.com/browse/FB8-172

Reference Patch: b90e801
Reference Patch: ecef4b2

There are cases where it is always unacceptable for a client to be doing full table scans. To prevent this from happening, add a new variable optimizer_full_scan. When it is turned off, it will return a new error code ER_FULL_SCAN_DISABLED.

For best results, use with optimizer_force_index_for_range. Currently, it does not fallback to alternative plans because we check at the end of query planning.

EXPLAIN queries with derived tables do not populate select_options in the JOIN structure properly. This means that we could error if we tried to run an explain on a query with derived tables.

Instead of reading from the JOIN struct, read from the LEX struct on the THD, as this is where the flag is original set on during query parsing.

This bug also means that in upstream, we increment some status variables tracking full table scans despite the fact that only an explain statement was done. This seems to have been fixed in 8.0 though.

### DIFF
--- a/mysql-test/t/optimizer_full_scan.test
+++ b/mysql-test/t/optimizer_full_scan.test
@@ -1,0 +1,64 @@
+CREATE TABLE t (i INT, j INT, KEY(i));
+INSERT INTO t VALUES (1, 1);
+INSERT INTO t VALUES (1, 2);
+INSERT INTO t VALUES (2, 1);
+INSERT INTO t VALUES (2, 2);
+INSERT INTO t VALUES (3, 1);
+INSERT INTO t VALUES (3, 2);
+
+ANALYZE TABLE t;
+
+--echo # Basic tests
+--replace_column 10 #
+EXPLAIN SELECT * FROM t;
+SET optimizer_full_scan = OFF;
+--error ER_FULL_SCAN_DISABLED
+SELECT * FROM t;
+SET optimizer_full_scan = ON;
+SELECT * FROM t;
+
+--replace_column 10 #
+EXPLAIN SELECT i FROM t FORCE INDEX (i);
+SET optimizer_full_scan = OFF;
+--error ER_FULL_SCAN_DISABLED
+SELECT i FROM t FORCE INDEX (i);
+SET optimizer_full_scan = ON;
+SELECT i FROM t FORCE INDEX (i);
+
+--replace_column 10 #
+EXPLAIN SELECT * FROM t a, t b WHERE a.i = b.i;
+SET optimizer_full_scan = OFF;
+--error ER_FULL_SCAN_DISABLED
+SELECT * FROM t a, t b WHERE a.i = b.i;
+SET optimizer_full_scan = ON;
+SELECT * FROM t a, t b WHERE a.i = b.i;
+
+--replace_column 10 #
+EXPLAIN SELECT * FROM t a STRAIGHT_JOIN t b WHERE a.i = 10;
+SET optimizer_full_scan = OFF;
+--error ER_FULL_SCAN_DISABLED
+SELECT * FROM t a STRAIGHT_JOIN t b WHERE a.i = 10;
+SET optimizer_full_scan = ON;
+SELECT * FROM t a STRAIGHT_JOIN t b WHERE a.i = 10;
+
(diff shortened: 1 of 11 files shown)


sql/sql_select.cc
```diff
@@ -2590,6 +2590,11 @@ bool make_join_readinfo(JOIN *join, uint no_jbuf_after) {
               join->thd->inc_status_select_scan();
             else
               join->thd->inc_status_select_full_join();
+            /* Block full table/index scans, if optimizer_full_scan is off. */
```
laurynas-biveinis, line 2589: Should be moved above the status var bumps, or they will be bumped for full table scans that did not actually happen

laurynas-biveinis (changes requested):

sql/sql_select.cc
inikep: It's done

mysql-test/t/optimizer_full_scan.test
```diff
@@ -0,0 +1,70 @@
+set @orig_optimizer_full_scan = @@optimizer_full_scan;
```
percona-ysorokin: no need to save/restore session vars.

mysql-test/t/optimizer_full_scan.test
```diff
@@ -0,0 +1,70 @@
...
+set @orig_optimizer_force_index_for_range = @@optimizer_force_index_for_range;
```
percona-ysorokin: no need to save/restore session vars.

mysql-test/t/optimizer_full_scan.test
```diff
@@ -0,0 +1,70 @@
...
+explain select i from t where i in (1, 2, 3) and j in (1, 2);
+--error ER_FULL_SCAN_DISABLED
+select i from t where i in (1, 2, 3) and j in (1, 2);
+--replace_column 10 #
+explain select i from t force index (primary) where i in (1, 2, 3) and j in (1, 2);
+select i from t force index (primary) where i in (1, 2, 3) and j in (1, 2);
+
+drop table t;
+
+set optimizer_full_scan = @orig_optimizer_full_scan;
```
percona-ysorokin: no need to save/restore session vars.

mysql-test/t/optimizer_full_scan.test
```diff
@@ -0,0 +1,70 @@
...
+set optimizer_force_index_for_range = @orig_optimizer_force_index_for_range;
```
percona-ysorokin: no need to save/restore session vars.

sql/sys_vars.cc
```diff
@@ -2752,6 +2752,10 @@ static Sys_var_bool Sys_optimizer_force_index_for_range(
     SESSION_VAR(optimizer_force_index_for_range), CMD_LINE(OPT_ARG),
     DEFAULT(false));
 
+static Sys_var_bool Sys_optimizer_full_scan(
+    "optimizer_full_scan", "Enable full table and index scans.",
+    SESSION_VAR(optimizer_full_scan), CMD_LINE(OPT_ARG), DEFAULT(true));
```
percona-ysorokin: Consider `HINT_UPDATEABLE SESSION_VAR(optimizer_full_scan)`.

percona-ysorokin (changes requested):

inikep:
All comments are addressed and PR was updated.

inikep:
PR was updated

percona-ysorokin (approved):
LGTM

share/errmsg-utf8.txt
```diff
@@ -18728,6 +18728,9 @@ ER_RPL_FAILED_IN_RLI_INIT_INFO
 ER_SEMISYNC_FORCE_SHUTDOWN
    eng "Force shutdown: Semi-sync master is being switched off while there are active un-acked transactions"
 
+ER_FULL_SCAN_DISABLED
```
lth: We should reuse the 5.6 error number: 50077.

inikep:
PR was updated with `ER_FULL_SCAN_DISABLED` moved to 50077

lth (approved):

PR #971, opened 2019-02-26
FB8-72: Added session statistics on number of bytes written to temporary tables data files

dutow:
Summary:

JIRA ticket: https://jira.percona.com/browse/FB8-72

Reference commit: https://github.com/facebook/mysql-5.6/commit/c03842f888b0ddd509c2d112f1b599b7ceaa2f5b
Reference commit: https://github.com/facebook/mysql-5.6/commit/672cf0e

-------- c03842f --------

This change adds a global statistic called 'Tmp_table_bytes_written' on the number of bytes written to temporary tables data files. This was done mostly by checking in the debugger what called my_pwrite/my_write during writes/updates/deletes.

The key files were trickier to handle, because they are only written to on eviction from the key cache, so this wasn't done. I could either make some data structure that mapped file descriptors to whether they belong to temporary tables or not, or I could tag every page in the key cache with a flag, so that I would know what to do on eviction.

-------- 672cf0e --------

The global status variable 'Tmp_table_bytes_written' is changed so
that it is now a session variable.

Also, its value is logged to the slow query log.

Test Plan: mysqltest.sh temp_table_size_counter.test

### DIFF
--- a/storage/innobase/handler/ha_innodb.cc
+++ b/storage/innobase/handler/ha_innodb.cc
@@ -7798,12 +7798,32 @@ int ha_innobase::intrinsic_table_write_row(uchar *record) {
       convert_error_code_to_mysql(err, m_prebuilt->table->flags, m_user_thd));
 }
 
+static ulint get_temp_tablespace_size(const ibt::Tablespace *ts) noexcept {
+  if (ts == nullptr) {
+    return 0;
+  }
+  fil_space_t *space = fil_space_get(ts->space_id());
+  if (space == nullptr) {
+    return 0;
+  }
+  const page_size_t page_size(space->flags);
+  return space->size * page_size.physical();
+}
+
+void ha_innobase::inc_tmp_table_bytes_written(ulint by_value) noexcept {
+  if (m_user_thd) m_user_thd->status_var.tmp_table_bytes_written += by_value;
+}
+
 /** Stores a row in an InnoDB database, to the table specified in this
  handle.
  @return error code */
 
 int ha_innobase::write_row(uchar *record) /*!< in: a row in MySQL format */
 {
+  const auto session = thd_to_innodb_session(m_user_thd);
+  const ulint temp_table_start_size =
+      get_temp_tablespace_size(session->get_usr_temp_tblsp_if_exists());
+
   dberr_t error;
   int error_result = 0;
   bool auto_inc_used = false;
@@ -7992,6 +8012,13 @@ report_error:
 func_exit:
 
   innobase_active_small();
+
+  const ulint temp_table_end_size =
+      get_temp_tablespace_size(session->get_usr_temp_tblsp_if_exists());
+
+  if (temp_table_end_size > temp_table_start_size) {
+    inc_tmp_table_bytes_written(temp_table_end_size - temp_table_start_size);
+  }
 
   DBUG_RETURN(error_result);
 }
@@ -8394,6 +8421,10 @@ if its index columns are updated!
 int ha_innobase::update_row(const uchar *old_row, uchar *new_row) {
   int err;
 
+  const auto session = thd_to_innodb_session(m_user_thd);
+  const ulint temp_table_start_size =
+      get_temp_tablespace_size(session->get_usr_temp_tblsp_if_exists());
+
   dberr_t error;
   trx_t *trx = thd_to_trx(m_user_thd);
   ib_uint64_t new_counter = 0;
@@ -8538,6 +8569,13 @@ func_exit:
 
   innobase_active_small();
 
+  const ulint temp_table_end_size =
+      get_temp_tablespace_size(session->get_usr_temp_tblsp_if_exists());
+
+  if (temp_table_end_size > temp_table_start_size) {
+    inc_tmp_table_bytes_written(temp_table_end_size - temp_table_start_size);
+  }
+
   DBUG_RETURN(err);
 }
 
@@ -8547,6 +8585,10 @@ func_exit:
 int ha_innobase::delete_row(
     const uchar *record) /*!< in: a row in MySQL format */
 {
+  const auto session = thd_to_innodb_session(m_user_thd);
+  const ulint temp_table_start_size =
+      get_temp_tablespace_size(session->get_usr_temp_tblsp_if_exists());
+
   dberr_t error;
   trx_t *trx = thd_to_trx(m_user_thd);
   TrxInInnoDB trx_in_innodb(trx);
@@ -8588,6 +8630,13 @@ int ha_innobase::delete_row(
   utility threads: */
 
   innobase_active_small();
+
+  const ulint temp_table_end_size =
+      get_temp_tablespace_size(session->get_usr_temp_tblsp_if_exists());
+
+  if (temp_table_end_size > temp_table_start_size) {
+    inc_tmp_table_bytes_written(temp_table_end_size - temp_table_start_size);
+  }
 
   DBUG_RETURN(
       convert_error_code_to_mysql(error, m_prebuilt->table->flags, m_user_thd));
--- a/mysql-test/suite/innodb/r/temp_table_size_counter.result
+++ b/mysql-test/suite/innodb/r/temp_table_size_counter.result
@@ -1,0 +1,60 @@
+create table t1 (i int, c char(255));
+create table t2 (i int, c varchar(255));
+insert into t1 values (0, lpad('a', 250, 'b'));
+insert into t1 select i+1,c from t1;
+insert into t1 select i+2,c from t1;
+insert into t1 select i+4,c from t1;
+insert into t1 select i+8,c from t1;
+insert into t1 select i+16,c from t1;
+insert into t1 select i+32,c from t1;
+insert into t1 select i+64,c from t1;
+insert into t1 select i+128,c from t1;
+insert into t1 select i+256,c from t1;
+insert into t1 select i+512,c from t1;
+insert into t1 select i+1024,c from t1;
+insert into t1 select i+2048,c from t1;
+insert into t1 select i+4096,c from t1;
+insert into t2 select * from t1;
+set session tmp_table_size=16384;
+set session max_heap_table_size=16384;
+create temporary table tm(i int, c char(255)) engine=innodb;
+include/assert.inc [The above statements should not have increased Tmp_table_bytes_written]
+insert into tm select * from t1;
+insert into tm select * from t1;
+insert into tm select * from t1;
+update tm set c = lpad('b', 250, 'a');
+include/assert.inc [The above statements should not have increased Tmp_table_bytes_written]
+drop temporary table tm;
+create temporary table tm(i int, c varchar(255)) engine=innodb;
+include/assert.inc [The above statements should not have increased Tmp_table_bytes_written]
+insert into tm select * from t2;
+include/assert.inc [The above statements should not have increased Tmp_table_bytes_written]
+update tm set c = lpad('b', 250, 'a');
+include/assert.inc [The above statements should not have increased Tmp_table_bytes_written]
+delete from tm where i < 100;
+include/assert.inc [The above statements should not have increased Tmp_table_bytes_written]
+select i, c, count(*) from t1 group by i, c having count(*) > 1;
+i	c	count(*)
+include/assert.inc [The above statements should not have increased Tmp_table_bytes_written]
+select i, c, count(*) from t2 group by i, c having count(*) > 1;
+i	c	count(*)
+include/assert.inc [The above statements should not have notincreased Tmp_table_bytes_written]
+drop temporary table tm;
+include/assert.inc [The above statements should not have increased Tmp_table_bytes_written]
+create temporary table tm(i int, c char(255)) engine=innodb;
+insert into tm select * from t1;
+insert into tm select * from t1;
+insert into tm select * from t1;
+insert into tm select * from t1;
+include/assert.inc [The above statements should not have increased session Tmp_table_bytes_written]
+drop temporary table tm;
+create temporary table tm(i int, c char(255)) engine=innodb;
+insert into tm select * from t1;
+insert into tm select * from t1;
+insert into tm select * from t1;
+insert into tm select * from t1;
+insert into tm select * from t1;
+insert into tm select * from t1;
+include/assert.inc [The above statements should not decrease global Tmp_table_bytes_written]
+drop table t1;
+drop table t2;
--- a/mysql-test/suite/innodb/t/temp_table_size_counter-master.opt
+++ b/mysql-test/suite/innodb/t/temp_table_size_counter-master.opt
@@ -1,0 +1,3 @@
+--innodb-max-dirty-pages-pct=0
+--innodb-max-dirty-pages-pct-lwm=0
+--innodb-adaptive-flushing=0
--- a/mysql-test/suite/innodb/t/temp_table_size_counter.test
+++ b/mysql-test/suite/innodb/t/temp_table_size_counter.test
@@ -1,0 +1,183 @@
+--source include/count_sessions.inc
+
+--let $prevSize = query_get_value(SHOW STATUS LIKE 'Tmp_table_bytes_written', Value, 1)
+
+create table t1 (i int, c char(255));
+create table t2 (i int, c varchar(255));
+
+insert into t1 values (0, lpad('a', 250, 'b'));
+insert into t1 select i+1,c from t1;
+insert into t1 select i+2,c from t1;
+insert into t1 select i+4,c from t1;
+insert into t1 select i+8,c from t1;
+insert into t1 select i+16,c from t1;
+insert into t1 select i+32,c from t1;
+insert into t1 select i+64,c from t1;
+insert into t1 select i+128,c from t1;
+insert into t1 select i+256,c from t1;
+insert into t1 select i+512,c from t1;
+insert into t1 select i+1024,c from t1;
+insert into t1 select i+2048,c from t1;
+insert into t1 select i+4096,c from t1;
+
+insert into t2 select * from t1;
+
+set session tmp_table_size=16384;
+set session max_heap_table_size=16384;
+
+create temporary table tm(i int, c char(255)) engine=innodb;
+
+--let $nextSize = query_get_value(show status like 'Tmp_table_bytes_written', Value, 1)
+--let $assert_text= The above statements should not have increased Tmp_table_bytes_written
+--let $assert_cond= $nextSize = $prevSize
+--source include/assert.inc
+--let $prevSize = $nextSize
+
+insert into tm select * from t1;
+insert into tm select * from t1;
+insert into tm select * from t1;
+
+--let $wait_condition = select variable_value > $prevSize from performance_schema.session_status where variable_name = 'Tmp_table_bytes_written'
+--source include/wait_condition.inc
+--let $prevSize = query_get_value(show status like 'Tmp_table_bytes_written', Value, 1)
+
+update tm set c = lpad('b', 250, 'a');
+
+# The current approach measures file size, and as such, it doesn't result in any increase for this update
+--let $nextSize = query_get_value(show status like 'Tmp_table_bytes_written', Value, 1)
+--let $assert_text= The above statements should not have increased Tmp_table_bytes_written
+--let $assert_cond= $nextSize = $prevSize
+--source include/assert.inc
+--let $prevSize = $nextSize
+
+drop temporary table tm;
+
+# The location where the byte count is increment is different for static and dynamic
+# records. Test with varchar to see if the code path for dynamic records also works. The
+# char version above only tests static records.
+create temporary table tm(i int, c varchar(255)) engine=innodb;
+
+# The above statements should not have increased Tmp_table_bytes_written
+--let $nextSize = query_get_value(SHOW STATUS LIKE 'Tmp_table_bytes_written', Value, 1)
+--let $assert_text= The above statements should not have increased Tmp_table_bytes_written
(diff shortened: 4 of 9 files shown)


mysql-test/t/temp_table_size_counter.test
```diff
@@ -0,0 +1,186 @@
...
+set session max_heap_table_size=16384;
+
+create temporary table tm(i int, c char(255)) engine=myisam;
+
+--let MYSQLD_TMPDIR = `select @@tmpdir`
+
+--let EXPECTED_FILE_SIZE = 0
+--source include/temp_table_size_counter_perl.inc
+
+# The above statements should not have increased Tmp_table_bytes_written
```
laurynas-biveinis: assert.inc here and in many places below

mysql-test/t/temp_table_size_counter.test
```diff
@@ -0,0 +1,186 @@
+--let $prevSize = query_get_value(SHOW STATUS LIKE 'Tmp_table_bytes_written', Value, 1)
```
laurynas-biveinis: --source include/count_sessions.inc and counterpart at the end

storage/myisam/mi_dynrec.cc
```diff
@@ -812,6 +815,7 @@ static int update_dynamic_record(MI_INFO *info, my_off_t filepos, uchar *record,
               if (info->s->file_write(info, (uchar *)del_block.header, 20,
                                       next_pos, MYF(MY_NABP)))
                 DBUG_RETURN(1);
+	      info->data_file_written+= MI_MAX_DYN_BLOCK_HEADER;
```
laurynas-biveinis: clang-format

storage/myisam/mi_statrec.cc
```diff
@@ -46,6 +46,7 @@ int _mi_write_static_record(MI_INFO *info, const uchar *record) {
     if (info->s->file_write(info, record, info->s->base.reclength, filepos,
                             MYF(MY_NABP)))
       goto err;
+    info->data_file_written+=info->s->base.reclength;
```
laurynas-biveinis: clang-format

storage/myisam/myisamdef.h
```diff
@@ -319,6 +319,7 @@ struct MI_INFO {
   THR_LOCK_DATA lock;
   uchar *rtree_recursion_state; /* For RTREE */
   int rtree_recursion_depth;
+   uint data_file_written; /* Track bytes written to data file */
```
laurynas-biveinis: clang-format

laurynas-biveinis (changes requested):

mysql-test/include/temp_table_size_counter_perl.inc
```diff
@@ -0,0 +1,22 @@
...
+my $found = 0;
+
+opendir(DIR, $dir) or die $!;
+while (my $file = readdir(DIR))
+{
+  next unless ($file =~ m/\.MYD$/);
+  $found = 1;
+  $size = -s "$dir/$file";
+  if ($size != $ENV{'EXPECTED_FILE_SIZE'}) {
+    print "Size size mismatch: $size != $ENV{'EXPECTED_FILE_SIZE'}\n";
```
percona-ysorokin: Typo
```suggestion
    print "Size mismatch: $size != $ENV{'EXPECTED_FILE_SIZE'}\n";
```

mysql-test/t/temp_table_size_counter.test
percona-ysorokin: Add `--source include/have_myisam.inc`

mysql-test/t/temp_table_size_counter.test
```diff
@@ -0,0 +1,186 @@
...
+
+--let MYSQLD_TMPDIR = `select @@tmpdir`
+
+--let EXPECTED_FILE_SIZE = 0
+--source include/temp_table_size_counter_perl.inc
+
+# The above statements should not have increased Tmp_table_bytes_written
+--let $nextSize = query_get_value(SHOW STATUS LIKE 'Tmp_table_bytes_written', Value, 1)
+--disable_query_log
+--eval select $nextSize = $prevSize as "no increase"
```
percona-ysorokin: Consider an assertion instead `--source include/assert.inc`

mysql-test/t/temp_table_size_counter.test
```diff
@@ -0,0 +1,186 @@
...
+--enable_query_log
+
+insert into tm select * from t1;
+
+--let $prevSize = $nextSize
+--let $nextSize = query_get_value(SHOW STATUS LIKE 'Tmp_table_bytes_written', Value, 1)
+--disable_query_log
+--eval select $nextSize > $prevSize as "increased"
```
percona-ysorokin: Consider an assertion instead `--source include/assert.inc`

mysql-test/t/temp_table_size_counter.test
```diff
@@ -0,0 +1,186 @@
...
+
+--let EXPECTED_FILE_SIZE = `select $nextSize - $prevSize`
+--source include/temp_table_size_counter_perl.inc
+
+update tm set c = lpad('b', 250, 'a');
+
+--let $prevSize = $nextSize
+--let $nextSize = query_get_value(SHOW STATUS LIKE 'Tmp_table_bytes_written', Value, 1)
+--disable_query_log
+--eval select $nextSize > $prevSize as "increased"
```
percona-ysorokin: Consider an assertion instead `--source include/assert.inc`

mysql-test/t/temp_table_size_counter.test
```diff
@@ -0,0 +1,186 @@
...
+--enable_query_log
+
+delete from tm where i < 100;
+
+--let $prevSize = $nextSize
+--let $nextSize = query_get_value(SHOW STATUS LIKE 'Tmp_table_bytes_written', Value, 1)
+--disable_query_log
+--eval select $nextSize > $prevSize as "increased"
```
percona-ysorokin: Consider an assertion instead `--source include/assert.inc`

mysql-test/t/temp_table_size_counter.test
```diff
@@ -0,0 +1,186 @@
...
+create temporary table tm(i int, c varchar(255)) engine=myisam;
+
+--let EXPECTED_FILE_SIZE = 0
+--source include/temp_table_size_counter_perl.inc
+
+# The above statements should not have increased Tmp_table_bytes_written
+--let $prevSize = $nextSize
+--let $nextSize = query_get_value(SHOW STATUS LIKE 'Tmp_table_bytes_written', Value, 1)
+--disable_query_log
+--eval select $nextSize = $prevSize as "no increase"
```
percona-ysorokin: Consider an assertion instead `--source include/assert.inc`

mysql-test/t/temp_table_size_counter.test
```diff
@@ -0,0 +1,186 @@
...
+--enable_query_log
+
+insert into tm select * from t2;
+
+--let $prevSize = $nextSize
+--let $nextSize = query_get_value(SHOW STATUS LIKE 'Tmp_table_bytes_written', Value, 1)
+--disable_query_log
+--eval select $nextSize > $prevSize as "increased"
```
percona-ysorokin: Consider an assertion instead `--source include/assert.inc`

mysql-test/t/temp_table_size_counter.test
```diff
@@ -0,0 +1,186 @@
...
+
+--let EXPECTED_FILE_SIZE = `select $nextSize - $prevSize`
+--source include/temp_table_size_counter_perl.inc
+
+update tm set c = lpad('b', 250, 'a');
+
+--let $prevSize = $nextSize
+--let $nextSize = query_get_value(SHOW STATUS LIKE 'Tmp_table_bytes_written', Value, 1)
+--disable_query_log
+--eval select $nextSize > $prevSize as "increased"
```
percona-ysorokin: Consider an assertion instead `--source include/assert.inc`

mysql-test/t/temp_table_size_counter.test
```diff
@@ -0,0 +1,186 @@
...
+--enable_query_log
+
+delete from tm where i < 100;
+
+--let $prevSize = $nextSize
+--let $nextSize = query_get_value(SHOW STATUS LIKE 'Tmp_table_bytes_written', Value, 1)
+--disable_query_log
+--eval select $nextSize > $prevSize as "increased"
```
percona-ysorokin: Consider an assertion instead `--source include/assert.inc`

mysql-test/t/temp_table_size_counter.test
```diff
@@ -0,0 +1,186 @@
...
+--enable_query_log
+
+# Test implicit tables.
+select i, c, count(*) from t1 group by i, c having count(*) > 1;
+
+--let $prevSize = $nextSize
+--let $nextSize = query_get_value(SHOW STATUS LIKE 'Tmp_table_bytes_written', Value, 1)
+--disable_query_log
+--eval select $nextSize > $prevSize as "increased"
```
percona-ysorokin: Consider an assertion instead `--source include/assert.inc`

mysql-test/t/temp_table_size_counter.test
```diff
@@ -0,0 +1,186 @@
...
+--enable_query_log
+
+select i, c, count(*) from t2 group by i, c having count(*) > 1;
+
+--let $prevSize = $nextSize
+--let $nextSize = query_get_value(SHOW STATUS LIKE 'Tmp_table_bytes_written', Value, 1)
+--disable_query_log
+--eval select $nextSize > $prevSize as "increased"
```
percona-ysorokin: Consider an assertion instead `--source include/assert.inc`

mysql-test/t/temp_table_size_counter.test
```diff
@@ -0,0 +1,186 @@
...
+--enable_query_log
+
+drop table t1;
+drop table t2;
+drop temporary table tm;
+
+--let $prevSize = $nextSize
+--let $nextSize = query_get_value(SHOW STATUS LIKE 'Tmp_table_bytes_written', Value, 1)
+--disable_query_log
+--eval select $nextSize = $prevSize as "no increase"
```
percona-ysorokin: Consider an assertion instead `--source include/assert.inc`

mysql-test/t/temp_table_size_counter.test
```diff
@@ -0,0 +1,186 @@
...
+# Second, do some work on default connection.
+connection default;
+insert into tm values (1, 'asdf');
+
+# Third, check variables on con1
+connection con1;
+--let $con1NextSize = query_get_value(SHOW STATUS LIKE 'Tmp_table_bytes_written', Value, 1)
+--let $globalNextSize = query_get_value(SHOW GLOBAL STATUS LIKE 'Tmp_table_bytes_written', Value, 1)
+--disable_query_log
+--eval select $con1NextSize = $con1PrevSize as "no increase"
```
percona-ysorokin: Consider an assertion instead `--source include/assert.inc`

mysql-test/t/temp_table_size_counter.test
```diff
@@ -0,0 +1,186 @@
...
+--eval select $globalNextSize > $globalPrevSize as "increased"
```
percona-ysorokin: Consider an assertion instead `--source include/assert.inc`

mysql-test/t/temp_table_size_counter.test
```diff
@@ -0,0 +1,186 @@
...
+# Record global status in default connection.
+--let $globalPrevSize = query_get_value(SHOW GLOBAL STATUS LIKE 'Tmp_table_bytes_written', Value, 1)
+
+# Do work in con1
+connection con1;
+create temporary table tm(i int, c char(255)) engine=myisam;
+insert into tm values (1, 'asdf');
+--let $globalNextSize = query_get_value(SHOW GLOBAL STATUS LIKE 'Tmp_table_bytes_written', Value, 1)
+--disable_query_log
+--eval select $globalNextSize > $globalPrevSize as "increased"
```
percona-ysorokin: Consider an assertion instead `--source include/assert.inc`

mysql-test/t/temp_table_size_counter.test
```diff
@@ -0,0 +1,186 @@
...
+--enable_query_log
+
+disconnect con1;
+--source include/wait_until_disconnected.inc
+
+connection default;
+--let $globalPrevSize = $globalNextSize
+--let $globalNextSize = query_get_value(SHOW GLOBAL STATUS LIKE 'Tmp_table_bytes_written', Value, 1)
+--disable_query_log
+--eval select $globalNextSize >= $globalPrevSize as "no decrease"
```
percona-ysorokin: Consider an assertion instead `--source include/assert.inc`

percona-ysorokin (changes requested):

dutow:
updated.

percona-ysorokin (approved):
LGTM

hermanlee:
Can this be rebased on to the latest 8.0.13? Thanks!

hermanlee (changes requested):
Given 8.0.16 will always use InnoDB for temp tables, could this patch be modified to track bytes written there? It looks like session temporary tablespaces are create in .ibt files. Maybe we could track logical bytes written to these space_ids using fsp_is_system_temporary()? 

We still want to keep tmp_table_bytes_written as the statistic being returned though.

dutow:
@hermanlee a few questions regarding that:

* Do you want to also keep the MyISAM support?
* By keeping `tmp_table_bytes_written` as the statistics, you mean I shouldn't rename it to `innodb_tmp_table_bytes_written`, like most other innodb statistics?

hermanlee:
Let's keep the myisam support until myisam gets dropped. This is in case specific applications switch to myisam for the tmp table.

Yes, keeping the tmp_table_bytes_written variable would be great. But if it makes more sense to use innodb_tmp_table_bytes_written for the innodb portion, I can update our scripts to aggregate the two.

dutow:
@hermanlee One more question: currently, for myisam, this diff measures the exact disk writes per session.

For innodb, I think this isn't possible.

* We can try to do some guesswork in the actual session's thread - possibly before actually writing anything to the disk, and possibly resulting in an inaccurate number. I wouldn't recommend this.
* Or we can aggregate the actual size of written data in the io threads, but at that point, we no longer know which session it belongs to (existing statistics, such as innodb_data_written follow this approach: these are global only)

Which leaves us with two options:

* if global statistics is enough, it's simple, I can make a separate global counter for innodb
* if the session part of this feature is important... then we can rely on an existing innodb session mapping (each innodb session has up to two tmp tablespaces: one for user and one for internal temporary tables). Based on this, we can look up the session from the space id, and increment per session statistics when the writes actually happen. Even this won't be 100% correct, as we can't use this hack to handle ibtmp1 (the rollback segments). I can include the ibtmp1 writes in the global counter, but I can't tell which sessions it belongs to, so those changes won't be visible in the session counters.

hermanlee:
The session part is important since we'd want to know which queries are generating large writes to temp tables. Not handling the undo logs for these tables are fine though. I suspect the size of the undo entries will be proportional to the actual temp table data. We're not looking for exact sizes, but the order of magnitude of writes on the current session. Let's go with the hacked approach then, unless the estimates within the session are close enough to the actual bytes written.

dutow:
@hermanlee for this latest update:

* innodb temporary tables are now correctly counted (at least as much as I was able to verify it)
* the innodb test itself is unstable. When I ran it locally, it works, on our jenkins, it fails. (I pushed it primarily to check this) The main issue here is again that I only update the statistics when the writes actually happen, which can be a bit later than the test checks it. I'll try to improve on this.

storage/innobase/buf/buf0buf.cc
```diff
@@ -4893,6 +4897,13 @@ bool buf_page_io_complete(buf_page_t *bpage, bool evict) {
       ut_error;
   }
 
+  if (io_type == BUF_IO_WRITE && fsp_is_system_temporary(bpage->id.space())) {
+    const auto it = innodb_temp_table_thds.find(bpage->id.space());
```
hermanlee: I think you need the lock to call find() since it's not thread-safe? But it would also lead to every temp table write synchronizing on this mutex. Maybe shard the lock based on space_id? Or if there's another way to track writes on the space_id and then update the session variable at a later point without having to acquire a lock here?

dutow: You are right, I missed this. I added this map to avoid a global lock (there is already an existing list which requires that in the codebase), but I think I can fix it to work without it. Will update the PR later today.

storage/innobase/include/srv0tmp.h
```diff
@@ -176,7 +176,24 @@ class Tablespace_pool {
     release();
   }
 
+  THD *get_thd_for_ts(space_id_t id) const noexcept {
+    /* No lock needed:
+
+       the map used is only modified in expand(), where first a new map is
```
hermanlee: I don't think this works either. innodb_temp_table_thds is a unique ptr. When expand() is called, the old one is deallocated. What happens if it gets deallocated while the find is running on the old map?

It seems this might be difficult to do without acquiring a lock to a global map in the I/O threads. @lth who worked on the original change has a suggestion for returning the size of the temp_table at the end of the query vs. returning bytes written. This is mainly a proxy for detecting slow queries that are trying to create large temp table files. If it's easier for the thd to pull out the size of the filespace after the query and record that in the session variable, we can do that too.

Thoughts?

dutow: That would work, as that information is already available and used in information_schema, could be easily added to any session variable  - but it would be a different number. Shouldn't it be a different variable then?

Or I still think this can be fixed - I wanted to simplify it with a single unique ptr and a map, but I missed this problem. I also missed that tablespace ids here are increasing from a reserved range, so they can't be any value, which simplifies things further.

We know the possible maximum size of this structure (the maximum of max_connections, or the size of the reserved tablespace range), so instead of a map, I could use a fixed std::array, which then can be modified without issues. The downside is fixed memory usage, which is quite high for servers with a few connections (1600KB). We could do a two-step version, let's say 1000 and max, only allocating the larger structure when we reach 1000 connections, and never free any. (There are some other structures in the codebase that do this, for the same reason. a few megabytes doesn't really matter if you already have more than 1000 connections)

(tablespace ids start with a fixed id and then increase by one. But we might have holes, as we only care about user tables).

hermanlee: With connection pooling, servers don't usually have many connections. So allocating a large map to track the space_id usage is probably not great. There are also possibly many small use cases of mysql that use temp tables, so a few megabytes of memory adds up.

Let's go with the size idea. We could record the before/after sizes of the temp tables and use that as "bytes" written. We lose the bytes written on updates, but that would probably be fine for now.

mysql-test/include/temp_table_size_counter_perl.inc
```diff
@@ -0,0 +1,22 @@
+--perl
+my $dir = $ENV{'MYSQLD_TMPDIR'};
+my $size;
+my $found = 0;
+
+opendir(DIR, $dir) or die $!;
+while (my $file = readdir(DIR))
+{
+  next unless ($file =~ m/\.MYD$/);
```
lth: How is this even working if we're using innodb storage engines for temptables?  Shouldn't it be ibt files?

storage/innobase/handler/ha_innodb.cc
```diff
@@ -7798,12 +7798,32 @@ int ha_innobase::intrinsic_table_write_row(uchar *record) {
...
+static ulint get_temp_tablespace_size(const ibt::Tablespace *ts) noexcept {
+  if (ts == nullptr) {
+    return 0;
+  }
+  fil_space_t *space = fil_space_get(ts->space_id());
+  if (space == nullptr) {
+    return 0;
+  }
+  const page_size_t page_size(space->flags);
+  return space->size * page_size.physical();
```
lth, line 7810: Do you know if innodb temporary tablespaces shrink after a query finishes execution? If it doesn't, then we could be miscounting. It would be nice to have a test case that proves this.

lth (commented):
I wonder if it makes sense to drop the MyISAM parts because in 8.0.16, it is no longer possible to create temporary tables with the MyISAM engine.

storage/innobase/handler/ha_innodb.cc
dutow: We are miscounting, as anything that doesn't increment the file size (e.g. many updates, delete/drop, etc) isn't counted.

Otherwise, I don't think the file size decreases during the lifetime of a session, and that's why this approximation should work.

mysql-test/include/temp_table_size_counter_perl.inc
dutow: It doesn't. But also, this will never work as it did for myisam: we have a buffer pool with a larger initial size, and then it increases in larger chunks. That's why I don't call it from the ib version of the test.

dutow:
Re: dropping myisam. I asked @hermanlee earlier about that, as I wanted to drop it. But his opinion was that we should keep it, as this PR is for 8.0.13 currently.

hermanlee:
Yes, given temp table for myisam is dropped from 8.0.16, I think we should remove the stats at this point. The main reason I was keeping it was for configs which could use myisam as a temp table.

storage/innobase/handler/ha_innodb.cc
hermanlee: With some testing, it looks like the tablesize gets reset when the session is closed and the temp table is deleted. But while the table/connection is active, it seems to behave like other tablespace, so the space is not reclaimed from the file on deletion, nor overwrites.

hermanlee:
We're going to resolve this one separately and can close this PR.

PR #972, opened 2019-02-27
FB8-120: Added db_metadata options to create database command

dutow:
JIRA ticket: https://jira.percona.com/browse/FB8-120

Reference commit: https://github.com/facebook/mysql-5.6/commit/2f08e3a
Reference commit: https://github.com/facebook/mysql-5.6/commit/a21e84f
Reference commit: https://github.com/facebook/mysql-5.6/commit/5983ba3181a
Reference commit: https://github.com/facebook/mysql-5.6/commit/414870abeee

-------- 2f08e3a --------

Added new option db_metadata to the create database
`CREATE DATABASE [IF NOT EXISTS] db_name DB_METADATA [=] metadata`

The new option is stored in the db.opt file along with other database options like charset and read_only.
The options accepts any string value. If not specified, it writes empty string as the default value in the db.opt file.
To retrieve the option value for a particular database, the show command can be used as follows:
`SHOW CREATE DATABASE db_name`

Example Usage:

create database test2 charset utf8 read_only = true db_metadata '{\'shard_name\':\'myshard_for_test2\'}';
show create database test2;

Output:

+----------+--------------------------------------------------------------------------------------------------------------------------+
| Database | Create Database                                                                                                          |
+----------+--------------------------------------------------------------------------------------------------------------------------+
| test2    | CREATE DATABASE `test2` /*!40100 DEFAULT CHARACTER SET utf8 READ_ONLY DB_METADATA {'shard_name':'myshard_for_test2'} */  |
+----------+--------------------------------------------------------------------------------------------------------------------------+


sql/dd/impl/types/schema_impl.cc
```diff
@@ -106,6 +106,13 @@ bool Schema_impl::set_options_raw(const String_type &options_raw) {
   return false;
 }
 
+bool Schema_impl::set_db_metadata(const String_type &metadata) {
+  m_metadata = metadata;
+  return true;
+}
+
+String_type const &Schema_impl::get_db_metadata() const { return m_metadata; }
```
percona-ysorokin:
```suggestion
String_type const &Schema_impl::get_db_metadata() const noexcept { return m_metadata; }
```

sql/dd/impl/types/schema_impl.h
```diff
@@ -127,6 +128,9 @@ class Schema_impl : public Entity_object_impl, public Schema {
 
   virtual bool set_options_raw(const String_type &options_raw);
 
+  virtual bool set_db_metadata(const String_type &metadata) override;
+  virtual String_type const &get_db_metadata() const override;
```
percona-ysorokin:
```suggestion
  virtual String_type const &get_db_metadata() const noexcept override;
```

sql/dd/impl/types/schema_impl.h
```diff
@@ -160,9 +164,9 @@ class Schema_impl : public Entity_object_impl, public Schema {
     sprintf(outbuf,
             "SCHEMA OBJECT: id= {OID: %lld}, name= %s, "
             "collation_id={OID: %lld},"
-            "m_created= %llu, m_last_altered= %llu",
+            "m_created= %llu, m_last_altered= %llu, m_metadata=%s",
             id(), name().c_str(), m_default_collation_id, m_created,
-            m_last_altered);
+            m_last_altered, m_metadata.c_str());
```
percona-ysorokin: Potential buffer overflow here with very long metadata.
Consider `snprintf()`.

sql/dd/types/schema.h
```diff
@@ -110,6 +110,9 @@ class Schema : virtual public Entity_object {
 
   virtual bool set_options_raw(const String_type &options_raw) = 0;
 
+  virtual bool set_db_metadata(const String_type &metadata) = 0;
+  virtual String_type const &get_db_metadata() const = 0;
```
percona-ysorokin:
```suggestion
  virtual String_type const &get_db_metadata() const noexcept = 0;
```

sql/sql_yacc.yy
```diff
@@ -6071,6 +6076,35 @@ read_only_opt:
...
+            if ($3.length > 0)
+            {
+              boost::property_tree::ptree db_metadata_root;
+              std::istringstream is($3.str);
+              try
+              {
+                boost::property_tree::json_parser::read_json(is,
+                                                             db_metadata_root);
+              }
+              catch (std::exception)
```
percona-ysorokin:
```suggestion
              catch (const std::exception&)
```

sql/dd/impl/types/schema_impl.h
```diff
@@ -66,7 +66,8 @@ class Schema_impl : public Entity_object_impl, public Schema {
       : m_created(0),
         m_last_altered(0),
         m_default_collation_id(INVALID_OBJECT_ID),
-        m_options(new Properties_impl()) {}
+        m_options(new Properties_impl()),
+        m_metadata("") {}
```
percona-ysorokin:
```suggestion
        m_metadata() {}
```

sql/dd/dd_schema.cc
```diff
@@ -58,14 +58,15 @@ bool schema_exists(THD *thd, const char *schema_name, bool *exists) {
...
-                   const CHARSET_INFO *charset_info) {
+                   const CHARSET_INFO *charset_info, const char *db_metadata) {
   // Create dd::Schema object.
   std::unique_ptr<dd::Schema> schema(dd::create_object<dd::Schema>());
 
   // Set schema name and collation id.
   schema->set_name(schema_name);
   DBUG_ASSERT(charset_info);
   schema->set_default_collation_id(charset_info->number);
+  schema->set_db_metadata(db_metadata ? db_metadata : "");
```
percona-ysorokin: no implicit conversions
```suggestion
  schema->set_db_metadata(db_metadata != nullptr ? db_metadata : "");
```

percona-ysorokin (changes requested):
Mostly stylistic comments + one potential buffer overflow problem.

Please, also extend commit message with some info that in 8.0 database metadata is stored in Data Dictionary rather than in an `.opt` file.

mysql-test/t/db_metadata.test
```diff
@@ -0,0 +1,163 @@
+# Test per-database database-metadata attribute
+
+--disable_query_log
```
laurynas-biveinis: I don't think there is need to disable query log here, please check

mysql-test/t/db_metadata.test
```diff
@@ -0,0 +1,163 @@
...
+create database test13 db_metadata = '{"sha\'\\"rd":"test13\'_sh\\"ard"}';
+show create database test13;
+--let $db_create=query_get_value(show create database test13, 'Create Database', 1)
+drop database test13;
+eval $db_create;
+show create database test13;
+
+
+# cleanup
+--disable_warnings
```
laurynas-biveinis: Is the warning disable required here? What are the warnings?

sql/sql_yacc.yy
```diff
@@ -1236,6 +1239,7 @@ void warn_about_deprecated_national(THD *thd)
 %token<keyword> GTID_SYM
 %token<keyword> GTID_EXECUTED                 /* MYSQL */
 %token<keyword> SUPER_READ_ONLY_SYM           /* MYSQL */
+%token  DB_METADATA_SYM
```
laurynas-biveinis, line 1242: 8.0 has some attempts to have parser token ID (and resulting query digest) stability, please check: might be a good idea to skip 100 IDs and start a FB-specific section

laurynas-biveinis (changes requested):

sql/sql_yacc.yy
dutow: The 3 tokens above this are also added by facebook - I simply added this one to the end of the existing list.

dutow:
updated

sql/sql_yacc.yy
laurynas-biveinis: Are you sure? All three look like MySQL ones

dutow: There is a comment at line 1236: "Tokens from FB MySQL"

laurynas-biveinis: You are right, sorry. And there is no provision to skip IDs, thus everything already is as good as it could be

percona-ysorokin (approved):
LGTM

lth:
You might want to follow how the DB read only feature was done, because I believe by changing the schema, you need to bump some data dictionary version or something. The FIELD_OPTIONS column is meant to be a generic place to add key/values, and I'm wondering if that's a better place for this.

https://github.com/facebook/mysql-5.6/commit/b9ec08aa20b70dcf80d64c3582def400097baf5a

dutow:
Updated: I removed the additional columns from the dd, and used the options field. Tests pass now.

share/errmsg-utf8.txt
```diff
@@ -18728,6 +18728,15 @@ ER_RPL_FAILED_IN_RLI_INIT_INFO
 ER_SEMISYNC_FORCE_SHUTDOWN
    eng "Force shutdown: Semi-sync master is being switched off while there are active un-acked transactions"
 
+ER_DB_METADATA_TOO_LONG
```
hermanlee: This should replace ER_PLACEHOLDER_50063.

share/errmsg-utf8.txt
```diff
@@ -18728,6 +18728,15 @@ ER_RPL_FAILED_IN_RLI_INIT_INFO
...
+  eng "Metadata for the database is too long. Max length is %d bytes"
+
+ER_DB_METADATA_INVALID_JSON
```
hermanlee: Replace ER_PLACEHOLDER_50064

share/errmsg-utf8.txt
```diff
@@ -18728,6 +18728,15 @@ ER_RPL_FAILED_IN_RLI_INIT_INFO
...
+  eng "Invalid JSON for DB_METADATA attribute: %s."
+
+ER_DB_METADATA_READ_ERROR
```
hermanlee: Replace ER_PLACEHOLDER_50065

sql/dd/impl/types/schema_impl.h
```diff
@@ -127,6 +128,9 @@ class Schema_impl : public Entity_object_impl, public Schema {
 
   virtual bool set_options_raw(const String_type &options_raw);
 
+  virtual bool set_db_metadata(const String_type &metadata) override;
```
hermanlee: We also compile with -Winconsistent-missing-override, so adding this override cause the compiler to warn about the rest of the object:

```
In file included from ../../sql/dd/impl/bootstrapper.cc:58:
../../sql/dd/impl/types/schema_impl.h:75:31: error: 'object_table' overrides a member function but is not marked 'override' [-Werror,-Winconsistent-missing-override]
  virtual const Object_table &object_table() const;
../../sql/dd/impl/types/weak_object_impl.h:50:31: note: overridden virtual function is here
  virtual const Object_table &object_table() const = 0;
                              ^
In file included from ../../sql/dd/impl/bootstrapper.cc:58:
../../sql/dd/impl/types/schema_impl.h:77:16: error: 'validate' overrides a member function but is not marked 'override' [-Werror,-Winconsistent-missing-override]
  virtual bool validate() const;
               ^
../../sql/dd/impl/types/weak_object_impl.h:52:16: note: overridden virtual function is here
  virtual bool validate() const = 0;
               ^
```

sql/sql_const.h
```diff
@@ -286,6 +286,9 @@ static const ulong EVENT_DEF_CACHE_MIN = 256;
 */
 #define MAX_TIME_ZONE_NAME_LENGTH (NAME_LEN + 1)
 
+#define DB_METADATA_MAX_LENGTH 1024 /* Max length for the db-metadata option \
```
hermanlee: Seems like "*/" should also fit onto this line too?

sql/dd/upgrade/schema.cc
```diff
@@ -155,7 +155,7 @@ bool migrate_schema_to_dd(THD *thd, const char *dbname) {
   // Disable autocommit option
   Disable_autocommit_guard autocommit_guard(thd);
 
-  if (dd::create_schema(thd, schema_name, schema_charset)) {
+  if (dd::create_schema(thd, schema_name, schema_charset, nullptr)) {
```
hermanlee, line 158: In 5.6, we store the db_metadata information in the db_opt file using the keyword `db-metadata`. Since there are in-place upgrade paths planned to 8.0, it should be fine to have nullptr here.

sql/dd/impl/types/schema_impl.cc
```diff
@@ -106,6 +106,25 @@ bool Schema_impl::set_options_raw(const String_type &options_raw) {
   return false;
 }
 
+bool Schema_impl::set_db_metadata(const String_type &metadata) {
+  if (metadata.empty()) {
+    if (options().exists("db_metadata")) {
```
hermanlee: Can "db_metadata" be made into a constant string, similar to `read_only_options_key`?

hermanlee (commented):
Code looks good overall. Just some small minor comments.

hermanlee (changes requested):
Requesting changes for the db_metadata key.

dutow:
Updated.

hermanlee (approved):

PR #975, opened 2019-02-28
Bug#84958 -  InnoDB's MVCC has O(N^2) behaviors

satya-bodapati:
(Also fixes PS-4712 - Assertion after bug#84958 fix on FB tree)

Problem:

If there're multiple row versions in InnoDB, reading one row
from PK may have O(N) complexity and reading from secondary keys may have
O(N^2) complexity.

Fix:

The problem occurs when there are many pending versions of the same row,
meaning that the primary key is the same, but a secondary key is different.
The slowdown occurs when the secondary index is traversed. This patch
creates a helper class for the function row_sel_get_clust_rec_for_mysql()
which can remember and re-use cached_clust_rec & cached_old_vers so that
rec_get_offsets() does not need to be called over and over for the
clustered record.

PS-4712 : Assertion after bug#84958 fix on FB tree

Problem:

The function row_sel_build_prev_vers_for_mysql() goes through the old versions
of the clustered index record and identifies the one which can be viewed by the
current transaction (MVCC).  The result is then cached in an object of
Row_sel_get_clust_rec_for_mysql.

The two things that are cached by the helper class
Row_sel_get_clust_rec_for_mysql are cached_clust_rec and cached_old_vers.  If
the latest record is cached_clust_rec, its old version for the given
transaction that is viewable is cached_old_vers. The important point here is
that the record offsets for these two records are not cached.

The record offsets for cached_clust_rec and the cached_old_vers need not be the
same.  This is confirmed by looking at row_sel_build_prev_vers_for_mysql().
The 6th argument which contains the record offsets is an input/output argument.
The input is the offsets of the latest version of clust_rec, and the output is
the record offsets of the older version.  So the record offsets of different
versions of clust_rec can change.

Solution:

Re-calculate the record offsets for the cached old_vers.

### DIFF
--- a/storage/innobase/row/row0sel.cc
+++ b/storage/innobase/row/row0sel.cc
@@ -3021,35 +3021,90 @@ row_sel_build_prev_vers_for_mysql(
 	return(err);
 }
 
-/*********************************************************************//**
-Retrieves the clustered index record corresponding to a record in a
+/** Helper class to cache clust_rec and old_ver bug 84958 */
+
+class
+Row_sel_get_clust_rec_for_mysql
+{
+	const rec_t*	cached_clust_rec;
+	rec_t*		cached_old_vers;
+public:
+	Row_sel_get_clust_rec_for_mysql() :
+		cached_clust_rec(NULL),
+		cached_old_vers(NULL) {}
+
+	/** Retrieves the clustered index record corresponding to a record in a
+	non-clustered index. Does the necessary locking. Used in the MySQL
+	interface.
+	@param[in]	prebuilt	prebuilt struct in the handle
+	@param[in]	sec_index	secondary index where rec resides
+	@param[in]	rec		record in a non-clustered index; if
+					this is a locking read, then rec is not
+					allowed to be delete-marked, and that would
+					not make sense either
+	@param[in]	thr		query thread
+	@param[out]	out_rec		clustered record or an old version of
+					it, NULL if the old version did not exist
+					in the read view, i.e., it was a fresh
+					inserted version
+	@param[in,out]	offsets		offsets returned by
+					rec_get_offsets(rec, sec_index);
+					out: offsets returned by
+					rec_get_offsets(out_rec, clust_index)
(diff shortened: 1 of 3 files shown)


laurynas-biveinis:
@satya-bodapati , so what is the difference in this patch from the previous reverted version? What is the extra fix? Has it been backported from 8.0?

satya-bodapati:
@laurynas-biveinis  Extra fix is this. and yes it is backported from 8.0
```
diff --git a/storage/innobase/row/row0sel.cc b/storage/innobase/row/row0sel.cc
index b0cd1ae90c3..bb19b402a93 100644
--- a/storage/innobase/row/row0sel.cc
+++ b/storage/innobase/row/row0sel.cc
@@ -3208,6 +3208,9 @@ Row_sel_get_clust_rec_for_mysql::operator()(
                        } else {
                                err = DB_SUCCESS;
                                old_vers = cached_old_vers;
+                               if (old_vers != NULL) {
+                                *offsets = rec_get_offsets(old_vers, clust_index, *offsets, ULINT_UNDEFINED, offset_heap);
+                               }
                        }

```

storage/innobase/row/row0sel.cc
```diff
@@ -3021,14 +3021,51 @@ row_sel_build_prev_vers_for_mysql(
 	return(err);
 }
 
+/*********************************************************************//**
```
laurynas-biveinis: Use new style InnoDB header comments

laurynas-biveinis (commented):
LGTM, just one very minor comment

storage/innobase/row/row0sel.cc
satya-bodapati: Ok, I thought it was done only from 5.7 or 8.0

Let me reformat anyway

satya-bodapati:
ready for review again

satya-bodapati:
@laurynas-biveinis says Ok with the new patch (on slack).

Ready for FB review

PR #980, opened 2019-03-05
FB8-104: enabling subsecond lock_wait_timeout

dutow:
For a first review, I only added the code changes to this pull request for two reasons:
* a full test is still running, and I'll have to check the results (I do see a few non output only failures because timeouts, so looks like I missed some nsec-sec conversion in this diff - I'll fix that and update). And checking output changes / rerecording will take time.
* compared to the 5.6 patch, the code change for 8.0 is huge, and affects the service api - reviewing only the code changes is probably a good idea

Jira ticket: https://jira.percona.com/browse/FB8-104

Reference commit: https://github.com/facebook/mysql-5.6/commit/c21adb7

Currently lock_wait_timeout is an integer type variable which only have
seconds resolution. We would want to extend the variable to double/decimal
type, so it can have milliseconds or microseconds resolution, such as
lock_wait_timeout = 1.23. The integral part should still mean seconds to
keep backward compatibility.

The range of valid values expanded to [ 0 .. 604800 ] to allow timeouts below
1 second; such as lock_wait_timeout = 0.005.  Values that are too small
will effectively mean not waiting for locks.

This update also changes the output of lock_wait_timeout to be decimal;
note the extra .000000 appended to the seconds.

    SELECT @global.lock_wait_timeout;
    @global.lock_wait_timeout
    604800.000000

### DIFF
--- a/components/mysql_server/mysql_backup_lock.h
+++ b/components/mysql_server/mysql_backup_lock.h
@@ -1,24 +1,24 @@
 /* Copyright (c) 2017, Oracle and/or its affiliates. All rights reserved.
 
-   This program is free software; you can redistribute it and/or modify
-   it under the terms of the GNU General Public License, version 2.0,
-   as published by the Free Software Foundation.
+  This program is free software; you can redistribute it and/or modify
+  it under the terms of the GNU General Public License, version 2.0,
+  as published by the Free Software Foundation.
 
-   This program is also distributed with certain software (including
-   but not limited to OpenSSL) that is licensed under separate terms,
-   as designated in a particular file or component or in included license
-   documentation.  The authors of MySQL hereby grant you an additional
-   permission to link the program and your derivative works with the
-   separately licensed software that they have included with MySQL.
+  This program is also distributed with certain software (including
+  but not limited to OpenSSL) that is licensed under separate terms,
+  as designated in a particular file or component or in included license
+  documentation.  The authors of MySQL hereby grant you an additional
+  permission to link the program and your derivative works with the
+  separately licensed software that they have included with MySQL.
 
-   This program is distributed in the hope that it will be useful,
-   but WITHOUT ANY WARRANTY; without even the implied warranty of
-   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-   GNU General Public License, version 2.0, for more details.
+  This program is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License, version 2.0, for more details.
 
-   You should have received a copy of the GNU General Public License
-   along with this program; if not, write to the Free Software
-   Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301  USA */
+  You should have received a copy of the GNU General Public License
+  along with this program; if not, write to the Free Software
+  Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301  USA */
 
 #ifndef MYSQL_BACKUP_LOCK_INCLUDED
 #define MYSQL_BACKUP_LOCK_INCLUDED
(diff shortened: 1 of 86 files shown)


sql/auth/sql_auth_cache.cc
```diff
@@ -3475,6 +3475,8 @@ void shutdown_acl_cache() {
 
 /* Constants used by Acl_cache_lock_guard */
 static const ulong ACL_CACHE_LOCK_TIMEOUT = 3600UL;
+static const ulong ACL_CACHE_LOCK_TIMEOUT_NSEC =
```
percona-ysorokin: `ulong` is not guaranteed to be 64-bit, and therefore may be truncated here as `ACL_CACHE_LOCK_TIMEOUT * 1000000000ULL >2^32`
```suggestion
static constexpr ulonglong ACL_CACHE_LOCK_TIMEOUT_NSEC =
```

sql/sys_vars.cc
```diff
@@ -1992,11 +1992,27 @@ static Sys_var_bool Sys_local_infile("local_infile",
...
     "lock_wait_timeout",
-    "Timeout in seconds to wait for a lock before returning an error.",
-    HINT_UPDATEABLE SESSION_VAR(lock_wait_timeout), CMD_LINE(REQUIRED_ARG),
-    VALID_RANGE(1, LONG_TIMEOUT), DEFAULT(LONG_TIMEOUT), BLOCK_SIZE(1));
+    "Timeout in seconds to wait for a lock before returning an error. "
+    "The argument will be treated as a decimal value with nanosecond "
+    "precision.",
+    // very small nanosecond values will effectively be no waiting
+    HINT_UPDATEABLE SESSION_VAR(lock_wait_timeout_double),
+    CMD_LINE(REQUIRED_ARG), VALID_RANGE(0, LONG_TIMEOUT), DEFAULT(LONG_TIMEOUT),
```
percona-ysorokin, line 2013: Is `0` a valid value now on purpose?

percona-ysorokin (changes requested):

sql/sys_vars.cc
dutow: Yes, to allow values below 1 second.

sql/auth/sql_auth_cache.cc
dutow: Thanks, this was one of the typos. I found another when rereading my changes, looks like there aren't any timeout errors so far with these changes.

Updated, and I'll do another update tomorrow with the test changes.

dutow:
Updated with changed tests.

percona-ysorokin (approved):
LGTM

components/mysql_server/mysql_backup_lock.h
```diff
@@ -1,24 +1,24 @@
...
-
-   This program is distributed in the hope that it will be useful,
-   but WITHOUT ANY WARRANTY; without even the implied warranty of
-   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-   GNU General Public License, version 2.0, for more details.
-
-   You should have received a copy of the GNU General Public License
-   along with this program; if not, write to the Free Software
-   Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301  USA */
+  This program is free software; you can redistribute it and/or modify
```
hermanlee, line 3: Could this indent change be reverted and the original indent spacing be preserved?

hermanlee (changes requested):
I'm seeing a number of test failures due to loading of locking_service.so.

CURRENT_TEST: main.locking_service
mysqltest: At line 34: Query 'CREATE FUNCTION service_get_read_locks  RETURNS INT SONAME "$LOCKING_SERVICE"' failed.
ERROR 1126 (HY000): Can't open shared library 'locking_service.so' (errno: 0 /data/users/herman/rocks-mysql/8.0/_build-8.0-Debug/plugin_output_directory/locking_
service.so: undefined symbol: _Z29acquire_lo)

I think some of the plugins and gunit tests need to change acquire_locking_service_locks to acquire_locking_service_locks_nsec?

hermanlee (approved):

PR #982, opened 2019-03-07
FB8-128: track Semisync_ack Engine_commit time in slow query log

inikep:
Jira issue: https://jira.percona.com/browse/FB8-128

Reference Patch: https://github.com/facebook/mysql-5.6/commit/2dd16b3

Closes https://github.com/facebook/mysql-5.6/pull/681

Originally Reviewed By: lth

### DIFF
--- a/mysql-test/t/slow_log_engine_fields.test
+++ b/mysql-test/t/slow_log_engine_fields.test
@@ -1,0 +1,16 @@
+SET SESSION long_query_time = 0;
+
+CREATE TABLE t1 (A INT);
+INSERT INTO t1 VALUES (1),(2),(3);
+
+--let SEARCH_FILE = $MYSQLTEST_VARDIR/mysqld.1/mysqld-slow.log
+
+# Check if Semisync_ack_time is included in slow log results
+--let SEARCH_PATTERN= Semisync_ack_time:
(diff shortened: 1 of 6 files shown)


mysql-test/t/slow_log_engine_fields.test
```diff
@@ -0,0 +1,16 @@
+SET SESSION long_query_time = 0;
+
+CREATE TABLE t1 (A INT);
+INSERT INTO t1 VALUES (1),(2),(3);
+
+--let SEARCH_FILE = $MYSQLTEST_VARDIR/mysqld.1/mysqld-slow.log
+
+# Check if Semisync_ack_time is included in slow log results
+--let SEARCH_PATTERN= Semisync_ack_time:
+--source include/search_pattern_in_file.inc
```
laurynas-biveinis, line 10: Can there be a pre-existing slow log, resulting in false positives?

inikep: This feature is compiled inside a server so all comments in slow log should have "Semisync_ack_time" or none. For this case "--repeat" doesn't matter.

percona-ysorokin (approved):
LGTM

hermanlee (approved):

PR #987, opened 2019-03-18
FB8-132: NUMA support

dutow:
Jira ticket: https://jira.percona.com/browse/FB8-132

Reference commit: https://github.com/facebook/mysql-5.6/commit/e50c754879828c1d0b4f40cc13475ec7adea9afc
Reference commit: https://github.com/facebook/mysql-5.6/commit/286e975

-------- e50c754 --------

WebScaleSQL Feature: NUMA Support

Credits for research and implementation: Jeremy Cole and Davi Arnaut

This patch provides startup options:
* flush-caches: Flush and purge buffers/caches
* numa-interleave: Run mysqld with its memory interleaved on all CPUs

It also provides a config option:
* innodb_buffer_pool_populate: pre-allocation of buffer pool
memory at start up:
  -- Use MAP_POPULATE if supported (Linux 2.6.23 and higher)
  -- Forced pre-allocation using memset

-------- 286e975 --------

Automation will pass in the right value to this option if
it needs to bind mysqld to a particular socket. That will happen
via systemd scripts.

This is only adding the capability on the mysqld_safe side to be able to
pass in numactl to the server. The turning on of this feature will need more
testing and slow roll, and that will happen in different diffs on the ops repo.

### DIFF
--- a/scripts/mysqld_safe.sh
+++ b/scripts/mysqld_safe.sh
@@ -17,6 +17,8 @@ MYSQLD=
 niceness=0
 mysqld_ld_preload=
 mysqld_ld_library_path=
+flush_caches=0
+numa_nodebind=
 
 # Initial logging status: error log is not open, and not using syslog
 logging=init
@@ -94,6 +96,9 @@ Usage: $0 [OPTIONS]
     timestamps=TYPE          system (ISO 8601 local time), hyphen
                              (hyphenated date a la mysqld 5.6), legacy
                              (legacy non-ISO 8601 mysqld_safe timestamps)
+  --flush-caches             Flush and purge buffers/caches
+  --numa-nodebind            Run mysqld with numa binding to one socket
+
 
 All other options are passed to the mysqld program.
 
@@ -287,7 +292,9 @@ parse_arguments() {
       --syslog) want_syslog=1 ;;
       --skip-syslog) want_syslog=0 ;;
       --syslog-tag=*) syslog_tag="$val" ;;
-      --timezone=*) TZ="$val"; export TZ; ;;
+      --timezone=*) TZ="$val"; export TZ; ;; 
+      --flush-caches) flush_caches=1 ;;
+      --numa-nodebind=*) numa_nodebind="$val" ;;
 
       --help) usage ;;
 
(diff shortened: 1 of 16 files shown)


percona-ysorokin (approved):
LGTM

dutow:
Note: I didn't do much testing - I have no numa capable hardware.

scripts/mysqld_safe.sh
```diff
@@ -17,6 +17,9 @@ MYSQLD=
 niceness=0
 mysqld_ld_preload=
 mysqld_ld_library_path=
+flush_caches=0
+numa_interleave=0
```
laurynas-biveinis: Isn't mysqld_safe.sh numa_interleave option redundant due to 5.7 core server --innodb-numa-interleave?

storage/innobase/handler/ha_innodb.cc
```diff
@@ -19826,6 +19826,13 @@ static MYSQL_SYSVAR_ULONG(
     NULL, 120, 1, 127, 0);
 #endif /* defined UNIV_DEBUG || defined UNIV_PERF_DEBUG */
 
+static MYSQL_SYSVAR_BOOL(
+    buffer_pool_populate, srv_buf_pool_populate,
+    PLUGIN_VAR_NOCMDARG | PLUGIN_VAR_READONLY,
+    "Preallocate (pre-fault) the page frames required for the mapping "
+    "established by the buffer pool memory region. Disabled by default.",
+    NULL, NULL, FALSE);
```
laurynas-biveinis:
```suggestion
    nullptr, nullptr, false);
```

storage/innobase/os/os0proc.cc
```diff
@@ -70,10 +83,25 @@ ulint os_proc_get_number(void) {
 #endif
 }
 
+/****************************************************************/ /**
```
laurynas-biveinis: New style header comment please

storage/innobase/os/os0proc.cc
```diff
@@ -70,10 +83,25 @@ ulint os_proc_get_number(void) {
...
+ Retrieve and compare operating system release.
+ @return	TRUE if the OS release is equal to, or later than release. */
```
laurynas-biveinis:
```suggestion
 @return	true if the OS release is equal to, or later than release. */
```

storage/innobase/os/os0proc.cc
```diff
@@ -70,10 +83,25 @@ ulint os_proc_get_number(void) {
...
+static bool os_compare_release(
+    /*===============*/
+    const char *release) /*!< in: OS release */
+{
+#if defined(UNIV_LINUX) && defined(_GNU_SOURCE)
+  struct utsname name;
+  return (uname(&name) == 0 && strverscmp(name.release, release) >= 0);
+#else
+  return (FALSE);
```
laurynas-biveinis:
```suggestion
  return (false);
```

storage/innobase/os/os0proc.cc
```diff
@@ -158,6 +187,26 @@ void *os_mem_alloc_large(ulint *n) {
...
   }
 #endif
+
+#if OS_MAP_ANON && OS_MAP_POPULATE
+  /* MAP_POPULATE is only supported for private mappings
+  since Linux 2.6.23. */
+  populate = populate && !os_compare_release("2.6.23");
+
+  if (ptr && populate) {
+    fprintf(stderr,
```
laurynas-biveinis: ib::warn()

laurynas-biveinis (changes requested):

hermanlee:
Can this be rebased onto the latest 8.0.13? Thanks!

storage/innobase/srv/srv0start.cc
```diff
@@ -1816,7 +1816,8 @@ dberr_t srv_start(bool create_new_db, const std::string &scan_directories) {
   ib::info(ER_IB_MSG_1130, size, unit, srv_buf_pool_instances, chunk_size,
            chunk_unit);
 
-  err = buf_pool_init(srv_buf_pool_size, srv_buf_pool_instances);
+  err = buf_pool_init(srv_buf_pool_size, srv_buf_pool_populate,
```
hermanlee, line 1819: I'll fix this, but populate and instances need to be swapped.

hermanlee (approved):

PR #990, opened 2019-03-19
FB8-92: Checking for NULL while preparing write set according to binlog_row_image

inikep:
Summary:

Jira issue: https://jira.percona.com/browse/FB8-92

Reference Patch: https://github.com/facebook/mysql-5.6/commit/f3dc487

Field::cmp_binary_offset returns 0 when comparing a NULL text
field and an empty string leading to the field not being logged when the
value is changed from NULL to empty string. This fix first checks if
both AI and BI fields are not NULL before comparing them.

Originally Reviewed By: santoshbanda


percona-ysorokin (approved):
LGTM

hermanlee (approved):

PR #992, opened 2019-03-19
FB8-75: Periodic fsync in select into outfile controlled by sysvar

dutow:
Jira ticket: https://jira.percona.com/browse/FB8-75

Reference commit: https://github.com/facebook/mysql-5.6/commit/14b1086

-------- 14b1086 --------

Added flush and fsync for SELECT INTO OUTFILE.

Two variables are added, GLOBAL variables with SESSION override.
select_into_file_fsync_size, which controls the size of the buffer
after which a fsync will be done. Needs to be a multiple of 1024

After each fsync you can optionally sleep for
select_into_file_fsync_timeout milliseconds.

TODO: Record main.all_persisted_variables +2

### DIFF
--- a/mysql-test/include/have_outfile.inc
+++ b/mysql-test/include/have_outfile.inc
@@ -1,5 +1,5 @@
 let $outfile= `SELECT LOAD_FILE(CONCAT(@tmpdir, '/outfile.test')) LIKE 'Outfile OK%'`;
-remove_file $MYSQLTEST_VARDIR/tmp/outfile.test;
+remove_file $MYSQL_TMP_DIR/outfile.test;
 
 if (!$outfile)
 {
--- a/mysql-test/include/test_outfile.inc
+++ b/mysql-test/include/test_outfile.inc
@@ -1,0 +1,1 @@
+eval select "Outfile OK" into outfile "$MYSQL_TMP_DIR/outfile.test";
--- a/mysql-test/t/select_outfile_fsync_debug.test
+++ b/mysql-test/t/select_outfile_fsync_debug.test
@@ -1,0 +1,47 @@
+-- source include/have_debug.inc
+
+--disable_query_log
+-- source include/test_outfile.inc
+eval set @tmpdir='$MYSQL_TMP_DIR';
+-- source include/have_outfile.inc
+--enable_query_log
+
+# Save the initial number of concurrent sessions
+--source include/count_sessions.inc
+
+SET SESSION select_into_file_fsync_size=262144;
+SET GLOBAL debug = '+d,print_select_file_fsync_stats';
+
+# Create a table and populate it with some data
+CREATE TABLE t2 (a TEXT, b TEXT) ENGINE=INNODB;
+
+# Here we end up with 32768 rows in the table
+--disable_query_log
+INSERT INTO t2 (a, b) VALUES (REPEAT('a', 512), REPEAT('b', 512));
+INSERT INTO t2 (a, b) VALUES (REPEAT('A', 512), REPEAT('B', 512));
+let $i=14;
+while ($i) {
+  INSERT INTO t2 (a, b) SELECT a, b FROM t2;
+  dec $i;
+}
+
+--eval SELECT * INTO OUTFILE "$MYSQL_TMP_DIR/t2.txt" FROM t2;
+--enable_query_log
+
+--let $assert_file= $MYSQLTEST_VARDIR/log/mysqld.1.err
+--let $assert_only_after = CURRENT_TEST: select_outfile_fsync_debug
+--let $assert_count = 1
+--let $assert_select = \[select_to_file\]\[fsync_count\] 128
+--let $assert_text = Found expected number of select_to_file
+--source include/assert_grep.inc
+
+DROP TABLE t2;
+
+SET GLOBAL debug = '-d,print_select_file_fsync_stats';
+
+# Wait till we reached the initial number of concurrent sessions
+--source include/wait_until_count_sessions.inc
+
+--disable_query_log
+--remove_file $MYSQL_TMP_DIR/t2.txt
+--enable_query_log
--- a/sql/query_result.cc
+++ b/sql/query_result.cc
@@ -605,12 +605,35 @@ bool Query_result_export::send_data(List<Item> &items) {
   if (my_b_write(&cache, (uchar *)exchange->line.line_term->ptr(),
                  exchange->line.line_term->length()))
     goto err;
+
+  /* fsync the file after every select_into_file_fsync_size bytes
+       optionally sleep */
+  if (thd->variables.select_into_file_fsync_size != 0) {
+    my_off_t cur_fsize = my_b_tell(&cache);
+    if (cur_fsize - last_fsync_off >=
+        thd->variables.select_into_file_fsync_size) {
+      if (flush_io_cache(&cache) || mysql_file_sync(cache.file, MYF(MY_WME)))
+        goto err;
+#ifndef DBUG_OFF
+      n_fsyncs++;
+#endif
+      last_fsync_off = cur_fsize;
+      if (thd->variables.select_into_file_fsync_timeout)
+        my_sleep(thd->variables.select_into_file_fsync_timeout * 1000);
+    }
+  }
+
   DBUG_RETURN(false);
 err:
   DBUG_RETURN(true);
 }
 
 void Query_result_export::cleanup() {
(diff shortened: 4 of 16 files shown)


mysql-test/t/select_outfile_fsync_debug.test
```diff
@@ -0,0 +1,63 @@
...
+enable_query_log;
+-- source include/have_outfile.inc
+-- source include/have_debug.inc
+
+# Save the initial number of concurrent sessions
+--source include/count_sessions.inc
+
+# Check for fsync's 
+--disable_warnings
+drop table if exists t2;
```
laurynas-biveinis: Redundant

mysql-test/t/select_outfile_fsync_debug.test
```diff
@@ -0,0 +1,63 @@
...
+--enable_warnings
+
+SET SESSION select_into_file_fsync_size=262144;
+SET GLOBAL debug = '+d,print_select_file_fsync_stats';
+
+let SEARCH_FILE = $MYSQLTEST_VARDIR/log/mysqld.1.err;
```
laurynas-biveinis: Restart with custom error log path

laurynas-biveinis (changes requested):

mysql-test/t/select_outfile_fsync_debug.test
inikep: or use `assert_grep.inc` with `CURRENT_TEST: select_outfile_fsync_debug` e.g.
```
--let $assert_file= $MYSQLTEST_VARDIR/log/mysqld.1.err
--let $assert_only_after = CURRENT_TEST: binlog.binlog_cache_disk_full
--let $assert_count = 1
--let $assert_select = Disk is full writing .* \(OS errno 28 - No space left on device\). Waiting for someone to free space...
--let $assert_text = Found the expected error: No space left on device.
--source include/assert_grep.inc
```

laurynas-biveinis: I think we discussed this before but I don't recall the outcome: is asset_only_after compatible with --repeat ?

inikep: This was Satya's idea. It is compatible with `--repeat` and is has some advantages:
- no need to restart server (faster)
- everything is in logs (and a custom error log will be deleted).

mysql-test/t/select_outfile_fsync_debug.test
```diff
@@ -0,0 +1,63 @@
+disable_query_log;
+-- source include/test_outfile.inc
+# Server are started in "var/master-data", so "../../tmp" will be "var/tmp"
```
percona-ysorokin: Server **is** started

mysql-test/t/select_outfile_fsync_debug.test
```diff
@@ -0,0 +1,63 @@
...
+eval set @tmpdir="../../tmp";
```
percona-ysorokin: 1. `eval` is not needed.
2. `@tmpdir` is not used anywhere.
3. May be just `$MYSQL_TMP_DIR` instead of `../../tmp`?

mysql-test/t/select_outfile_fsync_debug.test
```diff
@@ -0,0 +1,63 @@
...
+INSERT INTO t2 (a, b) VALUES (REPEAT('A', 512), REPEAT('B', 512));
+let $i=14;
+while ($i) {
+  INSERT INTO t2 (a, b) SELECT a, b FROM t2;
+  dec $i;
+}
+--enable_query_log
+
+SELECT *
+INTO OUTFILE "../../tmp/t2.txt"
```
percona-ysorokin:
```suggestion
INTO OUTFILE "$MYSQL_TMP_DIR/t2.txt"
```

mysql-test/t/select_outfile_fsync_debug.test
```diff
@@ -0,0 +1,63 @@
...
+EOF
+
+DROP TABLE t2;
+
+SET GLOBAL debug = '-d,print_select_file_fsync_stats';
+
+# Wait till we reached the initial number of concurrent sessions
+--source include/wait_until_count_sessions.inc
+
+--remove_file $MYSQLTEST_VARDIR/tmp/t2.txt
```
percona-ysorokin:
```suggestion
--remove_file $MYSQL_TMP_DIR/t2.txt
```

sql/query_result.cc
```diff
@@ -605,12 +605,37 @@ bool Query_result_export::send_data(List<Item> &items) {
   if (my_b_write(&cache, (uchar *)exchange->line.line_term->ptr(),
                  exchange->line.line_term->length()))
     goto err;
+
+  /* fsync the file after every select_into_file_fsync_size bytes
+       optionally sleep */
+  if (thd->variables.select_into_file_fsync_size) {
```
percona-ysorokin:
```suggestion
  if (thd->variables.select_into_file_fsync_size != 0) {
```

sql/query_result.cc
```diff
@@ -605,12 +605,37 @@ bool Query_result_export::send_data(List<Item> &items) {
...
+    my_off_t cur_fsize = my_b_tell(&cache);
+    if (cur_fsize - last_fsync_off >=
+        thd->variables.select_into_file_fsync_size) {
+      if (flush_io_cache(&cache) || mysql_file_sync(cache.file, MYF(MY_WME)))
+        goto err;
+      else {
```
percona-ysorokin: else and block are not needed here

sql/sys_vars.cc
```diff
@@ -1396,6 +1396,20 @@ static Sys_var_charptr Sys_character_sets_dir(
     READ_ONLY NON_PERSIST GLOBAL_VAR(charsets_dir), CMD_LINE(REQUIRED_ARG),
     IN_FS_CHARSET, DEFAULT(0));
 
+static Sys_var_ulong Sys_select_into_file_fsync_size(
+    "select_into_file_fsync_size",
+    "Do an fsync to disk when the buffer grows by these many bytes "
+    "for SELECT INTO OUTFILE",
```
percona-ysorokin: Also add *Set to 0 to disable*

percona-ysorokin (changes requested):

dutow:
Updated.

mysql-test/include/test_outfile.inc
```diff
@@ -0,0 +1 @@
+eval select "Outfile OK" into outfile "$MYSQLTEST_VARDIR/tmp/outfile.test";
```
percona-ysorokin: I believe this should be `$MYSQL_TMP_DIR` instead of `$MYSQLTEST_VARDIR/tmp`. There is a difference between them then MTR is run with --parallel=n
```suggestion
eval select "Outfile OK" into outfile "$MYSQL_TMP_DIR/outfile.test";
```

percona-ysorokin (approved):
LGTM with `$MYSQLTEST_VARDIR` issue addressed

percona-ysorokin (approved):
LGTM

mysql-test/include/have_outfile.inc
```diff
@@ -1,5 +1,5 @@
 let $outfile= `SELECT LOAD_FILE(CONCAT(@tmpdir, '/outfile.test')) LIKE 'Outfile OK%'`;
-remove_file $MYSQLTEST_VARDIR/tmp/outfile.test;
+remove_file $MYSQL_TMP_DIR/outfile.test;
```
hermanlee, line 2: This change causes main.outfile to fail since that test creates the outfile in $MYSQLTEST_VARDIR/tmp/outfile.test. Maybe fsync test should change to use $MYSQLTEST_VARDIR/tmp instead?

hermanlee (changes requested):

mysql-test/include/have_outfile.inc
dutow: We changed this as the MYSQL_TMP_DIR is safer in parallel test runs (see above comments).

I could update main.outfile to use TMP_DIR instead?

hermanlee: Yes, that would be fine too. Thanks!

hermanlee: I just checked on the $MYSQLTEST_VARDIR variable. In a parallel test run, the path is:

/data/users/herman/rocks-mysql/8.0/_build-8.0-Debug/mysql-test/var/1/

So I think it already includes the parallel worker id in the pathname.

PR #998, opened 2019-03-22
FB8-106: Add more information to duplicate key error

inikep:
Jira issue: https://jira.percona.com/browse/FB8-106

Reference Patch: https://github.com/facebook/mysql-5.6/commit/2c7f717
Reference Patch: https://github.com/facebook/mysql-5.6/commit/f3a0ed0

Summary: The current error information for duplicate key lists the key value and key name.  In some scenarios this may not be sufficient.  Add a variable that, when enabled, includes the tableame for the key and the query text (or up to 512 bytes of it) into the error message.  The new variable, improved_dup_key_error, will change the output error message to contain this new information.

### DIFF
--- a/sql/mysqld.cc
+++ b/sql/mysqld.cc
@@ -921,6 +921,7 @@ bool opt_general_log, opt_slow_log, opt_general_log_raw;
 ulonglong log_output_options;
 bool opt_log_queries_not_using_indexes = 0;
 ulong opt_log_throttle_queries_not_using_indexes = 0;
+bool opt_improved_dup_key_error = false;
 bool opt_disable_networking = 0, opt_skip_show_db = 0;
 bool opt_skip_name_resolve = 0;
 bool opt_character_set_client_handshake = 1;
--- a/storage/innobase/handler/ha_innodb.h
+++ b/storage/innobase/handler/ha_innodb.h
@@ -67,6 +67,8 @@ struct INNOBASE_SHARE {
 /** Prebuilt structures in an InnoDB table handle used within MySQL */
 struct row_prebuilt_t;
 
+struct ha_innobase_inplace_ctx;
+
 namespace dd {
 namespace cache {
 class Dictionary_client;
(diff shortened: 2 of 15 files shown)


storage/innobase/handler/ha_innodb.h
```diff
@@ -409,6 +411,10 @@ class ha_innobase : public handler {
                                   bool commit, const dd::Table *old_dd_tab,
                                   dd::Table *new_dd_tab);
   /** @} */
+  inline MY_ATTRIBUTE((warn_unused_result)) bool commit_try_rebuild(
```
percona-ysorokin: No definition available  - so not inline
```suggestion
  MY_ATTRIBUTE((warn_unused_result)) bool commit_try_rebuild(
```

storage/innobase/handler/handler0alter.cc
```diff
@@ -6442,7 +6442,7 @@ when rebuilding the table.
 @retval true Failure
 @retval false Success
 */
-inline MY_ATTRIBUTE((warn_unused_result)) bool commit_try_rebuild(
+inline MY_ATTRIBUTE((warn_unused_result)) bool ha_innobase::commit_try_rebuild(
```
percona-ysorokin:
```suggestion
MY_ATTRIBUTE((warn_unused_result)) bool ha_innobase::commit_try_rebuild(
```

percona-ysorokin (changes requested):

sql/mysqld.cc
```diff
@@ -921,6 +921,7 @@ bool opt_general_log, opt_slow_log, opt_general_log_raw;
 ulonglong log_output_options;
 bool opt_log_queries_not_using_indexes = 0;
 ulong opt_log_throttle_queries_not_using_indexes = 0;
+bool opt_improved_dup_key_error = 0;
```
laurynas-biveinis:
```suggestion
bool opt_improved_dup_key_error = false;
```

laurynas-biveinis (changes requested):

inikep:
PR was updated according to comments.

percona-ysorokin (approved):
LGTM

hermanlee (approved):

PR #1003, opened 2019-03-27
WIP FB8-159: Changing thread priorities

inikep:
Jira issue: https://jira.percona.com/browse/FB8-159

Reference Patch: https://github.com/facebook/mysql-5.6/commit/2971de1

Summary: To change os priorities of thread from mysql

Originally Reviewed By: anirbanr-fb

fbshipit-source-id: fd355ee

### DIFF
--- a/sql/sql_class.h
+++ b/sql/sql_class.h
@@ -2083,6 +2083,10 @@ class THD : public MDL_context_owner,
                          value that is only to be used for temporary THDs not present in
                          the global THD list.
                        */
+#ifdef HAVE_SYS_GETTID
+  pid_t system_thread_id;
+#endif
+
  private:
   my_thread_id m_thread_id;
 
@@ -3991,6 +3995,13 @@ class THD : public MDL_context_owner,
   */
   void claim_memory_ownership();
 
+  bool is_thd_priority_alt() { return thd_priority_alt; }
+  void mark_thd_priority_as_alt() {
+    mysql_mutex_lock(&LOCK_thd_data);
+    thd_priority_alt = true;
+    mysql_mutex_unlock(&LOCK_thd_data);
+  }
+
   bool is_a_srv_session() const { return is_a_srv_session_thd; }
   void mark_as_srv_session() { is_a_srv_session_thd = true; }
 #ifndef DBUG_OFF
@@ -4007,6 +4018,9 @@ class THD : public MDL_context_owner,
     aggregates THD.
   */
   bool is_a_srv_session_thd;
+
+  /*Variable to mark weather nice value of this thread changed or not*/
+  bool thd_priority_alt = false;
 
   /**
     Creating or dropping plugin native table through a plugin service.
--- a/sql/sys_vars.cc
+++ b/sql/sys_vars.cc
@@ -99,6 +99,7 @@
 #include "sql/log_event.h"  // MAX_MAX_ALLOWED_PACKET
 #include "sql/mdl.h"
 #include "sql/my_decimal.h"
+#include "sql/mysqld_thd_manager.h"  // Global_THD_manager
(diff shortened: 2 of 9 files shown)


sql/sql_class.h
```diff
@@ -3991,6 +3995,13 @@ class THD : public MDL_context_owner,
   */
   void claim_memory_ownership();
 
+  bool is_thd_priority_alt() { return thd_priority_alt; }
+  void mark_thd_priority_as_alt() {
+    mysql_mutex_lock(&LOCK_thd_data);
```
percona-ysorokin, line 4000: May be make `thd_priority_alt` an instance of `std::atomic_bool` to avoid mutex locks?

sql/sql_class.h
```diff
@@ -3991,6 +3995,13 @@ class THD : public MDL_context_owner,
   */
   void claim_memory_ownership();
 
+  bool is_thd_priority_alt() { return thd_priority_alt; }
```
percona-ysorokin, line 3998:
```suggestion
  bool is_thd_priority_alt() const noexcept { return thd_priority_alt; }
```

sql/sys_vars.cc
```diff
@@ -6467,3 +6468,83 @@ static Sys_var_charptr Sys_read_only_error_msg_extra(
...
+      }
+    }
+  }
+
+  bool parse_thread_nice_value(char *thd_id_nice_val) {
+    // parse input to get thread_id and nice_val
+    const std::string thd_id_nice_val_str(thd_id_nice_val);
+
+    const std::size_t delimpos = thd_id_nice_val_str.find(':');
+    if (delimpos == std::string::npos || delimpos == 0) return false;
```
percona-ysorokin, line 6498: This if does not cover a case with trailing colon, like `42:`
may be
```suggestion
    if (delimpos == std::string::npos || delimpos == 0 || delimpos == thd_id_nice_val_str.size() - 1) return false;
```

sql/sys_vars.cc
```diff
@@ -6467,3 +6468,83 @@ static Sys_var_charptr Sys_read_only_error_msg_extra(
...
     GLOBAL_VAR(opt_read_only_error_msg_extra), CMD_LINE(OPT_ARG),
     IN_SYSTEM_CHARSET, DEFAULT(""), NO_MUTEX_GUARD, NOT_IN_BINLOG);
+
+#ifdef HAVE_SYS_GETTID
+class Change_priority : public Do_THD_Impl {
+ public:
+  Change_priority()
+      : m_changed(false), m_found(false), m_nice_val(0), m_thread_id(0) {}
+
+  virtual void operator()(THD *thd) {
```
percona-ysorokin, line 6478:
```suggestion
  virtual void operator()(THD *thd) override {
```

sql/sys_vars.cc
```diff
@@ -6467,3 +6468,83 @@ static Sys_var_charptr Sys_read_only_error_msg_extra(
...
+    // Check weather nice_value is in a valid range
+    if (m_nice_val > 19 || m_nice_val < -20) {
+      LogErr(INFORMATION_LEVEL, ER_THRNICE_INVALID_RANGE, m_nice_val);
+      return false;
+    }
+
+    return true;
+  }
+
+  bool m_changed;
```
percona-ysorokin, line 6521: I would do the following 4 vars private with 2 read-only getters for `m_found` and `m_changed`

percona-ysorokin (changes requested):

laurynas-biveinis:
MySQL 8.0 has resource groups (https://www.percona.com/blog/2018/01/25/mysql-8-0-resource_group-overview/) which seem to cover this feature completely

PR #986, opened 2019-03-15
WIP FB8-144: Send events to async slaves after ACKed by at least one semi-sync slave

inikep:
```
Because of conflicts I removed changes in `mysql-test/t/all_persisted_variables.test`. 
This test has to be modified at `let $total_persistent_vars=XXX;` (+1 increase)
and it should be re-recorded.
```

Jira issue: https://jira.percona.com/browse/FB8-144

Reference Patch: https://github.com/facebook/mysql-5.6/commit/cc4e803
Reference Patch: https://github.com/facebook/mysql-5.6/commit/aad3ce2a54e

---------- https://github.com/facebook/mysql-5.6/commit/cc4e803 ----------

Send events to async slaves after ACKed by at least one semi-sync slave

Summary:
Added a new mysql variable `wait_for_semi_sync_ack` to control
sending binlog events to async slaves. When this variables is set,
events are sent to async slave only after it is ACKed by at least one
semi-sync slave (if any).

Originally Reviewed By: Tema

fbshipit-source-id: 9cad7e5

---------- https://github.com/facebook/mysql-5.6/commit/aad3ce2a54e ----------
rpl_semi_sync_master_crash_if_active_trxs
Fixed rpl_wait_for_semi_sync_ack

Summary:rpl_semi_sync_master_crash_if_active_trxs
Following changes are made to fix the feature
 - Condition variable is registered in dump thread's THD before waiting so that
   the thread can respond to kill command
 - rpl_wait_for_semi_sync_ack is respected only when semi-sync master is enabled
   (i.e rpl_semi_sync_master_enabled = 1)
 - Added a status variable Rpl_semi_sync_master_ack_waits which counts the
   number of times we waited for an ACK (useful for benchmarking)rpl_semi_sync_master_crash_if_active_trxs

Squash with D6022785

Originally Reviewed By: hermanlee

fbshipit-source-id: 64710f9

PR #1006, opened 2019-03-28
FB8-144: Send events to async slaves after ACKed by at least one semi-sync slave

inikep:
Jira issue: https://jira.percona.com/browse/FB8-144

Reference Patch: https://github.com/facebook/mysql-5.6/commit/cc4e803
Reference Patch: https://github.com/facebook/mysql-5.6/commit/aad3ce2a54e
Reference Patch: https://github.com/facebook/mysql-5.6/commit/b42b911fa15

---------- https://github.com/facebook/mysql-5.6/commit/cc4e803 ----------

Send events to async slaves after ACKed by at least one semi-sync slave

Summary:
Added a new mysql variable `wait_for_semi_sync_ack` to control
sending binlog events to async slaves. When this variables is set,
events are sent to async slave only after it is ACKed by at least one
semi-sync slave (if any).

Originally Reviewed By: Tema

fbshipit-source-id: 9cad7e5

---------- https://github.com/facebook/mysql-5.6/commit/aad3ce2a54e ----------

Fixed rpl_wait_for_semi_sync_ack

Summary:
Following changes are made to fix the feature
 - Condition variable is registered in dump thread's THD before waiting so that
   the thread can respond to kill command
 - rpl_wait_for_semi_sync_ack is respected only when semi-sync master is enabled
   (i.e rpl_semi_sync_master_enabled = 1)
 - Added a status variable Rpl_semi_sync_master_ack_waits which counts the
   number of times we waited for an ACK (useful for benchmarking)

Squash with D6022785

fbshipit-source-id: 64710f9

---------- https://github.com/facebook/mysql-5.6/commit/b42b911fa15 ----------

Fix rpl_wait_for_semi_sync_ack feature

Summary:
Fixed the following:
1. Initializing last acked position to what is retrived from engine during
server startup. This makes sure that lagging async slaves are able to catchup
until the last acked position after master restarts.
2. Resetting last acked posistion when `RESET MASTER` is issued. This makes sure
that after the binlogs are reset we wait for acks.
3. Signalling/updating last acked positions only on events that were actually
acked by the semi-sync slave (like the Xid event of the last trx in a group
commit). This is done by signalling inside of the plugin
(ReplSemiSyncMaster::reportReplyBinlog).
4. Signalling/updating on trxs skipped on semi-sync slave connection while
searching for first gtid connection

fbshipit-source-id: f8917a9c9ff

--------------------------------------------------------------------

TO DO: Because of conflicts I removed changes in `mysql-test/t/all_persisted_variables.test`. 
This test has to be modified at `let $total_persistent_vars=XXX;` (+1 increase)
and it should be re-recorded.


sql/binlog.h
```diff
@@ -883,6 +884,26 @@ class MYSQL_BIN_LOG : public TC_LOG {
...
+  my_off_t get_binlog_end_pos_async(bool async_slave) const {
+    mysql_mutex_assert_not_owner(&LOCK_log);
+    if (async_slave)
+      return last_acked.second;
+    return atomic_binlog_end_pos;
+  }
+
+  void signal_semi_sync_ack(const char *const log_file,
+                                                const my_off_t log_pos) {
+    fprintf(stderr, "%lu signal_semi_sync_ack log_file: %s, log_pos: %llu\n", pthread_self(), log_file, log_pos);
```
laurynas-biveinis: Debugging leftover?

sql/rpl_binlog_sender.cc
```diff
@@ -566,6 +575,8 @@ int Binlog_sender::send_events(File_reader *reader, my_off_t end_pos) {
         m_packet.length(tmp.length());
       }
 
+      fprintf(stderr, "%lu send_events semi_sync_slave=%d log_file=%s, log_pos=%llu\n", pthread_self(), semi_sync_slave, log_file, log_pos);
```
laurynas-biveinis: Debugging leftover?

sql/rpl_binlog_sender.cc
```diff
@@ -428,17 +428,24 @@ int Binlog_sender::send_binlog(File_reader *reader, my_off_t start_pos) {
...
+static bool is_semi_sync_slave() {
+  long long val = 0;
+  get_user_var_int("rpl_semi_sync_slave", &val, nullptr);
+  return val;
+}
+
 int Binlog_sender::get_binlog_end_pos(File_reader *reader, my_off_t *end_pos) {
   DBUG_ENTER("Binlog_sender::get_binlog_end_pos()");
   my_off_t read_pos = reader->position();
+  bool semi_sync_slave = is_semi_sync_slave();
```
laurynas-biveinis:
```suggestion
  const bool semi_sync_slave = is_semi_sync_slave();
```

sql/rpl_binlog_sender.cc
```diff
@@ -469,6 +477,7 @@ int Binlog_sender::send_events(File_reader *reader, my_off_t end_pos) {
   my_off_t log_pos = reader->position();
   my_off_t exclude_group_end_pos = 0;
   bool in_exclude_group = false;
+  bool semi_sync_slave = is_semi_sync_slave();
```
laurynas-biveinis:
```suggestion
  const bool semi_sync_slave = is_semi_sync_slave();
```

sql/sys_vars.cc
```diff
@@ -3409,6 +3409,24 @@ static Sys_var_int32 Sys_regexp_stack_limit(
     GLOBAL_VAR(opt_regexp_stack_limit), CMD_LINE(REQUIRED_ARG),
     VALID_RANGE(0, INT32_MAX), DEFAULT(8000000), BLOCK_SIZE(1));
 
+static bool update_rpl_wait_for_semi_sync_ack(sys_var *, THD *, enum_var_type) {
+  if (rpl_wait_for_semi_sync_ack == false)
```
```suggestion
  if (!rpl_wait_for_semi_sync_ack)
```

sql/sys_vars.cc
```diff
@@ -3409,6 +3409,24 @@ static Sys_var_int32 Sys_regexp_stack_limit(
...
+  }
+  return false;
+}
+
+static Sys_var_bool Sys_wait_semi_sync_ack(
+    "rpl_wait_for_semi_sync_ack",
+    "Wait for events to be acked by a semi-sync slave before sending them "
+    "to the async slaves",
+    GLOBAL_VAR(rpl_wait_for_semi_sync_ack), CMD_LINE(OPT_ARG), DEFAULT(false),
+    NO_MUTEX_GUARD, NOT_IN_BINLOG, ON_CHECK(0),
```
```suggestion
    NO_MUTEX_GUARD, NOT_IN_BINLOG, ON_CHECK(nullptr),
```

laurynas-biveinis (changes requested):

sql/binlog.h
```diff
@@ -387,6 +387,7 @@ class MYSQL_BIN_LOG : public TC_LOG {
   mysql_cond_t update_cond;
 
   std::atomic<my_off_t> atomic_binlog_end_pos;
+  std::pair<std::string, my_off_t> last_acked;
```
percona-ysorokin, line 390: typedef/using for a custom type

sql/rpl_binlog_sender.cc
```diff
@@ -456,6 +463,7 @@ int Binlog_sender::get_binlog_end_pos(File_reader *reader, my_off_t *end_pos) {
     if (!m_wait_new_events || flush_net()) DBUG_RETURN(1);
 
     if (unlikely(wait_new_events(read_pos))) DBUG_RETURN(1);
+
```
percona-ysorokin: formatting

percona-ysorokin (changes requested):

sql/binlog.h
inikep: Why? It's only place when this type is used.

percona-ysorokin: For me, personally, this is just a rule of thumb: if you need a complex type, declare a typedef/using for it - pretty big chances you will have to refer to it more than once. In this particular case it does not seem very important.
Leaving this up to you.

percona-ysorokin (approved):
LGTM

sql/binlog.cc
```diff
@@ -8167,6 +8168,13 @@ void MYSQL_BIN_LOG::process_after_commit_stage_queue(THD *thd, THD *first) {
       Thd_backup_and_restore switch_thd(thd, head);
       bool all = head->get_transaction()->m_flags.real_commit;
       (void)RUN_HOOK(transaction, after_commit, (head, all));
+
+      if (rpl_wait_for_semi_sync_ack) {
```
hermanlee: If signal_semi_sync_ack doesn't get called on every process_after_commit, I think there's a potential problem here. If no further writes come into the system, but rpl_wait_for_semi_sync_ack is enabled, wouldn't async slaves be held back at the last_acked checkpoint that is much much older than the latest semi-sync acked checkpoint?

It seems like we should update the last_acked position for each commit.

It also seems like the original code might have this issue too.

hermanlee (commented):
This approach is definitely simpler than the original one. I'd go with this.

sql/binlog.cc
abhinav04sharma: Is there a way we can push this if condition inside the plugin? Maybe in the after_commit callback?

sql/binlog.h
```diff
@@ -883,6 +884,23 @@ class MYSQL_BIN_LOG : public TC_LOG {
...
     return atomic_binlog_end_pos;
   }
+
+  my_off_t get_binlog_end_pos_async(bool async_slave) const {
+    mysql_mutex_assert_not_owner(&LOCK_log);
+    if (async_slave) return last_acked.second;
+    return atomic_binlog_end_pos;
+  }
+
+  void signal_semi_sync_ack(const char *const log_file,
```
abhinav04sharma: Is there a way to push this inside the plugin as well?

sql/binlog.h
```diff
@@ -883,6 +884,23 @@ class MYSQL_BIN_LOG : public TC_LOG {
     mysql_mutex_assert_not_owner(&LOCK_log);
     return atomic_binlog_end_pos;
   }
+
+  my_off_t get_binlog_end_pos_async(bool async_slave) const {
```
abhinav04sharma: Any reason why doing this in get_binlog_end_pos() would not be ideal?

sql/binlog.cc
inikep: You are right. I followed the original code and I assumed that async slaves will catch up when writes come into the system. I think that the reason behind checking `rpl_wait_for_semi_sync_ack` was not to unnecessary call `signal_semi_sync_ack()` when `rpl_wait_for_semi_sync_ack` is disabled and may not be enabled at all.
Should I remove `if (rpl_wait_for_semi_sync_ack)`?

sql/binlog.h
inikep: fixed

sql/binlog.cc
inikep: We can move this code and `signal_semi_sync_ack` to the plugin but we will also need to add a callback to get last acked position from the plugin. I'm not sure if it's worth doing.

sql/binlog.h
inikep: As above

sql/binlog.cc
abhinav04sharma: @hermanlee I'm not sure i understand this statement: " last_acked checkpoint that is much much older than the latest semi-sync acked checkpoint"

hermanlee: I think the scenario is this:

rpl_wait_for_semi_sync_ack is disabled
master processes writes and generates 1GB worth of binlog entries
update binlog entry is not called, so the async slave's last_acked checkpoint occurs before the new binlog entries
writes stop
rpl_wait_for_semi_sync_ack is enabled
async slaves connects, but last_acked checkpoint before the latest 1GB of binlog entries

abhinav04sharma: I think that case is handled in update_rpl_wait_for_semi_sync_ack() ?

hermanlee: The only place I see last_acked updated is in signal_semi_sync_ack(). How does update_rpl_wait_for_semi_sync_ack() trigger signal_semi_sync_ack()? The only calls I see to signal_semi_sync_ack() is through process_after_commit_stage_queue(), which only happens when writes come into the system.

Maybe update_rpl_wait_for_semi_sync_ack() should update last_ack somehow too?

abhinav04sharma: Won't `Binlog_sender::get_binlog_end_pos()` be called after `mysql_bin_log.signal_update()` is called in `update_rpl_wait_for_semi_sync_ack()`? If yes, then this call to `Binlog_sender::get_binlog_end_pos()` will now call `mysql_bin_log.get_binlog_end_pos()` with `async_slave = false` so it'll return the latest pos instead of the last acked pos after rpl_wait_for_semi_sync_ack is disabled. I don't think we need to update `last_acked` but I might be missing something.

sql/binlog.h
```diff
@@ -879,10 +880,32 @@ class MYSQL_BIN_LOG : public TC_LOG {
     It is called by the threads (e.g. dump thread, applier thread) which want
     to read hot log without LOCK_log protection.
   */
-  my_off_t get_binlog_end_pos() const {
+  my_off_t get_binlog_end_pos(bool need_lock = false,
+                              bool async_slave = false) {
```
abhinav04sharma: I think it'll be better to call this parameter wait_for_ack instead of async_slave.

sql/binlog.cc
hermanlee: I think I see where the disconnect is. The question is what should be the expected behavior when we transition from rpl_wait_for_semi_sync from on to off to on.

Here's the case: rpl_wait_for_semi_sync is on, semi-sync ackers are running, last_ack is updated, so all is working as intended.

Now, rpl_wait_for_semi_sync is disabled, semi_sync ackers are removed. Async slaves connected to the server can tail all binlogs.

The master server generates 1GB of binlog entries after rpl_wait_for_semi_sync is disabled. I think we agree last_ack is pointing to the binlog entry before the latest 1GB was created (assuming they are all in the same binlog file).

Now someone re-enables rpl_wait_for_semi_sync.

Is the expectation for async slaves that connect now to be waiting for the last 1GB of binlogs to be acked by a semi-sync acker before they can read them? From my understanding, last_ack is only updated when semi-sync ackers respond. Or is the expectation for async slaves to be able to read binlogs up to when rpl_wait_for_semi_sync is enabled, which was my expectation, but it might not be what is desired in all scenarios.

The code seems to support the former, but I was thinking of the latter.

abhinav04sharma: I see you concern. In the original FB patch this case was handled by updating the last acked position on initial semi-sync slave connection (it was updated to the point where the semi-sync slave is asking binlogs from because we can assume that all trxs were acked until that point).

hermanlee: It sounds like we'd want the same behavior? Keep last_acked as is until a new semi-sync slave starts asking for binlogs? Update last_ack to that position if needed? If there are no semi-sync ackers, then it sounds like the original was to keep last_ack untouched and block async slaves from anything that hasn't been acked by a semi-sync acker.

@abhinav04sharma Is this correct?

abhinav04sharma: That is correct.

sql/binlog.h
inikep: done

sql/binlog.cc
inikep: I did a lot of research and in 8.0 it's not possible to use information from initial semi-sync slave connection to update `last_ack`. With `gtid-mode=ON` the initial connection sends only an executed GTID set. I think that the best solution would be to remove `if (rpl_wait_for_semi_sync_ack)` and update `last_ack` unconditionally.

abhinav04sharma: @inikep even if we remove the rpl_wait_for_semi_sync_ack check the signalling will only happen when something is written to the binlog, correct? So will the async slaves not wait until something is new is written to the binlog (in @hermanlee example above) ?

abhinav04sharma: Nevermind, looks like last_acked in maintained in mysql_bin_log which is a global singleton, so this should work. I'll let @hermanlee comment on the proposed solution.

inikep: I have a new idea. What do you think about calling `signal_semi_sync_ack()` only when `semisync_master.so` is loaded (installed)? It can be easily done using `plugin_dl_find()` from `sql_plugin.cc`.

hermanlee: I think updating unconditionally would be fine.

@abhinav04sharma is the recent problem found with ABS an issue with this port to 8.0?

sql/binlog.h
```diff
@@ -879,10 +880,32 @@ class MYSQL_BIN_LOG : public TC_LOG {
...
     return atomic_binlog_end_pos;
   }
+
+  void signal_semi_sync_ack(const char *const log_file,
+                            const my_off_t log_pos) {
+    const auto acked = std::make_pair(std::string(log_file), log_pos);
+    lock_binlog_end_pos();
+    if (acked > last_acked) {
+      last_acked = acked;
+      signal_update();
```
hermanlee: We could update only if rpl_wait_for_semi_sync_ack is enabled. Maybe that would be slightly more performant?

sql/binlog.cc
abhinav04sharma: That issue was around group commit, it should not be a problem in this patch since we're signalling inside `ordered_commit()`.

Regarding when to call `signal_semi_sync_ack()`, i think even we always call it (i.e. even if rpl_wait_for_semi_sync_ack = 0) we'll still have the same problem if the master restarts since the last_acked will be reset and we'll need at least one write to start sending events to lagging async slaves... Not sure how we can solve this.

inikep:
PR was updated. 
In the event reading loop (i.e `send_events`) I added a check if a given event was already executed (included in `executed_gtids`). If it was executed then we may call `signal_semi_sync_ack`.
Currently `last_ack` will be updated only with `rpl_wait_for_semi_sync_ack` turned on when:
- new writes come into the system
- a new semi-sync slave starts asking for binlogs

hermanlee:
I talking to @abhinav04sharma, there is probably an issue if `RESET MASTER` is called. The binlog files/offsets get reset, but last_ack does not. So last_ack could be very far in the "future". I think we may also need to reset last_ack when `RESET MASTER` is performed.

sql/rpl_binlog_sender.cc
```diff
@@ -528,6 +537,11 @@ int Binlog_sender::send_events(File_reader *reader, my_off_t end_pos) {
     processlist_slave_offset(log_file, log_pos);
 
     if (before_send_hook(log_file, log_pos)) DBUG_RETURN(1);
+
+    if (rpl_wait_for_semi_sync_ack && m_is_semi_sync_slave &&
+        is_event_executed(event_ptr))
+      mysql_bin_log.signal_semi_sync_ack(log_file, log_pos);
```
abhinav04sharma: AFAIK (at least in 5.6) gtid_executed can contain un-acked transactions because it's updated after the flush stage in group commit. So this might not be correct.

abhinav04sharma: We're signalling on the GTID event, so the rest of the transaction will not be signaled, correct?

inikep: This code calls `signal_semi_sync_ack()` only for the following events:
```
SET @@SESSION.GTID_NEXT=`...`
```
It means that in case of server restart and a semi-sync slave starts asking for binlogs, async slaves will be updated only to the start of the last transaction. The rest of the transaction will be updated when new writes come into the system. 
Another issue is that when new writes come into the system async slaves can be updated up to the start of the last transaction which may be not acked yet.

inikep: The first issue can be fixed, but the second (updating up to the start of the last transaction which may be not acked yet) is hard to fix.

inikep:
I have 3 proposals how to solve a problem when the master restarts and the `last_acked` will be reset and we'll need at least one write to start sending events to lagging async slaves:
1) assume that all GTIDs in `gtid_executed` are commited except the last GTID (it should be true if semi-sync replication was enabled) - disadvantage: slaves will not get the last transaction until new writes come into the system
2) keep `last_acked` in a configuration file with all variables set with "SET PERSIST" (it can be updated when master stops or with every call to `signal_semi_sync_ack`)
3) switch back to a fully synchronized `wait_for_semi_sync_ack` which is already implemented at https://github.com/facebook/mysql-5.6/pull/986

hermanlee:
I think the case of the master restarting, it would be fine to set last_acked to the gtid_executed value once server recovery completes (prepared transactions in the storage engine were either rolled forward or rolled back). @abhinav04sharma , do you think this is would work. Since the storage engine and binlog will be in sync, and the executed set is just the binlog entries in the binlog, it would be fine to send those entries to the async slaves.

If a semi-sync client never acked the binlog entries, our plan is to trim the binlog entries on the server so that they match the storage engine position, at which point, the gtid_executed set would still be the set of acked binlogs.

abhinav04sharma:
I think that should be fine. I have a patch up for review which does this
along with handling RESET MASTER (in 5.6). I'll mention the patch in the PR
as soon as it lands.

~Abhinav

inikep:
Last changes in the PR:
1. Ported "Fix rpl_wait_for_semi_sync_ack feature" from https://github.com/facebook/mysql-5.6/commit/b42b911fa152a5449d10ca87aa3a5a6d7f996dc0
2. Resetting `last_acked` to the current binlog with a position of 0 for both: a restart/crash and "RESET MASTER".
3. Signalling/updating `last_acked` only in `process_after_commit_stage_queue`

sql/binlog.cc
```diff
@@ -8955,6 +8960,60 @@ inline void MYSQL_BIN_LOG::update_binlog_end_pos(my_off_t pos) {
...
+
+my_off_t MYSQL_BIN_LOG::get_last_acked_pos(bool &wait_for_ack, bool need_lock,
+                                           const char *sender_log_name) {
+  wait_for_ack = wait_for_ack && rpl_wait_for_semi_sync_ack &&
+                 rpl_semi_sync_master_enabled;
+
+  if (!wait_for_ack) return atomic_binlog_end_pos;
+
+  if (need_lock) lock_binlog_end_pos();
+  const int res = strcmp(last_acked.first.c_str(), sender_log_name);
```
abhinav04sharma: We should also check for file length here to compare binlog-1000 and binlog-999 correctly

sql/binlog.cc
```diff
@@ -8955,6 +8960,60 @@ inline void MYSQL_BIN_LOG::update_binlog_end_pos(my_off_t pos) {
...
+  return atomic_binlog_end_pos;
+}
+
+void MYSQL_BIN_LOG::signal_semi_sync_ack(const char *const log_file,
+                                         const my_off_t log_pos) {
+  if (!log_file) return;
+
+  const auto acked = std::make_pair(std::string(log_file), log_pos);
+  lock_binlog_end_pos();
+  if (acked > last_acked) {
```
abhinav04sharma: Same comment about comparing lengths

sql/binlog.cc
inikep: It's fixed now.

inikep:
Latest changes in the PR:
- check for file length to compare `master-bin.999999` (6 digits) and `master-bin.1000000` (7 digits) correctly

plugin/semisync/semisync_slave_plugin.cc
```diff
@@ -130,6 +131,14 @@ static int repl_semi_slave_read_event(Binlog_relay_IO_param *,
 static int repl_semi_slave_queue_event(Binlog_relay_IO_param *param,
                                        const char *, unsigned long, uint32) {
   if (rpl_semi_sync_slave_status && semi_sync_need_reply) {
+    DBUG_EXECUTE_IF("before_semi_sync_reply", {
+      static constexpr char act[] =
+          "now signal reached "
```
percona-ysorokin: I would add more unique names instead of `reached` and `continue` to avoid possible collisions.

mysql-test/suite/rpl_gtid/t/rpl_wait_for_semi_sync_ack.test
```diff
@@ -0,0 +1,289 @@
...
+--source include/rpl_connection_master.inc
+--echo "Store the last acked pos"
+--let $last_acked= query_get_value(show status like "Rpl_last_semi_sync_acked_pos", Value, 1)
+--send create table t1 (a int)
+
+--let $rpl_connection_name= semi_sync_slave
+--source include/rpl_connection.inc
+set debug_sync='now WAIT_FOR reached';
+
+--sleep 1
```
percona-ysorokin: Doesn't look reliable. May be something like `sync_slave_io_with_master.inc` instead?

mysql-test/suite/rpl_gtid/t/rpl_wait_for_semi_sync_ack.test
```diff
@@ -0,0 +1,289 @@
...
+set @@global.debug= '+d,before_semi_sync_reply';
+
+--source include/rpl_connection_master.inc
+--send create table t5 (a int)
+
+--let $rpl_connection_name= semi_sync_slave
+--source include/rpl_connection.inc
+set debug_sync='now WAIT_FOR reached';
+
+--sleep 1
```
percona-ysorokin: Doesn't look reliable. May be something like `sync_slave_io_with_master.inc` instead?

percona-ysorokin (commented):
Just a few comments regarding debug sync point names and suspicious sleeps

inikep:
Latest changes in the PR:
- improved rpl_wait_for_semi_sync_ack.test

PR was tested with Jenkins.

abhinav04sharma:
@inikep, are we handling skipped transactions during initial connection in the dump threads. I was not able to find that code.

inikep:
@abhinav04sharma During server startup, a binary log is rotated and all trxs in a previous binlog are already committed to the storage engine. Therefore setting `last_acked` to the current
binlog with a position of 0 is a good solution for both: a restart and a crash. Moreover, there are no trxs in a new binlog so there is no need to ack skipped transactions.

abhinav04sharma:
@inikep, what about the case where the master doesn't restart? There can be a case where a semi-sync slave has transactions that are not yet acked and then reconnects. In this case, since we're not updating the last acked pos by skipped transactions on reconnection, we'll have to wait of a single write to happen on the master for async slaves to move ahead. Let me know what you think about this (cc: @hermanlee )

inikep:
@abhinav04sharma According to https://www.percona.com/community-blog/2018/08/23/question-about-semi-synchronous-replication-answer-with-all-the-details/ :
When MySQL commits a transaction, it is going through the following steps:
1. Prepare the transaction in the storage engine (InnoDB),
2. Write the transaction to the binary logs,
3. Complete the transaction in the storage engine,
4. Return an acknowledgment to the client.

Semi-sync in MySQL 5.5 and 5.6 happens between step #3 and #4. 

Semi-sync in MySQL 8.0 happens between step #2 and #3. At this point, the transaction is not “completed” in the storage engine, so other clients do not see its data yet.

In this PR `signal_semi_sync_ack()` is called after #3 in `process_after_commit_stage_queue()`. It means that all trxs that are committed in the storage engine in the master will also update `last_acked`. While using the semi-sync plugin we can achieve #3 only with:
- at least one semisynchronous slave acknowledges that it has received all events for the transaction
- a timeout occurs

In the case you have described above, I expect a timeout if trxs were executed on the slave (`slave_gtid_executed` will be send at reconnect) or otherwise re-sending of trxs.

abhinav04sharma:
@inikep I don't think the trx will timeout, after the semi-sync slave reconnects it should continue to commit. I've written a small test case to help test this hypothesis (sorry, I wrote it for 5.6):

```
diff --git a/mysql-test/suite/rpl/t/rpl_wait_for_semi_sync_ack.test b/mysql-test/suite/rpl/t/rpl_wait_for_semi_sync_ack.test
index 6b10285940b..36ff46ebe27 100644
--- a/mysql-test/suite/rpl/t/rpl_wait_for_semi_sync_ack.test
+++ b/mysql-test/suite/rpl/t/rpl_wait_for_semi_sync_ack.test
@@ -318,6 +318,54 @@ echo "Async Slave: Should contain t6";
 show tables;
 
 
+# Case 6: We should update last_acked on skipped transactions
+
+echo "Stopping async slave to simulate lag";
+connection async_slave;
+source include/stop_slave.inc;
+
+echo "Generating traffic on the master";
+connection master;
+create table t7(a int);
+insert into t7 values(1);
+insert into t7 values(2);
+insert into t7 values(3);
+insert into t7 values(4);
+
+let $sync_slave_connection= semi_sync_slave;
+source include/sync_slave_sql_with_master.inc;
+
+echo "Blocking semi-sync slave just before sending an ack";
+connection semi_sync_slave;
+set @@global.debug= '+d,wait_before_semi_sync_reply';
+
+connection master;
+send insert into t7 values(5); # this transaction will be blocked
+
+connection semi_sync_slave;
+SET DEBUG_SYNC='now wait_for Reached';
+
+echo "Restarting the semi-sync slave";
+connection master;
+let $rpl_server_number=3;
+let $rpl_force_stop= 1;
+source include/rpl_restart_server.inc;
+
+echo "Starting async slave";
+connection async_slave;
+source include/start_slave.inc;
+
+connection master;
+reap;
+
+echo "Waiting till async slave is caught up";
+# The async slave will catchup only if we signal for skipped trxs on semi-sync
+# connection
+connection master;
+let $sync_slave_connection= async_slave;
+source include/sync_slave_sql_with_master.inc;
+
+
 # Cleanup
 
 connection master;
@@ -327,6 +375,7 @@ drop table t3;
 drop table t4;
 drop table t5;
 drop table t6;
+drop table t7;
 
 connection master;
 let $sync_slave_connection= async_slave;
diff --git a/sql/rpl_slave.cc b/sql/rpl_slave.cc
index 3c7b5c9a3f3..cb0eee5589f 100644
--- a/sql/rpl_slave.cc
+++ b/sql/rpl_slave.cc
@@ -5315,6 +5315,14 @@ Stopping slave I/O thread due to out-of-memory error from master");
 
       DBUG_EXECUTE_IF("error_before_semi_sync_reply", goto err;);
 
+      DBUG_EXECUTE_IF("wait_before_semi_sync_reply",
+                      {
+                      const char act[]= "now signal Reached wait_for signal.semi_sync_reply_continue";
+                      DBUG_ASSERT(opt_debug_sync_timeout > 0);
+                      DBUG_ASSERT(!debug_sync_set_action(current_thd,
+                                                         STRING_WITH_LEN(act)));
+                      });
+
       if (RUN_HOOK(binlog_relay_io, after_queue_event,
                    (thd, mi, event_buf, event_len, synced)))
       {
```

Can you check if this works in the current implementation?

inikep:
@abhinav04sharma I added your `Case #6` to `rpl_wait_for_semi_sync_ack.test` in 8.0. Everything works fine. 

I also debugged it. There are 2 cases:
1. An unacked trx was not commited on a semi-sync slave
2. An unacked trx was commited on a semi-sync slave

In 1st case: during reconnect the master is re-sending the trx as expected.
In 2nd case: during reconnect the master skips trx but calls `after_send_hook(log_pos)` (https://github.com/facebook/mysql-5.6/blob/fb-mysql-8.0.13/sql/rpl_binlog_sender.cc#L600) which calls `skipSlaveReply()` in the plugin (https://github.com/facebook/mysql-5.6/blob/fb-mysql-8.0.13/plugin/semisync/semisync_master_plugin.cc#L178) that is equal to an ack from a slave

plugin/semisync/semisync_slave_plugin.cc
```diff
@@ -129,7 +130,23 @@ static int repl_semi_slave_read_event(Binlog_relay_IO_param *,
 
 static int repl_semi_slave_queue_event(Binlog_relay_IO_param *param,
                                        const char *, unsigned long, uint32) {
+  DBUG_EXECUTE_IF("before_semi_sync_event", {
```
abhinav04sharma, line 133: Can we use `before_semi_sync_reply` below instead of creating a new sync point?

abhinav04sharma (approved):

plugin/semisync/semisync_slave_plugin.cc
inikep: `before_semi_sync_event` is the same as your `wait_before_semi_sync_reply`.

We can replace `before_semi_sync_event` with `before_semi_sync_reply` and MTR test should work, but the behavior is different:
a) `before_semi_sync_event` - An unacked trx was not yet commited on a semi-sync slave
b) `before_semi_sync_reply` - An unacked trx was already commited on a semi-sync slave

inikep:
PR was updated:
1. conflicts after merging https://github.com/facebook/mysql-5.6/commit/b5d9d0426353162a1529dcf22406c23334203b16 were resolved
2. `binlog_nogtid.binlog_persist_variables` and `binlog_nogtid.binlog_persist_only_variables` were re-recorded

PR #1007, opened 2019-03-28
FB8-183: Implement system status variable to track duration of global read lock (Metadata lock) and the COMMIT lock

inikep:
Jira issue: https://jira.percona.com/browse/FB8-183

Reference Patch: https://github.com/facebook/mysql-5.6/commit/ba953ba

Summary:
In this change, we are adding status variable Global_Read_Only_Lock_WaitTime status variable which keep tracks of how long either global readonly lock or commit lock is held.
This is going to help with automation of failover.

Originally Reviewed By: Tema

### DIFF
--- a/mysql-test/t/commit_lock_wait_time.test
+++ b/mysql-test/t/commit_lock_wait_time.test
@@ -1,0 +1,33 @@
+--source include/have_debug_sync.inc
+--source include/count_sessions.inc
+
+--connect (con1,localhost,root)
+--connection default
+create table t1 (a int) engine=innodb;
+insert into t1 values(1);
+show status like 'super_read_only_block_microsec';
+
+insert into t1 values(1);
+set debug_sync='RESET';
+set debug_sync='ha_commit_trans_after_acquire_commit_lock WAIT_FOR go';
+--send insert into t1 values(1)
+
+--connection con1
+select sleep(1.1);
+--let $exec = `show status like 'super_read_only_block_microsec'`
+--let $assert_text = super_read_only_block_microsec is bigger than 1100000
+--let $assert_cond = [select substring("$exec", 32)+0] > 1100000
+--source include/assert.inc
+
+set debug_sync='now SIGNAL go';
+select sleep(1.1);
+show status like 'super_read_only_block_microsec';
+
+set debug_sync= 'RESET';
+--disconnect con1
+
+--connection default
+--reap
+drop table t1;
+
+--source include/wait_until_count_sessions.inc
--- a/mysql-test/t/show_super_read_only_block.test
+++ b/mysql-test/t/show_super_read_only_block.test
@@ -1,0 +1,24 @@
+create database testdb1;
+
+use testdb1;
+
+create table t1 (a int);
+
+insert into t1 values(1);
+
+show status like 'super_read_only_block_microsec';
+
+lock table t1 write;
+
+select sleep(1.1);
+
+--let $exec = `show status like 'super_read_only_block_microsec'`
+--let $assert_text = super_read_only_block_microsec is bigger than 1100000
+--let $assert_cond = [select substring("$exec", 32)+0] > 1100000
+--source include/assert.inc
+
+unlock tables;
+
+show status like 'super_read_only_block_microsec';
+
+drop database testdb1;
--- a/sql/mdl.cc
+++ b/sql/mdl.cc
@@ -51,6 +51,7 @@
 #include "mysqld_error.h"
 #include "prealloced_array.h"
 #include "sql/debug_sync.h"
+#include "sql/mysqld.h"
 #include "sql/thr_malloc.h"
 
 extern MYSQL_PLUGIN_IMPORT CHARSET_INFO *system_charset_info;
@@ -3362,6 +3363,19 @@ bool MDL_context::acquire_lock_nsec(MDL_request *mdl_request,
   /* Do some work outside the critical section. */
   set_timespec_nsec(&abs_timeout, lock_wait_timeout_nsec);
 
+#if !defined(EXTRA_CODE_FOR_UNIT_TESTING)
+  if (mdl_request->key.mdl_namespace() == MDL_key::GLOBAL &&
+      mdl_request->type != MDL_SHARED) {
+    unsigned long long expected = 0ULL;
+    init_global_rolock_timer.compare_exchange_strong(expected, my_timer_now());
+  }
+  if (mdl_request->key.mdl_namespace() == MDL_key::COMMIT &&
+      mdl_request->type != MDL_SHARED) {
+    unsigned long long expected = 0ULL;
+    init_commit_lock_timer.compare_exchange_strong(expected, my_timer_now());
+  }
+#endif
+
   if (try_acquire_lock_impl(mdl_request, &ticket)) return true;
 
   if (mdl_request->ticket) {
@@ -4056,6 +4070,17 @@ void MDL_context::release_lock(enum_mdl_duration duration, MDL_ticket *ticket) {
   DBUG_PRINT("enter", ("db=%s name=%s", lock->key.db_name(), lock->key.name()));
 
   DBUG_ASSERT(this == ticket->get_ctx());
+
+#if !defined(EXTRA_CODE_FOR_UNIT_TESTING)
+  if (lock->key.mdl_namespace() == MDL_key::GLOBAL &&
+      ticket->get_type() != MDL_SHARED)
+    init_global_rolock_timer = 0;
+
+  if (lock->key.mdl_namespace() == MDL_key::COMMIT &&
+      ticket->get_type() != MDL_SHARED)
+    init_commit_lock_timer = 0;
+#endif
+
   mysql_mutex_assert_not_owner(&LOCK_open);
 
   /*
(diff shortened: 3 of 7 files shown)


mysql-test/t/commit_lock_wait_time.test
```diff
@@ -0,0 +1,34 @@
...
+
+insert into t1 values(1);
+set debug_sync='RESET';
+set debug_sync='ha_commit_trans_after_acquire_commit_lock WAIT_FOR go';
+--send insert into t1 values(1)
+
+--connection con1
+select sleep(1.1);
+--let $exec = `show status like 'super_read_only_block_microsec'`
+if (`select substring('$exec', 32)+0 > 1100000`)
```
laurynas-biveinis: include/assert.inc

mysql-test/t/show_super_read_only_block.test
```diff
@@ -0,0 +1,25 @@
...
+insert into t1 values(1);
+
+show status like 'super_read_only_block_microsec';
+
+lock table t1 write;
+
+select sleep(1.1);
+
+let $exec = `show status like 'super_read_only_block_microsec'`;
+if (`select SUBSTRING('$exec', 32)+0 > 1100000`)
```
laurynas-biveinis: Likewise

sql/mysqld.cc
```diff
@@ -7957,6 +7960,34 @@ static int show_connection_errors_net_ER_NET_WRITE_INTERRUPTED(THD *, SHOW_VAR *
   return 0;
 }
 
+static inline ulonglong get_super_read_only_block_microsec(
+    std::atomic_ullong &lock_block_timer) {
+  ulonglong lock_wait_microsec =
+      lock_block_timer.load(std::memory_order_seq_cst);
```
laurynas-biveinis: This could be memory_order_relaxed, couldn't it?

laurynas-biveinis (changes requested):

mysql-test/t/commit_lock_wait_time.test
```diff
@@ -0,0 +1,34 @@
...
+{
+--echo -- super_read_only_block_microsec is higher than 1100000
+}
+
+set debug_sync='now SIGNAL go';
+select sleep(1.1);
+show status like 'super_read_only_block_microsec';
+
+set debug_sync= 'RESET';
+drop table t1;
```
percona-ysorokin: Dropping `t1` in connection `con1` while `--send insert into t1 values(1)` may still be in progress in connection `default`.
This may cause instabilities. I suggest to move `drop table t1;` after
```
--connection default
--reap
```

sql/mysqld.h
```diff
@@ -374,6 +374,13 @@ inline double my_timer_to_microseconds(ulonglong when) {
   ret /= (double)(my_timer.frequency);
   return ret;
 }
+/* Convert native timer units in a ulonglong into microseconds in a ulonglong */
+inline ulonglong my_timer_to_microseconds_ulonglong(ulonglong when) {
+  ulonglong ret = (ulonglong)(when);
```
percona-ysorokin: when is already of type `ulonglong`
```suggestion
  ulonglong ret = when;
```

sql/mysqld.h
```diff
@@ -374,6 +374,13 @@ inline double my_timer_to_microseconds(ulonglong when) {
...
+  ret *= 1000000;
```
percona-ysorokin:
```suggestion
  ret *= 1000000ULL;
```

sql/mysqld.h
```diff
@@ -374,6 +374,13 @@ inline double my_timer_to_microseconds(ulonglong when) {
...
+  ret = (ulonglong)((ret + my_timer.frequency - 1) / my_timer.frequency);
```
percona-ysorokin:
```suggestion
  ret = static_cast<ulonglong>((ret + my_timer.frequency - 1ULL) / my_timer.frequency);
```

sql/mysqld.cc
```diff
@@ -7957,6 +7960,34 @@ static int show_connection_errors_net_ER_NET_WRITE_INTERRUPTED(THD *, SHOW_VAR *
...
+  var->type = SHOW_LONGLONG;
+  var->value = buff;
+  const ulonglong super_read_only_block_microsec =
+      get_super_read_only_block_microsec(init_global_rolock_timer);
+
+  const ulonglong commit_lock_wait =
+      get_super_read_only_block_microsec(init_commit_lock_timer);
+
+  if (commit_lock_wait > super_read_only_block_microsec)
+    *((longlong *)buff) = commit_lock_wait;
```
percona-ysorokin:
```suggestion
    *(reinterpret_cast<longlong *>(buff)) = commit_lock_wait;
```

sql/mysqld.cc
```diff
@@ -7957,6 +7960,34 @@ static int show_connection_errors_net_ER_NET_WRITE_INTERRUPTED(THD *, SHOW_VAR *
...
+  else
+    *((longlong *)buff) = super_read_only_block_microsec;
```
percona-ysorokin:
```suggestion
    *(reinterpret_cast<longlong *>(buff)) = super_read_only_block_microsec;
```

sql/mysqld.cc
```diff
@@ -7957,6 +7960,34 @@ static int show_connection_errors_net_ER_NET_WRITE_INTERRUPTED(THD *, SHOW_VAR *
   return 0;
 }
 
+static inline ulonglong get_super_read_only_block_microsec(
+    std::atomic_ullong &lock_block_timer) {
+  ulonglong lock_wait_microsec =
+      lock_block_timer.load(std::memory_order_seq_cst);
+  if (lock_wait_microsec)
```
percona-ysorokin:
```suggestion
  if (lock_wait_microsec != 0ULL)
```

sql/mdl.cc
```diff
@@ -3362,6 +3363,16 @@ bool MDL_context::acquire_lock(MDL_request *mdl_request,
   /* Do some work outside the critical section. */
   set_timespec(&abs_timeout, lock_wait_timeout);
 
+#if !defined(EXTRA_CODE_FOR_UNIT_TESTING)
+  if (mdl_request->key.mdl_namespace() == MDL_key::GLOBAL &&
+      mdl_request->type != MDL_SHARED && init_global_rolock_timer == 0)
+    init_global_rolock_timer = my_timer_now();
```
percona-ysorokin: Consider compare_exchange_strong here
```c++
  if (mdl_request->key.mdl_namespace() == MDL_key::GLOBAL &&
      mdl_request->type != MDL_SHARED) {
    unsigned long long expected = 0ULL;
    init_global_rolock_timer.compare_exchange_strong(expected, my_timer_now());
  }
```

sql/mdl.cc
```diff
@@ -3362,6 +3363,16 @@ bool MDL_context::acquire_lock(MDL_request *mdl_request,
...
+
+  if (mdl_request->key.mdl_namespace() == MDL_key::COMMIT &&
+      mdl_request->type != MDL_SHARED && init_commit_lock_timer == 0)
+    init_commit_lock_timer = my_timer_now();
```
percona-ysorokin: Likewise

sql/mysqld.cc
percona-ysorokin: @laurynas-biveinis We could only if we don't care too much about accuracy of these counters. In other words, with `memory_order_relaxed` here, there is a chance that even when `init_global_rolock_timer` / `init_commit_lock_timer` are set inside one thread in `MDL_context::acquire_lock()` with `memory_order_seq_cst` (which is a default memory order for `std::atomic<>::operator =` ), another thread executing `show status like 'super_read_only_block_microsec'` may read these values as zeros.
I am not 100% sure that this is unacceptable, but personally, I would go with Sequentially-consistent memory order here for simplicity.
In theory, the most correct way to do this would be using `std::memory_order_release` for setting these vars inside `MDL_context::acquire_lock()` and reading them here with `std::memory_order_acquire`, but because the order in which they are set is not determined (`init_global_rolock_timer` is set in one call of `MDL_context::acquire_lock()` whereas `init_commit_lock_timer` in another), I wouldn't go this road.

So, just to simplify
```suggestion
      lock_block_timer;
```

percona-ysorokin (changes requested):

sql/mysqld.cc
percona-ysorokin: BTW, I'have recently discovered a very nice tool "CppMem: Interactive C/C++ memory model"
http://svr-pes20-cppmem.cl.cam.ac.uk/cppmem/
which is supposed to statically check race conditions for cases like this.
However, it's very sophisticated and I must admit I understand only 10% of it :)

sql/mdl.cc
inikep: I like this solution.

inikep:
PR was updated.

percona-ysorokin (approved):
LGTM

inikep:
PR was updated: resolved a conflict in `mysqld.cc`.

hermanlee:
Will close this one for now, but we have the patch tracked internally. I think querying the performance_schema tables might be a way to implement this:

```
mysql> select metadata_locks.object_type, threads.processlist_time from metadata_locks join threads join events_waits_current on metadata_locks.object_type in ("GLOBAL", "COMMIT") and threads.thread_id=metadata_locks.owner_thread_id and threads.thread_id = events_waits_current.thread_id\G
*************************** 1. row ***************************
     object_type: COMMIT
processlist_time: 1177
*************************** 2. row ***************************
     object_type: GLOBAL
processlist_time: 1177
2 rows in set (0.00 sec)
```

PR #1010, opened 2019-04-01
FB8-180: Implemented TTL base cache eviction policy for "open table" and "table description" caches

inikep:
Jira issue: https://jira.percona.com/browse/FB8-180

Reference Patch:  https://github.com/facebook/mysql-5.6/commit/ec9e122

Summary:
MySQL has variable "flush_time". If it is set to a nonzero value, all tables are closed every "flush_time" seconds to free up resources and synchronize unflushed data to disk. This option is best used only on systems with minimal resources.
I added variable "flush_only_old_cache_entries" which were open more than "flush_time" seconds.

Originally Reviewed By: anirbanr-fb

### DIFF
--- a/sql/sql_base.cc
+++ b/sql/sql_base.cc
@@ -537,6 +537,9 @@ static TABLE_SHARE *process_found_table_share(THD *thd MY_ATTRIBUTE((unused)),
     share->prev = 0;
   }
 
+  /* update access time */
+  share->set_last_access_time();
+
   /* Free cache if too big */
   while (table_def_cache->size() > table_def_size && oldest_unused_share->next)
     table_def_cache->erase(to_string(oldest_unused_share->table_cache_key));
@@ -832,6 +835,9 @@ TABLE_SHARE *get_table_share(THD *thd, const char *db, const char *table_name,
 #else
   share->m_psi = NULL;
 #endif
+
+  /* update access time */
+  share->set_last_access_time();
 
   DBUG_PRINT("exit", ("share: %p  ref_count: %u", share, share->ref_count()));
 
@@ -9526,7 +9532,34 @@ err:
 
 void tdc_flush_unused_tables() {
   table_cache_manager.lock_all_and_tdc();
-  table_cache_manager.free_all_unused_tables();
+
+  if (flush_only_old_table_cache_entries) {
+    DBUG_EXECUTE_IF("tdc_flush_unused_tables", {
+      static constexpr char act[] = "now SIGNAL signal.enabled";
+      DBUG_ASSERT(!debug_sync_set_action(current_thd, STRING_WITH_LEN(act)));
+    };);
+
+    const time_point flush_cutpoint =
+        std::chrono::system_clock::now() - std::chrono::seconds(flush_time);
+    table_cache_manager.free_old_unused_tables(flush_cutpoint);
+
+    /* Free table shares which were not freed implicitly by loop above. */
+    for (TABLE_SHARE *s = oldest_unused_share; s->next; s = s->next) {
+      if (should_be_evicted(s->last_accessed, flush_cutpoint)) {
+        table_def_cache->erase(to_string(s->table_cache_key));
+      }
+    }
+  } else {
+    DBUG_EXECUTE_IF("tdc_flush_unused_tables", {
+      static constexpr char act[] = "now SIGNAL signal.disabled";
+      DBUG_ASSERT(!debug_sync_set_action(current_thd, STRING_WITH_LEN(act)));
+    };);
+
+    table_cache_manager.free_all_unused_tables();
+    /* Free table shares which were not freed implicitly by loop above. */
+    while (oldest_unused_share->next)
+      table_def_cache->erase(to_string(oldest_unused_share->table_cache_key));
+  }
   table_cache_manager.unlock_all_and_tdc();
 }
 
--- a/sql/sql_manager.cc
+++ b/sql/sql_manager.cc
@@ -50,6 +50,7 @@
 #include "sql/log.h"
 #include "sql/mysqld.h"    // flush_time
 #include "sql/sql_base.h"  // tdc_flush_unused_tables
+#include "sql/sql_class.h"
 
 static bool volatile manager_thread_in_use;
 static bool abort_manager;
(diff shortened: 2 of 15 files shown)


sql/table.cc
```diff
@@ -7774,3 +7777,17 @@ void TABLE::set_pos_in_table_list(TABLE_LIST *table_list) noexcept {
...
+
+void TABLE_SHARE::set_last_access_time() noexcept {
+  this->last_accessed = get_time_now();
+}
+
+void TABLE::set_last_access_time() noexcept {
+  this->last_accessed = get_time_now();
+}
+
+bool should_be_evicted(time_point last_accessed, time_point cutpoint) {
```
laurynas-biveinis: static

sql/table.cc
```diff
@@ -7774,3 +7777,17 @@ void TABLE::set_pos_in_table_list(TABLE_LIST *table_list) noexcept {
 }
 
 //////////////////////////////////////////////////////////////////////////
+
+const time_point get_time_now() { return std::chrono::system_clock::now(); }
```
laurynas-biveinis: IMHO this wrapper is not worth it, system_clock::now() can be just called directly.
But otherwise "static"

laurynas-biveinis (commented):
Zero test coverage for the feature, even at the most basic "does not crash when enabled" level?
I'd personally also prefer more of C++11 AAA (almost always auto) style

sql/table.cc
inikep: The function is used in `table_cache.cc` and `table.cc`.

inikep:
PR was updated. I added `mysql-test/t/flush_only_old_table_cache_entries.test` which checks if a server does not crash when the option is disabled and enabled.

sql/sql_base.cc
```diff
@@ -9526,7 +9532,38 @@ bool mysql_rm_tmp_tables(void) {
...
+      static constexpr char act[] = "now SIGNAL signal.enabled";
+      DBUG_ASSERT(!debug_sync_set_action(current_thd, STRING_WITH_LEN(act)));
+    };);
+
+    const time_point flush_cutpoint =
+        std::chrono::system_clock::now() - std::chrono::seconds(flush_time);
+    table_cache_manager.free_old_unused_tables(flush_cutpoint);
+
+    /* Free table shares which were not freed implicitly by loop above. */
+    std::vector<TABLE_SHARE *> old_entries;
```
percona-ysorokin: typedef/using for a custom container

sql/sql_base.cc
```diff
@@ -9526,7 +9532,38 @@ bool mysql_rm_tmp_tables(void) {
...
+    for (TABLE_SHARE *s = oldest_unused_share; s->next; s = s->next) {
+      if (should_be_evicted(s->last_accessed, flush_cutpoint)) {
+        old_entries.push_back(s);
```
percona-ysorokin: Do we really need to save table share pointers to be evicted in a separate container.
Can they be deleted in-place?
```suggestion
        table_def_cache->erase(to_string(s->table_cache_key));
```

sql/sql_manager.cc
```diff
@@ -95,9 +101,10 @@ static void *handle_manager(void *arg MY_ATTRIBUTE((unused))) {
     }
   }
   manager_thread_in_use = 0;
-  DBUG_LEAVE;  // Can't use DBUG_RETURN after my_thread_end
+  thd->release_resources();
+  delete thd;
   my_thread_end();
-  return (NULL);
+  DBUG_RETURN(nullptr);
```
percona-ysorokin: There was a comment `Can't use DBUG_RETURN after my_thread_end` and this is exactly the case.

sql/table_cache.cc
```diff
@@ -129,6 +129,26 @@ void Table_cache::free_all_unused_tables() {
   }
 }
 
+void Table_cache::free_old_unused_tables(time_point cutpoint) {
+  assert_owner();
+
+  if (!m_unused_tables) return;
+  std::vector<TABLE *> old_tables;
```
percona-ysorokin, line 136: typedef/using for a custom container.

sql/table_cache.cc
```diff
@@ -129,6 +129,26 @@ void Table_cache::free_all_unused_tables() {
...
+
+  for (TABLE *t = m_unused_tables;;) {
+    if (should_be_evicted(t->last_accessed, cutpoint)) {
+      old_tables.push_back(t);
```
percona-ysorokin, line 140: Likewise. Why collecting tables first in a custom container and delete later?

sql/table.cc
```diff
@@ -7774,3 +7777,15 @@ void TABLE::set_pos_in_table_list(TABLE_LIST *table_list) noexcept {
...
+
+void TABLE_SHARE::set_last_access_time() noexcept {
+  this->last_accessed = std::chrono::system_clock::now();
+}
+
+void TABLE::set_last_access_time() noexcept {
+  this->last_accessed = std::chrono::system_clock::now();
+}
+
+bool should_be_evicted(time_point last_accessed, time_point cutpoint) {
```
percona-ysorokin:
```suggestion
bool should_be_evicted(time_point last_accessed, time_point cutpoint) noexcept {
```

sql/table.h
```diff
@@ -3974,4 +3989,7 @@ bool create_table_share_for_upgrade(THD *thd, const char *path,
                                     bool is_fix_view_cols_and_deps);
 //////////////////////////////////////////////////////////////////////////
 
+const time_point get_time_now();
+bool should_be_evicted(time_point last_accessed, time_point cutpoint);
```
percona-ysorokin:
```suggestion
bool should_be_evicted(time_point last_accessed, time_point cutpoint) noexcept;
```

sql/table.h
```diff
@@ -3974,4 +3989,7 @@ bool create_table_share_for_upgrade(THD *thd, const char *path,
                                     bool is_fix_view_cols_and_deps);
 //////////////////////////////////////////////////////////////////////////
 
+const time_point get_time_now();
```
percona-ysorokin: A function without definition
```suggestion
```

percona-ysorokin (changes requested):

sql/table_cache.cc
inikep: We have a circular double-linked list, whose head can be removed and reaching the head is our stop condition.

sql/table_cache.cc
inikep: Are there any advantages of `using`? This type occurs only once. For me it doesn't look better:
```
  using table_vector = std::vector<TABLE *>;
  table_vector old_tables;
```

percona-ysorokin (approved):
LGTM

PR #1011, opened 2019-04-02
FB8-98, FB8-126, FB8-127: Introducing high prpiority DDLs

dutow:
Jira ticket: https://jira.percona.com/browse/FB8-98

Reference commit: 2b59a06
Reference commit: f816fee
Reference commit: 397fa65
Reference commit: 00323f7
Reference commit: bffb8d7
Reference commit: e9cb575
Reference commit: b145f26
Reference commit: 7e41852
Reference commit: 434d496

-------- 2b59a06 --------

This diff introduces high priority DDLs. If enabled, DDLs will kill the
blocking connections that hold shared metadata locks. Below are the details:

- Added high_priority_ddl system variable that will allow DDL to kill blocking
  connections which hold shared locks. High priority DDLs are only available to
  admin users. By default the functionality is turned off.
- Only the shared locks (read/write) will be affected, which are DML queries
  such as select, insert, update, etc. Any lock higher than upgradable MDL
  (i.e. from conflicting DDL) will not be affected.
- When killing the blocking connections, their transactions will automatically
  be rolled back. This is the same behavior as `kill processlist_id` command.
- The DDL will kill blocking connections at the end of the timeout instead of
  instantly. Admin can control the waiting period by setting the session
  lock_wait_timeout before issuing the DDL.

-------- 397fa65 --------

Adding high_priority syntax in DDL commands

This diff introduces `HIGH_PRIORITY` syntax in some DDL commands and optimize
command. The purpose is the same as the system variable `high_priority_ddl`
which is described in D4784076.

The list of commands that support `HIGH_PRIORITY` syntax is:
- create/drop/alter/truncate table
- create/drop index
- create/drop trigger
- optimize table

With this diff, there are two ways of issuing high priority DDLs:
  e.g. `CREATE HIGH_PRIORITY INDEX ...`

Recap of the high_priority ddl feature:

- Only the shared locks (read/write) will be affected, which are DML queries
  such as select, insert, update, etc. Any lock higher than upgradable MDL
  (i.e. from conflicting DDL) will not be affected.
- When killing the blocking connections, their transactions will automatically
  be rolled back. This is the same behavior as `kill processlist_id` command.
- The DDL will kill blocking connections at the end of the timeout instead of
  instantly. Admin can control the waiting period by setting the session
  lock_wait_timeout before issuing the DDL.

-------- b145f26 --------

Add LOCK TABLES as a supported command for high priority ddl

Although LOCK TABLES is not a ddl command, it is useful to have it support taking mdl locks with high priority, via the high_priority_ddl session variable.

-------- 7e41852 --------

Add high_priority_lock_wait_timeout system variable

The original lock_wait_timeout is used for all queries and also the long
running slave threads. It'll be better that high_priority commands (DDLs) use
a separate system variable for timeout, which can be used across the
master/slaves without messing with lock_wait_timeout.

This diff introduces a new system variable high_priority_lock_wait_timeout that
can be set for high_priority command's lock wait timeout, after which the
command's thread will kill shared locks that are blocking the command.

The new variable doesn't apply to regular queries.


mysql-test/include/rpl_ddl_high_priority.inc
```diff
@@ -0,0 +1,73 @@
...
+# Slave applier parameters for the failed retry
+#set @@global.innodb_lock_wait_timeout=1;
+#set @@global.slave_transaction_retries=2;
+#--source include/restart_slave_sql.inc
+
+connection master;
+
+# create t1
+--disable_warnings
+drop table if exists t1;
```
percona-ysorokin: Old idiom

mysql-test/include/rpl_ddl_high_priority.inc
```diff
@@ -0,0 +1,73 @@
...
+--source include/rpl_connect.inc
+
+#--echo connection slave
+#connection slave;
+# Slave applier parameters for the failed retry
+#set @@global.innodb_lock_wait_timeout=1;
+#set @@global.slave_transaction_retries=2;
+#--source include/restart_slave_sql.inc
+
+connection master;
```
percona-ysorokin:
```suggestion
--source include/rpl_connection_master.inc
```

mysql-test/include/rpl_ddl_high_priority.inc
```diff
@@ -0,0 +1,73 @@
...
+show create table t1;
+insert into t1 values (1), (2), (3);
+
+sync_slave_with_master;
+
+--echo connection slave
+connection slave;
+select * from t1;
+
+--echo connection slave_block
```
percona-ysorokin:
```suggestion
--let $rpl_connection_name= slave_block
--source include/rpl_connect.inc
```

mysql-test/include/rpl_ddl_high_priority.inc
```diff
@@ -0,0 +1,73 @@
...
+
+# create t1
+--disable_warnings
+drop table if exists t1;
+--enable_warnings
+create table t1 (i int);
+show create table t1;
+insert into t1 values (1), (2), (3);
+
+sync_slave_with_master;
```
percona-ysorokin:
```suggestion
--source include/sync_slave_sql_with_master.inc
```

mysql-test/include/rpl_ddl_high_priority.inc
```diff
@@ -0,0 +1,73 @@
...
+
+--echo connection slave
```
percona-ysorokin:
```suggestion
--source include/rpl_connection_slave.inc
```

mysql-test/include/rpl_ddl_high_priority.inc
```diff
@@ -0,0 +1,73 @@
...
+select * from t1;
+
+--echo connection slave_block
+connection slave_block;
+
+# issue a bloking read from slave_block connection
+lock tables t1 read;
+
+--echo connection master;
+connection master;
```
percona-ysorokin:
```suggestion
--source include/rpl_connection_master.inc
```

mysql-test/include/rpl_ddl_high_priority.inc
```diff
@@ -0,0 +1,73 @@
...
+
+# issue a high_pri command from master
+create high_priority index idx1 on t1 (i);
+show create table t1;
+
+--echo connection slave;
```
percona-ysorokin:
```suggestion
--source include/rpl_connection_slave.inc
```

mysql-test/include/rpl_ddl_high_priority.inc
```diff
@@ -0,0 +1,73 @@
...
+connection slave;
+
+select * from performance_schema.metadata_locks;
+
+connection master;
```
percona-ysorokin:
```suggestion
--source include/rpl_connection_master.inc
```

mysql-test/include/rpl_ddl_high_priority.inc
```diff
@@ -0,0 +1,73 @@
...
+# sync. high_priority_lock_wait_timeout is 1 second by default
+sync_slave_with_master;
+
+--echo connection slave;
+connection slave;
```
percona-ysorokin:
```suggestion
--source include/rpl_connection_slave.inc
```

mysql-test/include/rpl_ddl_high_priority.inc
```diff
@@ -0,0 +1,73 @@
...
+
+# the high_pri ddl replication should go through
+show create table t1;
+
+
+##
+## clean up
+##
+
+connection master;
```
percona-ysorokin:
```suggestion
--source include/rpl_connection_master.inc
```

mysql-test/include/rpl_ddl_high_priority.inc
```diff
@@ -0,0 +1,73 @@
...
+drop table t1;
+
+--source include/sync_slave_sql_with_master.inc
```
percona-ysorokin: Redundant

mysql-test/suite/rpl/t/rpl_ddl_high_priority.test
```diff
@@ -0,0 +1 @@
+--source include/rpl_ddl_high_priority.inc
```
percona-ysorokin: Why this additional level of indirection. Can the content of `rpl_ddl_high_priority.inc` be put here inline?

mysql-test/t/ddl_high_priority_sync.test
```diff
@@ -0,0 +1,28 @@
+--source include/have_debug_sync.inc
+
```
percona-ysorokin, line 3: `--source include/count_sessions.inc`

mysql-test/t/ddl_high_priority_sync.test
```diff
@@ -0,0 +1,28 @@
...
+--source include/wait_until_count_sessions.inc
+
+set debug_sync='now WAIT_FOR parked';
+set debug_sync='now SIGNAL go';
+let $wait_condition= select count(*) > 0 from information_schema.tables where table_name = 'new_t1';
+--source include/wait_condition.inc
+
+DROP TABLE new_t1;
+
+disconnect con2;
```
percona-ysorokin, line 29: --source include/wait_until_count_sessions.inc

sql/mdl.cc
```diff
@@ -653,6 +657,8 @@ class MDL_lock {
       m_strategy->m_notify_conflicting_locks(ctx, this);
   }
 
+  bool kill_conflicting_locks(MDL_context *ctx) { return m_strategy->m_kill_conflicting_locks ? m_strategy->m_kill_conflicting_locks(ctx, this) :  false; }
```
percona-ysorokin: Line too long

sql/mdl.cc
```diff
@@ -2170,6 +2179,7 @@ const MDL_lock::MDL_lock_strategy MDL_lock::m_scoped_lock_strategy = {
       locks.
     */
     NULL,
+    NULL,
```
percona-ysorokin:
```suggestion
    nullptr
```

sql/mdl.cc
```diff
@@ -2911,7 +2922,7 @@ bool MDL_context::try_acquire_lock_impl(MDL_request *mdl_request,
                MYF(0));
       return true;
     }
-    ticket->m_hton_notified = true;
+      ticket->m_hton_notified = true;
```
percona-ysorokin: Formatting

sql/mdl.cc
```diff
@@ -3341,6 +3353,26 @@ void MDL_lock::object_lock_notify_conflicting_locks(MDL_context *ctx,
   }
 }
 
+  bool MDL_lock::object_lock_kill_conflicting_locks(MDL_context *ctx,
```
percona-ysorokin: Formatting

sql/sql_class.h
```diff
@@ -2582,6 +2582,19 @@ class THD : public MDL_context_owner,
...
+    blocking threads which hold shared MDL locks.
+    Invoked when acquiring an exclusive lock, for each thread that
+    has a conflicting shared metadata lock.
+    @note The current thread must hold an upgradable MDL lock, and must be a
+    super user to kill other threads.
+    @param ctx_in_use           The MDL context owner (thread) to wake up.
+    @retval  TRUE   if the blocking thread is killed
+    @retval  FALSE  No thread is killed
+   */
+  virtual bool kill_shared_locks(MDL_context_owner *in_use);
```
percona-ysorokin:
```suggestion
  virtual bool kill_shared_locks(MDL_context_owner *in_use) override;
```

sql/sql_lex.h
```diff
@@ -3612,6 +3612,10 @@ struct LEX : public Query_tables_list {
   /* Partition info structure filled in by PARTITION BY parse part */
   partition_info *part_info;
 
+
+    /* High priority DDL */
```
percona-ysorokin: Formatting

sql/sql_parse.cc
```diff
@@ -1051,6 +1051,29 @@ bool is_explainable_query(enum enum_sql_command command) {
   return (sql_command_flags[command] & CF_CAN_BE_EXPLAINED) != 0;
 }
 
+bool support_high_priority(enum enum_sql_command command)
```
percona-ysorokin:
```suggestion
bool support_high_priority(enum enum_sql_command command) noexcept
```

sql/sql_parse.h
```diff
@@ -93,6 +93,7 @@ bool merge_sp_var_charset_and_collation(const CHARSET_INFO **to,
 bool check_host_name(const LEX_CSTRING &str);
 bool mysql_test_parse_for_slave(THD *thd);
 bool is_update_query(enum enum_sql_command command);
+bool support_high_priority(enum enum_sql_command command);
```
percona-ysorokin:
```suggestion
bool support_high_priority(enum enum_sql_command command) noexcept;
```

unittest/gunit/test_mdl_context_owner.h
```diff
@@ -53,6 +53,8 @@ class Test_MDL_context_owner : public MDL_context_owner {
   virtual void notify_hton_post_release_exclusive(const MDL_key *) {}
 
   virtual uint get_rand_seed() { return 0; }
+
+     virtual bool kill_shared_locks(MDL_context_owner * /* unused */) { return false; }
```
percona-ysorokin: Formatting
```suggestion
     virtual bool kill_shared_locks(MDL_context_owner * /* unused */) override { return false; }
```

percona-ysorokin (changes requested):

mysql-test/include/ddl_high_priority.inc
```diff
@@ -0,0 +1,194 @@
...
+--let $cmd = alter table t1 modify i bigint;
+--let $high_priority_cmd = alter high_priority table t1 modify i bigint;
+
+--source include/ddl_high_priority_module.inc
+
+disconnect con2;
+
+# Ensure con2 is closed
+--source include/wait_until_count_sessions.inc
+# Secondary check to ensure con2 is closed
```
laurynas-biveinis: The "primary" check checks the code path which is already past the "secondary" check point. Consider removing the secondary check

mysql-test/include/ddl_high_priority.inc
```diff
@@ -0,0 +1,194 @@
...
+}
+
+
+##
+## clean up
+##
+
+drop user test_user1@localhost;
+drop user test_user2@localhost;
+--disable_warnings
```
laurynas-biveinis: No need for warnings and if exists

laurynas-biveinis (changes requested):

mysql-test/suite/rpl/t/rpl_ddl_high_priority.test
dutow: The testcase also has a rocksdb variant.

percona-ysorokin: Oh, I see now.

dutow:
Updated.

percona-ysorokin (approved):
LGTM

mysql-test/suite/rpl/t/rpl_ddl_high_priority.test
hermanlee: Is there a .result file for this test?

mysql-test/include/rpl_ddl_high_priority.inc
```diff
@@ -0,0 +1,56 @@
...
+create table t1 (i int);
+insert into t1 values (1), (2), (3);
+
+--source include/sync_slave_sql_with_master.inc
+
+--source include/rpl_connection_slave.inc
+select * from t1;
+
+--let $rpl_connection_name= slave_block
+--source include/rpl_connect.inc
```
hermanlee: I think this should be `include/rpl_connection.inc`

sql/mdl.cc
```diff
@@ -3341,6 +3357,23 @@ void MDL_lock::object_lock_notify_conflicting_locks(MDL_context *ctx,
...
 }
 
+bool MDL_lock::object_lock_kill_conflicting_locks(MDL_context *ctx,
+                                                  MDL_lock *lock) {
+  Ticket_iterator it(lock->m_granted);
+  MDL_ticket *conflicting_ticket;
+
+  while ((conflicting_ticket = it++)) {
+    if (conflicting_ticket->get_ctx() != ctx &&
+        conflicting_ticket->get_type() <= MDL_SHARED_NO_WRITE) {
```
hermanlee, line 3367: This seems to have changed from MDL_SHARED_UPGRADEABLE in the original diff to MDL_SHARED_NO_WRITE.

In looking through the code and testing, it looks like `lock tables <table> read;` has change the enumeration of the lock acquired to MDL_SHARED__READ_ONLY, which is higher than MDL_SHARED_UPGRADEABLE. I assume that is why this was done?

hermanlee (changes requested):
I don't think PR is working as intended. A simple test, even with regular DML conflicting with DDL, does not pass. It looks like there are some changes in MDL_context::acquire_lock_nsec which is causing the high priority ddl request to loop around notify_conflicting_locks(), and never getting to issuing the kill.

In my test case:

```
connection 1:
create table t1 (a int, b int, primary key (a)) engine=innodb;
insert into t1 values (1,1);
begin;
select * from t1;

--connection 2:
 create high_priority index idx1 on t1 (b); /* Blocks */
```

Here's the output from performance_schema.metadata_locks:

```
mysql> select * from performance_schema.metadata_locks;
+-------------------+--------------------+----------------+-------------+-----------------------+---------------------+---------------+-------------+--------------------+-----------------+----------------+
| OBJECT_TYPE       | OBJECT_SCHEMA      | OBJECT_NAME    | COLUMN_NAME | OBJECT_INSTANCE_BEGIN | LOCK_TYPE           | LOCK_DURATION | LOCK_STATUS | SOURCE             | OWNER_THREAD_ID | OWNER_EVENT_ID |
+-------------------+--------------------+----------------+-------------+-----------------------+---------------------+---------------+-------------+--------------------+-----------------+----------------+
| TABLE             | test               | t1             | NULL        |       140736473807280 | SHARED_READ_ONLY    | TRANSACTION   | GRANTED     | sql_parse.cc:6125  |              42 |          37794 |
| SCHEMA            | test               | NULL           | NULL        |       140736473813200 | INTENTION_EXCLUSIVE | TRANSACTION   | GRANTED     | dd_schema.cc:108   |              42 |          37797 |
| COLUMN STATISTICS | test               | t1             | a           |       140736473814240 | SHARED_READ         | STATEMENT     | GRANTED     | sql_base.cc:573    |              42 |          38549 |
| COLUMN STATISTICS | test               | t1             | b           |       140736473815680 | SHARED_READ         | STATEMENT     | GRANTED     | sql_base.cc:573    |              42 |          38549 |
| GLOBAL            | NULL               | NULL           | NULL        |       140736461202128 | INTENTION_EXCLUSIVE | STATEMENT     | GRANTED     | sql_base.cc:5412   |              46 |           1588 |
| SCHEMA            | test               | NULL           | NULL        |       140736461202688 | INTENTION_EXCLUSIVE | TRANSACTION   | GRANTED     | sql_base.cc:5399   |              46 |           1588 |
| TABLE             | test               | t1             | NULL        |       140736461201888 | SHARED_UPGRADABLE   | TRANSACTION   | GRANTED     | sql_parse.cc:6125  |              46 |           1590 |
| BACKUP LOCK       | NULL               | NULL           | NULL        |       140736461202288 | INTENTION_EXCLUSIVE | TRANSACTION   | GRANTED     | sql_base.cc:5419   |              46 |           1591 |
| TABLESPACE        | NULL               | test/t1        | NULL        |       140736461202528 | INTENTION_EXCLUSIVE | TRANSACTION   | GRANTED     | lock.cc:806        |              46 |           1697 |
| TABLE             | test               | #sql-339fa7_c  | NULL        |       140736461201968 | EXCLUSIVE           | STATEMENT     | GRANTED     | sql_table.cc:14687 |              46 |           1722 |
| TABLE             | test               | t1             | NULL        |       140736461203088 | EXCLUSIVE           | TRANSACTION   | PENDING     | mdl.cc:3720        |              46 |           1732 |
| TABLE             | performance_schema | metadata_locks | NULL        |       140736473816240 | SHARED_READ         | TRANSACTION   | GRANTED     | sql_parse.cc:6125  |              42 |          38921 |
+-------------------+--------------------+----------------+-------------+-----------------------+---------------------+---------------+-------------+--------------------+-----------------+----------------+
```

sql/mdl.cc
dutow: Yes - and I think the issue was more generic than lock tables, but it's been 4 weeks since I debugged this issue.

mysql-test/suite/rpl/t/rpl_ddl_high_priority.test
dutow: There should be, as I remember creating it, and also that it was working. This and the blocking issue is probably related to a bad rebasing/amending or recreating the entire commit from a diff later, as I did that multiple times during this. I'll look into this and fix it.

dutow:
@hermanlee Fixed - I'm still running a full test suite to see if there's any issues, but the high priority part is working again. I left out a part of one of the commits when I rebased on the separate lock wait timeout commit, and that caused these issues.

sql/sql_yacc.yy
```diff
@@ -2908,15 +2918,15 @@ create_index_stmt:
...
+        | create_opt FULLTEXT_SYM INDEX_SYM ident ON_SYM table_ident
           '(' key_list_with_expression ')' opt_fulltext_index_options opt_index_lock_and_algorithm
           {
             $$= NEW_PTN PT_create_index_stmt(YYMEM_ROOT, KEYTYPE_FULLTEXT, $4,
                                              NULL, $6, $8, $10,
                                              $11.algo.get_or_default(),
                                              $11.lock.get_or_default());
           }
-        | CREATE SPATIAL_SYM INDEX_SYM ident ON_SYM table_ident
+        | create_opt SPATIAL_SYM INDEX_SYM ident ON_SYM table_ident
```
hermanlee: I'm seeing bison register another conflict:

/home/herman/rocks-mysql/8.0/sql/sql_yacc.yy: error: shift/reduce conflicts: 110 found, 109 expected

The conflict comes from the above rule and rule for CREATE SPATIAL_SYM REFERENCE_SYM". Changing "CREATE SPATIAL_SYM REFERENCE_SYM" to "create_opt SPATIAL_SYM REFERENCE_SYM" for the create_srs_stmt production resolves this, but also changing the create_srs_stmt -> CREATE OR_SYM REPLACE_SYM SPATIAL_SYM REFERENCE_SYM causes another conflict.

What is the best way to resolve this, if any?

sql/system_variables.h
```diff
@@ -228,6 +228,7 @@ struct System_variables {
   ulonglong histogram_generation_max_mem_size;
   ulong join_buff_size;
   ulonglong lock_wait_timeout_nsec;
+  ulonglong high_priority_lock_wait_timeout_nsec;
```
hermanlee, line 231: I noticed the original patch wasn't updated this initializing this correctly, resulting in it being 0 unless it's explicitly set. The fix is to initialize it similar to lock_wait_timeout_nsec in mysqld.cc:get_options().

mysql-test/include/rpl_ddl_high_priority.inc
```diff
@@ -0,0 +1,50 @@
...
+
+create table t1 (i int);
+insert into t1 values (1), (2), (3);
+
+--source include/sync_slave_sql_with_master.inc
+
+--source include/rpl_connection_slave.inc
+select * from t1;
+
+--source include/rpl_connection_slave.inc
```
hermanlee: This needs to be another connection to the slave. It gets killed by high priority ddl and the subsequent sync_slave_with_master call on it will fail.

mysql-test/include/rpl_ddl_high_priority.inc
```diff
@@ -0,0 +1,50 @@
...
+
+--source include/rpl_connection_master.inc
+
+# issue a high_pri command from master
+create high_priority index idx1 on t1 (i);
+show create table t1;
+
+--source include/rpl_connection_slave.inc
+
+select * from performance_schema.metadata_locks;
```
hermanlee: It's not clear to me this output is deterministic..

sql/sql_yacc.yy
dutow: I would approach this from a different angle:

* do you want to add the high priority option to other commands?
* if yes, do you actually want those to be logical? (e.g. if one variant gets it, others get it too)

I'm not sure if it can be correctly resolved if one of the answers is no. (and the conflict was there in 5.6 too)

hermanlee: Yes. We should add high priority ddl to those statements too and having them be logical would be fine.

dutow: To clarify: this statement is the `CREATE SPATIAL REFERENCE SYSTEM`, not the `CREATE SPATIAL INDEX`, and I don't think high priority ddl will do anything for it?

But if I do add the high priority option to it, the warning is fixable, so I updated my PR with that.

If I do not add the option to it, the conflict is only resolvable with a larger refactoring, which I don't think we should do.

hermanlee: Yes, adding high priority DDL to those statements to fix the warning without a major refactoring is fine. That's the temporary fix I'm using at the moment to remove the warning. Since high priority ddl is basically a noop for it, I think it would be okay.

mysql-test/include/ddl_high_priority.inc
```diff
@@ -0,0 +1,188 @@
...
+--let $high_priority_cmd = truncate high_priority t1;
+
+--source include/ddl_high_priority_module.inc
+
+##
+## high_priority create/drop trigger
+##
+
+#--let $blocking_sql = lock tables t1 read;
+#--let $cmd = create trigger ins_sum before insert on t1 for each row set @sum = @sum + new.i;
```
hermanlee, line 137: I just noticed this test was disabled and it looks like create trigger always runs as high priority ddl even when high priority ddl is not specified. Have you had a chance to debug this and determine what is causing it?

hermanlee: After discussion with the team, I think we're going to remove the high_priority syntax from the create statements and rely on other means to support this functionality over replication. So no need to investigate this further. Thanks!

dutow: So I should remove the syntax related parts of this commit, and keep only the session variable?

hermanlee: Yes, that would be great. It'll simplify the PR. We will rely on the session variable to enable hi priority ddl.

PR #1012, opened 2019-04-03
FB8-117: Add thread ID option for SQL SET command

inikep:
Summary:

Jira issue: https://jira.percona.com/browse/FB8-117

Reference Patch: https://github.com/facebook/mysql-5.6/commit/19f5aad

Added thread ID option to the SQL SET command. The syntax of the command is as follows:

`SET SESSION [THREAD_ID] var_name = var_value [, [SESSION [THREAD_ID]] var_name = var_value]`

Notes:
- The thread ID can be specified only with the SESSION keyword
- If no thread ID is provided, it is assumed to be the current thread
- Similarly, if own thread ID is provided, it is equivalent to `SET var_name = var_value`
- If the thread with given ID does not exist, it will fail
- For commands like `SET SESSION THREAD_ID var1 = var_value1, var2 = var_value2, var3 = var_value3` the SESSION THREAD_ID will apply to all three variables as explained in https://dev.mysql.com/doc/refman/5.6/en/set-variable.html

### DIFF
--- a/mysql-test/t/set_thread_var.test
+++ b/mysql-test/t/set_thread_var.test
@@ -1,0 +1,152 @@
+--source include/count_sessions.inc
+
+#
+# test SET VARIABLES with thread id
+#
+
+--let $conn_root = `SELECT CONNECTION_ID()`
+
+# Create a regular user
+CREATE USER mysqluser1;
+
+--connect (user1,localhost,mysqluser1)
+--let $conn_mysqluser1 = `SELECT CONNECTION_ID()`
+
+# Set a variable for current thd
+SET max_join_size = 1000;
+
+--let $assert_text = Check if max_join_size is correct
+--let $assert_cond = @@max_join_size = 1000
+--source include/assert.inc
+
+
+# Connect as root
+--connection default
+
+# Set a variable for current thd
+SET max_join_size = 2000;
+
+--let $assert_text = Check if max_join_size is correct
+--let $assert_cond = @@max_join_size = 2000
+--source include/assert.inc
+
+# Set variable for user1 thd from root thd
+--replace_regex /SESSION [0-9]*/SESSION $mysqluser1/
+--eval SET SESSION $conn_mysqluser1 max_join_size = 1001
+
+--let $assert_text = Check if max_join_size is correct
+--let $assert_cond = @@max_join_size = 2000
+--source include/assert.inc
+
+# Set variable for root thd from root thd
+--replace_regex /SESSION [0-9]*/SESSION $conn_root/
+--eval SET SESSION $conn_root max_join_size = 2001
+
+--let $assert_text = Check if max_join_size is correct
+--let $assert_cond = @@max_join_size = 2001
+--source include/assert.inc
+
+# Set variable for root thd from root thd without session ID
+SET SESSION max_join_size = 2002;
+
+--let $assert_text = Check if max_join_size is correct
+--let $assert_cond = @@max_join_size = 2002
+--source include/assert.inc
+
+
+--connection user1
+
+--let $assert_text = Check if max_join_size is correct
+--let $assert_cond = @@max_join_size = 1001
+--source include/assert.inc
+
+
+# Reconnect as root
+--connection default
+
+# Set variable for both current thd and user1 from root thd
+--replace_regex /SESSION [0-9]* max_join_size = 1002, SESSION [0-9]* max_join_size = 2003/SESSION $conn_mysqluser1 max_join_size = 1002, SESSION $conn_root max_join_size = 2003/
+--eval SET SESSION $conn_mysqluser1 max_join_size = 1002, SESSION $conn_root max_join_size = 2003
+
+--let $assert_text = Check if max_join_size is correct
+--let $assert_cond = @@max_join_size = 2003
+--source include/assert.inc
+
+
+--connection user1
+
+--let $assert_text = Check if max_join_size is correct
+--let $assert_cond = @@max_join_size = 1002
+--source include/assert.inc
+
+# Set variable for root thd from user1 thd
+--replace_regex /SESSION [0-9]*/SESSION $conn_root/ /thread [0-9]*/thread $root/
+--error ER_SPECIFIC_ACCESS_DENIED_ERROR
+--eval SET SESSION $conn_mysqluser1 max_join_size = 1003, SESSION $conn_root max_join_size = 2004
+
+--let $assert_text = Check if max_join_size is correct
+--let $assert_cond = @@max_join_size = 1002
+--source include/assert.inc
+
+# Set variable for user1 thd from user1 thd
+--replace_regex /SESSION [0-9]*/SESSION $mysqluser1/
+--eval SET SESSION $conn_mysqluser1 max_join_size = 1003
+
+--let $assert_text = Check if max_join_size is correct
+--let $assert_cond = @@max_join_size = 1003
+--source include/assert.inc
+
+# Set variable for user1 thd from user1 thd without session ID
+SET SESSION max_join_size = 1004;
+
+--let $assert_text = Check if max_join_size is correct
+--let $assert_cond = @@max_join_size = 1004
+--source include/assert.inc
+
+
+--connection default
+
+--let $assert_text = Check if max_join_size is correct
+--let $assert_cond = @@max_join_size = 2003
+--source include/assert.inc
+
+# Set multiple variables for both current thd and user1 from root thd
+--replace_regex /SESSION [0-9]* max_join_size = 1005/SESSION $conn_mysqluser1 max_join_size = 1005/ /SESSION [0-9]* max_join_size = 2004/SESSION $conn_root max_join_size = 2004/
+--eval SET SESSION $conn_mysqluser1 max_join_size = 1005, max_error_count = 30, SESSION $conn_root max_join_size = 2004, max_error_count = 31
+
+--let $assert_text = Check if max_join_size is correct
+--let $assert_cond = @@max_join_size = 2004 AND @@max_error_count = 31
+--source include/assert.inc
+
+
+--connection user1
+
+--let $assert_text = Check if max_join_size is correct
+--let $assert_cond = @@max_join_size = 1005 AND @@max_error_count = 30
+--source include/assert.inc
+
+--error ER_NO_SUCH_THREAD
+--eval SET SESSION 999 max_join_size = 9999
+
+--error ER_PARSE_ERROR
+--eval SET GLOBAL 1 max_join_size = 1000
+
+--error ER_PARSE_ERROR
+--eval SET SESSION -1 max_join_size = 1000
+
+--error ER_PARSE_ERROR
+--eval SET SESSION 1 TRANSACTION READ WRITE
+
+--connection default
+SET GLOBAL TRANSACTION READ ONLY;
+SET SESSION TRANSACTION READ ONLY;
+--let $assert_text = Check if SET SESSION/GLOBAL TRANSACTION READ ONLY works correctly
+--let $assert_cond = @@GLOBAL.transaction_read_only = 1 AND @@transaction_read_only = 1
+--source include/assert.inc
+SET GLOBAL TRANSACTION READ WRITE;
+SET SESSION TRANSACTION READ WRITE;
+
+# cleanup
+DROP USER mysqluser1;
+--disconnect user1
+--source include/wait_until_count_sessions.inc
--- a/sql/parse_tree_helpers.cc
+++ b/sql/parse_tree_helpers.cc
@@ -164,6 +164,7 @@ bool set_system_variable(THD *thd, struct sys_var_with_base *var_with_base,
                                         &var_with_base->base_name, val)))
     return true;
 
+  var->thd_id = lex->thread_id_opt;
   return lex->var_list.push_back(var);
 }
 
@@ -358,12 +359,14 @@ bool sp_create_assignment_instr(THD *thd, const char *expr_end_ptr) {
   }
 
   /* Remember option_type of the currently parsed LEX. */
-  enum_var_type inner_option_type = lex->option_type;
+  const enum_var_type inner_option_type = lex->option_type;
+  const ulong thd_id_opt = lex->thread_id_opt;
 
   if (sp->restore_lex(thd)) return true;
 
   /* Copy option_type to outer lex in case it has changed. */
   thd->lex->option_type = inner_option_type;
+  thd->lex->thread_id_opt = thd_id_opt;
 
   return false;
 }
--- a/sql/parse_tree_nodes.h
+++ b/sql/parse_tree_nodes.h
@@ -1162,19 +1162,21 @@ class PT_option_value_no_option_type_password_for
 };
 
 class PT_option_value_type : public Parse_tree_node {
-  typedef Parse_tree_node super;
-
-  enum_var_type type;
+  enum_var_type option_type;
+  ulong thread_id;
   PT_option_value_following_option_type *value;
 
  public:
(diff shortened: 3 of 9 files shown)


sql/set_var.cc
```diff
@@ -801,6 +803,59 @@ int sql_set_variables(THD *thd, List<set_var_base> *var_list, bool opened) {
...
+  */
+  std::unordered_map<ulong, List<set_var_base>> thd_var_map;
+
+  while ((var = it++)) {
+    ulong thd_id_opt = var->thd_id;
+    if (thd_id_opt == current_thd_id) thd_id_opt = 0;
+    thd_var_map[thd_id_opt].push_back(var);
+  }
+
+  for (auto var_iter : thd_var_map) {
```
laurynas-biveinis:
```suggestion
  for (const auto &var_iter : thd_var_map) {
```

sql/sql_yacc.yy
```diff
@@ -450,7 +450,7 @@ void warn_about_deprecated_national(THD *thd)
   1. We do not accept any reduce/reduce conflicts
   2. We should not introduce new shift/reduce conflicts any more.
 */
-%expect 108
+%expect 109
```
laurynas-biveinis: I guess here (and in the high priority DDL PR) no easy way to avoid this?

laurynas-biveinis (changes requested):

mysql-test/t/set_thread_var.test
```diff
@@ -0,0 +1,148 @@
+--source include/count_sessions.inc
+
+#
+# test SET VARIABLES with thread id
+#
+
+--let $conn_root = query_get_value(SELECT * FROM information_schema.processlist WHERE USER = 'root',ID,1)
```
percona-ysorokin: There is a simpler way
```suggestion
--let $conn_root = `SELECT CONNECTION_ID()`
```

mysql-test/t/set_thread_var.test
```diff
@@ -0,0 +1,148 @@
...
+
+# Create a regular user
+CREATE USER mysqluser1;
+
+--connect (user1,localhost,mysqluser1)
+--let $conn_mysqluser1 = query_get_value(SELECT * FROM information_schema.processlist WHERE USER = 'mysqluser1',ID,1)
```
percona-ysorokin: There is a simpler way
```suggestion
--let $conn_mysqluser1 = `SELECT CONNECTION_ID()`
```

mysql-test/t/set_thread_var.test
```diff
@@ -0,0 +1,148 @@
...
+
+--let $assert_text = Check if max_join_size is correct
+--let $assert_cond = @@max_join_size = 1002
+--source include/assert.inc
+
+# Set variable for root thd from user1 thd
+--replace_regex /SESSION [0-9]*/SESSION $conn_root/ /thread [0-9]*/thread $root/
+--error ER_SHOW_DENIED_ERROR
+--eval SET SESSION $conn_root max_join_size = 2004
+
```
percona-ysorokin, line 90: Also check what happens when you try to set 2 vars simultaneously (one for connection default(root) and one for connection user1) from a non-privileged user.
```
--eval SET SESSION $conn_root max_join_size = XXX, SESSION $conn_mysqluser1 max_join_size  = YYY
```
Is this going to be:
1. `ER_SHOW_DENIED_ERROR` and no side effects.
2. `ER_SHOW_DENIED_ERROR` and `max_join_size` changed for $conn_mysqluser1.
3. No error and `max_join_size` changed for $conn_mysqluser1.

According to https://dev.mysql.com/doc/refman/5.6/en/set-variable.html
```
If any variable assignment in a SET statement fails, the entire statement fails and no variables are changed.
```

mysql-test/t/set_thread_var.test
```diff
@@ -0,0 +1,148 @@
...
+--eval SET GLOBAL 1 max_join_size = 1000
+
+--error ER_PARSE_ERROR
+--eval SET SESSION -1 max_join_size = 1000
+
+--error ER_PARSE_ERROR
+--eval SET SESSION 1 TRANSACTION READ WRITE
+
+--connection default
+SET GLOBAL TRANSACTION READ ONLY;
```
percona-ysorokin, line 141: @hermanlee 
Do we also need `SET SESSION [THREAD_ID] TRANSACTION ...` syntax extensions for completeness?

sql/parse_tree_nodes.h
```diff
@@ -1161,20 +1161,33 @@ class PT_option_value_no_option_type_password_for
...
-
+class PT_option_type_with_thread_id : public Parse_tree_node {
+  ulong thread_id;
   enum_var_type type;
+
+ public:
+  PT_option_type_with_thread_id(ulong thread_id_arg, enum_var_type type_arg)
+      : thread_id(thread_id_arg), type(type_arg) {}
+
+  virtual bool contextualize(Parse_context *pc) {
```
percona-ysorokin:
```suggestion
  virtual bool contextualize(Parse_context *pc) override {
```

sql/set_var.cc
percona-ysorokin: Because of the `unordered_map`, variables may be set not in the order they are specified and not even in the order of <thread_id>. Is this OK?
For instance, in
`SET SESSION <thread_id1> xxx = kkk, SESSION <thread_id2> yyy = nnn;`, where `SET SESSION <thread_id2> yyy = nnn` is supposed to fail and `SET SESSION <thread_id1> xxx = kkk` is supposed to succeed,
The order in the `thd_var_map` container is important because we may get different output:
1. `xxx` is changed, `yyy` is not.
2. nothing is changed.

sql/set_var.cc
```diff
@@ -801,6 +803,59 @@ int sql_set_variables(THD *thd, List<set_var_base> *var_list, bool opened) {
...
+    This map stores a list of variables to be updated for each thread specified
+    by the user, essentially grouping variabled by thread ID. Variable updates
+    are processed per thread.
+  */
+  std::unordered_map<ulong, List<set_var_base>> thd_var_map;
+
+  while ((var = it++)) {
+    ulong thd_id_opt = var->thd_id;
+    if (thd_id_opt == current_thd_id) thd_id_opt = 0;
+    thd_var_map[thd_id_opt].push_back(var);
```
percona-ysorokin: Grouping variables by <thread_id>, although seems like a good optimization from the performance point of view, may have side effects - vars will be set not in the order they are specified.
It may be important if one of the assignments is supposed to fail.

sql/set_var.cc
```diff
@@ -801,6 +803,59 @@ int sql_set_variables(THD *thd, List<set_var_base> *var_list, bool opened) {
...
+        if (!sctx->check_access(SUPER_ACL) &&
+            !sctx->has_global_grant(STRING_WITH_LEN("SYSTEM_VARIABLES_ADMIN"))
+                 .first) {
+          error = 1;
+          my_error(ER_SHOW_DENIED_ERROR, MYF(0), var_iter.first);
+        } else {
+          error =
+              sql_set_variables_internal(change_thd, &var_iter.second, false);
+        }
+        mysql_mutex_unlock(&change_thd->LOCK_thd_data);
```
percona-ysorokin: I don't see any corresponding `mysql_mutex_lock(&change_thd->LOCK_thd_data);` here.
Was it acquired implicitly inside `Global_THD_manager::get_instance()->find_thd(&find_thd_with_id);` ?

percona-ysorokin (changes requested):

sql/set_var.cc
inikep: Yes, it's locked at https://github.com/facebook/mysql-5.6/blob/fb-mysql-8.0.13/sql/mysqld_thd_manager.cc#L62

sql/set_var.cc
inikep: It's fixed

sql/set_var.cc
inikep: It's fixed

inikep:
PR was updated.

mysql-test/t/set_thread_var.test
inikep: `SET SESSION [THREAD_ID] TRANSACTION` returns `ER_PARSE_ERROR` with both this and original patches. We can add `SET SESSION [THREAD_ID] TRANSACTION` but this will be an additional feature.

percona-ysorokin (approved):
LGTM

sql/set_var.cc
```diff
@@ -733,6 +735,61 @@ sys_var *intern_find_sys_var(const char *str, size_t length) {
...
+      Find_thd_with_id find_thd_with_id(thd_id);
+      change_thd =
+          Global_THD_manager::get_instance()->find_thd(&find_thd_with_id);
+      if (!change_thd) {
+        my_error(ER_NO_SUCH_THREAD, MYF(0), thd_id);
+        DBUG_RETURN(1);
+      }
+
+      Security_context *sctx = thd->security_context();
+      if (!sctx->check_access(SUPER_ACL) &&
```
hermanlee, line 770: Would it make sense to cache the result of the first security_context check? It doesn't seem like the query result could change if it's a pass. Setting multiple variables is probably rare anyway, so maybe this is not worth considering.

hermanlee (commented):
Most of the code looks good. The changes to the parse tree code seems to work, but I'm wondering if it would be simpler to add the thread_id value to the option_type structure directly instead of create a new one.

sql/parse_tree_nodes.h
```diff
@@ -1161,20 +1161,33 @@ class PT_option_value_no_option_type_password_for
   virtual bool contextualize(Parse_context *pc);
 };
 
-class PT_option_value_type : public Parse_tree_node {
-  typedef Parse_tree_node super;
-
+class PT_option_type_with_thread_id : public Parse_tree_node {
```
hermanlee: I seem to have lost some review comments.

It seems thread_id could be pushed into PT_option_value_type directly and the constructor can support an extra argument for thread_id (enum_var_type, ulong, PT_option_value_following_option_type*)

Then the grammer rule for option_value could be modified to:

```
option_value:
          option_type option_value_following_option_type
          {
            $$= NEW_PTN PT_option_value_type($1, 0, $2);
          }
          | SESSION_SYM opt_var_tid option_value_following_option_type
          {
            $$= NEW_PTN PT_option_value_type(OPT_SESSION, $2, $3);
          }        
          | LOCAL_SYM opt_var_tid option_value_following_option_type
          {
            $$= NEW_PTN PT_option_value_type(OPT_SESSION, $2, $3);
          }
          | option_value_no_option_type { $$= $1; }
          ;
```

It seems then the rest of the grammer rules wouldn't have to change.

sql/set_var.cc
inikep: This code is called only with `SET SESSION THREAD_ID` and it's not called without `THREAD_ID`. The optimization will make sense only for `SET SESSION THREAD_ID var_name1 = var_value1, SET SESSION THREAD_ID var_name2 = var_value2` or more values.

sql/parse_tree_nodes.h
inikep: `option_type_with_thread_id:` is used for both `option_value:` and `start_option_value_list_following_option_type:`. 
IMHO merging `option_value:` with `option_type_with_thread_id` as you've proposed will make everything more complex.

sql/set_var.cc
hermanlee: Yes. The use case I was thinking of would have been trying to update one variable in all of outstanding sessions, so the call would be:

SET SESSION thread_id1 var_name = var_value, SESSION thread_id2 var_name = var_value, SESSION thread_id3 var_name = var_value;

But it's minor anyhow.

sql/parse_tree_nodes.h
hermanlee: Makes sense. The grammer rules for the `start_option_value_list:` case would need to be extended to support the cases where option_type is SESSION,  but disabled for the the SESSION TRANSACTION subset, which is probably more complicated.

hermanlee (approved):

sql/set_var.cc
```diff
@@ -733,6 +735,61 @@ sys_var *intern_find_sys_var(const char *str, size_t length) {
...
+          !sctx->has_global_grant(STRING_WITH_LEN("SYSTEM_VARIABLES_ADMIN"))
+               .first) {
+        mysql_mutex_unlock(&change_thd->LOCK_thd_data);
+        my_error(ER_SHOW_DENIED_ERROR, MYF(0), thd_id);
```
lth: I don't know if the change of behaviour here is intentional, but previously, we just checked if the user was an owner of the thread, instead of checking for specific privileges. Now the error is kind of misleading, because you'll get "You are not owner of thread", when we should really be getting something from `ER_SPECIFIC_ACCESS_DENIED_ERROR`.

sql/sql_yacc.yy
lth: We can fix this by reworking the grammar rules so that the thread id in option_type_with_thread_id is non-optional. It looks like we're confusing the `start_option_value_list_following_option_type` rule, because it doesn't know if it needs to reduce `option_type` to `option_type_with_thread_id`.

sql/parse_tree_helpers.cc
```diff
@@ -358,12 +359,14 @@ bool sp_create_assignment_instr(THD *thd, const char *expr_end_ptr) {
   }
 
   /* Remember option_type of the currently parsed LEX. */
-  enum_var_type inner_option_type = lex->option_type;
+  const enum_var_type inner_option_type = lex->option_type;
```
lth, line 362: Not too important, but I wonder if this code path actually works? It'd be nice to have simple test for set statements in stored procedures.

lth (commented):
I mainly have a question about the change in behaviour in the security checks.

sql/set_var.cc
lth: I spoke with @hermanlee and the change in behaviour is probably okay.

However, I think we should use `ER_SPECIFIC_ACCESS_DENIED_ERROR` instead so that the error is less confusing.

sql/sql_yacc.yy
hermanlee: @inikep, as per @lth's suggestion, can something like the following grammar rules be used instead and can the parse tree be able to support it easily? It basically removes option_type from option_type_with_thread_id and explicitly adds productions for option_type_with_thread_id to the appropriate locations that have productions for option_type. Changes to your existing rules would be:

```
option_type_with_thread_id:
          LOCAL_SYM opt_var_tid 
        | SESSION_SYM opt_var_tid
        ;

start_option_value_list_following_option_type:
          option_type option_value_following_option_type option_value_list_continued
        | option_type_with_thread_id option_value_following_option_type option_value_list_continued
        | option_type TRANSACTION_SYM transaction_characteristics
        ;

option_value:
          option_type option_value_following_option_type
        | option_type_with_thread_id option_value_following_option_type
        | option_value_no_option_type
        ;
```

Note I'm only including the grammar rules here without any code associated on those rules. The above change, when compiled with bison, does not have any additional shift/reduce conflicts.

inikep: Done

sql/set_var.cc
inikep: Done

inikep:
PR was updated with the following changes:
1. `ER_SHOW_DENIED_ERROR` was replaced with `ER_SPECIFIC_ACCESS_DENIED_ERROR`
2. The grammar was changed according to the proposal from @hermanlee and appropriate classes were updated

sql/parse_tree_helpers.cc
inikep: It doesn't work because of an upstream bug:
https://bugs.mysql.com/bug.php?id=95734

PR #1013, opened 2019-04-04
Fix for FB8-139: TraceID support in MySQLD

inikep:
Summary:

Reference Patch: https://github.com/facebook/mysql-5.6/commit/40ef633540662d0753bc04094fe10f85b79019ae

1. Remove redundant EOL in query attributes
2. Detect a broken slow log in `main.slow_log_extra_big`

Squash with D14387874


percona-ysorokin (approved):
LGTM

PR #1017, opened 2019-04-25
FB8-181: Maintain a whitelist of semi-sync slaves which can ACK transactions

inikep:
Summary:

Jira issue: https://jira.percona.com/browse/FB8-181

Reference Patch: https://github.com/facebook/mysql-5.6/commit/ee761fc
Reference Patch: https://github.com/facebook/mysql-5.6/commit/68f34521e82
Reference Patch: https://github.com/facebook/mysql-5.6/commit/dfb038d48d0
Reference Patch: https://github.com/facebook/mysql-5.6/commit/de1e8c7bfe7
Reference Patch: https://github.com/facebook/mysql-5.6/commit/26835a39459

A whitelist of semi-sync slaves can be maintained using a new sys-var
`rpl_semi_sync_master_whitelist`. The value of this variable is a comma
separated list of UUIDs. ACK from any slave which is not in the whitelist will
be ignored. The whitelist is persisted in file
`semi_sync_master_whitelist.info` to survive server restarts.

Porting notice:
You have to use `set @@persist.rpl_semi_sync_master_whitelist` to keep
value across server restarts.

TO DO: Because of conflicts I removed changes in `mysql-test/t/all_persisted_variables.test`. 
This test has to be modified at `let $total_persistent_vars=XXX;` (+1 increase)
and it should be re-recorded.

Originally Reviewed By: Tema


plugin/semisync/semisync_master.cc
```diff
@@ -513,13 +523,29 @@ void ReplSemiSyncMaster::lock() { mysql_mutex_lock(&LOCK_binlog_); }
 
 void ReplSemiSyncMaster::unlock() { mysql_mutex_unlock(&LOCK_binlog_); }
 
-void ReplSemiSyncMaster::add_slave() {
+void ReplSemiSyncMaster::add_slave(int server_id) {
   lock();
   rpl_semi_sync_master_clients++;
+  rpl_semi_sync_master_whitelist_set_lock.lock();
```
percona-ysorokin: Use `std::lock_guard` here as well, probably wrap into a new code block `{...}`
```suggestion
  std::lock_guard<std::mutex> guard(rpl_semi_sync_master_whitelist_set_lock);
```

plugin/semisync/semisync_master.cc
```diff
@@ -513,13 +523,29 @@ void ReplSemiSyncMaster::lock() { mysql_mutex_lock(&LOCK_binlog_); }
...
+  } else
+    semisync_info.emplace(server_id, semisync_info_t(is_any, get_slave_uuid()));
+
+  rpl_semi_sync_master_whitelist_set_lock.unlock();
   unlock();
 }
 
-void ReplSemiSyncMaster::remove_slave() {
+void ReplSemiSyncMaster::remove_slave(int server_id) {
+  rpl_semi_sync_master_whitelist_set_lock.lock();
```
percona-ysorokin: Likewise
```suggestion
std::lock_guard<std::mutex> guard(rpl_semi_sync_master_whitelist_set_lock);
```

plugin/semisync/semisync_master.cc
```diff
@@ -554,10 +580,214 @@ bool ReplSemiSyncMaster::is_semi_sync_slave() {
...
+
+  // case: temp file exists, delete it
+  if (!my_access(tmp_file_path, F_OK) &&
+      my_delete(tmp_file_path, MYF(MY_WME))) {
+    LogErr(ERROR_LEVEL, ER_SEMISYNC_FAILED_DELETE_TMP_WHITELIST_FILE,
+           tmp_file_path);
+    return 1;
+  }
+
+  std::ifstream info_file(info_file_path, std::ofstream::in);
```
percona-ysorokin:
```suggestion
  std::ifstream info_file(info_file_path, std::ios_base::in);
```

plugin/semisync/semisync_master.cc
```diff
@@ -554,10 +580,214 @@ bool ReplSemiSyncMaster::is_semi_sync_slave() {
...
+    local_whitelist_set.erase(str);
+  }
+  // case: full comma separated string is specified
+  else {
+    local_whitelist_set = split_into_set(wlist, ',');
+  }
+
+  // re-calculate wlist from the set
+  wlist = "";
+  for (const auto uuid : local_whitelist_set) {
```
percona-ysorokin: Is the order here important?

plugin/semisync/semisync_master.cc
```diff
@@ -554,10 +580,214 @@ bool ReplSemiSyncMaster::is_semi_sync_slave() {
...
+      crc32(0L, reinterpret_cast<const uchar *>(wlist.data()), wlist.length());
+
+  char tmp_file_path[FN_REFLEN];
+  char info_file_path[FN_REFLEN];
+  fn_format(tmp_file_path, SEMI_SYNC_MASTER_WHITELIST_TMP_FILE.c_str(),
+            mysql_data_home, "", MY_UNPACK_FILENAME | MY_RETURN_REAL_PATH);
+  fn_format(info_file_path, SEMI_SYNC_MASTER_WHITELIST_FILE.c_str(),
+            mysql_data_home, "", MY_UNPACK_FILENAME | MY_RETURN_REAL_PATH);
+
+  std::ofstream tmp_info_file(tmp_file_path, std::ofstream::out);
```
percona-ysorokin:
```suggestion
  std::ifstream info_file(info_file_path, std::ios_base::in);
```

plugin/semisync/semisync_master.cc
```diff
@@ -554,10 +580,214 @@ bool ReplSemiSyncMaster::is_semi_sync_slave() {
...
+  if (!tmp_info_file.is_open()) {
+    LogErr(ERROR_LEVEL, ER_SEMISYNC_FAILED_TO_OPEN_WHITELIST_TMP_FILE,
+           tmp_file_path);
+    return false;
+  }
+  tmp_info_file << wlist << std::endl << checksum;
```
percona-ysorokin:
```suggestion
  tmp_info_file << wlist << "\n" << checksum;
```

plugin/semisync/semisync_master.h
```diff
@@ -626,6 +627,23 @@ class ReplSemiSyncMaster : public ReplSemiSyncBase {
 
   AckContainer ack_container_;
 
+  /* Whitelist related members */
+  std::mutex rpl_semi_sync_master_whitelist_set_lock;
+  std::unordered_set<std::string> rpl_semi_sync_master_whitelist_set;
+  std::atomic<ulonglong> rpl_semi_sync_master_whitelist_ver{1};
+  const std::string SEMI_SYNC_MASTER_WHITELIST_FILE =
```
percona-ysorokin: static constexpr?

plugin/semisync/semisync_master.h
```diff
@@ -626,6 +627,23 @@ class ReplSemiSyncMaster : public ReplSemiSyncBase {
...
+      "semi_sync_master_whitelist.info";
+  const std::string SEMI_SYNC_MASTER_WHITELIST_TMP_FILE =
```
percona-ysorokin: static constexpr?

plugin/semisync/semisync_master_plugin.cc
```diff
@@ -217,6 +225,22 @@ static void fix_rpl_semi_sync_master_wait_for_slave_count(MYSQL_THD thd,
                                                           void *ptr,
                                                           const void *val);
 
+static void update_whitelist(THD *, SYS_VAR *, void *var_ptr,
+                             const void *save) {
+  auto wlist_buf = *const_cast<char **>(
+      const_cast<const char **>(static_cast<const char *const *>(save)));
```
percona-ysorokin: Oogh! Looks like too many casts :)

plugin/semisync/semisync_master.h
```diff
@@ -626,6 +627,23 @@ class ReplSemiSyncMaster : public ReplSemiSyncBase {
 
   AckContainer ack_container_;
 
+  /* Whitelist related members */
+  std::mutex rpl_semi_sync_master_whitelist_set_lock;
+  std::unordered_set<std::string> rpl_semi_sync_master_whitelist_set;
```
percona-ysorokin: Name for a custom container? It is used at least 5 times.

sql/handler.cc
```diff
@@ -7868,6 +7868,30 @@ void handler::unlock_shared_ha_data() {
...
+  size_t start = 0;
+  std::unordered_set<std::string> elems;
+
+  // Find next delimiter
+  while ((pos = input.find(delimiter, start)) != std::string::npos) {
+    // If there is any data since the last delimiter add it to the list
+    if (pos > start) elems.insert(input.substr(start, pos - start));
+
+    // Set our start position to the character after the delimiter
+    start = pos + 1;
```
percona-ysorokin, line 7885: Is the case with trailing comma `delimiter` handled properly?

plugin/semisync/semisync_master.cc
percona-ysorokin:
```suggestion
  for (const auto& uuid : local_whitelist_set) {
```

plugin/semisync/semisync_master.cc
```diff
@@ -554,10 +580,214 @@ bool ReplSemiSyncMaster::is_semi_sync_slave() {
...
+  local_whitelist_set = rpl_semi_sync_master_whitelist_set;
+
+  // case: add a single uuid to the whitelist (value starts with +)
+  if (wlist[0] == '+') {
+    const auto str = wlist.substr(1);
+    // case: +ANY specified, start with a clean slate
+    if (str == any_slave) {
+      local_whitelist_set.clear();
+    }
+    local_whitelist_set.insert(str);
```
percona-ysorokin: We still add "ANY" in this case. Is this intentional?

plugin/semisync/semisync_master.cc
```diff
@@ -554,10 +580,214 @@ bool ReplSemiSyncMaster::is_semi_sync_slave() {
...
+  }
+  // case: remove a single uuid to the whitelist (value starts with -)
+  else if (wlist[0] == '-') {
+    const auto str = wlist.substr(1);
+    // case: -ANY specified, start with a clean slate
+    if (str == any_slave) {
+      local_whitelist_set.clear();
+    }
+    local_whitelist_set.erase(str);
```
percona-ysorokin:
```suggestion
  else
     local_whitelist_set.erase(str);
```

plugin/semisync/semisync_master.cc
```diff
@@ -554,10 +580,214 @@ bool ReplSemiSyncMaster::is_semi_sync_slave() {
...
+
+  // remove all spaces
+  wlist.erase(std::remove(wlist.begin(), wlist.end(), ' '), wlist.end());
+
+  std::unordered_set<std::string> local_whitelist_set;
+  local_whitelist_set = rpl_semi_sync_master_whitelist_set;
+
+  // case: add a single uuid to the whitelist (value starts with +)
+  if (wlist[0] == '+') {
+    const auto str = wlist.substr(1);
```
percona-ysorokin: Should we also check that `str` is not a list.
Is `+<guid1>,<guid2>` syntax supported?

plugin/semisync/semisync_master.cc
```diff
@@ -554,10 +580,214 @@ bool ReplSemiSyncMaster::is_semi_sync_slave() {
...
+    // case: +ANY specified, start with a clean slate
+    if (str == any_slave) {
+      local_whitelist_set.clear();
+    }
+    local_whitelist_set.insert(str);
+  }
+  // case: remove a single uuid to the whitelist (value starts with -)
+  else if (wlist[0] == '-') {
+    const auto str = wlist.substr(1);
```
percona-ysorokin: Likewise
Is `-<guid1>,<guid2>` syntax supported?

mysql-test/suite/rpl/t/rpl_semi_sync_master_whitelist.test
```diff
@@ -0,0 +1,277 @@
...
+--source include/assert.inc
+
+--let $sync_slave_connection= slave2
+--source include/sync_slave_sql_with_master.inc
+--let $assert_text= The t1 table should have 5 rows
+--let $assert_cond= count(*) = 5 from t1
+--source include/assert.inc
+
+--source include/rpl_connection_slave1.inc
+start slave;
```
percona-ysorokin:
```suggestion
--source include/start_slave.inc
```

mysql-test/suite/rpl/t/rpl_semi_sync_master_whitelist.test
```diff
@@ -0,0 +1,277 @@
...
+
+--source include/rpl_connection_master.inc
+replace_result $slave1_uuid slave1_uuid $slave2_uuid slave2_uuid;
+select @@global.rpl_semi_sync_master_whitelist;
+
+
+# Check if trx fails when nobody from whitelist ACKs
+# stop all ackers
+--source include/rpl_connection_slave1.inc
+stop slave;
```
percona-ysorokin:
```suggestion
--source include/stop_slave.inc
```

mysql-test/suite/rpl/t/rpl_semi_sync_master_whitelist.test
```diff
@@ -0,0 +1,277 @@
...
+--source include/rpl_connection_slave2.inc
+stop slave;
```
percona-ysorokin:
```suggestion
--source include/stop_slave.inc
```

mysql-test/suite/rpl/t/rpl_semi_sync_master_whitelist.test
```diff
@@ -0,0 +1,277 @@
...
+--let $wait_condition= select count(*) = 1 from information_schema.processlist where state = "Waiting for semi-sync ACK from slave"
+--source include/wait_condition.inc
+
+--let $assert_text= The t1 table should have 5 rows
+--let $assert_cond= count(*) = 5 from t1
+--source include/assert.inc
+
+# start ackers
+--source include/rpl_connection_slave1.inc
+start slave;
```
percona-ysorokin:
```suggestion
--source include/start_slave.inc
```

mysql-test/suite/rpl/t/rpl_semi_sync_master_whitelist.test
```diff
@@ -0,0 +1,277 @@
...
+--source include/rpl_connection_slave2.inc
+start slave;
```
percona-ysorokin:
```suggestion
--source include/start_slave.inc
```

mysql-test/suite/rpl/t/rpl_semi_sync_master_whitelist.test
```diff
@@ -0,0 +1,277 @@
...
+--let $wait_condition= select count(*) = 2 from information_schema.processlist where command = "Binlog Dump"
+--source include/wait_condition.inc
+
+# remove slave2 from whitelist
+--replace_result $slave1_uuid slave1_uuid $slave2_uuid slave2_uuid
+--eval set @@global.rpl_semi_sync_master_whitelist='-$slave2_uuid'
+
+# stop the valid semi-sync slave so that the ACK is received from invalid slave
+--source include/rpl_connection_slave1.inc
+stop slave;
```
percona-ysorokin:
```suggestion
--source include/stop_slave.inc
```

mysql-test/suite/rpl/t/rpl_semi_sync_master_whitelist.test
```diff
@@ -0,0 +1,277 @@
...
+
+--source include/rpl_connection_master.inc
+# wait for the dump thread's death
+--let $wait_timeout= 20
+--let $wait_condition= select count(*) <= 1 from information_schema.processlist where command = "Binlog Dump"
+--source include/wait_condition.inc
+
+# start the valid slave
+--source include/rpl_connection_slave1.inc
+start slave;
```
percona-ysorokin:
```suggestion
--source include/start_slave.inc
```

percona-ysorokin (changes requested):

laurynas-biveinis (changes requested):
semi_sync_master_whitelist.info persistence should be replaced by SET PERSIST

plugin/semisync/semisync_master.cc
inikep: no, `wlist` is used just to replace value of `rpl_semi_sync_master_whitelist`

plugin/semisync/semisync_master.cc
inikep: no

plugin/semisync/semisync_master.cc
inikep: no

plugin/semisync/semisync_master.cc
inikep: Yes, it's required to set `wlist` and then `rpl_semi_sync_master_whitelist`

sql/handler.cc
inikep: yes

inikep:
All comments including `SET PERSIST` were addressed.

percona-ysorokin (approved):
LGTM

plugin/semisync/semisync_master.cc
```diff
@@ -513,13 +522,29 @@ void ReplSemiSyncMaster::lock() { mysql_mutex_lock(&LOCK_binlog_); }
 
 void ReplSemiSyncMaster::unlock() { mysql_mutex_unlock(&LOCK_binlog_); }
 
-void ReplSemiSyncMaster::add_slave() {
+void ReplSemiSyncMaster::add_slave(int server_id) {
```
hermanlee: The server_id comes from the slave, but uniqueness is to be enforced by an external agent. What happens if this is not true and two slaves have matching server_id, but different UUIDs? It seems depending on this value to separate out slaves could be risky, especially since it could lead to whitelisting of slaves whose UUID is not in the list, but whose server_id matched of a slave whose UUID is in the list.

One could make the argument about UUIDs conflicting too and assume the same system that assigned server_id is the same as that for UUID, so either both server_id+UUID conflict or neither conflict. But there is probably some corner case where server_id conflict, but UUID does not and vice versa.

plugin/semisync/semisync_master.cc
```diff
@@ -554,10 +579,121 @@ bool ReplSemiSyncMaster::is_semi_sync_slave() {
...
+  rpl_semi_sync_master_whitelist_set = local_whitelist_set;
+  // remove ANY from set
+  rpl_semi_sync_master_whitelist_set.erase(any_slave);
+  // NOTE: make sure to change the version only after the updating the whitelist
+  ++rpl_semi_sync_master_whitelist_ver;
+
+  LogErr(INFORMATION_LEVEL, ER_SEMISYNC_WHITELIST_UPDATED,
+         rpl_semi_sync_master_whitelist, wlist.c_str());
+
+  rpl_semi_sync_master_whitelist_is_any = !strcmp(wlist.c_str(), any_slave);
```
hermanlee: I think technically, this cached value of the whitelist should be updated before the version variable is bumped up. Otherwise, it's possible the slave would skip the any_slave check in verify_against_whitelist(), see the new version number, check the list (which is "ANY"), but not find its UUID in there, and be denied access. Though this problem would be resolved on the next connect.

plugin/semisync/semisync_master_plugin.cc
```diff
@@ -165,6 +166,11 @@ static int repl_semi_before_send_event(Binlog_transmit_param *param,
                                        const char *log_file, my_off_t log_pos) {
   if (!is_semi_sync_dump()) return 0;
 
+  if (param->server_id > 0 &&
+      !repl_semisync->verify_against_whitelist(param->server_id)) {
```
hermanlee: This check wasn't in the original patch. Is it necessary to check against the whitelist prior to sending? It seems we check on both send and receive then for the common use case where the semi-sync slave is in the whitelist.

plugin/semisync/semisync_master.cc
inikep: Both `verify_against_whitelist` and `update_whitelist` are protected with 
```
std::lock_guard<std::mutex> guard(rpl_semi_sync_master_whitelist_set_lock)
```

hermanlee: Sorry, I missed that the guard was moved outside the if check in verify_against_whitelist(). The original code used the version comparison as an optimization to avoid having to grab the lock on every verify_against_whitelist() call though.  I think it would be better to follow the original implementation for this.

plugin/semisync/semisync_master_plugin.cc
inikep: This check is not necessary, but I wanted to stop binlog sender before calling `updateSyncHeader()`  in `repl_semi_before_send_event()`.

plugin/semisync/semisync_master.cc
inikep: Using `current_thd` (and `current_thd->semisync_whitelist_ver`) is not possible from the AckReceiver thread therefore I moved `whitelist_ver` to  a new `semisync_info_t` struct which should be protected with `rpl_semi_sync_master_whitelist_set_lock`.

inikep:
PR was updated: using server_uuid instead of server_id in the semi-sync plugin

plugin/semisync/semisync_master_plugin.cc
inikep: It's removed now.

plugin/semisync/semisync_master.cc
hermanlee: Ah, okay. Hm... comparing UUID would be more expensive than server_id then for access to the map.

Looking at this code, can we get rid of rpl_semi_sync_master_whitelist_set_lock if update_whitelist() also called lock() to acquire the LOCK_binlog_? It already looks like all of the other calls to update/modify the white list already acquire this lock, so we could use it to protect the set.

I think the whitelist set could be part of AckReceiver. We could add a verified field to struct Slave that caches the check. Whenever the whitelist is changed, go through the list of m_slaves to check the UUIDs. Whenever AckReceiver::add_slave() is called, compare it with the whitelist set. If the whitelist set is only accessed with the m_lock mutex in AckReceiver, we can remove the whitelist_set mutex.

This might be a simpler solution?

inikep:
PR was updated:
1. I moved most of code to the `AckReciver` class
2. `verify_against_whitelist()` is now called only for `AckReciver::add_slave` and `update_whitelist`.
3. `semisync_whitelist_ver` and `rpl_semi_sync_master_whitelist_set_lock` were no longer needed.

plugin/semisync/semisync_master.h
```diff
@@ -387,21 +388,23 @@ class ActiveTranx : public Trace {
...
   char binlog_name[FN_REFLEN];
   unsigned long long binlog_pos = 0;
 
   AckInfo() { clear(); }
 
   void clear() { binlog_name[0] = '\0'; }
   bool empty() const { return binlog_name[0] == '\0'; }
-  bool is_server(int server_id) const { return this->server_id == server_id; }
+  bool is_server(std::string slave_uuid) const {
+    return this->slave_uuid == slave_uuid;
```
hermanlee: I realize I'm waffling here, but with this change to use the ack receiver, we can use server_id. The whitelist on the ack_receiver verifies the UUID of a server and only lets it be added after verification. At that point, we can use whatever server_id is presented by the client, since we would never get acks from non-approved servers that may have conflicting server_id.

It'll also be cheaper to compare server_id than uuid. Thoughts? I can revert these changes back before pushing if it's too much effort to do it here.

mysql-test/suite/sys_vars/t/rpl_semi_sync_master_whitelist_basic.test
```diff
@@ -0,0 +1,104 @@
+--source include/count_sessions.inc
+
+SET @start_global_value = @@GLOBAL.rpl_semi_sync_master_whitelist;
+
+--let $assert_text= rpl_semi_sync_master_whitelist should be "ANY"
+--let $assert_cond= @@GLOBAL.rpl_semi_sync_master_whitelist = "ANY"
+--source include/assert.inc
+
+SET @@GLOBAL.rpl_semi_sync_master_whitelist = NULL;
```
hermanlee, line 9: I'm seeing a crash in my test build:

Thread 46 "mysqld" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7ffff5af8700 (LWP 3858613)]
__strcmp_sse2_unaligned () at ../sysdeps/x86_64/multiarch/strcmp-sse2-unaligned.S:31
31	../sysdeps/x86_64/multiarch/strcmp-sse2-unaligned.S: No such file or directory.
(gdb) bt
#0  __strcmp_sse2_unaligned () at ../sysdeps/x86_64/multiarch/strcmp-sse2-unaligned.S:31
#1  0x00007ffff7fa4631 in Ack_receiver::disconnect_non_whitelisted_slaves (this=0x7ffff569a620) at /home/herman/rocks-mysql/8.0/plugin/semisync/semisync_master_ack_receiver.cc:220
#2  0x00007ffff7fa28c2 in update_whitelist (var_ptr=0x7ffff7fb5f10 <rpl_semi_sync_master_whitelist>, save=0x7ffff5af0228) at /home/herman/rocks-mysql/8.0/plugin/semisync/semisync_master_plugin.cc:239
#3  0x000000000299997f in plugin_var_memalloc_global_update (thd=0x7fffbf207000, var=0x7ffff7fb5e08 <mysql_sysvar_whitelist>, dest=0x7ffff7fb5f10 <rpl_semi_sync_master_whitelist>, value=0x0)
    at /home/herman/rocks-mysql/8.0/sql/sql_plugin_var.cc:74
#4  0x000000000299ac91 in sys_var_pluginvar::global_update (this=0x7fffea561f20, thd=0x7fffbf207000, var=0x7fffbf255c00) at /home/herman/rocks-mysql/8.0/sql/sql_plugin_var.cc:412
(gdb) f 1
#1  0x00007ffff7fa4631 in Ack_receiver::disconnect_non_whitelisted_slaves (this=0x7ffff569a620) at /home/herman/rocks-mysql/8.0/plugin/semisync/semisync_master_ack_receiver.cc:220
220	  if (strcmp(rpl_semi_sync_master_whitelist, any_slave) == 0)
(gdb) p rpl_semi_sync_master_whitelist
$6 = 0x0

Adding `rpl_semi_sync_master_whitelist` as part of the strcmp() check seems to work.

plugin/semisync/semisync_master.cc
```diff
@@ -554,8 +558,24 @@ bool ReplSemiSyncMaster::is_semi_sync_slave() {
...
+  const auto it = thd->user_vars.find("slave_uuid");
+  if (it != thd->user_vars.end() && it->second->length() > 0) {
+    result = std::string(it->second->ptr(), it->second->length());
+  }
+
+  mysql_mutex_unlock(&thd->LOCK_thd_data);
+  return result;
+}
+
+int ReplSemiSyncMaster::reportReplyBinlog(const char *log_file_name,
```
hermanlee: The void -> int change can now be reverted?

plugin/semisync/semisync_master.h
```diff
@@ -439,7 +443,7 @@ class AckContainer : public Trace {
     if (m_ack_array) {
       for (unsigned i = 0; i < m_size; ++i) {
         m_ack_array[i].clear();
-        m_ack_array[i].server_id = 0;
+        m_ack_array[i].slave_uuid = "";
```
hermanlee: I'm also seeing a crash here. I think it's because m_ack_array is allocated through my_malloc(), and the constructors are never called on it? Though if we keep server_id for this structure, then these changes and issues go away.

inikep:
PR was updated.
1. Crashes are fixed
2. `slave_uuid` is used only in `AckReciver::add_slave` and cached in `struct Slave` for usage in `update_whitelist`.

plugin/semisync/semisync_master_ack_receiver.cc
```diff
@@ -131,13 +133,17 @@ void Ack_receiver::stop() {
...
 }
 
-bool Ack_receiver::add_slave(THD *thd) {
+bool Ack_receiver::add_slave(THD *thd, uint32 server_id,
+                             std::string slave_uuid) {
   Slave slave;
   const char *kWho = "Ack_receiver::add_slave";
   function_enter(kWho);
 
+  if (!rpl_semi_sync_master_whitelist) return function_exit(kWho, true);
```
hermanlee: I think this needs to be checked under m_lock and update_whitelist() needs to grab the m_lock when setting the rpl_semi_sync_master_whitelist variable? Otherwise, how do you guarantee rpl_semi_sync_master_whitelist doesn't get set to nullptr after this check and before the strcmp()?

hermanlee (commented):
Overall, the changes look good, but I think there are still some races with checking the whitelist variable?

inikep:
PR was updated.
All occurrences of `rpl_semi_sync_master_whitelist` are protected with `Ack_receiver::m_mutex`.

plugin/semisync/semisync_master_ack_receiver.cc
inikep: You're right. I protected all occurences of `rpl_semi_sync_master_whitelist_set` but forgot about `rpl_semi_sync_master_whitelist`.

plugin/semisync/semisync_master_ack_receiver.cc
```diff
@@ -203,6 +214,91 @@ void Ack_receiver::remove_slave(THD *thd) {
...
+
+  if (rpl_semi_sync_master_whitelist &&
+      strcmp(rpl_semi_sync_master_whitelist, any_slave) == 0) {
+    return function_exit(kWho);
+  }
+
+  for (Slave_vector_it it = m_slaves.begin(); it != m_slaves.end(); ++it) {
+    if (!rpl_semi_sync_master_whitelist ||
+        !verify_against_whitelist(it->slave_uuid))
+      vio_shutdown(it->vio);
```
hermanlee, line 231: Since this is happening on the connection executing the SET statement, it seems like this is not thread safe. Similar calls to vio_shutdown() in sql_class.cc THD::shutdown_active_vio asserts the thd lock is held. Do we need to stash the thd pointer so that the thread can be killed vs. calling vio_shutdown() directly?

inikep: It seems that:
1. Only `THD::active_vio` field is protected by `LOCK_thd_data` (look at `THD::set_active_vio`)
2. `Ack_receiver::run` uses `net.vio` (e.g. `my_net_read(&net)`) without any additional mutex
3. `struct Vio` can be thread-safe with `std::atomic_flag poll_shutdown_flag`

hermanlee: This is probably fine then. The read code doesn't acquire mutex, so killing the thread probably shuts down the vio in a similar way. I wasn't sure if the perf schema access code prior to poll_shutdown_flag is thread-safe, but looking through it, it seems like it should be. In any case, it's probably good enough.

PR #1041, opened 2019-08-26
Issue790: MultiGet-MRR implementation

spetrunia:
First implementation.
Supports scans on full primary key
Cost-based choice whether to use MRR/BKA is not supported (set mrr=on,mrr_cost_based=off)

### DIFF
--- a/mysql-test/suite/rocksdb/t/rocksdb_mrr.test
+++ b/mysql-test/suite/rocksdb/t/rocksdb_mrr.test
@@ -1,8 +1,7 @@
+#
+#  Test for MyRocks' MRR implementation based on MultiGet.
+#
 --source include/have_rocksdb.inc
-
---disable_warnings
-drop table if exists t0,t1,t2,t3,t4,t5,t6,t7,t8,t9,t10,t11;
---enable_warnings
 
 create table t0(a int);
 insert into t0 values (0),(1),(2),(3),(4),(5),(6),(7),(8),(9);
@@ -172,8 +171,89 @@ having
 
 drop table t11,t12;
 
---echo # Same as above but for MRR (not BKA):
---echo # This will not use MRR:
+--echo #
+--echo # Check the userstat counters
+--echo #
+show variables like '%user_table_stat%';
+set optimizer_switch='mrr=off';
+flush statistics;
+
+select * from t3,t0 where t3.col1=t0.a;
+
+--echo # COUNTERS FOR: mrr=OFF, secondary index lookups
+query_vertical
+select
+  rows_inserted, rows_updated, rows_deleted, rows_read, rows_requested
+from
+  information_schema.table_statistics
+where table_name = 't3';
+
+set optimizer_switch='mrr=on';
+flush statistics;
+
+select * from t3,t0 where t3.col1=t0.a;
+
+--echo # COUNTERS FOR: mrr=ON, secondary index lookups
+query_vertical
+select
+  rows_inserted, rows_updated, rows_deleted, rows_read, rows_requested
+from
+  information_schema.table_statistics
+where table_name = 't3';
+
+set optimizer_switch='mrr=off';
+flush statistics;
+explain
+select * from t3,t0 where t3.pk1=t0.a and t3.pk2=t0.a;
+select * from t3,t0 where t3.pk1=t0.a and t3.pk2=t0.a;
+
+--echo # COUNTERS FOR: mrr=OFF, primary index lookups
+query_vertical
+select
+  rows_inserted, rows_updated, rows_deleted, rows_read, rows_requested
+from
+  information_schema.table_statistics
+where table_name = 't3';
+
+set optimizer_switch='mrr=on';
+flush statistics;
+explain
+select * from t3,t0 where t3.pk1=t0.a and t3.pk2=t0.a;
+select * from t3,t0 where t3.pk1=t0.a and t3.pk2=t0.a;
+
+--echo # COUNTERS FOR: mrr=ON, primary index lookups
+query_vertical
+select
+  rows_inserted, rows_updated, rows_deleted, rows_read, rows_requested
+from
+  information_schema.table_statistics
+where table_name = 't3';
+
+--echo #
+--echo # Check how the counters are incremented when SQL
+--echo # layer doesn't read all of the MultiGet results
+--echo #
+create table t20 (a int);
+insert into t20 values (1);
+set global rocksdb_force_flush_memtable_now=1;
+
+explain
+select a, a+20 in (select t2.filler from t2,t0 where t2.pk=t0.a+20) from t20;
+
+create temporary table t11 as
+select * from information_schema.global_status where variable_name like 'rocksdb_num%';
+flush statistics;
+select a, a+20 in (select t2.filler from t2,t0 where t2.pk=t0.a+20) from t20;
+--echo # COUNTERS FOR: mrr=ON, primary index lookups
+query_vertical
+select rows_read, rows_requested
+from information_schema.table_statistics
+where table_name = 't2';
+
+drop table t20;
+--echo #
+--echo # Check how MRR works without BKA
+--echo #
 explain select t3.col1 from t3 where t3.col1=20 or t3.col1 between 25 and 28;
 
 --echo # This will use MRR:
--- a/mysql-test/suite/rocksdb/t/rocksdb_mrr.test
+++ b/mysql-test/suite/rocksdb/t/rocksdb_mrr.test
@@ -26,8 +26,9 @@ select * from t2,t0 where t2.pk=t0.a;
 select * from t2,t0 where t2.pk=t0.a;
 
 --echo # Check the counters
-create table t10 as 
-  select * from information_schema.global_status where variable_name like 'rocksdb%multiget%';
+create temporary table t10 as 
+select * from information_schema.global_status
+where variable_name like 'rocksdb_num%' or variable_name='rocksdb_rows_read';
 
 flush status; # for Handler_mrr_init
 
@@ -35,15 +36,17 @@ select * from t2,t0 where t2.pk=t0.a;
 
 show status like 'Handler_mrr_init';
 
-create table t11 as 
-  select * from information_schema.global_status where variable_name like 'rocksdb%multiget%';
+create temporary table t11 as 
+select * from information_schema.global_status
+where variable_name like 'rocksdb_num%' or variable_name='rocksdb_rows_read';
 
 select 
   variable_name, 
   t11.variable_value - t10.variable_value as DIFF 
 from 
-  t10 join t11 using (VARIABLE_NAME);
-
+  t10 join t11 using (VARIABLE_NAME)
+having
+  DIFF>0;
 drop table t10, t11;
 
 --echo # This will use MRR:
@@ -154,12 +157,14 @@ select * from t3,t0 where t3.col1=t0.a;
 
 --echo # Now, run the query and check the counters
 create temporary table t11 as
-select * from information_schema.global_status where variable_name like 'rocksdb_num%';
+select * from information_schema.global_status 
+where variable_name like 'rocksdb_num%' or variable_name='rocksdb_rows_read';
 
 select * from t3,t0 where t3.col1=t0.a;
 
 create temporary table t12 as
-select * from information_schema.global_status where variable_name like 'rocksdb_num%';
+select * from information_schema.global_status 
+where variable_name like 'rocksdb_num%' or variable_name='rocksdb_rows_read';
 
 select 
   variable_name, 
@@ -239,10 +244,32 @@ set global rocksdb_force_flush_memtable_now=1;
 explain
 select a, a+20 in (select t2.filler from t2,t0 where t2.pk=t0.a+20) from t20;
 
-create temporary table t11 as
-select * from information_schema.global_status where variable_name like 'rocksdb_num%';
+create temporary table t10 as 
+select * from information_schema.global_status
+where variable_name like 'rocksdb_num%' or variable_name='rocksdb_rows_read';
 flush statistics;
+
 select a, a+20 in (select t2.filler from t2,t0 where t2.pk=t0.a+20) from t20;
+
+create temporary table t11 as 
+select * from information_schema.global_status
+where variable_name like 'rocksdb_num%' or variable_name='rocksdb_rows_read';
+
+--echo # The below shows ROCKSDB_ROWS_READ=13. 
+--echo #  1 is from table t20
+--echo # 10 are from table t0, BKA reads all its rows into buffer
+--echo #  2 are from table t2.  BKA code reads two rows before it figures that
+--echo #    the subquery has a match and no more rows are needed.
+--echo #    (note that MultiGet call has read all 10 rows, but SQL layer only has read 2)
+select 
+  variable_name, 
+  t11.variable_value - t10.variable_value as DIFF 
+from 
+  t10 join t11 using (VARIABLE_NAME)
+having
+  DIFF>0;
+drop table t10, t11;
+
 --echo # COUNTERS FOR: mrr=ON, primary index lookups
 query_vertical
 select rows_read, rows_requested
--- a/mysql-test/suite/rocksdb/t/rocksdb_mrr.test
+++ b/mysql-test/suite/rocksdb/t/rocksdb_mrr.test
@@ -154,4 +154,60 @@ select * from t3,t0 where t3.col1=t0.a;
 select * from t3,t0 where t3.col1=t0.a;
 
 drop table t0,t1,t2,t3,t4;
+
+--echo #
+--echo # Multi-keypart testcase
+--echo #
+create table t0(a int);
+insert into t0 values (0),(1),(2),(3),(4),(5),(6),(7),(8),(9);
+
+CREATE TABLE t2 (
+  id bigint(20) unsigned NOT NULL DEFAULT '0',
+  type tinyint(3) unsigned NOT NULL DEFAULT '0',
+  c1 int(10) unsigned NOT NULL DEFAULT '0',
+  c2 int(10) unsigned NOT NULL DEFAULT '0',
+  time_updated int(10) unsigned NOT NULL DEFAULT '0',
+  id2 bigint(20) unsigned NOT NULL DEFAULT '0',
+  c3 int(10) unsigned NOT NULL DEFAULT '0',
+  c_id bigint(20) unsigned NOT NULL DEFAULT '0',
+  a_id bigint(20) unsigned NOT NULL DEFAULT '0',
+  PRIMARY KEY (id,`type`),
+  KEY time_updated (time_updated),
+  KEY id2 (id2),
+  KEY a_idx (a_id),
+  KEY c_a_idx (c_id,`a_id`)
+) ENGINE=ROCKSDB DEFAULT CHARSET=latin1;
+
+
+insert into t2 (id, type, id2, time_updated)
+ select a, 1, 1, 1565888298 
+from t0;
+
+insert into t2 (id, type, id2, time_updated)
+ select a, 2, 2, 1565888298 
+from t0;
+
+set optimizer_trace=1;
+--echo # The following should have type=range, and "Using MRR":
+--replace_column 9 #
+explain
+SELECT 
+  id2, type, SUM(c1), SUM(c2) 
+FROM t2 force index (id2) 
+WHERE 
+  id2 IN (1) AND 
+  time_updated > 1565888297 AND 
+  time_updated <= UNIX_TIMESTAMP() 
+GROUP BY id2, type;
+
+select 
+  MID(trace, locate('attached_conditions_computation', trace), 360) 
+    as TRACE_FRAGMENT
+from
+  information_schema.optimizer_trace;
+
+set optimizer_trace=0;
+
+drop table t0, t2;
+
 set optimizer_switch=@save_optimizer_switch;
--- a/mysql-test/suite/rocksdb/t/rocksdb_mrr.test
+++ b/mysql-test/suite/rocksdb/t/rocksdb_mrr.test
@@ -1,0 +1,55 @@
+--source include/have_rocksdb.inc
+
+--disable_warnings
+drop table if exists t0,t1,t2,t3,t4,t5,t6,t7,t8,t9,t10;
+--enable_warnings
+
+create table t0(a int);
+insert into t0 values (0),(1),(2),(3),(4),(5),(6),(7),(8),(9);
+
+create table t1(a int);
+insert into t1 select A.a + B.a* 10 + C.a * 100 from t0 A, t0 B, t0 C;
+
+create table t2 (
+  pk int primary key,
+  col1 int,
+  filler char(32)
+) engine=rocksdb;
+
+insert into t2 select a,a,a from t1;
+set global rocksdb_force_flush_memtable_now=1;
+
+set @save_optimizer_switch=@@optimizer_switch;
+set optimizer_switch='mrr_cost_based=off,batched_key_access=on';
+
+explain 
+select * from t2,t0 where t2.pk=t0.a;
+select * from t2,t0 where t2.pk=t0.a;
+
+create table t3 (
+  pk1 int,
+  pk2 int,
+  col1 int,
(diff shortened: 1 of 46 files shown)


storage/rocksdb/ha_rocksdb.cc
```diff
@@ -15076,6 +15118,373 @@ void rdb_tx_multi_get(Rdb_transaction *tx,
...
+  if (res == HA_POS_ERROR || m_lock_rows != RDB_LOCK_NONE || !mrr_enabled)
+    return res;
+
+  if (keyno == table->s->primary_key) {
+    bool all_eq_ranges = true;
+    KEY_MULTI_RANGE range;
+    range_seq_t seq_it;
+    seq_it = seq->init(seq_init_param, n_ranges, *flags);
+    while (!seq->next(seq_it, &range)) {
+      if (!(range.range_flag & UNIQUE_RANGE)) {
```
yoshinorim, line 15148: Could you check if thd is marked as killed or not in the loop, and returning errors if killed, like this?

```
      if (thd && thd->killed) {
        return HA_ERR_QUERY_INTERRUPTED;
      }
```

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -15076,6 +15118,373 @@ void rdb_tx_multi_get(Rdb_transaction *tx,
...
+  buf += sizeof(rocksdb::PinnableSlice) * n_elements;
+
+  mrr_range_ptrs = (char **)buf;
+  buf += sizeof(char *) * n_elements;
+
+  ssize_t elem = -1;
+  int key_size;
+  char *range_ptr;
+  while ((key_size = mrr_rowid_reader->get_next_rowid((uchar*)buf, &range_ptr)) > 0 ) {
+
```
yoshinorim: thd->killed check, same as above

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -15076,6 +15118,373 @@ void rdb_tx_multi_get(Rdb_transaction *tx,
...
+  mrr_values = nullptr;
+}
+
+int ha_rocksdb::multi_range_read_next(char **range_info) {
+  if (mrr_uses_default_impl) {
+    return handler::multi_range_read_next(range_info);
+  }
+
+  while (1) {
+    if (mrr_read_index >= mrr_n_elements) {
```
yoshinorim: thd->killed check, same as above

storage/rocksdb/ha_rocksdb.cc
spetrunia: Done

spetrunia:
Ok,  now the code is ready for the review.

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -15076,6 +15114,467 @@ void rdb_tx_multi_get(Rdb_transaction *tx,
...
+    while (!seq->next(seq_it, &range)) {
+      if (!(range.range_flag & UNIQUE_RANGE)) {
+        all_eq_ranges = false;
+        break;
+      }
+      if (table->in_use->killed) return HA_POS_ERROR;
+    }
+
+    if (all_eq_ranges) {
+      // Indicate that we will use Mutlit-Get MRR
```
lth: "Multi-Get" or "MultiGet"?

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -15076,6 +15114,467 @@ void rdb_tx_multi_get(Rdb_transaction *tx,
...
+      *flags &= ~HA_MRR_USE_DEFAULT_IMPL;
+      *flags |= HA_MRR_CONVERT_REF_TO_RANGE;
```
lth: Does the `HA_MRR_CONVERT_REF_TO_RANGE` flag make sense for primary keys? It's not like we're going to use MRR with ref plans on primary key anyway?

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -15076,6 +15114,467 @@ void rdb_tx_multi_get(Rdb_transaction *tx,
...
+
+        if (table->in_use->killed) return HA_ERR_QUERY_INTERRUPTED;
+
+        if ((rc = mrr_fill_buffer())) {
+          if (rc == HA_ERR_END_OF_FILE) table->status = STATUS_NOT_FOUND;
+          return rc;
+        }
+      }
+      // Skip the "is not found" errors
+      if (mrr_statuses[mrr_read_index].ok()) break;
```
lth, line 15557: Perhaps we should check explicitly for "not found" vs other errors here.

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -15076,6 +15114,467 @@ void rdb_tx_multi_get(Rdb_transaction *tx,
...
+    if (m_pk_descr->has_ttl() &&
+        should_hide_ttl_rec(*m_pk_descr, m_retrieved_record,
+                            tx->m_snapshot_timestamp)) {
+      m_retrieved_record.Reset();
+      mrr_values[cur_key].Reset();
+      continue;
+    }
+
+    rc = convert_record_from_storage_format(&rowkey, table->record[0]);
+    m_retrieved_record.Reset();
```
lth: Is this safe to do for TEXT/BLOBs? From what I remember, we cannot call Reset right away because it may be pointing to blob data.

If you look at the existing usage of PinnableSlice, we just call Reset right before we issue a Get:
https://github.com/facebook/mysql-5.6/blob/7bdd211c8eb75740dc881289ef3b3514235f432c/storage/rocksdb/ha_rocksdb.cc#L3466

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -3781,6 +3791,24 @@ class Rdb_writebatch_impl : public Rdb_transaction {
     return get(column_family, key, value);
   }
 
+  void multi_get(const rocksdb::ReadOptions &read_options,
+                 rocksdb::ColumnFamilyHandle *column_family,
+                 const size_t num_keys, const rocksdb::Slice *keys,
+                 rocksdb::PinnableSlice *values, rocksdb::Status *statuses,
+                 bool sorted_input) override {
+    // todo: could we just read the committed content from the DB here?
```
lth: What's wrong with just calling `MultiGetFromBatchAndDB` like below? (That said, I don't think this code path will be exercised).

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -7991,8 +8022,6 @@ int ha_rocksdb::read_row_from_secondary_key(uchar *const buf,
   int rc = 0;
   uint pk_size;
 
-  /* Get the key columns and primary key value */
-  const rocksdb::Slice &rkey = m_scan_it->key();
   const rocksdb::Slice &value = m_scan_it->value();
```
lth: `covered_lookup` uses  `kd.covers_lookup` which determines results on a per-value basis. Currently, it looks like we only calculate `covered_lookup` once.

That said, the covered key feature (https://github.com/facebook/mysql-5.6/pull/639) is currently off, but it's probably easier/safer just to fix this.

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -15076,6 +15114,467 @@ void rdb_tx_multi_get(Rdb_transaction *tx,
...
+      return HA_ERR_END_OF_FILE; //  At eof already
+
+    KEY_MULTI_RANGE range;
+    if ((mrr_ranges_eof = self->mrr_funcs.next(mrr_seq_it, &range)))
+      return HA_ERR_END_OF_FILE; //  Got eof now
+
+    key_part_map all_parts_map =
+        (key_part_map(1) << self->m_pk_descr->get_key_parts()) - 1;
+    DBUG_ASSERT(range.start_key.keypart_map == all_parts_map);
+    DBUG_ASSERT(range.end_key.keypart_map == all_parts_map);
```
lth, line 15244: Might as well assert on flags as well?

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -15076,6 +15114,467 @@ void rdb_tx_multi_get(Rdb_transaction *tx,
...
+                                SQL layer which lookup key the returned record
+                                matches with (**)
+    {PK lookup value}[N]      - The rowid (Primary Key) to lookup. The
+                                corresponding rocksdb::Slice object points to
+                                this key.
+
+  (*) The memory for rows is allocated somewhere inside RocksDB, there's no
+      way to make it use the user-supplied buffer.
+  (**) The engine could specify HA_MRR_NO_ASSOCIATION which would mean "we
+      cannot tell which key the returned records match" we don't do this.
```
lth, line 15397: It seems like we do have some sort of handling below for `HA_MRR_NO_ASSOCIATION`, and this flag gets passed in `get_quick_select_for_ref` which seems to be used only for filesorts? (I couldn't tell which plans called `get_quick_select_for_ref` though)

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -15076,6 +15114,467 @@ void rdb_tx_multi_get(Rdb_transaction *tx,
...
+// secondary index and returning rowids from the index records
+//
+class Mrr_sec_key_rowid_source : public Mrr_rowid_source {
+  ha_rocksdb *self;
+  int err;
+ public:
+  Mrr_sec_key_rowid_source(ha_rocksdb *self_arg) : self(self_arg), err(0) {
+  }
+
+  int init(RANGE_SEQ_IF *seq, void *seq_init_param,
```
lth, line 15271: Do we have to handle the callbacks passed into `RANGE_SEQ_IF`, like `skip_index_tuple` or `skip_record`.

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -15076,6 +15114,467 @@ void rdb_tx_multi_get(Rdb_transaction *tx,
...
+  if (keyno == table->s->primary_key) {
+    // We need all ranges to be single-point lookups using full PK values.
+    // (Range scans, like "pk BETWEEN 10 and 20" or restrictions on PK prefix
+    //  cannot be used)
+    bool all_eq_ranges = true;
+    KEY_MULTI_RANGE range;
+    range_seq_t seq_it;
+    seq_it = seq->init(seq_init_param, n_ranges, *flags);
+    while (!seq->next(seq_it, &range)) {
+      if (!(range.range_flag & UNIQUE_RANGE)) {
```
lth, line 15148: Maybe add an assert that it is EQ_RANGE too?

lth (commented):
It's kind of strange how the same sql level switch will give different behaviour on innodb vs rocksdb tables. I wonder if we want to add an extra rocksdb switch for people running with two engines. At least for us, we mainly run with single engine so it's not a big deal.

storage/rocksdb/ha_rocksdb.cc
spetrunia: Fixed

storage/rocksdb/ha_rocksdb.cc
spetrunia: You're right, this flag won't make any difference here. Lookups on full constant PK will use eq_ref, not ref...

storage/rocksdb/ha_rocksdb.cc
spetrunia: Added asserts.

storage/rocksdb/ha_rocksdb.cc
spetrunia: In my opinion, it would be redundant. Looking at the definition of EQ_RANGE and UNIQUE_RANGE in include/my_base.h, one can see that UNIQUE_RANGE is a stronger requirement than EQ_RANGE.
The code that sets these flags in opt_range.cc is fairly straightforward so I don't think there's any risk of getting a UNIQUE_RANGE that's not a EQ_RANGE.

storage/rocksdb/ha_rocksdb.cc
spetrunia: Nothing is wrong, and I think we can do with just one `multi_get` function now (since the bypass' function switched to using the same function signature.

storage/rocksdb/ha_rocksdb.cc
spetrunia: Agree. Changed accordingly.

storage/rocksdb/ha_rocksdb.cc
spetrunia: Yes for `skip_index_tuple`, no for `skip_record` but I've added handling for both  ( notes taken during investigation are here: https://github.com/facebook/mysql-5.6/issues/790#issuecomment-541188976)

storage/rocksdb/ha_rocksdb.cc
spetrunia: `get_quick_select_for_ref` is called when one has
- ref(const) access on the first table 
- "Using filesort" (w/o "Using temporary).

In this case, the query plan is to get filesort() to read the first table and
sort it according to the sort criteria.  filesort is unable to use ref access
as the input stream, it can either read from a quick select or do a full table
scan. Because of that, get_quick_select_for_ref is used.

spetrunia: The code here doesn't really handle `HA_MRR_NO_ASSOCIATION`. We just increment the counter, in the same way like DsMrr_impl::dsmrr_init  does:
```
  is_mrr_assoc= !MY_TEST(mode & HA_MRR_NO_ASSOCIATION);
  if (is_mrr_assoc)
    status_var_increment(table->in_use->status_var.ha_multi_range_read_init_count);
```
This allows to use the counter value as a counter of how many times BKA was used.

Some background:
Initially, HA_MRR_NO_ASSOCIATION was introduced for MRR/NDB . NDB returns records but it cannot tell which lookup key each record is for. 

Now, HA_MRR_NO_ASSOCATION is used as follows:
* BKA code does need the storage engine to tell which lookup key the record is for. It doesn't pass HA_MRR_NO_ASSOCATION. 
* Range access doesn't need this, it just needs the record. It passes HA_MRR_NO_ASSOCATION flag.

So, for an MRR implementation, HA_MRR_NO_ASSOCATION means that it doesn't need to keep track which result record matches which range. For  MyRocks/MRR, we won't need the `mrr_range_ptrs` array.  This allows to save a bit of buffer space (8 bytes per range) ...  but you can only do that when you are using range access, and typically don't have many ranges.

Should I implement HA_MRR_NO_ASSOCATION still?

storage/rocksdb/ha_rocksdb.cc
spetrunia: Yes, fixed.

spetrunia:
@lth , I have addressed (or replied to) all the input.  The patch is ready for another review pass.

storage/rocksdb/ha_rocksdb.cc
hermanlee: It doesn't seem like supporting HA_MRR_NO_ASSOCIATION is really needed? It seems to be passed in mainly as an optimization to save on the space? It looks like the bka_unique_range_seq_skip_record callback requires !NO_ASSOCIATION and that appears to be used mainly for filtering. 

I think we don't need to implement the HA_MRR_NO_ASSOCIATION feature buffer unless saving 8 bytes per range is really critical.

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -15076,6 +15089,497 @@ void rdb_tx_multi_get(Rdb_transaction *tx,
...
+  while (1) {
+    while (1) {
+      if (mrr_read_index >= mrr_n_elements) {
+        if (mrr_rowid_reader->eof() || !mrr_n_elements) {
+          table->status = STATUS_NOT_FOUND;  // not sure if this is necessary?
+          mrr_free_rows();
+          return HA_ERR_END_OF_FILE;
+        }
+
+        if (table->in_use->killed) return HA_ERR_QUERY_INTERRUPTED;
```
hermanlee: This should be checked on each row being processed, but it's a minor point.

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -10560,6 +10569,8 @@ int ha_rocksdb::index_end() {
   active_index = MAX_KEY;
   in_range_check_pushed_down = FALSE;
 
+  mrr_free();
```
hermanlee: Minor point: can this be conditioned on mrr_uses_default_impl? Otherwise, it seems like every index_end() call always goes through mrr_free(), even though mrr is never invoked. It might be safer to always call mrr_free(), but it seems unnecessary most of the time.

storage/rocksdb/ha_rocksdb.cc
spetrunia: As far as I understand, there is a check for `thd->killed` for each row processed (actually, there are several). 
My reasons for putting the check into this loop were:
* the loop iterates through the arrays returned by MultiGet(), so it enumerates O(#rows_read). 
* Some subtle bug could cause execution to get stuck in the loop (after all, there are several "continue;" statements in this loop).

Should this check be kept or removed?

storage/rocksdb/ha_rocksdb.cc
spetrunia: Yes. I think it's better to condition on `mrr_rowid_reader`  (which is always present when using a non-default MRR implementation). Done.

storage/rocksdb/ha_rocksdb.cc
hermanlee: We should keep the check, but I was thinking it should be moved outside of the 'if (mrr_read_index >= mrr_n_elements) conditional. If, for whatever reason, mrr_read_index is never incremented or mrr_n_elements is really large, we can still loop without hitting this check.

Maybe move the check right after the second while (1)?

Also, is it possible when executing query plans for table->in_use to not be the same as ha_rocksdb::ha_thd()?

storage/rocksdb/ha_rocksdb.cc
lth: > This allows to use the counter value as a counter of how many times BKA was used.

Ah okay, this makes sense a lot more sense now. We're just using `HA_MRR_NO_ASSOCATION` to tell us if BKA was used or not (and the Handler_mrr_init counter is meant to only count MRR used for BKA).

I don't think we need to implement `HA_MRR_NO_ASSOCIATION` either. I was just curious why we bumped this counter only sometimes with `HA_MRR_NO_ASSOCIATION`.

storage/rocksdb/ha_rocksdb.cc
lth: Maybe I missed it ... but I don't actually see where this was addressed? (or maybe it wasn't an issue after all?)

spetrunia: Yes... I remember addressing it, but the change is not there... will fix now

spetrunia: Ok, I've added the change. I wasn't able to construct a testcase for this, though - I couldn't find a way to cause RocksDB make the underlying data be released before the SQL layer tries to access it.  The code looks fairly clear, though.

storage/rocksdb/ha_rocksdb.cc
spetrunia: Ok

storage/rocksdb/ha_rocksdb.cc
spetrunia: > is it possible when executing query plans for table->in_use to not be the same as ha_rocksdb::ha_thd()?

As far as I know, no.

mysql-test/suite/rocksdb/t/rocksdb_mrr.test
```diff
@@ -0,0 +1,309 @@
+--source include/have_rocksdb.inc
+
+--disable_warnings
+drop table if exists t0,t1,t2,t3,t4,t5,t6,t7,t8,t9,t10,t11;
+--enable_warnings
```
lth: We don't use the `drop table if exists` pattern anymore because mtr should ensure correct cleanup for us (and also for other tests).

lth (commented):
One minor issue: It looks like we're not updating the rows read counters (both the status var and the counters for user/table stats). Can we update the counter as well?

It's also unfortunate that we'd have to pin our data in block cache until the mrr buffer finishes reading just because of blobs, but this doesn't seem to be an issue for now.

storage/rocksdb/ha_rocksdb.cc
spetrunia: (now the testcase is constructed and pushed)

mysql-test/suite/rocksdb/t/rocksdb_mrr.test
spetrunia: Removed.

lth:
Some other stats may be missing as well (eg. rows_requested)

mysql-test/suite/rocksdb/t/rocksdb_mrr.test
```diff
@@ -0,0 +1,389 @@
...
+  t11 join t12 using (VARIABLE_NAME)
+having
+  DIFF>0;
+
+drop table t11,t12;
+
+--echo #
+--echo # Check the userstat counters
+--echo #
+show variables like '%user_table_stat%';
```
lth: This isn't a useful variable to check. This controls the population of the `information_schema.user_table_statistics` table (not table_statistics). We can probably just remove this.

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -15076,6 +15094,513 @@ void rdb_tx_multi_get(Rdb_transaction *tx,
...
+    }
+
+    m_last_rowkey.copy((const char *)rowkey.data(), rowkey.size(),
+                       &my_charset_bin);
+
+    *range_info = mrr_range_ptrs[cur_key];
+
+    m_retrieved_record.Reset();
+    m_retrieved_record.PinSlice(mrr_values[cur_key], &mrr_values[cur_key]);
+    mrr_values[cur_key].Reset();
```
lth: It looks like this was newly added, but I don't think it actually does anything.

In 
```
    m_retrieved_record.PinSlice(mrr_values[cur_key], &mrr_values[cur_key]);
```

We're calling

```
  inline void PinSlice(const Slice& s, Cleanable* cleanable) {
    assert(!pinned_);
    pinned_ = true;
    data_ = s.data();
    size_ = s.size();
    cleanable->DelegateCleanupsTo(this);
    assert(pinned_);
  }
```

and `DelegateCleanupsTo` will actually set the cleanup functions to nullptr on the origin slice. Effectively, we've already "move"d the Slice away, so `mrr_values[cur_key] ` should be empty already.

lth (commented):
I think we're still missing the call to `update_row_stats` for ROWS_READ. This will update the `rocksdb_rows_read` status var.

mysql-test/suite/rocksdb/t/rocksdb_mrr.test
spetrunia: Removed

storage/rocksdb/ha_rocksdb.cc
spetrunia: Agree. I misread what PinnableSlice is doing. Reverted the change

spetrunia:
Hi @lth,

Thanks for the input so far,

I've followed it and made `rocksdb_rows_read` to show the number of rows the SQL layer has read from MyROcks.

However, I've previously made `I_S.table_statistics.rows_read` to be the number of rows that MyRocks has read from RocksDB:
 https://github.com/spetrunia/mysql-5.6/blob/fb-mysql-5.6.35-issue790/storage/rocksdb/ha_rocksdb.cc#L15518

These two numbers are typically the same, but they can differ if the SQL layer stops the MRR scan before it has read all the rows.

An example of this can be found here:
https://github.com/spetrunia/mysql-5.6/blob/fb-mysql-5.6.35-issue790/mysql-test/suite/rocksdb/r/rocksdb_mrr.result#L385
It shows that for table t2, I_S.table_statistics.rows_read=10 while t2's contribution to ROCKSDB_ROWS_READ is 2.

Is the current state fine, or the counters should be made to agree? If we need them to agree, which one should be changed?

lth:
@spetrunia Thanks, the current implementation is fine.

lth (approved):

PR #1046, opened 2019-09-07
Support Create_time and Update_time in MyRocks table status

spetrunia:
- Create_time is read from frm file, like InnoDB. Prior to this diff, create_time was NULL.
- Update_time is in-memory only (like in InnoDB).

### DIFF
--- a/mysql-test/suite/rocksdb/t/show_table_status.test
+++ b/mysql-test/suite/rocksdb/t/show_table_status.test
@@ -1,5 +1,5 @@
 --source include/have_rocksdb.inc
-
+--source include/have_partition.inc
 # 
 # SHOW TABLE STATUS statement
 #
@@ -24,7 +24,7 @@ set global rocksdb_force_flush_memtable_now = true;
 
 CREATE TABLE t3 (a INT, b CHAR(8), pk INT PRIMARY KEY) ENGINE=rocksdb CHARACTER SET utf8;
 
---replace_column 6 # 7 #
+--replace_column 6 # 7 # 12 # 13 #
 SHOW TABLE STATUS WHERE name IN ( 't1', 't2', 't3' );
 
 # Some statistics don't get updated as quickly.  The Data_length and
@@ -48,7 +48,7 @@ set global rocksdb_force_flush_memtable_now = true;
 
 # We expect the number of rows to be 10000. Data_len and Avg_row_len
 # may vary, depending on built-in compression library.
---replace_column 6 # 7 #
+--replace_column 6 # 7 # 12 # 13 #
 SHOW TABLE STATUS WHERE name LIKE 't2';
 DROP TABLE t1, t2, t3;
 
@@ -62,3 +62,114 @@ CREATE TABLE `t1_new..............................................end`(a int) en
 INSERT INTO `t1_new..............................................end` VALUES (1);
 --query_vertical SELECT TABLE_SCHEMA, TABLE_NAME FROM information_schema.table_statistics WHERE TABLE_NAME = 't1_new..............................................end'
 DROP DATABASE `db_new..............................................end`;
+
+--echo #
+--echo # MDEV-17171: Bug: RocksDB Tables do not have "Creation Date"
+--echo #
+
+use test;
+create table t1 (a int) engine=rocksdb;
+
+select create_time is not null, update_time, check_time 
+from information_schema.tables where table_schema=database() and table_name='t1';
+
+insert into t1 values (1);
+select create_time is not null, update_time is not null, check_time 
+from information_schema.tables where table_schema=database() and table_name='t1';
+
+flush tables;
+select create_time is not null, update_time is not null, check_time 
+from information_schema.tables where table_schema=database() and table_name='t1';
+
+select create_time, update_time into @create_tm, @update_tm
+from information_schema.tables 
+where table_schema=database() and table_name='t1';
+
+select sleep(3);
+insert into t1 values (2);
+
+--vertical_results
+select 
+  create_time=@create_tm /* should not change */ , 
+  timestampdiff(second, @update_tm, update_time) > 2,
+  check_time
+from information_schema.tables 
+where table_schema=database() and table_name='t1';
+
+--echo #
+--echo # Check how create_time survives ALTER TABLE.
+--echo # First, an ALTER TABLE that re-creates the table:
+alter table t1 add b int;
+select sleep(2);
+
+select
+  create_time<>@create_tm /* should change */,
+  create_time IS NOT NULL,
+  update_time IS NULL
+from information_schema.tables 
+where table_schema=database() and table_name='t1';
+
+insert into t1 values (5,5);
+
+select create_time, update_time into @create_tm, @update_tm
+from information_schema.tables 
+where table_schema=database() and table_name='t1';
+
+select sleep(2);
+--echo # Then, an in-place ALTER TABLE:
+alter table t1 add key (a);
+
+--echo # create_time will change as .frm file is rewritten:
+select
+  create_time=@create_tm,
+  update_time
+from information_schema.tables 
+where table_schema=database() and table_name='t1';
+
+--echo # Check TRUNCATE TABLE
+insert into t1 values (10,10);
+select create_time, update_time into @create_tm, @update_tm
+from information_schema.tables 
+where table_schema=database() and table_name='t1';
+
+select sleep(2);
+truncate table t1;
+
+select
+  create_time=@create_tm /* should not change */,
+  update_time
+from information_schema.tables 
+where table_schema=database() and table_name='t1';
+
+
+--echo #
+--echo # Check what is left after server restart
+--echo #
+
+--echo # Save t1's creation time
+create table t2 as
+select create_time
+from information_schema.tables
+where table_schema=database() and table_name='t1';
+
+--source include/restart_mysqld.inc
+
+select sleep(2);
+select
+  create_time=(select create_time from t2)  /* should not change */,
+  update_time
+from information_schema.tables
+where table_schema=database() and table_name='t1';
+
+drop table t1, t2;
+
+--echo #
+--echo # Check how it works for partitioned tables
+--echo #
+create table t1 (pk int primary key) partition by hash(pk) partitions 2;
+insert into t1 values (1);
+
+select create_time IS NOT NULL , update_time IS NOT NULL
+from information_schema.tables 
+where table_schema=database() and table_name='t1';
+drop table t1;
--- a/storage/rocksdb/ha_rocksdb.cc
+++ b/storage/rocksdb/ha_rocksdb.cc
@@ -2464,6 +2464,8 @@ class Rdb_transaction {
   bool m_is_delayed_snapshot = false;
   bool m_is_two_phase = false;
 
+  std::unordered_set<Rdb_tbl_def*> modified_tables;
+
  private:
   /*
     Number of write operations this transaction had when we took the last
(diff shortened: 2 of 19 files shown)


spetrunia:
Addressed the review input. Ready for another review pass.

spetrunia:
Take-aways from the last discussion:

- Put the timestamp into the DDL_ENTRY_INDEX_START_NUMBER.
- This will require increasing the format version for that record. Old servers will not be able to start on a datadir where a new server has run a DDL.
- In order to allow for reversible upgrades, introduce a variable similar to innodb_file_format.

spetrunia:
Suggestions for the variable:

The variable will be named `rocksdb_file_format`, type integer.
It will be a global read-only variable, only settable in my.cnf/server args.

* Current data format is denoted by rocksdb_file_format=1.
* The data format after this patch is rocksdb_file_format=2.
* The default setting after this patch will be `rocksdb_file_format=2`.  (Those willing to do an un-doable upgrade will have to manually set rocksdb_file_format=1 in my.cnf)

spetrunia:
... Implemented the variable, and switched the patch back to the "variant 2" of the on-disk format.  The PR is ready for review.

storage/rocksdb/rdb_datadic.h
```diff
@@ -506,7 +523,10 @@ class Rdb_key_def {
   // Data dictionary schema version. Introduce newer versions
   // if changing schema layout
   enum {
-    DDL_ENTRY_INDEX_VERSION = 1,
+    DDL_ENTRY_INDEX_VERSION_1 = 1,
```
hermanlee: These enums should probably reference ROCKSDB_DATADIC_FORMAT_INITIAL and ROCKSDB_DATADIC_FORMAT_CREATE_TIMESTAMP.

storage/rocksdb/rdb_datadic.h
```diff
@@ -203,6 +203,22 @@ enum {
...
 
+
+/*
+  Global user-visible data dictionary format version.
+  The server will read the data of any version, but will not write data
+  structures that were introduced after the version in
+  rocksdb_table_dictionary_format.
+  This way, one can keep the on-disk data backward-compatible.
+*/
+const uint ROCKSDB_DATADIC_FORMAT_INITIAL = 1;
```
hermanlee: Maybe rename FORMAT_INDEX_ONLY?

storage/rocksdb/rdb_datadic.cc
```diff
@@ -4015,27 +4028,52 @@ bool Rdb_ddl_manager::init(Rdb_dict_manager *const dict_arg,
...
+          Rdb_key_def::DDL_ENTRY_INDEX_VERSION_2, version);
       return true;
     }
+    int real_val_size = val.size() - Rdb_key_def::VERSION_SIZE;
+
+    if (version == Rdb_key_def::DDL_ENTRY_INDEX_VERSION_2) {
+      if (real_val_size < Rdb_key_def::TABLE_CREATE_TIMESTAMP_SIZE) {
+        // NO_LINT_DEBUG
+        sql_print_error( "RocksDB: DDL ENTRY V2 doesn't have timestamp");
+        delete tdef;
```
hermanlee: The other error return paths should probably delete tdef too for correctness, though if rocksdb doesn't load, the server terminates anyway. It might prevent asan leak issues though.

storage/rocksdb/rdb_datadic.cc
```diff
@@ -3514,8 +3514,21 @@ bool Rdb_tbl_def::put_dict(Rdb_dict_manager *const dict,
                            const rocksdb::Slice &key) {
   StringBuffer<8 * Rdb_key_def::PACKED_SIZE> indexes;
   indexes.alloc(Rdb_key_def::VERSION_SIZE +
+                Rdb_key_def::TABLE_CREATE_TIMESTAMP_SIZE +
                 m_key_count * Rdb_key_def::PACKED_SIZE * 2);
-  rdb_netstr_append_uint16(&indexes, Rdb_key_def::DDL_ENTRY_INDEX_VERSION);
+
+  if (rocksdb_table_dictionary_format <
```
hermanlee: It might be better if this was a switch:

```
switch (rocksdb_table_dictionary_format) {
  case ROCKSDB_DATADIC_FORMAT_INITIAL:
    rdb_netstr_append_uint16(&indexes, Rdb_key_def::DDL_ENTRY_INDEX_VERSION_1);
    set_create_time(0);
    break;
  case ROCKSDB_DATADIC_FORMAT_CREATE_TIMESTAMP:
    rdb_netstr_append_uint16(&indexes, Rdb_key_def::DDL_ENTRY_INDEX_VERSION_2);
    rdb_netstr_append_uint64(&indexes, create_time);
    break;
  default:
    DBUG_ASSERT(0);
    rdb_netstr_append_uint16(&indexes, Rdb_key_def::DDL_ENTRY_INDEX_VERSION_1);
    set_create_time(0);
    break;
}
```

This would be easier to extend. At minimum for the else {}, we should assert rocksdb_table_dictionary_format == ROCKSDB_DATADIC_FORMAT_CREATE_TIMESTAMP for this build.

storage/rocksdb/rdb_datadic.h
```diff
@@ -203,6 +203,22 @@ enum {
...
+const uint ROCKSDB_DATADIC_FORMAT_CREATE_TIMESTAMP = 2;
+
+// Maximum possible value:
+const uint ROCKSDB_DATADIC_FORMAT_MAX = 2;
+const uint ROCKSDB_DATADIC_FORMAT_DEFAULT = ROCKSDB_DATADIC_FORMAT_MAX;
```
hermanlee: Can the default be ROCKSDB_DATADIC_FORMAT_INITIAL instead so that MyRocks and PerconaRocks don't automatically change behaviors? Can you commit a separate change to update it to 2 on the MariaDB branch?

mysql-test/suite/rocksdb/t/show_table_status.test
```diff
@@ -62,3 +62,80 @@ CREATE TABLE `t1_new..............................................end`(a int) en
 INSERT INTO `t1_new..............................................end` VALUES (1);
 --query_vertical SELECT TABLE_SCHEMA, TABLE_NAME FROM information_schema.table_statistics WHERE TABLE_NAME = 't1_new..............................................end'
 DROP DATABASE `db_new..............................................end`;
+--echo #
+--echo # MDEV-17171: Bug: RocksDB Tables do not have "Creation Date"
+--echo #
+use test;
+create table t1 (a int) engine=rocksdb;
```
hermanlee, line 71: Add a test for partitioned tables?

hermanlee (commented):
Looks good overall. Just a few minor comments.

mysql-test/suite/rocksdb/t/show_table_status.test
spetrunia: Done

storage/rocksdb/rdb_datadic.cc
spetrunia: Added `delete tdef` to other paths.

storage/rocksdb/rdb_datadic.h
spetrunia: Rephrased.

storage/rocksdb/rdb_datadic.cc
spetrunia: Done

storage/rocksdb/rdb_datadic.h
spetrunia: I don't mind renaming, but I don't understand the reasoning behind the name?

(Before this patch, all data dictionary records had version =1 so I thought _INITIAL was an obvious choice of name...)

spetrunia:
@hermanlee, I have now addressed all input except the question about rename to FORMAT_INDEX_ONLY about which I have a question posted above.

storage/rocksdb/rdb_datadic.h
hermanlee: Initial is fine. I thought index_only would be more descriptive when compared to format_create_timestamp.

storage/rocksdb/rdb_datadic.h
hermanlee: Does rocksdb_table_dictionary_format also control the version of CF_DEFINITION, BINLOG_INFO_INDEX_NUMBER, DDL_DROP_INDEX_ONGOING version?

Or is it mainly to control the DDL_ENTRY? If it's to control DDL_ENTRY, then DDL_ENTRY_INDEX_VERSION_1 can be set directly to ROCKSDB_DATADIC_FORMAT_INITIAL.

hermanlee (approved):
Code looks fine from the functional perspective. It's fine to be checked in.

lth:
I don't see timestamp being preserved in `::truncate` so I assume that we update create time on every truncate. Is that expected behaviour?

At least on InnoDB, it seems like truncate will not update timestamp (but InnoDB just returns filesystem ctime of the .frm file).

https://github.com/facebook/mysql-5.6/blob/9b7f226385031c3237977b9ec2971cb883b6e50a/storage/innobase/handler/ha_innodb.cc#L12535-L12537

I also wonder if just checking the .frm ctime is a simpler diff since we won't have to change the data dictionary format. I understand it's not perfect though (eg. other things could change file metadata, or maybe fs does not provide it).

george-lorch:
@lth checking .frm file might be OK here in 5.6, but is not portable to 8.0, so I would caution against it.

hermanlee:
@georgelorchpercona I think 8.0 intends to store this in the new data dictionary though. There is a FIELD_CREATED "created TIMESTAMP NOT NULL" field in the table definition for table.

george-lorch:
@hermanlee correct, there is a facility within 8.0 to handle this.  Create time is handled in the DD, but update an check times seem to need engine participation (still on our 8.0 MyRocks todo to figure this out).  My point, for whatever it is worth to FB MyRocks, is that basing anything on .frm is throw away for 8.0 and beyond.

lth:
@georgelorchpercona My point is that both solutions 1. checking .frm files, and 2. changing the engine data dictionary format to include creation time is throwaway work for 8.0+ because new data dictionary handles it for us.

Given that this is temporary, I feel like checking .frm files is a much simpler solution for 5.6

hermanlee:
@georgelorchpercona Yes, update time support is still needed for this patch, but it doesn't require any data dictionary changes. Since create time is supported by DD, any change we make here is considered throw away. If that is the case, it would probably be better not to modify the data dictionary and do what InnoDB does, query the .frm file. When that is ported to 8.0, the code gets removed.

spetrunia:
Ok, updated the patch:
- Create_time is now taken from the .frm file's timestamp. This means that the patch doesn't make any changes to the on-disk data format anymore.
- (followup to @lth 's comment) TRUNCATE TABLE now keeps the Create_time intact.  Update_time is set to NULL (or should I set it to the current time?)  (interesting: in InnoDB `TRUNCATE TABLE` does not change the Update_time at all: https://gist.github.com/spetrunia/7b4125ea4be4e365db20565e4ce445a8)  
- Made some cleanup in the testcase
 
And the patch is now ready for another review round.

spetrunia:
... Wait, there will be more changes

spetrunia:
Now, ready for another pass.

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -7549,6 +7576,7 @@ int ha_rocksdb::create_table(const std::string &table_name,
 
   /* Create table/key descriptions and put them into the data dictionary */
   m_tbl_def = new Rdb_tbl_def(table_name);
+  //m_tbl_def->set_create_time(time(nullptr));
```
hermanlee: Most of these changes can be removed?

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -7696,8 +7726,9 @@ int ha_rocksdb::create(const char *const name, TABLE *const table_arg,
       DBUG_RETURN(HA_ERR_ROCKSDB_CORRUPT_DATA);
     }
   }
-
-  DBUG_RETURN(create_table(str, table_arg, create_info->auto_increment_value));
+  int rc = create_table(str, table_arg, create_info->auto_increment_value);
```
hermanlee: This change can be removed?

storage/rocksdb/ha_rocksdb.cc
spetrunia: Yes, now done.  (they were actually gone in MariaDB version. I'm sorry for mixing something up and still having them there)

storage/rocksdb/ha_rocksdb.cc
spetrunia: Yes, done.

spetrunia:
@hermanlee , ok, now all input is addressed.

PR #1051, opened 2019-09-23
FB8-236, FB8-237: fix compilation issues

inikep:
FB8-236: abi_check CMake target fails after introducing SUM_ROWS_DELETED
https://jira.percona.com/browse/FB8-236
Fix missing variables m_rows_deleted, m_rows_inserted, m_rows_updated in psi_abi_statement_v1.h.pp

FB8-237: merge_small_tests CMake target fails after introducing high priority DDLs
https://jira.percona.com/browse/FB8-237
Disable high priority DDL for unit tests.

### DIFF
--- a/unittest/gunit/fake_mdl.cc
+++ b/unittest/gunit/fake_mdl.cc
@@ -1,0 +1,33 @@
+/* Copyright (c) 2019 Percona LLC and/or its affiliates. All rights reserved.
+
+   This program is free software; you can redistribute it and/or modify
+   it under the terms of the GNU General Public License, version 2.0,
+   as published by the Free Software Foundation.
+
+   This program is also distributed with certain software (including
+   but not limited to OpenSSL) that is licensed under separate terms,
+   as designated in a particular file or component or in included license
+   documentation.  The authors of MySQL hereby grant you an additional
+   permission to link the program and your derivative works with the
+   separately licensed software that they have included with MySQL.
+
+   This program is distributed in the hope that it will be useful,
+   but WITHOUT ANY WARRANTY; without even the implied warranty of
+   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+   GNU General Public License, version 2.0, for more details.
+
+   You should have received a copy of the GNU General Public License
+   along with this program; if not, write to the Free Software
+   Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301  USA */
+
+#include <sql/auth/sql_security_ctx.h>
+#include <include/my_sqlcommand.h>
+
+/**
+  We need it for merge_test_small.cc
+*/
+bool slave_high_priority_ddl = 0;
+ulonglong slave_high_priority_lock_wait_timeout_nsec = 1.0;
+
+bool support_high_priority(enum enum_sql_command) noexcept { return false; };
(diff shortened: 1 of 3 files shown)


sql/mdl.cc
```diff
@@ -3394,8 +3394,9 @@ bool MDL_lock::object_lock_kill_conflicting_locks(MDL_context *ctx,
 
 bool MDL_context::acquire_lock_nsec(MDL_request *mdl_request,
                                     ulonglong lock_wait_timeout_nsec) {
-  THD *thd = get_thd();
   bool is_high_priority_ddl = false;
+#if !defined(EXTRA_CODE_FOR_UNIT_TESTING)
```
dutow: This looks like a hack to me: this definition should be used to add extra code to testing, not to remove some functionality. It just hides the real issue - why we get those undefined externals.

inikep: `merge_small_tests` at  `unittest/gunit/CMakeLists.txt` links only:
```
SET(SQL_GUNIT_LIB_SOURCE 
  ${CMAKE_SOURCE_DIR}/sql/filesort_utils.cc 
  ${CMAKE_SOURCE_DIR}/sql/mdl.cc
  ${CMAKE_SOURCE_DIR}/sql/sql_list.cc
  ${CMAKE_SOURCE_DIR}/sql/stateless_allocator.cc
  ${CMAKE_SOURCE_DIR}/sql-common/sql_string.cc
  ${CMAKE_SOURCE_DIR}/sql/thr_malloc.cc
  )
```
These missing functions are defined in `sql_parse.cc`. When I added `sql_parse.cc` to `SQL_GUNIT_LIB_SOURCE` there is a lot more `error: undefined reference to ...`.

dutow: There are still other choices:

* these unresolved symbols don't really depend on anything, it would be possible to extract them to somewhere else, which can be added to the unit tests
* or the unit tests already use dummy implementations of some functions/variables, those could also be extended with these

inikep: Solution 1 requires to introduce a new file or move undefined functions to `mdl.cc`.
Solution 2 works:
```
--- a/sql/mdl.cc
+++ b/sql/mdl.cc
@@ -4971,3 +4971,14 @@ String timeout_message(const char *command, const char *name1,
   }
   return msg;
 }
+
+
+#if defined(EXTRA_CODE_FOR_UNIT_TESTING)
+
+bool slave_high_priority_ddl = 0;
+ulonglong slave_high_priority_lock_wait_timeout_nsec = 1.0;
+
+bool support_high_priority(enum enum_sql_command) noexcept { return false; };
+bool Security_context::check_access(ulong, bool) { return false; };
+
+#endif
```
but I'm not convinced it's a better solution because it "hides" that high priority DDL are disabled for unit tests.

unittest/gunit/fake_mdl.cc
```diff
@@ -0,0 +1,33 @@
+/* Copyright (c) 2019, Oracle and/or its affiliates. All rights reserved.
```
dutow: Shouldn't this be Percona? Or facebook? Not sure which, but I don't think we should add oracle copyright to files added by us.

inikep: updated

dutow (approved):

PR #1061, opened 2019-12-04
Add a new option to make MySQL server truncate binlogs during crash recovery.

percona-ysorokin:
Summary:
This diff adds a new option to trim binlog during recovery. The following happens if this option is set:
  1. All prepared trasnactions in the engine are rolled back during engine recovery
  2. The binlog file (which was marked 'IN USE') will be truncated to the position as reported by engine as the last commited gtid position in the binlog

Originally Reviewed By: abhinav04sharma

fbshipit-source-id: 6573e88b3bf


percona-ysorokin:
Jenkins build link
https://fb.cd.percona.com/job/fb-mysql-server-8.0-param/84/
`main.archive` - a known regression
`main.all_persisted_variables` - a known regression (just needs to be updated/re-recorded)
`rpl_nogtid.rpl_semi_sync_deadlock` - a known regression (test case times out)

hermanlee (approved):

PR #1066, opened 2019-12-12
FB8-229: MTS using table and row level dependencies (8.0.17)

inikep:
This PR consist of:
1. 34 commits ported from 5.6 at https://github.com/facebook/mysql-5.6/pull/1066/commits/70d144faa1324316053ceff21e6c6b442dd6566e
2. Improvements based on Abhinav's 8.0.11 patch at https://github.com/facebook/mysql-5.6/pull/1066/commits/392a9b66a3ec6d67d9b466dc3d3580b5de3a6f74, https://github.com/facebook/mysql-5.6/pull/1066/commits/e9f44a8332f8efb0ddb219c700e6198ed21bd4fd and https://github.com/facebook/mysql-5.6/pull/1066/commits/73d6ad305abe4233140af84418600b29cc0ae889
3. Upstream MTR parallel replication tests moved to `rpl_mts_gtid`, `rpl_mts_nogtid`, and `rpl_mts` (dependency replication).
4. Some upstream MTR parallel replication tests were fixed: https://bugs.mysql.com/bug.php?id=97694 https://bugs.mysql.com/bug.php?id=97708 https://bugs.mysql.com/bug.php?id=97713 https://bugs.mysql.com/bug.php?id=97715
5. Added `rpl_mts_dependency_validate` to check correctness of dependency-based replication


sql/rpl_gtid_state.cc
```diff
@@ -482,7 +483,17 @@ enum_return_status Gtid_state::generate_automatic_gtid(
...
+      MDL_DB_Name_List db_names;
+      thd->mdl_context.get_locked_object_db_names(db_names);
+      if (use_db_uuid && db_names.size()) {
+        rpl_sid db_uuid;
+        compute_md5_hash((char *)db_uuid.bytes, (*db_names.begin()).c_str(),
+                         (*db_names.begin()).length());
+        automatic_gtid.sidno = global_sid_map->add_sid(db_uuid);
+      } else
+        automatic_gtid.sidno = get_server_sidno();
+    }
```
abhinav04sharma: Is this change related to parallel replication?

sql/rpl_mts_submode.cc
```diff
@@ -164,7 +166,7 @@ int Mts_submode_database::wait_for_workers_to_finish(Relay_log_info *rli,
     }
     // resources relocation
     mts_move_temp_tables_to_thd(thd, entry->temporary_tables);
-    entry->temporary_tables = nullptr;
+    entry->temporary_tables = NULL;
```
abhinav04sharma: Is this change required?

sql/rpl_mts_submode.cc
```diff
@@ -958,3 +960,440 @@ Mts_submode_logical_clock::get_server_and_thread_id(TABLE *table) {
...
+    mysql_cond_wait(&dep_empty_cond, &dep_lock);
+    --num_workers_waiting;
+  }
+
+  if (!dep_queue.empty()) {
+    ret = dep_queue.front();
+    dep_queue.pop_front();
+  } else {
+    ret = nullptr;
+  }
```
abhinav04sharma: nit: this should not be required.

sql/rpl_reporting.cc
```diff
@@ -86,8 +86,13 @@ int Slave_reporting_capability::has_temporary_error(THD *thd, uint error_arg,
     wait timeout (innodb_lock_wait_timeout exceeded).
     Notice, the temporary error requires slave_trans_retries != 0)
   */
+  // Since slave_gtid_info table is not used if
+  // slave_use_idempotent_for_recovery is enabled, we don't consider COMMIT
+  // errors temporary if idempotent recovery is on.
```
abhinav04sharma: Since we haven't ported `slave_gtid_info` into 8.0 this comment is a little misleading. Actually, I'm not sure if we really need this change...

sql/log_event.h
```diff
@@ -2843,6 +2905,21 @@ class Rows_log_event : public virtual binary_log::Rows_event, public Log_event {
...
   void set_readwriteset_from_col_names(TABLE *table, table_def *tabledef,
                                        MY_BITMAP *after_image);
+
+ public:
+  bool get_keys(Relay_log_info *rli, std::shared_ptr<Log_event_wrapper> &ev,
+                std::deque<Dependency_key> &keys);
+
+ protected:
+  bool parse_keys(Relay_log_info *rli, std::shared_ptr<Log_event_wrapper> &ev,
+                  TABLE *table, std::deque<Dependency_key> &keys);
```
abhinav04sharma: Any reason for this method to be protected instead of private?

sql/log_event_wrapper.cc
```diff
@@ -0,0 +1,65 @@
...
+void Log_event_wrapper::set_raw_ev_worker(Log_event *ev) {
+  raw_ev_worker = static_cast<Slave_worker *>(ev->worker);
+}
+
+std::shared_ptr<Log_event_wrapper> Log_event_wrapper::next() {
+  if (unlikely(is_end_event)) return nullptr;
+
+  mysql_mutex_lock(&mutex);
+  Slave_worker *worker = raw_ev_worker;
+  DBUG_ASSERT(worker);
```
abhinav04sharma, line 39: Currently this assert is failing for `Rows_query_log_event` because `Rows_query_log_event::do_apply_event()` sets the worker to nullptr.

sql/mysqld.cc
```diff
@@ -1177,6 +1178,7 @@ const char *binlog_error_action_list[] = {"IGNORE_ERROR", "ABORT_SERVER",
 uint32 gtid_executed_compression_period = 0;
 bool opt_log_unsafe_statements;
 bool opt_log_global_var_changes;
+bool use_db_uuid = false;
```
abhinav04sharma: Is this change related to this feature?

sql/mysqld.cc
```diff
@@ -11219,6 +11306,9 @@ PSI_stage_info stage_suspending= { 0, "Suspending", 0, PSI_DOCUMENT_ME};
 PSI_stage_info stage_starting= { 0, "starting", 0, PSI_DOCUMENT_ME};
 PSI_stage_info stage_waiting_for_no_channel_reference= { 0, "Waiting for no channel reference.", 0, PSI_DOCUMENT_ME};
 PSI_stage_info stage_hook_begin_trans= { 0, "Executing hook on transaction begin.", 0, PSI_DOCUMENT_ME};
+PSI_stage_info stage_slave_waiting_for_dependencies= { 0, "Waiting for dependencies to be satisfied", 0, PSI_DOCUMENT_ME};
+PSI_stage_info stage_slave_waiting_for_dependency_workers= { 0, "Waiting for dependency workers to finish", 0, PSI_DOCUMENT_ME};
+PSI_stage_info stage_slave_waiting_semi_sync_ack= { 0, "Waiting for an ACK from semi-sync ACKers", 0, PSI_DOCUMENT_ME};
```
abhinav04sharma: Looks like this change also does not belong to this PR.

sql/rpl_rli_pdb.cc
```diff
@@ -2724,3 +2750,147 @@ const uint *Slave_worker::get_table_pk_field_indexes() {
...
+      diff_timespec(&worker->ts_exec[0], &worker->ts_exec[1]);
+  /* Adapting to possible new Format_description_log_event */
+  ptr_g = rli->gaq->get_job_group(ev->mts_group_idx);
+  if (ptr_g->new_fd_event) {
+    worker->set_rli_description_event(ptr_g->new_fd_event);
+    ptr_g->new_fd_event = NULL;
+  }
+
+  error = worker->slave_worker_exec_event(ev);
+  ev_wrap->set_raw_ev_worker(ev);
```
abhinav04sharma: Can we pass `worker` to this method instead? Or not maintain a separate `worker` member in `Log_event_wrapper` and just use the one in `Log_event`? FYI, there is an issue with the current code where worker will be null at this point for `Rows_query_log_event` (I've commented about this above).

sql/rpl_slave.cc
```diff
@@ -6930,6 +6964,33 @@ extern "C" void *handle_slave_sql(void *arg) {
...
+      static_cast<enum_tx_isolation>(slave_tx_isolation);
+
+  if (rli->mts_dependency_order_commits &&
+      is_mts_parallel_type_dependency(rli) &&
+      rli->opt_slave_parallel_workers > 0 && opt_bin_log &&
+      opt_log_slave_updates)
+    commit_order_mngr =
+        new Commit_order_manager(rli->opt_slave_parallel_workers);
+
+  rli->set_commit_order_manager(commit_order_mngr);
```
abhinav04sharma: This causes a mem-leak when both `slave_preserve_commit_order` and `mts_dependency_order_commits` are enabled (check line 6949 above). Maybe get rid of `mts_dependency_order_commits` completely?

sql/rpl_slave_commit_order_manager.h
```diff
@@ -1,27 +1,21 @@
...
-   documentation.  The authors of MySQL hereby grant you an additional
-   permission to link the program and your derivative works with the
-   separately licensed software that they have included with MySQL.
+   it under the terms of the GNU General Public License as published by
+   the Free Software Foundation; version 2 of the License.
 
    This program is distributed in the hope that it will be useful,
    but WITHOUT ANY WARRANTY; without even the implied warranty of
    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-   GNU General Public License, version 2.0, for more details.
```
abhinav04sharma: Is this change required?

sql/rpl_rli_pdb.cc
inikep: We can't use `Log_event` as it may be already freed in `Log_event_wrapper::next()`. I changed the function to accept `worker`.

sql/log_event.cc
```diff
@@ -14521,3 +14560,421 @@ size_t my_strmov_quoted_identifier_helper(int q, char *buffer,
...
+                  ? enum_row_image_type::DELETE_BI
+                  : enum_row_image_type::UPDATE_BI;
+    bool has_value_options =
+        (get_type_code() == binary_log::PARTIAL_UPDATE_ROWS_EVENT);
+
+    // This is needed only to satisfy some debug asserts, not strictly required
+    // for key parsing because upack_row takes cols as an argument instead of
+    // relying on read and write sets
+    bitmap_set_all(table->read_set);
+    bitmap_set_all(table->write_set);
```
abhinav04sharma: Can these lines be moved outside the loop?

inikep: It seems to work fine outside the loop. For example `rpl_mts_dependency_type_conv` fails without `bitmap_set_all` but works with `bitmap_set_all` moved outside the loop.

sql/rpl_reporting.cc
inikep: It was removed

inikep:
The changes in the updated patch:
1. All above comments were addressed
2. This patch was rebased to current `fb-mysql-8.0.17`
3. `rocksdb_rpl.rpl_mts_dependency_unique_key_conflicts` was enabled
4. Removed `mts_dependency_order_commits` in favor of `slave_preserve_commit_order` (warning: `slave_preserve_commit_order` is disabled by default)
5. Reverted https://github.com/facebook/mysql-5.6/commit/8fbbaff3f779 

I only tested (`--big-test --suite=rpl_mts`) the patch locally with clang-9 because `fb-mysql-8.0.17` has issues with gcc-9 compilation.

sql/log_event.cc
```diff
@@ -14535,3 +14574,424 @@ size_t my_strmov_quoted_identifier_helper(int q, char *buffer,
...
+
+  // This is needed only to satisfy some debug asserts, not strictly required
+  // for key parsing because upack_row takes cols as an argument instead of
+  // relying on read and write sets
+  bitmap_set_all(table->read_set);
+  bitmap_set_all(table->write_set);
+
+  for (i = 0, curr_row = m_rows_buf; curr_row != m_rows_end;
+       curr_row = curr_row_end, i++) {
+    MY_BITMAP *cols = &m_cols;  // init with before image bitmap
```
abhinav04sharma, line 14739: Is this needed?

inikep: Yes. I updated `Rows_log_event::parse_keys` for the case when the after image has a different set of columns than the before image. I'm not able to generate such a case but `Rows_log_event::force_write_to_binlog` supports it.

abhinav04sharma (approved):

PR #1067, opened 2019-12-16
Fix merge_small_tests-t, merge_large_tests-t, and routertest_component_bootstrap

inikep:
Fix unit tests after merging https://github.com/facebook/mysql-5.6/commit/3847e40e2641 and https://github.com/facebook/mysql-5.6/commit/9e2db8e1c0db

Changes:
1. Allow `MDL_context::acquire_lock_nsec` to work with `thd == nullptr`
2. Remove an upstream hack in `Test_MDL_context_owner::get_thd` that returns `(THD *)1`
3. Update error code for `BootstrapFailWhenServerResponseExceedsReadTimeout`
4. Fix `Mysys.CreateTempFile` when access to `/tmp` is denied


inikep:
I've updated this patch with added a fix for `Mysys.CreateTempFile`.

hermanlee (approved):

PR #1069, opened 2019-12-16
Fix counter leak in nonsuper_connections

percona-ysorokin:
Summary:
If a non-super user connects to mysqld with an invalid database name,
nonsuper_connections is incremented, but never decremented.

Originally Reviewed By: yizhang82

fbshipit-source-id: a9083cce39f

8.0 porting notes:
Since nonsuper_connections is already decremented by calling
release_user_connection(), changes to the code are not needed. Test is
ported though to make sure the leak won't appear again.

### DIFF
--- a/mysql-test/r/max_nonsuper_connections.result
+++ b/mysql-test/r/max_nonsuper_connections.result
@@ -51,6 +51,27 @@ disconnect con4;
 disconnect con3;
 disconnect con2;
 disconnect con1;
+Verifying nonsuper_connections is decremented when db is denied
+ERROR 42000: Access denied for user 'test_user'@'localhost' to database 'bogus_db'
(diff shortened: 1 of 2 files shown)


percona-ysorokin:
nonsuper_connections is incremented in https://github.com/facebook/mysql-5.6/blob/761c5426c8683ff195aaf0301dd7dceee7eb3a9e/sql/auth/sql_authentication.cc#L3531.
Then, as a requested database doesn't exist, release_user_connection() is called at https://github.com/facebook/mysql-5.6/blob/761c5426c8683ff195aaf0301dd7dceee7eb3a9e/sql/auth/sql_authentication.cc#L3561, which in turn decrements the counter: https://github.com/facebook/mysql-5.6/blob/761c5426c8683ff195aaf0301dd7dceee7eb3a9e/sql/sql_connect.cc#L282. That's why the test succeeds, and no code changes are necessary.

PR #1075, opened 2019-12-27
Add Information_schema.SOCKET_DIAG_SLAVES table for TCP socket diagnostic information for slave threads

percona-ysorokin:
Reference patch: 010e963

Summary: Add TCP diagnostic information from netlink sockets for replication threads. Most of the table columns should be self-explanatory.

fbshipit-source-id: 5a76c1a

---
Fix dangling SLAVE_INFO/THD in slave_list in case of slave communication failure

Reference patch: 1865e9d

Summary:
The global slave_list in master is inserted when a slave calls REGISTER_SLAVE on master, and removed when COM_BINLOG_DUMP_GTID is finished execution. If for some reason, the slave fails to communicate with the master, it is possible the slave_list didn't get a chance to be updated, and therefore leaking the SLAVE_INFO with a dangling THD pointer (that gets deleted earlier when the connection disconnects). If the slave later reconnects with the same server_id, it could ovewrite the same entry by unregister and registering again, and end up fixing this leak. However, this is still problematic because:
1. You'd have a "brief" window where slave_list contains the leaked SLAVE_INFO with dangling THD,
2. If the slave connects with a different server_id or never reconnect, you'd stuck with the leaked SLAVE_INFO and dangling THD pointer.

In either case, socket_diag_slaves or show slave hosts could access this dangling THD pointer and crash. In the actual crash dump, the wormhole is the "slave" and it always reconnects with a new random server_id, which is effectively #2.

Given that the slave may never reconnect again, the right fix is to have the THD unregister itself only if the THD in the list with the same server_id matches this particular THD (so that it won't race with the unregister/register in case the slave reconnects with a new THD).

The unregistration can happen after the cleanup during LOCK_thd_data. This way, the fill_socket_diag_slaves code would either see the empty VIO and stop proceeding, or seeing the slave_list data before the THD data gets released. The slave_list could still be "dangling" and the data could be stale, but it is "valid" enough that the only pointer are THD and rest are character arrays/integer/etc, so fill_socket_diag_slave can proceed safely.

NOTE: I've avoided changing SHOW SLAVE HOSTS for now and it needs to have the same set of locking l(locking the global THD lock, individual THD_data lock, etc) like fill_socket_diag_slaves do given that it touches THD as well. But in order to lower the risk of this change I'm avoiding SHOW SLAVE HOSTS command for now and will address it separately.

Originally Reviewed By: lth

fbshipit-source-id: ae15f99

---
Avoid DNS lookup in i_s.socket_diag_slaves table

Reference patch: 20aaaf8

Summary: DNS lookup can become quite expensive in some scenarios and as a result it may hold global mutex lock_thd_remove and thread-local mutex LOCK_thd_data for too long, which can lead to other secondary effects such as kill_zombie_dump_thread will hold LOCK_thread_count while trying to take LOCK_thd_data, which gets blocked and therefore preventing new connections getting established. This removes the unnecessary DNS lookup - they were added as a 'feature' but turns out that wasn't a great idea.

Originally Reviewed By: luqun

fbshipit-source-id: 0530f128066

---
8.0 porting notes:
* In 'rpl_master.h'
  - Added 'is_semi_sync_slave_nolock()'.
  - 'copy_slave_map()' renamed to 'copy_slaves()', underlying container
    changed from 'std::map<>' to 'malloc_unordered_map<>'.
  - 'get_slave_uuid()' extended with additional parameter 'need_lock'
    (default 'true') that allows it to be called from contexts where
    'LOCK_thd_data' is already locked.
* In 'sql_show.cc'
  - 'socket_diag_hashmap' changed from 'std::unordered_map<>' to 'malloc_unordered_map<>'.
  - Removed designated initializers that are non-standard gcc / clang c++ extensions.
  - As in 8.0 'LOCK_thd_remove' is made private, 'information_schema.socket_diag_slaves'
    is now populated via 'Fill_fill_socket_diag_slaves' class inherited from 'Do_THD_Impl'.
    On first iteration it populates internal 'm_slaves' member with 'copy_slaves()'.
    On each iteration it checks whether 'iteration_thd' is in the 'm_slaves' container
    and tries to populate IS table record only if it is found.
* Instead of 'remove_global_thread()' 'unregister_slave()' is now called inside
  'Global_THD_manager::remove_thd()'.
* Fixed 'main.socket_diag_slaves*' MTR test cases - we now try to execute a dummy DDL
  on master when slave is stopped to make sure disconnected state is detected.
* Removed unnecessary '--force-restart' from the MTR test case '.opt' files.


sql/mysqld_thd_manager.cc
```diff
@@ -205,6 +206,18 @@ void Global_THD_manager::add_thd(THD *thd) {
...
+  /*
+    It is possible when a slave connection dies after register_slave,
+    it may end up leaking SLAVE_INFO structure with a dangling THD pointer
+    In most cases it may be freed by the same slave with the same server_id
+    when it register itself again with the same server_id, but if the
+    server_id ends up being different the entry could be leaked forever and
+    with an invalid THD pointer
+    To address this issue, we let the THD remove itself from the slave list
+  */
+  unregister_slave(thd, true, true);
```
hermanlee, line 219: This line seems to cause the merge_large_tests-t unit test to fail with:

```
[ RUN      ] ThreadManagerTest.AddRemoveTHDWithGuard 
merge_large_tests-t: /home/herman/rocks-mysql/8.0/include/thr_mutex.h:186: int my_mutex_lock(my_mutex_t *, const char *, uint): Assertion `mp->m_u.m_safe_ptr != __null' failed.
                                                    
Thread 1 "merge_large_tes" received signal SIGABRT, Aborted.
__GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
50      ../sysdeps/unix/sysv/linux/raise.c: No such file or directory.
(gdb) bt                                                                                                
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
#1  0x00007ffff7a21935 in __GI_abort () at abort.c:90
#2  0x00007ffff7a30107 in __assert_fail_base (fmt=0x7ffff7ba2200 "%s%s%s:%u: %s%sAssertion `%s' failed.\n%n", assertion=assertion@entry=0x342c865 "mp->m_u.m_safe_ptr != __null", 
    file=file@entry=0x3a27fcf "/home/herman/rocks-mysql/8.0/include/thr_mutex.h", line=line@entry=186, function=function@entry=0x354ff6a "int my_mutex_lock(my_mutex_t *, const char *, uint)") at assert.c:92
#3  0x00007ffff7a30173 in __GI___assert_fail (assertion=0x342c865 "mp->m_u.m_safe_ptr != __null", file=0x3a27fcf "/home/herman/rocks-mysql/8.0/include/thr_mutex.h", line=186, 
    function=0x354ff6a "int my_mutex_lock(my_mutex_t *, const char *, uint)") at assert.c:101
#4  0x0000000005c76704 in my_mutex_lock (mp=0x8867f60 <LOCK_slave_list>, file=0x35581e7 "/home/herman/rocks-mysql/8.0/sql/rpl_master.cc", line=198) at /home/herman/rocks-mysql/8.0/include/thr_mutex.h:186
#5  0x0000000005c71676 in inline_mysql_mutex_lock (that=0x8867f60 <LOCK_slave_list>, src_file=0x35581e7 "/home/herman/rocks-mysql/8.0/sql/rpl_master.cc", src_line=198)
    at /home/herman/rocks-mysql/8.0/include/mysql/psi/mysql_mutex.h:267
#6  0x0000000005c716df in unregister_slave (thd=0x7fffffffa030, only_mine=true, need_lock_slave_list=true) at /home/herman/rocks-mysql/8.0/sql/rpl_master.cc:198
#7  0x0000000005637c70 in Global_THD_manager::remove_thd (this=0x7ffff52be000, thd=0x7fffffffa030) at /home/herman/rocks-mysql/8.0/sql/mysqld_thd_manager.cc:219
#8  0x000000000519138f in thd_manager_unittest::ThreadManagerTest_AddRemoveTHDWithGuard_Test::TestBody (this=0x7ffff53fc340) at /home/herman/rocks-mysql/8.0/unittest/gunit/thd_manager-t.cc:79
#9  0x000000000805c01e in testing::internal::HandleSehExceptionsInMethodIfSupported<testing::Test, void> (object=0x7ffff53fc340, method=&virtual testing::Test::TestBody(), location=0x39f1894 "the test body")
    at /mnt/gvfs/third-party2/googletest/4e0d1647e5289c4a821397e0d728b8478b9546ce/master/src/googletest/googletest/src/gtest.cc:2417
#10 0x000000000804d6fe in testing::internal::HandleExceptionsInMethodIfSupported<testing::Test, void> (object=0x7ffff53fc340, method=&virtual testing::Test::TestBody(), location=0x39f1894 "the test body")
    at /mnt/gvfs/third-party2/googletest/4e0d1647e5289c4a821397e0d728b8478b9546ce/master/src/googletest/googletest/src/gtest.cc:2453
#11 0x0000000008033b63 in testing::Test::Run (this=0x7ffff53fc340) at /mnt/gvfs/third-party2/googletest/4e0d1647e5289c4a821397e0d728b8478b9546ce/master/src/googletest/googletest/src/gtest.cc:2490
#12 0x00000000080349a3 in testing::TestInfo::Run (this=0x7ffff52976c0) at /mnt/gvfs/third-party2/googletest/4e0d1647e5289c4a821397e0d728b8478b9546ce/master/src/googletest/googletest/src/gtest.cc:2677
#13 0x000000000803507f in testing::TestCase::Run (this=0x7ffff529be00) at /mnt/gvfs/third-party2/googletest/4e0d1647e5289c4a821397e0d728b8478b9546ce/master/src/googletest/googletest/src/gtest.cc:2802
#14 0x000000000803d457 in testing::internal::UnitTestImpl::RunAllTests (this=0x7ffff522e000)
    at /mnt/gvfs/third-party2/googletest/4e0d1647e5289c4a821397e0d728b8478b9546ce/master/src/googletest/googletest/src/gtest.cc:4737
#15 0x000000000805eb7e in testing::internal::HandleSehExceptionsInMethodIfSupported<testing::internal::UnitTestImpl, bool> (object=0x7ffff522e000, 
    method=(bool (testing::internal::UnitTestImpl::*)(class testing::internal::UnitTestImpl * const)) 0x803cfa0 <testing::internal::UnitTestImpl::RunAllTests()>, 
    location=0x30e88dd "auxiliary test code (environments or event listeners)") at /mnt/gvfs/third-party2/googletest/4e0d1647e5289c4a821397e0d728b8478b9546ce/master/src/googletest/googletest/src/gtest.cc:2417
#16 0x000000000804f744 in testing::internal::HandleExceptionsInMethodIfSupported<testing::internal::UnitTestImpl, bool> (object=0x7ffff522e000, 
    method=(bool (testing::internal::UnitTestImpl::*)(class testing::internal::UnitTestImpl * const)) 0x803cfa0 <testing::internal::UnitTestImpl::RunAllTests()>, 
    location=0x30e88dd "auxiliary test code (environments or event listeners)") at /mnt/gvfs/third-party2/googletest/4e0d1647e5289c4a821397e0d728b8478b9546ce/master/src/googletest/googletest/src/gtest.cc:2453
#17 0x000000000803cf53 in testing::UnitTest::Run (this=0x894b590 <testing::UnitTest::GetInstance()::instance>)
    at /mnt/gvfs/third-party2/googletest/4e0d1647e5289c4a821397e0d728b8478b9546ce/master/src/googletest/googletest/src/gtest.cc:4331
#18 0x0000000006ca7771 in RUN_ALL_TESTS () at /mnt/gvfs/third-party2/googletest/4e0d1647e5289c4a821397e0d728b8478b9546ce/master/src/googletest/googletest/include/gtest/gtest.h:2255
#19 0x0000000006ca7428 in main (argc=0, argv=0x7fffffffdbe0) at /home/herman/rocks-mysql/8.0/unittest/gunit/gunit_test_main_server.cc:91
```

hermanlee: It's mainly because init_slave_list() is not called to initialize LOCK_slave_list before the test calls remove_thd() and unregister_slave().

sql/sql_show.cc
```diff
@@ -1872,12 +1884,12 @@ class Fill_authinfo_list : public Do_THD_Impl {
                    ? nullptr
                    : thd->security_context()->priv_user().str) {}
 
-  virtual void operator()(THD *current_thd) override {
-    const auto current_sctx = current_thd->security_context();
+  virtual void operator()(THD *iteration_thd) override {
```
hermanlee: Does the rename help much? I think this can trigger conflicts during future rebases.

percona-ysorokin: @hermanlee It does not, but `current_thd` is a special global var in MySQL (even a macro in some flavors), so I did this change just to avoid potential magic behavior which would be hard to understand.
This change should have been done as part of the
FB8-53: Expose user certificate details to command line
https://jira.percona.com/browse/FB8-53
https://github.com/facebook/mysql-5.6/commit/eda838a6063afed06f273f486cf7900b1ac1e00f

hermanlee: Ah, okay. We can fix the current_thd in a future rebase for that patch. I'll separate this change out into another diff so that it can be squashed together.

sql/mysqld_thd_manager.cc
percona-ysorokin: Fixed

percona-ysorokin:
PR updated
* Rebased on a newer fb-mysql-8.0.17 top.
* Reverted `current_thd -> iteration_thd` in the `Fill_authinfo_list` class.
* Fixed crash in `ThreadManagerTest.AddRemoveTHDWithGuard` unit test.

hermanlee (approved):

PR #1078, opened 2020-01-07
MRR should honor buffer size limit specify by read_rnd_buffer_size

luqun:
in ha_rocksdb::multi_range_read_info_const(), when computing required buff size for MRR, it ignore buffer size limit specified in pass in bufsz. This ignorance will cause mysql allocate a lot memory(sometime Gig bytes memory) for some MRR queries.

The change is to honor pass in bufsz value and only allocate memory min(buffer size limit, calculate buffer size)

### DIFF
--- a/mysql-test/suite/rocksdb_sys_vars/r/all_vars.result
+++ b/mysql-test/suite/rocksdb_sys_vars/r/all_vars.result
@@ -9,5 +9,7 @@ There should be *no* long test name listed below:
 select variable_name as `There should be *no* variables listed below:` from t2
 left join t1 on variable_name=test_name where test_name is null ORDER BY variable_name;
 There should be *no* variables listed below:
+ROCKSDB_MRR_BATCH_SIZE
+ROCKSDB_MRR_BATCH_SIZE
 drop table t1;
 drop table t2;
--- a/storage/rocksdb/ha_rocksdb.cc
+++ b/storage/rocksdb/ha_rocksdb.cc
@@ -682,7 +682,7 @@ static my_bool rocksdb_select_bypass_log_failed = FALSE;
 static uint32_t rocksdb_select_bypass_debug_row_delay = 0;
 static unsigned long long  // NOLINT(runtime/int)
     rocksdb_select_bypass_multiget_min = 0;
-
+static uint32_t rocksdb_mrr_batch_size = 10;
 std::atomic<uint64_t> rocksdb_row_lock_deadlocks(0);
 std::atomic<uint64_t> rocksdb_row_lock_wait_timeouts(0);
 std::atomic<uint64_t> rocksdb_snapshot_conflict_errors(0);
@@ -2143,6 +2143,12 @@ static MYSQL_SYSVAR_ULONGLONG(
     "MultiGet",
     nullptr, nullptr, SIZE_T_MAX, /* min */ 0, /* max */ SIZE_T_MAX, 0);
 
+static MYSQL_SYSVAR_UINT(
+    mrr_batch_size, rocksdb_mrr_batch_size,
+    PLUGIN_VAR_RQCMDARG,
+    "maximum number of keys to fetch during each MRR",
+    nullptr, nullptr, rocksdb_mrr_batch_size, /* min */ 0, /* max */ INT_MAX, 0);
+
 static const int ROCKSDB_ASSUMED_KEY_VALUE_DISK_SIZE = 100;
 
 static struct st_mysql_sys_var *rocksdb_system_variables[] = {
@@ -2317,6 +2323,7 @@ static struct st_mysql_sys_var *rocksdb_system_variables[] = {
     MYSQL_SYSVAR(select_bypass_log_rejected),
     MYSQL_SYSVAR(select_bypass_debug_row_delay),
     MYSQL_SYSVAR(select_bypass_multiget_min),
+    MYSQL_SYSVAR(mrr_batch_size),
     nullptr};
 
 static rocksdb::WriteOptions rdb_get_rocksdb_write_options(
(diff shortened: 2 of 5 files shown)


hermanlee (approved):
The changes look fine. Could you clang-format your changes? Also push this to 8.0 too.

spetrunia (commented):
Hello,

Item #1: the patch needs a testcase. I was using something like this when looking at it: https://gist.github.com/spetrunia/64a49d8d9698abb2b6dba1fc907de057

spetrunia:
Part #2: Do we know that Batched Key Access is not affected by @@read_rnd_buffer_size setting:

proof (run this after the paste from the previous comment)

```
create table hundred(a int primary key);
insert into hundred select A.a + B.a* 10 from ten A, ten B;
analyze table hundred, t1;

explain
select * from hundred t0, t1 where t1.pk=t0.a;
select variable_value into @val1 from information_schema.global_status where variable_name='ROCKSDB_NUMBER_MULTIGET_GET';
select * from hundred t0, t1 where t1.pk=t0.a;
# This will show that one MultiGet call was done:
select variable_value-@val1 from information_schema.global_status where variable_name='ROCKSDB_NUMBER_MULTIGET_GET';
```

spetrunia:
I see no other issues.
If we assume that BKA is outside of the scope of this patch, then it looks good to push

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -15501,6 +15518,9 @@ int ha_rocksdb::mrr_fill_buffer() {
     return HA_ERR_INTERNAL_ERROR;
   }
 
+  THD *thd = table->in_use;
+  ssize_t elements_limit = THDVAR(thd, mrr_batch_size);
+  n_elements = std::min(n_elements, elements_limit);
```
spetrunia: This will cause a buffer overrun when one sets @@rocksdb_mrr_batch_size=0 (which the patch allows)
There's no meaningful reason to use this setting but this should not result in a buffer overrun anyway...

I think these lines should be moved to be above the `if (n_elements) < 1` above.

I think we should either adjust the variable limit, or have multi_range_read_info() switch to default MRR implementation (i.e. MRR-to-non-MRR converter) in this case.

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -15229,16 +15238,24 @@ ha_rows ha_rocksdb::multi_range_read_info_const(uint keyno, RANGE_SEQ_IF *seq,
...
-      *flags &= ~HA_MRR_SUPPORT_SORTED; //  Non-sorted mode
+      *flags &= ~HA_MRR_SUPPORT_SORTED;  //  Non-sorted mode
       *flags &= ~HA_MRR_USE_DEFAULT_IMPL;
       *flags |= HA_MRR_CONVERT_REF_TO_RANGE;
-      *bufsz = mrr_get_length_per_rec() * res * 1.1 + 1;
+      uint calculated_buf = mrr_get_length_per_rec() * res * 1.1 + 1;
+      // The passed in bufsz contains maximum available buff size --- by
+      // default, its value is specify by session variable read_rnd_buff_size,
+      // final bufsz shouldn't bigger than its limit
+      *bufsz = std::min(*bufsz, calculated_buf);
```
spetrunia: Should this also be capped by `@@rocksdb_mrr_batch_size`? This would allow to use a smaller buffer.

(note: for the BKA, there seems to be no way to let it know we'll use a smaller buffer. See JOIN_CACHE_BKA::aux_buffer_incr(),  where they assume that MRR buffer usage is proportional to #rows read, and it consumes `tab->file->stats.mrr_length_per_rec` per row read)

spetrunia (changes requested):
The above two changes need to be addresed. The rest looks good.

storage/rocksdb/ha_rocksdb.cc
luqun: @lth also mentioned this offline. let me update this code to get min of three data

lth (approved):
Looks good to me!

spetrunia (commented):
You'll also need to add a test for the variable in the rocksdb_sys_vars test
suite. Right now, the test is missing and this causes rocksdb_sys_vars.all_vars test
to fail.

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -15494,6 +15517,10 @@ int ha_rocksdb::mrr_fill_buffer() {
   // The buffer has space for this many elements:
   ssize_t n_elements = (mrr_buf.buffer_end - mrr_buf.buffer) / element_size;
 
+  THD *thd = table->in_use;
+  ssize_t elements_limit = THDVAR(thd, mrr_batch_size);
+  n_elements = std::min(n_elements, elements_limit);
+
```
spetrunia, line 15524: Ok, now the memory overrun for @@rocksdb_mrr_batch_size=0 is gone but it will hit the assertion failure a few lines below.

I think this also needs to be fixed. Could you please also add the check for
`@@rocksdb_mrr_batch_size!=0` into this check into ha_rocksdb::multi_range_read_init
```
  if (!current_thd->optimizer_switch_flag(OPTIMIZER_SWITCH_MRR) ||
      (mode & HA_MRR_USE_DEFAULT_IMPL) ||
      (buf->buffer_end - buf->buffer < mrr_get_length_per_rec()) ||
      (active_index != table->s->primary_key && (mode & HA_MRR_SORTED))) {
```
The check already has the logic "if nothing fits into the buffer, switch to  the default implementation" (on line#3). I think it makes sense to apply it for rocksdb_mrr_batch_size=0, too.

luqun: Good point! I added it.

mysql-test/suite/rocksdb_sys_vars/r/all_vars.result
```diff
@@ -9,5 +9,7 @@ There should be *no* long test name listed below:
 select variable_name as `There should be *no* variables listed below:` from t2
 left join t1 on variable_name=test_name where test_name is null ORDER BY variable_name;
 There should be *no* variables listed below:
+ROCKSDB_MRR_BATCH_SIZE
+ROCKSDB_MRR_BATCH_SIZE
```
spetrunia, line 13: No, this is not the right way to do it. As the above line says, "There should be *no* variables listed below"

spetrunia: This is how to fix it: https://gist.github.com/spetrunia/0e2cd3359990e9d14705aaf24f57baac

spetrunia (changes requested):

mysql-test/suite/rocksdb_sys_vars/r/all_vars.result
luqun: shame on myself. let me fix it in another PR.

luqun: Use https://github.com/facebook/mysql-5.6/pull/1085 to track this issue.

PR #1079, opened 2020-01-09
copy truncated binlog file to tmp dir

inikep:
Summary:
If opt_trim_binlog is set, then mysqld could trim the latest binlog file to
match the engine-commited position. This diff makes mysqld copy the binlog file
to tmpdir (as set in sysvar `tmpdir`) for debugging and safety purpose.

Reference Patch: https://github.com/facebook/mysql-5.6/commit/182d59c4ecc

Originally Reviewed By: yoshinorim

fbshipit-source-id: 1c8d7601174

### DIFF
--- a/mysql-test/suite/rocksdb/t/binlog_truncate_backup.test
+++ b/mysql-test/suite/rocksdb/t/binlog_truncate_backup.test
@@ -1,0 +1,75 @@
+#
+# verify that when a binlog gets truncated, a backup gets stored in the tmpdir
+#
+--source include/not_valgrind.inc
+--source include/have_log_bin.inc
+--source include/have_debug.inc
+--source include/have_rocksdb.inc
+
+CALL mtr.add_suppression("Taking backup from .*");
+CALL mtr.add_suppression("Error reading GTIDs from binary log");
+
+CREATE TABLE t1(c1 INT) ENGINE=rocksdb;
+INSERT INTO t1 VALUES(1);
+COMMIT;
+
+# Simulate after writing to binlog, but before commiting to engine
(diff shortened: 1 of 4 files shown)


mysql-test/suite/rocksdb/t/binlog_truncate_backup.test
```diff
@@ -0,0 +1,75 @@
...
+
+--enable_reconnect
+--echo # Restart the master server
+--exec echo "restart" > $MYSQLTEST_VARDIR/tmp/mysqld.1.expect
+--source include/wait_until_connected_again.inc
+--disable_reconnect
+
+--echo # Verify that the trx was rolled back
+--let $assert_text = t1 should have 1 row
+--let $assert_cond = [SELECT COUNT(*) from t1] = 1
```
hermanlee, line 35: Is it better to use assert_cond than to print out the return values from select * from t1? I think dumping out the row contents could also clue into bugs where the wrong rows were recovered, even if the row counts match.

hermanlee (commented):
The changes look good, but I had a question about the test change.

mysql-test/suite/rocksdb/t/binlog_truncate_backup.test
inikep: In Percona we have a rule that all `SELECT COUNT(*)/SELECT *` should be replaced with `assert.inc`. The reason behind it is that it's easy to overlook such an error (a different output of `SELECT`) when re-recording an MTR test because of other changes.

hermanlee (approved):

PR #1081, opened 2020-01-09
Fix compilation after 3339788

spetrunia:
Fix compilation after patch 3339788

Streaming compression

Summary:
Extend protocol to support streaming compression. ...


spetrunia:
It was failing like so https://gist.github.com/spetrunia/25f6bae72a1b2151edd414cf6ae768c5

PR #1084, opened 2020-01-10
support high precision time in show processlist (I_S.processlist)

inikep:
Summary:
The time information from 'show processlist' and in information_schema.processlist is at the granularity of a second
This diff provides an override (system variable) to display the time at the granularity of a micro-second (e.g, 0.001234)
The variable is session settable

Reference patch: https://github.com/facebook/mysql-5.6/commit/b840ad76249

fbshipit-source-id: 7d58e50e076

**TO DO: Please re-record `mysql-test/t/all_persisted_variables.test` with
`let $total_persistent_vars=XXX + 1;` (a new variable).**


inikep:
This PR was updated:
- `main.dd_is_compatibility_cs` has been re-recorded

sql/sql_show.cc
```diff
@@ -4116,7 +4131,7 @@ ST_FIELD_INFO processlist_fields_info[] = {
     {"HOST", HOST_AND_PORT_LENGTH - 1, MYSQL_TYPE_STRING, 0, 0, "Host", 0},
     {"DB", NAME_CHAR_LEN, MYSQL_TYPE_STRING, 0, 1, "Db", 0},
     {"COMMAND", 16, MYSQL_TYPE_STRING, 0, 0, "Command", 0},
-    {"TIME", 7, MYSQL_TYPE_LONG, 0, 0, "Time", 0},
+    {"TIME", MAX_DOUBLE_STR_LENGTH, MYSQL_TYPE_DOUBLE, 0, 0, "Time", 0},
```
hermanlee, line 4136: Any particular reason for double vs. decimal as per the original diff?

inikep: I noticed some issues for 5.6:
1. `INFORMATION_SCHEMA.PROCESSLIST` "TIME" column defined as `int(7)`
(i.e. "display width = 7" assumes numbers up to `9999999`) was replaced
with `decimal(9,6)` which can address only numbers from `-999.999999` to
`999.999999`
2. `store_result_field_processlist()` uses the `double` data type in
5.6 to store time with
```protocol->store(my_timeval_to_double(&tm_delta), 6, &tmp_str);```
instead of `protocol->store_decimal()` that should be used for `decimal(9,6)`

I fixed these issues in my 8.0 port using:
`'TIME' double NOT NULL DEFAULT '0'` for INFORMATION_SCHEMA.PROCESSLIST

File: sql/sql_show.cc [hunk 2]
```diff
@@ -2264,12 +2277,14 @@ class Fill_process_list : public Do_THD_Impl {
...
-                             false);
+    long long time_delta =
+        my_micro_time() -
+        my_timeval_to_micro_time(
+            inspect_thd->query_start_timeval_trunc(DATETIME_MAX_DECIMALS));
+    if (inspect_thd->variables.high_precision_processlist)
+      table->field[5]->store((double)time_delta / 1000000);
     else
-      table->field[5]->store(0, false);
+      table->field[5]->store((long)(time_delta / 1000000), false);
```
hermanlee: Should this also be a double to match the schema of processlist_fields_info? However, this still seems to be interpreted correctly from my testing.

inikep: We have 2 cases it this patch. The first case is `SHOW PROCESSLIST` which calls `mysqld_list_processes()`. It works fine because a result field type is send along with data:
```
field_list.push_back(new Item_return_int("Time", 6, MYSQL_TYPE_DOUBLE));
table->field[5]->store((double)time_delta / 1000000);
```
and
```
field_list.push_back(field = new Item_return_int("Time", 7, MYSQL_TYPE_LONG));
table->field[5]->store((long)(time_delta / 1000000), false);
```
The second case is `SELECT * FROM information_schema.processlist` that calls `Fill_process_list::operator()`.
In this case we send a `double` or a `long` which is not correct according to a definition of the `TIME` filed of `INFORMATION_SCHEMA.PROCESSLIST`.

inikep:
I updated the patch that `TIME` in `INFORMATION_SCHEMA.PROCESSLIST` is always `double` but with a different precision:
```
--- a/sql/sql_show.cc
+++ b/sql/sql_show.cc
@@ -2282,9 +2282,9 @@ class Fill_process_list : public Do_THD_Impl {
         my_timeval_to_micro_time(
             inspect_thd->query_start_timeval_trunc(DATETIME_MAX_DECIMALS));
     if (inspect_thd->variables.high_precision_processlist)
-      table->field[5]->store((double)time_delta / 1000000);
+      table->field[5]->store((double)time_delta / 1000000); // precision in microseconds
     else
-      table->field[5]->store((long)(time_delta / 1000000), false);
+      table->field[5]->store((double)(time_delta / 1000000)); // precision in seconds
```

PR #1086, opened 2020-01-13
FB8-242: Force minimal required version of liblz4 (1.8.0) and libzstd (1.4.0)

inikep:
https://github.com/facebook/mysql-5.6/commit/e9f1b50940b introduced changes that require newer versions of liblz4 (1.8.0) and libzstd (1.4.0).
This causes compilation errors with older libs e.g. for Ubuntu Bionic which has liblz4-dev 1.7.1 and libzstd-dev 1.3.3.
More information at https://jira.percona.com/browse/FB8-242

This patch solves the issues with the following changes:
1. Check minimal required version of liblz4 (1.8.0) and libzstd (1.4.0)
2. Update bundled liblz4 to 1.8.0
3. Add bundled libzstd 1.4.0


hermanlee:
Rather than add the libraries as extras to mysql, could we update the cmake files to validate the version requirements for lz4 and zstd are met?

How easily is it for Ubuntu Bionic platforms to get the latest zstd installed?

cc: @lth @abal147

inikep:
@hermanlee 
With this patch for both `LZ4` and `zstd` we have 3 options `system|bundled|3rdparty`.
For all options cmake validates the version requirement for `LZ4` and `zstd`.
We need `bundled` for our Ubuntu Bionic which has `liblz4-dev 1.7.1` and `libzstd-dev 1.3.3` as newest packages available. 
Moreover bundled libzstd was added by upstream in 8.0.18 (libzstd-dev 1.3.3) and upgraded in 8.0.19
to 1.4.3.

hermanlee:
Ah, okay. It looks like zstd could be dropped then when we port this forward, but lz4 and the cmake checks would still needed;

inikep:
The cmake `zstd` checks would still be needed for `system`. With this patch it can fail with an error
`Required libzstd 1.4.0 and installed version is 1.3.3`. We can change it to auto-switch to `bundled` in case if required version is not found.

PR #1087, opened 2020-01-14
handle binlog trimming across file rotation

inikep:
---------- https://github.com/facebook/mysql-5.6/commit/ec039b1a46a ----------
Summary:
Fix 1:
If the following sequence of events happen, then binlog trimming will leave some
extraneous gtid in the binlog
1. Binlog file is rotated
2. A few transactions successfully flushes to binlog file, but not yet commited
to engine
3. mysqld crashes

On crash recovery, engine will point to the position in the old binlog file.
Which means extraneous trxs in the current binlog file will not be trimmed and
gtid_executed will contain these trxs. This is not an ideal scenario.

The diff fixes it by taking into account the file name stored in engine and the
current binlog file name being recovered.
1. If both file names are the same, then trimming is based solely on the
position.
2. If file names are different and the file stored in engine is older (as
noticed by the index of the file name), then we trim the current binlog file to
the begining position of the first gtid event i.e all transactions in the
current binlog file are trimmed.

Note that this does not handle the case of file names wrapping around and this
can be a rare scenario which can be fixed by instance replacements.

Fix 2:
Binlog rotation is considered a sync point for existing transactions to flush.
This is achieved by making binlog rotation block until all prepared
transactions are committed. However, the trxs that do not generate XID are not
considered 'prepared' and hence binlog rotation is not blocked by such
transactions.This could cause some extraneous GTID to be present in a
previously rotated binlog file which never got committed. This diff fixes the
problem by making binlog rotation to block on transactions without xid.

Originally Reviewed By: abhinav04sharma

fbshipit-source-id: c4d98f06025

---------- https://github.com/facebook/mysql-5.6/commit/9420260b936 ----------
    
fix failure to start
    
Summary:
Allow for recovery to proceed without special trimming if engine and recovering
binlog file prefix do not match
    
Originally Reviewed By: abhinav04sharma
    
fbshipit-source-id: fa9a466

### DIFF
--- a/mysql-test/suite/binlog/r/binlog_rotate_block_on_trxs_rbr.result
+++ b/mysql-test/suite/binlog/r/binlog_rotate_block_on_trxs_rbr.result
@@ -1,0 +1,49 @@
+CALL mtr.add_suppression("Timeout waiting for reply of binlog *");
+SET @save.rpl_semi_sync_master_timeout = @@global.rpl_semi_sync_master_timeout;
+SET @save.rpl_semi_sync_master_enabled = @@global.rpl_semi_sync_master_enabled;
+[connection default]
+CREATE TABLE t1(c1 INT PRIMARY KEY);
+CREATE TABLE blackhole(c1 INT PRIMARY KEY) ENGINE=blackhole;
+INSERT INTO t1 VALUES(1);
+COMMIT;
+# Create a 20 sec semisync timeout
+SET global rpl_semi_sync_master_timeout=20000;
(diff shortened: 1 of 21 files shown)


inikep:
This patch can be squashed with https://github.com/facebook/mysql-5.6/commit/9420260b936
Should I do it?

hermanlee:
Yes, please squash those two together.

inikep:
9420260 was squashed with this patch.

hermanlee (approved):

PR #1088, opened 2020-01-15
initialize mrr_iter for RANGE_SEQ_IF::skip_record scenario

luqun:
Summary:
mrr_iter must be initialized before using in skip_record scenario. otherwise, mysqld will crash during [bka_range_seq_skip_record](https://github.com/facebook/mysql-5.6/blob/42a5444d52f264682c7805bf8117dd884095c476/sql/sql_join_buffer.cc#L2278)(rseq point to a null)
```
call stack:
#0 bka_range_seq_skip_record
#1 myrocks::ha_rocksdb::multi_range_read_next(https://github.com/facebook/mysql-5.6/blob/fb-mysql-5.6.35/storage/rocksdb/ha_rocksdb.cc#L15675)
#2 JOIN_CACHE_BKA::join_matching_records
#3 JOIN_CACHE::join_records 
```
Test Plan:
MTR

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/mysql-test/suite/rocksdb/r/rocksdb_mrr.result
+++ b/mysql-test/suite/rocksdb/r/rocksdb_mrr.result
@@ -652,6 +652,20 @@ select * from t0 left join t3 on t3.col1=t0.a where t3.pk1 is null;
 a	pk1	pk2	col1	filler
 3	NULL	NULL	NULL	NULL
 5	NULL	NULL	NULL	NULL
+explain 
+select * from t0 left join t4 using (a) where t4.a is null;
+id	select_type	table	type	possible_keys	key	key_len	ref	rows	Extra
+1	SIMPLE	t0	ALL	NULL	NULL	NULL	NULL	10	NULL
+1	SIMPLE	t4	eq_ref	PRIMARY	PRIMARY	4	test.t0.a	1	Using where; Not exists; Using index; Using join buffer (Batched Key Access)
+select * from t0 left join t4 using (a) where t4.a is null;
+a
+0
+1
+2
+3
+5
+7
+9
 set optimizer_switch='mrr=off';
 select * from t0 left join t3 on t3.col1=t0.a where t3.pk1 is null;
 a	pk1	pk2	col1	filler
--- a/mysql-test/suite/rocksdb/t/rocksdb_mrr.test
+++ b/mysql-test/suite/rocksdb/t/rocksdb_mrr.test
@@ -352,6 +352,11 @@ delete from t3 where col1 in (3,5);
 explain
 select * from t0 left join t3 on t3.col1=t0.a where t3.pk1 is null;
 select * from t0 left join t3 on t3.col1=t0.a where t3.pk1 is null;
+
+explain 
+select * from t0 left join t4 using (a) where t4.a is null;
+select * from t0 left join t4 using (a) where t4.a is null;
+
 set optimizer_switch='mrr=off';
 select * from t0 left join t3 on t3.col1=t0.a where t3.pk1 is null;
 set optimizer_switch='mrr=on';
--- a/storage/rocksdb/ha_rocksdb.cc
+++ b/storage/rocksdb/ha_rocksdb.cc
@@ -15319,14 +15319,13 @@ class Mrr_rowid_source {
 // Rowid source that produces rowids by enumerating a sequence of ranges
 //
 class Mrr_pk_scan_rowid_source : public Mrr_rowid_source {
-  range_seq_t mrr_seq_it;
   bool mrr_ranges_eof;  // true means we've got eof when enumerating the ranges.
   ha_rocksdb *self;
  public:
   Mrr_pk_scan_rowid_source(ha_rocksdb *self_arg, void *seq_init_param,
                            uint n_ranges, uint mode) :
     mrr_ranges_eof(false), self(self_arg) {
-    mrr_seq_it = self->mrr_funcs.init(seq_init_param, n_ranges, mode);
+    self->mrr_iter= self->mrr_funcs.init(seq_init_param, n_ranges, mode);
   }
 
   int get_next_rowid(uchar *buf, int *size, char **range_ptr) override {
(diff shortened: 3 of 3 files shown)


spetrunia (changes requested):
When I debug the testcase from this patch, I can see that the call to mrr_funcs.init is made.
It is made here:

```
  #0  bka_range_seq_init (init_param=0x7fffb8014ea0, n_ranges=10, flags=1936) at /home/psergey/dev-git/mysql-5.6-rocksdb/sql/sql_join_buffer.cc:2203
  #1  0x0000555556d26322 in myrocks::Mrr_pk_scan_rowid_source::Mrr_pk_scan_rowid_source (this=0x7fffb80050e0, self_arg=0x7fffb8011c50, seq_init_param=0x7fffb8014ea0, n_ranges=10, mode=1936) at /home/psergey/dev-git/mysql-5.6-rocksdb/storage/rocksdb/ha_rocksdb.cc:15329
  #2  0x0000555556d116bb in myrocks::ha_rocksdb::multi_range_read_init (this=0x7fffb8011c50, seq=0x7ffff45aae80, seq_init_param=0x7fffb8014ea0, n_ranges=10, mode=1936, buf=0x7fffb8014f78) at /home/psergey/dev-git/mysql-5.6-rocksdb/storage/rocksdb/ha_rocksdb.cc:15447
  #3  0x000055555682d898 in JOIN_CACHE_BKA::init_join_matching_records (this=0x7fffb8014ea0, seq_funcs=0x7ffff45aae80, ranges=10) at /home/psergey/dev-git/mysql-5.6-rocksdb/sql/sql_join_buffer.cc:2446
  #4  0x000055555682d5c0 in JOIN_CACHE_BKA::join_matching_records (this=0x7fffb8014ea0, skip_last=false) at /home/psergey/dev-git/mysql-5.6-rocksdb/sql/sql_join_buffer.cc:2353
  #5  0x000055555682c40c in JOIN_CACHE::join_records (this=0x7fffb8014ea0, skip_last=false) at /home/psergey/dev-git/mysql-5.6-rocksdb/sql/sql_join_buffer.cc:1709
  #6  0x000055555682fb19 in JOIN_CACHE::end_send (this=0x7fffb8014ea0) at /home/psergey/dev-git/mysql-5.6-rocksdb/sql/sql_join_buffer.h:475
  #7  0x00005555565c4b69 in sub_select_op (join=0x7fffb80073e8, join_tab=0x7fffb8014a98, end_of_records=true) at /home/psergey/dev-git/mysql-5.6-rocksdb/sql/sql_executor.cc:1094
  #8  0x00005555565c4cb9 in sub_select (join=0x7fffb80073e8, join_tab=0x7fffb8014798, end_of_records=true) at /home/psergey/dev-git/mysql-5.6-rocksdb/sql/sql_executor.cc:1250
```

The code that makes the call: 
```c++
class Mrr_pk_scan_rowid_source : public Mrr_rowid_source {
...
  Mrr_pk_scan_rowid_source(ha_rocksdb *self_arg, void *seq_init_param,
                           uint n_ranges, uint mode) :
    mrr_ranges_eof(false), self(self_arg) {
    mrr_seq_it = self->mrr_funcs.init(seq_init_param, n_ranges, mode);
```
The problem is that its result is not saved in $ha_rocksdb_object->mrr_iter object. It is stored somewhere else.

Mrr_rowid_source has another implementation, Mrr_sec_key_rowid_source. Mrr_sec_key_rowid_source::Mrr_sec_key_rowid_source makes this call

```c++
    return self->handler::multi_range_read_init(seq, seq_init_param, n_ranges,
                                                mode, nullptr);
```

which will set the ha_rocksdb_object->mrr_funcs_it. So, Mrr_sec_key_rowid_source is already fine. 

Getting back to the Mrr_pk_scan_rowid_source.  If this patch is applied in the current form, the code will make two calls to `mrr_funcs.init(...)` for each scan. This doesn't look good.

My suggestion would be to:
- make Mrr_pk_scan_rowid_source::Mrr_pk_scan_rowid_source update the ha_rocksdb_object->mrr_iter like so:
 
```c++
    self->mrr_iter= self->mrr_funcs.init(seq_init_param, n_ranges, mode);
```
- Then, remove Mrr_pk_scan_rowid_source::mrr_seq_it (and use self->mrr_iter wherever it was used).

This way:
- There will be one place to store mrr_iter, handler::mrr_iter (it looks like introducing Mrr_pk_scan_rowid_source::mrr_seq_it was a mistake)
- mrr_funcs.init() call will be made exactly once for each scan.

Any thoughts?

spetrunia (approved):
Looks good now.

lth:
@spetrunia Just out of curiosity, is there a reason why we have to call skip_record twice? It looks like we call skip_record in both `Mrr_sec_key_rowid_source::get_next_rowid` during buffer fill and in `ha_rocksdb::multi_range_read_next`. 

I also wonder if we have testing that covers the skip_index_tuple callback case.

spetrunia:
@lth 
> I also wonder if we have testing that covers the skip_index_tuple callback case.

We do. Search for these lines in `rocksdb_mrr.test`:
```
--echo # Test for BKA's variant of Index Condition Pushdown. With BKA,
--echo # pushed index conditions that refer to preceding tables are 
--echo # handled in a special way because there's no clear concept of 
--echo # "current row" for the preceding table(s)
```

spetrunia:
@lth 
> @spetrunia Just out of curiosity, is there a reason why we have to call skip_record twice? It looks like we call skip_record in both Mrr_sec_key_rowid_source::get_next_rowid during buffer fill and in ha_rocksdb::multi_range_read_next.

TLDR: Not much.

Let me start by saying that calling skip_record() is totally optional.

(as opposed to skip_index_tuple, which one *must* call, because of SQL layer relying on the storage engine to check the pushed index condition and not checking it itself)

Functions that skip_record may point to:
bka_unique_range_seq_skip_record, bka_range_seq_skip_record.

Both basically do the cache->get_match_flag_by_pos(...) check.

Now, if we look at JOIN_CACHE_BKA::join_matching_records() (or at JOIN_CACHE_BKA_UNIQUE::join_matching_records), we can see this function
* calls file->multi_range_read_next() to read a row
* then calls get_match_flag_by_pos(...) itself.
ergo, the check in the storage engine call is not mandatory.

It can be an optimization though, for the cases where we are only interested in the reading just one match, e.g.

SELECT * FROM t1 LEFT JOIN t2 ON t2.key=t1.col WHERE t2.not_null_col IS NULL
SELECT * FROM t1 WHERE t1.col IN (SELECT t2.key FROM t2 ....)

Here, for a row t1.ROW_X we only need one matching record in t2.

So, 
- the call in Mrr_sec_key_rowid_source::get_next_rowid() will prevent us from   reading a row of t2 which is match for t1.ROW_X for which we've already returned a match.

- the call in ha_rocksdb::multi_range_read_next is not as useful (we've read the t2's row already. It only saves the convert_record_from_storage_format() call, and the overhead of going to the SQL layer and back).

lth (approved):

PR #1089, opened 2020-01-16
Fix bug with COM_RESET_CONNECTION not respecting CLIENT_INTERACTIVE

percona-ysorokin:
Summary:
The COM_RESET_CONNECTION command is not currently respecting
the CLIENT_INTERACTIVE flag. The CLIENT_INTERACTIVE flag is tied to the
connection, rather than the session. When COM_RESET_CONNECTION resets
session state, it's expected that wait_timeout will be set to interactive_timeout when CLIENT_INTERACTIVE is set.

Reference Patch: facebook/mysql-5.6@5de8b68

Originally Reviewed By: jkedgar

8.0 porting notes:
* The `init ()` call in the `THD` constructor is delayed until client capabilities are set

### DIFF
--- a/client/mysqltest.cc
+++ b/client/mysqltest.cc
@@ -236,6 +236,9 @@ static bool can_handle_expired_passwords = true;
 static bool use_async_client = false;
 static bool enable_async_client = false;
 
+// To enable sending CLIENT_INTERACTIVE in client flags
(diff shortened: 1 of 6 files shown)


percona-ysorokin:
PR updated - rebased and fixed commit authors

PR #1090, opened 2020-01-16
Fixed githash.pl call parameters after introducing RocksDB

percona-ysorokin:
In cases when Facebook MySQL Server is compiled from a source tree that was not
retrieved from git (or just does not have the '.git' directory) MYSQL_GITHASH and
MYSQL_GITDATE CMake options can be defined to skip some calls to git
executable that identifies current revision.
This used to work fine until RocksDB storage engine code was merged. Now git
executable is unconditionally executed to determine revision of the rocksdb
submodule.
Fixed by passing "--rocksdb_githash=${ROCKSDB_GITHASH}" and
"--rocksdb_gitdate=${ROCKSDB_GITDATE})" to "githash.pl" script as in 5.6.35.

Reference patch: facebook/mysql-5.6@38c5ced

### DIFF
--- a/sql/CMakeLists.txt
+++ b/sql/CMakeLists.txt
@@ -1069,7 +1069,9 @@ IF(DEFINED MYSQL_GITHASH)
   # the perl script
   SET(GITHASH_PERL_OPTS ${GITHASH_PERL_OPTS}
                         --mysql_githash=${MYSQL_GITHASH}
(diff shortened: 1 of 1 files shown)


percona-ysorokin:
This change has already been imported to fb-mysql-8.0.17
https://github.com/facebook/mysql-5.6/commit/408ea32ed21
This PR should be closed.

PR #1094, opened 2020-01-21
removed hack to set opt_core_file in mysqld.cc

inikep:
removed hack to set opt_core_file in mysqld.cc

Summary: avoid dumping core files in case the server was aborted because of a write error

fbshipit-source-id: 736a451c418

Reference Patch: https://github.com/facebook/mysql-5.6/commit/e04edc791c2

--------------------------------------------------------------------

Suppressing core file generation on SHUTDOWN

Summary:
Forced SHUTDOWN (exiting with non zero error code) has
cases that existing or upcoming connections hit  errors like assertions
before completing exit. In that case, generating core files should be skipped.

fbshipit-source-id: 2f472e0

Reference Patch: https://github.com/facebook/mysql-5.6/commit/70278a5700f

**TO DO: Probably `main.all_persisted_variables` has to be re-recorded.**


percona-ysorokin (approved):
LGTM

inikep:
This PR ports also `Make mysqld "core_files" sys-var dynamic` at https://github.com/facebook/mysql-5.6/commit/4e26365d2bc2 which was required dependency

PR #1095, opened 2020-01-22
LOAD DATA INFILE with COMPRESSED option

percona-ysorokin:
Summary: Loading data into the table from ZSTD compressed text file

Reference Patch: facebook/mysql-5.6@784cc50

Originally Reviewed By:Reviewed By: lth

8.0 porting notes:
* Added more RAII for resorce management inside 'decompressor' class.
* Added a new check in 'main.load_data_infile_compressed' MTR
  test case that verifies that an error is triggered when user tries
  to read data from a file outside '@@secure_file_priv' directory.


percona-ysorokin:
PR updated - rebased and fixed commit authors

PR #1097, opened 2020-01-23
Replacing boost ptree by rapidjson and Better handle bad db_metadata value

percona-ysorokin:


percona-ysorokin:
PR updated - rebased and fixed commit authors.

PR #1099, opened 2020-01-28
Big files slow removal

percona-ysorokin:
Summary:
Original algorithm
https://github.com/midom/slowrm

When INNODB executes command DROP TABLE t, it deletes table and file.
If table is big, file removal may take a lot of time on SSD.
So here, first we move file to special directory MYSQLD_DATADIR/.slowrm and then slowly truncate it in background thread.

Reference Patch: https://github.com/facebook/mysql-5.6/commit/cc9b09ba97c

Originally Reviewed By: lth

fbshipit-source-id: c428c462c0b

### DIFF
--- a/storage/innobase/os/os0file.cc
+++ b/storage/innobase/os/os0file.cc
@@ -849,16 +849,6 @@ static bool os_file_handle_error(const char *name, const char *operation);
 @return DB_SUCCESS or error code */
 dberr_t os_file_punch_hole(os_file_t fh, os_offset_t off, os_offset_t len);
 
-/**
-Does error handling when a file operation fails.
-@param[in]	name		File name or NULL
-@param[in]	operation	Name of operation e.g., "read", "write"
-@param[in]	on_error_silent	if true then don't print any message to the log.
-@return true if we should retry the operation */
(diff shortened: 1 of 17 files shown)


storage/innobase/os/os0file.cc
```diff
@@ -3216,24 +3206,26 @@ bool os_file_create_directory(const char *pathname, bool fail_if_exists) {
...
+    if (errno != ENOENT || handle_nodir)
+      os_file_handle_error_no_exit(path, "opendir", false);
     return (false);
   }
 
   entry = readdir(directory);
 
   while (entry != nullptr) {
-    scan_cbk(path, entry->d_name);
+    if (!scan_cbk(path, entry->d_name)) break;
```
hermanlee, line 3228: Do we need the callback to determine whether to break out early? It seems even with slowrm, if one entry is not found, then the loop can immediately moved on to the next entry. With this change, the patch uses the next iteration of slowrm thread to pick up where it left off.

Is there much value to stopping early and resuming? I'm thinking of minimizing the API changes involved around the callback function, since this generates conflicts for rebases, and I don't see much value for breaking out early here.

PR #1111, opened 2020-02-17
make sys_vars.all_vars more robust

inikep:
Summary:
1. sys_vars.all_sys_vars should pretty any variables. Removing those variable name in the result file. And add all the missing *_basic.test files.
2. change skip_master_info_error_msg_extra test file name so that the file name match the real variable name.
3. change max variable length threshold.

Reference Patch: https://github.com/facebook/mysql-5.6/commit/130c294f977

Originally Reviewed By: luqun

fbshipit-source-id: 133f604ae58


percona-ysorokin (approved):
LGTM

PR #1117, opened 2020-03-24
FB8-246 main.implicit_commit fails in debug compilation

percona-ysorokin:
Summary:
`main.implicit_commit` crashes in debug build

Crashes due to the fact that `Diagnostics_area::m_status` is not set to other
than the default value (`DA_EMPTY`) on error. The error occurs due to the lack
of implementation of `Env::GetThreadList()` for rockdb.
Any `Diagnostics_area::m_status` must be set in `mysql_execute_command()` after
executing sql commands. In case of any error `Diagnostics_area::m_status` should
be set to `DA_ERROR` using `my_error()`.

### DIFF
--- a/sql/sql_parse.cc
+++ b/sql/sql_parse.cc
@@ -3505,17 +3505,26 @@ int mysql_execute_command(THD *thd, bool first_level, ulonglong *last_timer) {
     }
     case SQLCOM_SHOW_ENGINE_STATUS: {
       if (check_global_access(thd, PROCESS_ACL)) goto error;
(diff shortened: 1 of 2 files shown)


hermanlee (commented):
It seems InnoDB would hit similar issues if innodb_show_status() returned true as well?

Would it be better for rocksdb_show_status() to set my_error() if res is going to be true? The fix would have less impact on upstream code and if InnoDB does hit similar issues, we should file a bug and ask for a fix.

PR #1118, opened 2020-03-24
Fixes of innodb.mysqld_core_dump_without_buffer_pool*

percona-ysorokin:
Summary:
- innodb.mysqld_core_dump_without_buffer_pool* MTR test cases expect
  core pattern to be simple 'core' and core files located in the var dir
- `mysqld` is killed to dump a core using SIGSEGV


hermanlee:
It looks like upstream added the checkers to validate the core pattern is core. Can you file a bug with Oracle and request this be upstreamed?

PR #1125, opened 2020-04-20
Issue#1026: MyRocks may return "Can't find record" which is unexpected

spetrunia:
MyRocks generally uses Snapshot Checking for transactional isolation:
read rows through snapshot, if we need to modify a row - get a lock
on it and check that it hasn't been modified after snapshot was taken.

The exception is unique secondary key check which uses "READ-COMMITED":
get a lock on the secondary key value, then scan the latest committed
data to see if there are duplicate rows.

This works, except for INSERT ... ON DUPLICATE KEY UPDATE, where
the SQL layer expects to be able to read the duplicate row if it has
got HA_ERR_FOUND_DUPP_KEY error from the storage engine.  If it
doesn't see it, the user gets a "record not found" error.

Fixed this by returning "Snapshot conflict" in the case when the
duplicate row is present but is not visible in the current
transaction's snapshot.

### DIFF
--- a/storage/rocksdb/ha_rocksdb.cc
+++ b/storage/rocksdb/ha_rocksdb.cc
@@ -6394,7 +6394,7 @@ ha_rocksdb::ha_rocksdb(my_core::handlerton *const hton,
       m_lock_rows(RDB_LOCK_NONE),
       m_keyread_only(false),
       m_insert_with_update(false),
-      m_dup_pk_found(false),
+      m_dup_key_found(false),
       mrr_rowid_reader(nullptr),
       mrr_n_elements(0),
       mrr_enabled_keyread(false),
@@ -8494,6 +8494,23 @@ int ha_rocksdb::index_read_map_impl(uchar *const buf, const uchar *const key,
         kd.get_key_parts()) {
       using_full_key = false;
     }
+
+    if (m_insert_with_update && m_dup_key_found &&
+        active_index == m_dupp_errkey) {
+      /*
+        We are in INSERT ... ON DUPLICATE KEY UPDATE, and this is a read
+        that SQL layer does to read the duplicate key.
+        Its rowid is saved in m_last_rowkey. Get the full record and return it.
+      */
+
+      DBUG_ASSERT(m_dup_key_retrieved_record.length() >= packed_size);
+      DBUG_ASSERT(memcmp(m_dup_key_retrieved_record.ptr(), m_sk_packed_tuple,
+                         packed_size) == 0);
+
+      rc = get_row_by_rowid(buf, m_last_rowkey.ptr(), m_last_rowkey.length());
+      DBUG_RETURN(rc);
+    }
+
   } else {
     packed_size = kd.pack_index_tuple(table, m_pack_buffer, m_sk_packed_tuple,
                                       key, keypart_map);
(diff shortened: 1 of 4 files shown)


hermanlee:
For read-committed mode, it seems we have different behavior between PK and unique SK?

```
--source include/have_debug_sync.inc                                                                        
                                                                                                            
CREATE table t1 (pk int primary key, val int default 0) engine=rocksdb;                                     

insert into t1 (pk) values (20), (30);                                                                      
                                                                                                            
connect (con1, localhost, root,,);                                                                          

connection con1;                                                                                            
set tx_isolation='READ-COMMITTED';                                                                          
begin;
set debug_sync='rocksdb.after_unique_pk_check SIGNAL here WAIT_FOR go';                                     
send insert into t1 (pk) values (1), (2) on duplicate key update val = val + 1;                             
  
connection default;                                                                                         
set debug_sync='now WAIT_FOR here';                                                                         
begin;
insert into t1 (pk) values (2);                                                                             
commit;
set debug_sync='now SIGNAL go';                                                                             

connection con1;
reap;
commit;

connection default;
disconnect con1;                                                                                            

select * from t1;                                                                                           

drop table t1;                                                                                              

set debug_sync='RESET';          
```

The resulting select is:

```
select * from t1;
pk      val
1       0
2       1
20      0
30      0
```

For read-committed mode, do we want similar behavior? get_for_update() has do_validate, which is used to mimic InnoDB's read-committed behavior. So the snapshot set on the statement is temporarily ignored in the function.

```
          // If snapshot is set, and if skipping validation,
          // call GetForUpdate without validation and set back old snapshot
          auto saved_snapshot = m_read_opts.snapshot;
          m_read_opts.snapshot = nullptr;
          s = m_rocksdb_tx->GetForUpdate(m_read_opts, column_family, key, value,
                                         exclusive, false);
          m_read_opts.snapshot = saved_snapshot;
```

spetrunia:
Indeed...
A modified testcase 
https://gist.github.com/spetrunia/b3120fe66e81fc60f3762b7bdb4a7bcf
and this patch:
```diff
diff --git a/storage/rocksdb/ha_rocksdb.cc b/storage/rocksdb/ha_rocksdb.cc
index a83169e5a7d..3fa690f7014 100644
--- a/storage/rocksdb/ha_rocksdb.cc
+++ b/storage/rocksdb/ha_rocksdb.cc
@@ -9914,6 +9914,7 @@ int ha_rocksdb::check_uniqueness_and_lock(
       }
     } else {
       rc = check_and_lock_sk(key_id, row_info, &found);
+        DEBUG_SYNC(ha_thd(), "rocksdb.after_unique_sk_check");
     }
 
     if (rc != HA_EXIT_SUCCESS) {
```
still causes "Record not found" error:
https://gist.github.com/spetrunia/1a4e19cbc0f86ddc4ab34347acf1e3e7

spetrunia:
A related patch: https://github.com/facebook/mysql-5.6/commit/dfbb8ffd7b0d6f1c8f3cb211edcc446ced5badee , " Skip snapshot validation on Read Committed (#920) "

spetrunia:
So, when using READ-COMMITTED:
- the write path reads the latest committed data and doesn't do Snapshot Checking.
- the read path (that is, when the SQL layer calls `ha_rocksdb::index_read_map()`) will still use a read snapshot (which is refreshed at each statement iirc).

And this is why the issue is still there.

Solution #1:

Step around the 

```
  const bool do_validate =
      my_core::thd_tx_isolation(ha_thd()) > ISO_READ_COMMITTED;
```
logic in ha_rocksdb::check_and_lock_sk() by calling Rdb_transaction::get_for_update() and explicitly request that validation is done:

```diff
@@ -9869,7 +9869,8 @@ int ha_rocksdb::check_and_lock_sk(const uint key_id,
     */
     rocksdb::Slice found_key= iter->key();
     const rocksdb::Status s =
-        get_for_update(row_info.tx, kd.get_cf(), found_key, nullptr);
+        row_info.tx->get_for_update(kd.get_cf(), found_key, nullptr,
+                                    true /*exclusive*/, true /*do_validate*/);
 
     if (!s.ok()) {
       DBUG_ASSERT(!s.IsNotFound());
```
This fixes the "Record not found" error but doesn't fix the difference between PK/SK behavior

spetrunia:
Another relevant patch:  https://github.com/facebook/mysql-5.6/commit/469134bea110a7f1c888eb6ebbe4bce5a85b69be , "Avoid 2nd GetForUpdate on INSERT on DUPLICATE KEY. "
AFAIU its idea is to save the duplicate that ha_rocksdb::[update]_write_row() has found and then return it when the lookup is made.  Could we follow this approach with SK?

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -8494,6 +8494,23 @@ int ha_rocksdb::index_read_map_impl(uchar *const buf, const uchar *const key,
         kd.get_key_parts()) {
       using_full_key = false;
     }
+
+    if (m_insert_with_update && m_dup_key_found &&
+        active_index == m_dup_key_found) {
```
hermanlee: This conditional check looks wrong. active_index is uint and m_dup_key_found is a bool. Is this meant to be m_dupp_errkey?

spetrunia: Yes. Good catch. Fixed

PR #1129, opened 2020-05-05
Adding options to do blind deletes and not to use tx api

oleksandr-kachan:
Reference Patch: https://github.com/facebook/mysql-5.6/commit/1245598bf1e

---------- https://github.com/facebook/mysql-5.6/commit/1245598bf1e ----------
Summary:
This diff adds two session variables to allow some operations faster.
- rocksdb_blind_delete_primary_key -- Executing deletes without calling
GetForUpdate(). To use this feature, target table must have primary key
only (no secondary key). No multi-table operations are allowed.
- rocksdb_master_skip_tx_api -- Not using transaction api on master.
This means no row locks are held during row modifications. It's user's
responsibility to make operations safe.

Benchmark results -- deleting 100k rows with bulk deletes (1000 row
deletes per statement).
 1) Normal -- 1.448s
 2) rocksdb_blind_delete_primary_key -- 0.671s
 3) 2) + rocksdb_master_skip_tx_api -- 0.324s

fbshipit-source-id: fa5c138862e

### DIFF
--- a/mysql-test/suite/rocksdb/t/blind_delete_without_tx_api.test
+++ b/mysql-test/suite/rocksdb/t/blind_delete_without_tx_api.test
@@ -1,15 +1,9 @@
 --source include/have_rocksdb.inc
---source include/master-slave.inc;
+
+source include/master-slave.inc;
 
 --source include/rpl_connection_master.inc
 
-eval SET SESSION TRANSACTION ISOLATION LEVEL $trx_isolation;
-set @save_rocksdb_blind_delete_primary_key=@@session.rocksdb_blind_delete_primary_key;
-set @save_rocksdb_master_skip_tx_api=@@session.rocksdb_master_skip_tx_api;
-
---disable_warnings
-DROP TABLE IF EXISTS t1,t2;
---enable_warnings
 create table t1 (id int primary key, value int, value2 varchar(200)) engine=rocksdb;
 create table t2 (id int primary key, value int, value2 varchar(200), index(value)) engine=rocksdb;
 
@@ -42,7 +36,6 @@ SELECT count(*) FROM t1;
 
 --source include/sync_slave_sql_with_master.inc
 --source include/rpl_connection_slave.inc
-eval SET SESSION TRANSACTION ISOLATION LEVEL $trx_isolation;
 SELECT count(*) FROM t1;
 --source include/rpl_connection_master.inc
 
@@ -109,9 +102,11 @@ call mtr.add_suppression("Slave: Can't find record in 't1'.*");
 --source include/wait_for_slave_sql_error.inc
 
 --source include/rpl_connection_slave.inc
-set @save_rocksdb_read_free_rpl=@@global.rocksdb_read_free_rpl;
-set global rocksdb_read_free_rpl=PK_SK;
-START SLAVE;
+set @save_rocksdb_read_free_rpl_tables=@@global.rocksdb_read_free_rpl_tables;
+set global rocksdb_read_free_rpl_tables="t.*";
+set global sql_slave_skip_counter=1;
+--source include/start_slave.inc
+
 --source include/rpl_connection_master.inc
 --source include/sync_slave_sql_with_master.inc
 --source include/rpl_connection_slave.inc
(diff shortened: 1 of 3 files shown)


mysql-test/suite/rocksdb/t/blind_delete_without_tx_api.test
```diff
@@ -0,0 +1,131 @@
+--source include/have_rocksdb.inc
+
+source include/master-slave.inc;
+
+--source include/rpl_connection_master.inc
+
+set @save_rocksdb_blind_delete_primary_key=@@session.rocksdb_blind_delete_primary_key;
```
percona-ysorokin: There is no need to save/restore SESSION variables. This technique is only needed for GLOBAL ones.

mysql-test/suite/rocksdb/t/blind_delete_without_tx_api.test
```diff
@@ -0,0 +1,131 @@
...
+set @save_rocksdb_master_skip_tx_api=@@session.rocksdb_master_skip_tx_api;
```
percona-ysorokin: Likewise

mysql-test/suite/rocksdb/t/blind_delete_without_tx_api.test
```diff
@@ -0,0 +1,131 @@
...
+
+--disable_warnings
+DROP TABLE IF EXISTS t1,t2;
```
percona-ysorokin: Ancient MTR idiom - no need to delete tables before creating them. MTR now always start from a clean state.

mysql-test/suite/rocksdb/t/blind_delete_without_tx_api.test
```diff
@@ -0,0 +1,131 @@
...
+--source include/rpl_connection_slave.inc
+SELECT count(*) FROM t1;
+--source include/rpl_connection_master.inc
+
+
+# cleanup
+--source include/rpl_connection_slave.inc
+set global rocksdb_read_free_rpl_tables=@save_rocksdb_read_free_rpl_tables;
+--source include/rpl_connection_master.inc
+set session rocksdb_blind_delete_primary_key=@save_rocksdb_blind_delete_primary_key;
```
percona-ysorokin: No need to save/restore SESSION variables

mysql-test/suite/rocksdb/t/blind_delete_without_tx_api.test
```diff
@@ -0,0 +1,131 @@
...
+set session rocksdb_master_skip_tx_api=@save_rocksdb_master_skip_tx_api;
```
percona-ysorokin: Likewise

percona-ysorokin (changes requested):

percona-ysorokin (approved):
LGTM

hermanlee (commented):
It looks like there are some later changes to the tests which were not in the original list.

Could you update this patch to include the test changes from

[mysql-5.6][PR] Skip snapshot validation on Read Committed
https://github.com/facebook/mysql-5.6/commit/dfbb8ffd7b0d6f1c8f3cb211edcc446ced5badee

too? It looks like the test gets refactored to support both read-committed and repeatable-read isolation levels.

oleksandr-kachan:
> Could you update this patch to include the test changes from
> 
> [mysql-5.6][PR] Skip snapshot validation on Read Committed
> [dfbb8ff](https://github.com/facebook/mysql-5.6/commit/dfbb8ffd7b0d6f1c8f3cb211edcc446ced5badee)
> 
> too? It looks like the test gets refactored to support both read-committed and repeatable-read isolation levels.

Just to recheck. Isn't this change already done for 8.0.17 in https://github.com/facebook/mysql-5.6/commit/d433e614f566ff83f31219a3f3cd93a80e689b7f#diff-924972e9288665ca701c6a69ccf484fc

hermanlee:
Ah, yes, it has been ported. That change had renamed the rocksdb.blind_delete_without_tx_api test to rocksdb.blind_delete_rc, but this change re-introduces rocksdb.blind_delete_without_tx_api, which is what confused me. It looks like this patch is already in 8.0.17 then, but renamed to the blind_delete_r[rc] tests. We should be able to close this request then. Thanks!

PR #1130, opened 2020-05-05
PS-7000: Add "rocksdb.type_varchar_packing" to validate padding of [VAR]CHAR collations

inikep:
Add `rocksdb.type_varchar_packing` to validate padding of `CHAR` and `VARCHAR` collations.
Only multi-level collations (`latin2_czech_cs` and `cp1250_czech_cs`) are not supported.


inikep:
I added a support for `binary`. This MTR test checks if all collations except `binary` are space padded.

hermanlee:
This looks to be a good test to have for 5.6. Will there be an equivalent one for 8.0?

hermanlee (approved):

inikep:
> This looks to be a good test to have for 5.6. Will there be an equivalent one for 8.0?

Yes, I will also create an improved version for 8.0

PR #1131, opened 2020-05-08
Fix bug 1704195 / 87065 / TDB-83 (Stop ANALYZE TABLE from flushing ta…

oleksandr-kachan:
…ble definition cache)

Reference Patch: https://github.com/facebook/mysql-5.6/commit/dbc7c09e252

---------- https://github.com/facebook/mysql-5.6/commit/dbc7c09e252 ----------

Summary:
This is adapted from https://github.com/percona/percona-server/pull/1977

  Make ANALYZE TABLE stop flushing affected tables from the table
  definition cache, which has the effect of not blocking any subsequent
  new queries involving the table if there's a parallel long-running
  query:

  - new table flag HA_ONLINE_ANALYZE, return it for InnoDB and TokuDB
    tables;
  - in mysql_admin_table, if we are performing ANALYZE TABLE, and the
    table flag is set, do not remove the table from the table
    definition cache, do not invalidate query cache;
  - in partitioning handler, refresh the query optimizer statistics
    after ANALYZE if the underlying handler supports HA_ONLINE_ANALYZE;
  - new testcases main.percona_nonflushing_analyze_debug,
    parts.percona_nonflushing_abalyze_debug and a supporting debug sync
    point.

  For TokuDB, this change exposes bug TDB-83 (Index cardinality stats
  updated for handler::info(HA_STATUS_CONST), not often enough for
  tokudb_cardinality_scale_percent). TokuDB may return different
  rec_per_key values depending on dynamic variable
  tokudb_cardinality_scale_percent value. The server does not have a way
  of knowing that changing this variable invalidates the previous
  rec_per_key values in any opened table shares, and so does not call
  info(HA_STATUS_CONST) again. Fix by updating rec_per_key for both
  HA_STATUS_CONST and HA_STATUS_VARIABLE. This also forces a re-record
  of tokudb.bugs.db756_card_part_hash_1_pick, with the new output
  seeming to be more correct.

fbshipit-source-id: 659933a3408

### DIFF
--- a/mysql-test/collections/disabled_rocksdb.def
+++ b/mysql-test/collections/disabled_rocksdb.def
@@ -27,7 +27,6 @@ rocksdb.autoinc_crash_safe_partition : BUG#888001 partition feature is missing i
 rocksdb.explicit_snapshot : BUG#0000 missing patch 0b671a35cc7 SQL to create and manage explicit snapshots
 rocksdb.innodb_i_s_tables_disabled : BUG#0000 missing patch
 rocksdb_rpl.rpl_rocksdb_schema_change : BUG#0000 missing 031b5cec711 Use column names while applying row changes on slave
-rocksdb.percona_nonflushing_analyze_debug : BUG#0000 missing dbc7c09e252 Fix bug 1704195 / 87065 / TDB-83 (Stop ANALYZE...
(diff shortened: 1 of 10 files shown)


percona-ysorokin (approved):
LGTM

PR #1132, opened 2020-05-13
Dump core for mysqld without large memory buffer

oleksandr-kachan:
Reference Patch: https://github.com/facebook/mysql-5.6/commit/e31a21144b8

---------- https://github.com/facebook/mysql-5.6/commit/e31a21144b8 ----------
Summary:
Core files can be very helpful for debugging production crashes but core dump has been disabled on production due to multiple reasons. One of the reasons is that the core file can be huge due to the huge InnoDB buffer pool within the process. This diff introduces a new command line argument '--innodb-dump-core-without-large-mem-buf', when it is set, all the large memory allocation by function os_mem_alloc_large() in InnoDB won't be dumped in the core file by calling Linux API madvise() with argument MADV_DONTDUMP. It can be disabled by '--skip-innodb-dump-core-without-large-mem-buf'. os_mem_alloc_large() is used to allocate large memory chunk for InnoDB buffer pool, file data merge, row log buffer, etc. Buffer pool is the main one. With this feature, the core size will be less than 3GB when the buffer pool size is 60GB so it won't impact production.
-- This sys var only take effect when core dump is enabled by setting 'core-file' in my.cnf.
-- By default '--innodb-dump-core-without-large-mem-buf' is set in prod to dump smaller cores, but it is not set by default in MTR so that full core file can be dumped.
-- This diff also makes sys var core_file readonly.
-- There are other bugs that prevent mysqld from dumping cores, e.g. when mysqld switching user from root to a non-root user, etc. Those will be addressed in separated diffs.

Originally Reviewed By: jtolmer

fbshipit-source-id: 119924337ab

### DIFF
--- a/mysql-test/suite/innodb/t/mysqld_core_dump_without_large_mem_buf.test
+++ b/mysql-test/suite/innodb/t/mysqld_core_dump_without_large_mem_buf.test
@@ -1,0 +1,18 @@
+################################################################################
+# This test is to test if mysqld can dump a core without large memory buffers.
+# See opt file for the config:
+#   (1) --innodb-dump-core-without-large-mem-buf is set
+#   (2) the buffer pool is set to be large so that without dropping the large
+#       memory buffers the core size will be much greater than 3.5GB (the actual
+#       core size is less than 3GB now but set the limit to 3.5GB here in case
+#       the memory footprint increases in the future)
+
+--source include/big_test.inc
+--source include/not_valgrind.inc
+--source include/linux_core_pattern.inc
+
+--let $restart_parameters = restart:--log-error=$MYSQLTEST_VARDIR/log/core_dump.err
+--source include/restart_mysqld_no_echo.inc
+
+--let $expected_max_core_size = 3584
+--source include/mysqld_core_dump.inc
--- a/mysql-test/suite/innodb/t/mysqld_core_dump_without_large_mem_buf_with_resizing.test
+++ b/mysql-test/suite/innodb/t/mysqld_core_dump_without_large_mem_buf_with_resizing.test
@@ -1,0 +1,51 @@
+################################################################################
+# This test is to test if mysqld can dump a core without large memory buffers.
+# See opt file for the config:
+#   (1) --innodb-dump-core-without-large-mem-buf is set
+#   (2) the buffer pool is set to be large initially, shrink it, then expand
+#       it back to the original large size, without dropping the large memory
+#       buffers the core size will be much greater than 5.5GB (the actual
+#       core size is less than 5GB now but set the limit to 5.5GB here in case
+#       the memory footprint increases in the future)
+
+--source include/big_test.inc
+--source include/not_valgrind.inc
+--source include/linux_core_pattern.inc
+
+--let $restart_parameters = restart:--log-error=$MYSQLTEST_VARDIR/log/core_dump.err
+--source include/restart_mysqld_no_echo.inc
+
+let $wait_timeout = 600;
+let $wait_condition =
+  SELECT SUBSTR(variable_value, 1, 34) = 'Completed resizing buffer pool at '
+  FROM performance_schema.global_status
+  WHERE LOWER(variable_name) = 'innodb_buffer_pool_resize_status';
+
+--disable_query_log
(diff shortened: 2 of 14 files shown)


mysql-test/suite/innodb/t/mysqld_core_dump_without_large_mem_buf.test
```diff
@@ -0,0 +1,22 @@
...
+--source include/not_valgrind.inc
+--source include/linux_core_pattern.inc
+
+--echo # Shutdown server
+--source include/shutdown_mysqld.inc
+
+--echo # Restart server with --log-error
+--exec echo "restart:--log-error=$MYSQLTEST_VARDIR/log/core_dump.err" > $MYSQLTEST_VARDIR/tmp/mysqld.1.expect
+--enable_reconnect
+--source include/wait_until_connected_again.inc
```
percona-ysorokin: Use `restart_mysqld.inc` / `restart_mysqld_no_echo.inc` instead
```suggestion
--let $restart_parameters = restart:--log-error=$MYSQLTEST_VARDIR/log/core_dump.err
--source include/restart_mysqld.inc
```

mysql-test/suite/innodb/t/mysqld_core_dump_without_large_mem_buf_with_resizing.test
```diff
@@ -0,0 +1,55 @@
...
+--source include/not_valgrind.inc
+--source include/linux_core_pattern.inc
+
+--echo # Shutdown server
+--source include/shutdown_mysqld.inc
+
+--echo # Restart server with --log-error
+--exec echo "restart:--log-error=$MYSQLTEST_VARDIR/log/core_dump_with_resizing.err" > $MYSQLTEST_VARDIR/tmp/mysqld.1.expect
+--enable_reconnect
+--source include/wait_until_connected_again.inc
```
percona-ysorokin: Likewise

mysql-test/suite/innodb/t/mysqld_core_dump_without_large_mem_buf.test
```diff
@@ -0,0 +1,22 @@
...
+# See opt file for the config:
+#   (1) --innodb-dump-core-without-large-mem-buf is set
+#   (2) the buffer pool is set to be large so that without dropping the large
+#       memory buffers the core size will be much greater than 3.5GB (the actual
+#       core size is less than 3GB now but set the limit to 3.5GB here in case
+#       the memory footprint increases in the future)
+
+--source include/not_valgrind.inc
+--source include/linux_core_pattern.inc
+
```
percona-ysorokin, line 13: Consider making this test `--source include/big_test.inc` as it request 60GB buffer pool

mysql-test/suite/innodb/t/mysqld_core_dump_without_large_mem_buf_with_resizing.test
```diff
@@ -0,0 +1,55 @@
...
+#   (1) --innodb-dump-core-without-large-mem-buf is set
+#   (2) the buffer pool is set to be large initially, shrink it, then expand
+#       it back to the original large size, without dropping the large memory
+#       buffers the core size will be much greater than 5.5GB (the actual
+#       core size is less than 5GB now but set the limit to 5.5GB here in case
+#       the memory footprint increases in the future)
+
+--source include/not_valgrind.inc
+--source include/linux_core_pattern.inc
+
```
percona-ysorokin, line 14: Consider making this test `--source include/big_test.inc` as it request 60GB buffer pool

percona-ysorokin (changes requested):

percona-ysorokin (approved):
LGTM

hermanlee:
Sorry, we made a mistake adding this change to the list. The change was already upstreamed to Oracle and in 8.0.14:

commit 891460995137598a6e0ae3684ba1cc6ccd0c3ca3
Author: Jakub Łopuszański <jakub.lopuszanski@oracle.com>
Date:   Wed Aug 29 13:27:14 2018 +0200

    Bug #27724476   CONTRIBUTION BY FACEBOOK: DETACH INNODB BUFFER POOL BEFORE CREATING CORE

Since it's already been ported, I'll close out this PR.

PR #1134, opened 2020-05-14
Issue#250: MyRocks/Innodb different output from query with order by o…

oleksandr-kachan:
…n table with index and decimal type

Reference Patch: https://github.com/facebook/mysql-5.6/commit/4ea109268b6

---------- https://github.com/facebook/mysql-5.6/commit/4ea109268b6 ----------

Summary:
Make open_binary_frm() set TABLE_SHARE::primary_key before it computes
"handler->index_flags(keyno, ...) & HA_KEYREAD_ONLY".
In MyRocks, the value of this expression depends on whether keyno is a
clustered PK or not, so it's essential that MyRocks sees the correct
value in TABLE_SHARE::primary_key.

fbshipit-source-id: d2c17e5c5c5


percona-ysorokin (approved):
LGTM

PR #1138, opened 2020-05-25
Add lock information to lock timeout error message

oleksandr-kachan:
Reference Patch: https://github.com/facebook/mysql-5.6/commit/1e50e3fe7bf

Notice: Only outdated TODOs were removed in this commit

---------- https://github.com/facebook/mysql-5.6/commit/1e50e3fe7bf ----------
Summary:
Currently, on lock timeout, we only give
  ERROR 1205 (HY000): Lock wait timeout exceeded; try restarting transaction
 which has no information on what type of lock timed out.

This change augments the message with the lock type to say something like
  ERROR 1205 (HY000): Lock wait timeout exceeded; try restarting transaction: Timeout on index: test/t1.GEN_CLUST_INDEX
which shows which type of lock timed out as well as the associated object.

There two cases to handle here. For error messages coming from the storage engine, the error is stored onto the transaction, which will get forwarded to the SQL layer for display through `handler::get_error_message`.. For MDL locks in the SQL layer, we are directly using the information in the MDL lock to display the error message.

I didn't touch NDB because I didn't bother getting that to build.

Originally Reviewed By: jtolmer

fbshipit-source-id: 014f0bc407f

### DIFF
--- a/mysql-test/suite/rocksdb/t/hermitage.inc
+++ b/mysql-test/suite/rocksdb/t/hermitage.inc
@@ -125,8 +125,6 @@ if ($trx_isolation == "READ COMMITTED")
 }
 if ($trx_isolation == "REPEATABLE READ")
 {
-  # TODO(yzha) - Update error message once following patch is ported
-  # 1e50e3fe7bf Add lock information to lock timeout error message
   --error ER_LOCK_DEADLOCK
   reap;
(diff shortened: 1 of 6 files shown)


percona-ysorokin (approved):
LGTM

oleksandr-kachan:
It looks like all code is already in the 8.0.17 branch. This commit only removes some TODOs in code. Was there anything else planned in scope of this patch?

PR #1139, opened 2020-06-02
MultiGet-MRR implementation

oleksandr-kachan:
MultiGet-MRR implementation

Reference Patch: https://github.com/facebook/mysql-5.6/commit/45110dcda26
Reference Patch: https://github.com/facebook/mysql-5.6/commit/be12c8cfb44
Reference Patch: https://github.com/facebook/mysql-5.6/commit/d0c98c1e6a5
Reference Patch: https://github.com/facebook/mysql-5.6/commit/e99310c1197
Reference Patch: https://github.com/facebook/mysql-5.6/commit/888d47893f1
Reference Patch: https://github.com/facebook/mysql-5.6/commit/40d6766e57b

---------- https://github.com/facebook/mysql-5.6/commit/45110dcda26 ----------
Issue790: MultiGet-MRR implementation (#1041)

Summary:
First implementation.
Supports scans on full primary key
Cost-based choice whether to use MRR/BKA is not supported (set mrr=on,mrr_cost_based=off)
Pull Request resolved: #1041

fbshipit-source-id: dba1bfa
---------- https://github.com/facebook/mysql-5.6/commit/be12c8cfb44 ----------
Stabilize rocksdb.rocksdb_mrr

Summary:
This test was flaky because whenever flush/compaction runs, some rocksdb counters may have changed, and the sql query to find difference in counters could return additional changed counters.

To fix this, just disable background work when we run the counters collection query. Also, some unflushed stats on the background thread could also modify these counters. Force a flush using ANALYZE.

Squash with: D17058718

fbshipit-source-id: 1156d08

---------- https://github.com/facebook/mysql-5.6/commit/d0c98c1e6a5 ----------
Fix rocksdb.rocksdb_mrr

Summary:
Previous attempt at stabilizing (be12c8c) did not work because rocksdb sometimes acquires superversion for multigets depending on whether the thread local version is stale or not. This will depend on whether compaction/flush has happened or not, which is non deterministic.

Instead of trying to fix this, we will just query for the variables that we care about. I feel like querying for all variables with rocksdb_num prefix can potentially get very noisy anyway.

Originally Reviewed By: yizhang82

fbshipit-source-id: 8001e0d
---------- https://github.com/facebook/mysql-5.6/commit/e99310c1197 ----------
Summary:
in ha_rocksdb::multi_range_read_info_const(), when computing required buff size for MRR, it ignore buffer size limit specified in pass in bufsz. This ignorance will cause mysql allocate a lot memory(sometime Gig bytes memory) for some MRR queries.

The change is to honor pass in bufsz value and only allocate memory min(buffer size limit, calculate buffer size)
Pull Request resolved: #1078

Test Plan: MTR

Pulled By: luqun

fbshipit-source-id: dcb9e94
---------- https://github.com/facebook/mysql-5.6/commit/888d47893f1 ----------
Summary:
Previous PR #1078 incorrectly rebase  all_vars.results. As pointed out by sergey, the correct way is to add variable basic testcase.
Pull Request resolved: #1085

Test Plan: MTR

Pulled By: luqun

fbshipit-source-id: 2ad01878c0c
---------- https://github.com/facebook/mysql-5.6/commit/40d6766e57b ----------
Summary:
mrr_iter must be initialized before using in skip_record scenario. otherwise, mysqld will crash during [bka_range_seq_skip_record](https://github.com/facebook/mysql-5.6/blob/42a5444d52f264682c7805bf8117dd884095c476/sql/sql_join_buffer.cc#L2278)(rseq point to a null)

call stack:
```
```
#3 JOIN_CACHE::join_records
Pull Request resolved: #1088

Test Plan: MTR

Pulled By: luqun

fbshipit-source-id: 666dcd911d6

### DIFF
--- a/mysql-test/suite/rocksdb/t/rocksdb_mrr.test
+++ b/mysql-test/suite/rocksdb/t/rocksdb_mrr.test
@@ -1,0 +1,352 @@
+#
+#  Test for MyRocks' MRR implementation based on MultiGet.
+#
+--source include/have_rocksdb.inc
+
+create table t0(a int);
+insert into t0 values (0),(1),(2),(3),(4),(5),(6),(7),(8),(9);
+
+create table t1(a int);
+insert into t1 select A.a + B.a* 10 + C.a * 100 from t0 A, t0 B, t0 C;
+
+create table t2 (
+  pk int primary key,
+  col1 int,
+  filler char(32)
+) engine=rocksdb default charset=latin1;
+
+insert into t2 select a,a,a from t1;
+set global rocksdb_force_flush_memtable_now=1;
+
+set @save_optimizer_switch=@@optimizer_switch;
+set optimizer_switch='mrr=on,mrr_cost_based=off,batched_key_access=on';
+
+explain 
+select * from t2,t0 where t2.pk=t0.a;
+select * from t2,t0 where t2.pk=t0.a;
+
+--echo # Check the counters
+create temporary table t10 as 
+select * from performance_schema.global_status
+where variable_name in ('ROCKSDB_ROWS_READ', 'ROCKSDB_NUMBER_DB_NEXT', 'ROCKSDB_NUMBER_DB_NEXT_FOUND', 'ROCKSDB_NUMBER_DB_SEEK', 'ROCKSDB_NUMBER_DB_SEEK_FOUND', 'ROCKSDB_NUMBER_MULTIGET_BYTES_READ', 'ROCKSDB_NUMBER_MULTIGET_GET', 'ROCKSDB_NUMBER_MULTIGET_KEYS_READ');
+
+flush status; # for Handler_mrr_init
+
+select * from t2,t0 where t2.pk=t0.a;
+
+show status like 'Handler_mrr_init';
+
+create temporary table t11 as 
+select * from performance_schema.global_status
+where variable_name in ('ROCKSDB_ROWS_READ', 'ROCKSDB_NUMBER_DB_NEXT', 'ROCKSDB_NUMBER_DB_NEXT_FOUND', 'ROCKSDB_NUMBER_DB_SEEK', 'ROCKSDB_NUMBER_DB_SEEK_FOUND', 'ROCKSDB_NUMBER_MULTIGET_BYTES_READ', 'ROCKSDB_NUMBER_MULTIGET_GET', 'ROCKSDB_NUMBER_MULTIGET_KEYS_READ');
+
+select 
+  variable_name, 
+  t11.variable_value - t10.variable_value as DIFF 
+from 
+  t10 join t11 using (VARIABLE_NAME)
+having
+  DIFF>0;
+drop table t10, t11;
+
+--echo # This will use MRR:
+explain select * from t2 where pk in (1,2,3,4,5);
+
+--echo # Check how optimizer_switch flags effect it:
+set optimizer_switch='mrr_cost_based=on';
+
+--echo # No MRR/BKA:
+explain select * from t2,t0 where t2.pk=t0.a;
+explain select * from t2 where pk in (1,2,3,4,5);
+
+set optimizer_switch='mrr_cost_based=off,mrr=off';
+--echo # No MRR/BKA:
+explain select * from t2,t0 where t2.pk=t0.a;
+explain select * from t2 where pk in (1,2,3,4,5);
+
+set optimizer_switch='mrr_cost_based=off,mrr=on';
+--echo # Have MRR/BKA:
+explain select * from t2,t0 where t2.pk=t0.a;
+explain select * from t2 where pk in (1,2,3,4,5);
+
+--echo # This will not use MRR, as one of the ranges is a non-singlepoint:
+explain select * from t2 where pk in (1,2,3,4,5) or pk between 6 and 8;
+
+create table t3 (
+  pk1 int,
+  pk2 int,
+  col1 int,
+  filler char(32),
+  primary key(pk1, pk2),
+  key(col1)
+) engine=rocksdb default charset=latin1;
+
+insert into t3 select a,a, a,a from t1;
+set global rocksdb_force_flush_memtable_now=1;
+
+--echo # This must not use BKA as it uses key prefix:
+explain 
+select * from t3,t0 where t3.pk1=t0.a;
+
+--echo # This will use BKA as it uses full key:
+explain 
+select * from t3,t0 where t3.pk1=t0.a and t3.pk2=t0.a+1;
+
+--echo # This will use MRR as it uses full key:
+explain select * from t3 where pk1 in (1,2,3,4,5) and pk2 in (3,4,5);
+
+--echo # This will not use MRR as there are non-singlepoint ranges:
+explain select * from t3 where (pk1 in (1,2,4,5) and pk2 in (4,5)) or pk1=3;
+
+--echo # Due to HA_MRR_SUPPORT_SORTED, the following will use MRR, and not require filesort:
+explain select * from t3 where pk1 in (1,2,3,4,5) and pk2 in (3,4,5) order by pk1, pk2;
+
+#--echo #
+#--echo # More basic tests
+#--echo #
+
+--echo # Test for lookups returning nothing
+delete from t2 where pk in (3,5,7,9,10) or pk between 100 and 200;
+
+explain
+select * from t2,t0 where t2.pk=t0.a;
+select * from t2,t0 where t2.pk=t0.a;
+
+
+--echo # Test for a scan returning nothing at all:
+explain
+select * from t2,t0 where t2.pk=t0.a+500100;
+select * from t2,t0 where t2.pk=t0.a+500100;
+
+--echo # Make the scan finish before it has returned all the data:
+select * from t2,t0 where t2.pk=t0.a limit 1;
+
+--echo # Re-execute MRR scan in a subquery:
+create table t4(a int primary key);
+insert into t4 values (4),(6),(8);
+
+--replace_column 10 # 11 #
+explain
+select a ,
+       (select concat(t0.a,'-',t2.pk) 
+        from t2,t0 
+        where t2.pk=t0.a and t2.col1>=t4.a
+        limit 1) as SUBQ
+from t4;
+
+select a ,
+       (select concat(t0.a,'-',t2.pk) 
+        from t2,t0 
+        where t2.pk=t0.a and t2.col1>=t4.a
+        limit 1) as SUBQ
+from t4;
+
+--echo #
+--echo # MRR/BKA on secondary keys
+--echo #
+
+--echo # This won't use BKA because it's an index-only scan:
+explain
+select t3.col1 from t3,t0 where t3.col1=t0.a;
+
+--echo # This will use BKA: 
+explain
+select * from t3,t0 where t3.col1=t0.a;
+
+select * from t3,t0 where t3.col1=t0.a;
+
+--echo # Now, run the query and check the counters
+create temporary table t11 as
+select * from performance_schema.global_status
+where variable_name in ('ROCKSDB_ROWS_READ', 'ROCKSDB_NUMBER_DB_NEXT', 'ROCKSDB_NUMBER_DB_NEXT_FOUND', 'ROCKSDB_NUMBER_DB_SEEK', 'ROCKSDB_NUMBER_DB_SEEK_FOUND', 'ROCKSDB_NUMBER_MULTIGET_BYTES_READ', 'ROCKSDB_NUMBER_MULTIGET_GET', 'ROCKSDB_NUMBER_MULTIGET_KEYS_READ');
+
+select * from t3,t0 where t3.col1=t0.a;
+
+create temporary table t12 as
+select * from performance_schema.global_status
+where variable_name in ('ROCKSDB_ROWS_READ', 'ROCKSDB_NUMBER_DB_NEXT', 'ROCKSDB_NUMBER_DB_NEXT_FOUND', 'ROCKSDB_NUMBER_DB_SEEK', 'ROCKSDB_NUMBER_DB_SEEK_FOUND', 'ROCKSDB_NUMBER_MULTIGET_BYTES_READ', 'ROCKSDB_NUMBER_MULTIGET_GET', 'ROCKSDB_NUMBER_MULTIGET_KEYS_READ');
+
+select 
+  variable_name, 
+  t12.variable_value - t11.variable_value as DIFF 
+from 
+  t11 join t12 using (VARIABLE_NAME)
+having
+  DIFF>0;
+
+drop table t11,t12;
+
+--echo #
+--echo # Check how the counters are incremented when SQL
+--echo # layer doesn't read all of the MultiGet results
+--echo #
+create table t20 (a int);
+insert into t20 values (1);
+set global rocksdb_force_flush_memtable_now=1;
+
+explain
+select a, a+20 in (select t2.filler from t2,t0 where t2.pk=t0.a+20) from t20;
+
+create temporary table t10 as 
+select * from performance_schema.global_status
+where variable_name in ('ROCKSDB_ROWS_READ', 'ROCKSDB_NUMBER_DB_NEXT', 'ROCKSDB_NUMBER_DB_NEXT_FOUND', 'ROCKSDB_NUMBER_DB_SEEK', 'ROCKSDB_NUMBER_DB_SEEK_FOUND', 'ROCKSDB_NUMBER_MULTIGET_BYTES_READ', 'ROCKSDB_NUMBER_MULTIGET_GET', 'ROCKSDB_NUMBER_MULTIGET_KEYS_READ');
+
+select a, a+20 in (select t2.filler from t2,t0 where t2.pk=t0.a+20) from t20;
+
+create temporary table t11 as 
+select * from performance_schema.global_status
+where variable_name in ('ROCKSDB_ROWS_READ', 'ROCKSDB_NUMBER_DB_NEXT', 'ROCKSDB_NUMBER_DB_NEXT_FOUND', 'ROCKSDB_NUMBER_DB_SEEK', 'ROCKSDB_NUMBER_DB_SEEK_FOUND', 'ROCKSDB_NUMBER_MULTIGET_BYTES_READ', 'ROCKSDB_NUMBER_MULTIGET_GET', 'ROCKSDB_NUMBER_MULTIGET_KEYS_READ');
+
+--echo # The below shows ROCKSDB_ROWS_READ=13. 
+--echo #  1 is from table t20
+--echo # 10 are from table t0, BKA reads all its rows into buffer
+--echo #  2 are from table t2.  BKA code reads two rows before it figures that
+--echo #    the subquery has a match and no more rows are needed.
+--echo #    (note that MultiGet call has read all 10 rows, but SQL layer only has read 2)
+select 
+  variable_name, 
+  t11.variable_value - t10.variable_value as DIFF 
+from 
+  t10 join t11 using (VARIABLE_NAME)
+having
+  DIFF>0;
+drop table t10, t11;
+drop table t20;
+
+--echo #
+--echo # Check how MRR works without BKA
+--echo #
+explain select t3.col1 from t3 where t3.col1=20 or t3.col1 between 25 and 28;
+
+--echo # This will use MRR:
+explain 
+select * from t3 where t3.col1=20 or t3.col1 between 25 and 28;
+select * from t3 where t3.col1=20 or t3.col1 between 25 and 28;
+
+--echo # Check if Index Condition Pushdown works
+explain 
+select * from t3 where (t3.col1=20 or t3.col1 between 25 and 28) and mod(t3.col1,2)=0;
+select * from t3 where (t3.col1=20 or t3.col1 between 25 and 28) and mod(t3.col1,2)=0;
+
+select * from t3 use index() where (t3.col1=20 or t3.col1 between 25 and 28) and mod(t3.col1,2)=0;
+
+explain
+select pk1,pk2,col1, filler,mod(t3.col1,2) from t3
+where (t3.col1=20 or t3.col1 between 25 and 28) and mod(t3.col1,2)=0;
+
+select pk1,pk2,col1, filler,mod(t3.col1,2) from t3
+where (t3.col1=20 or t3.col1 between 25 and 28) and mod(t3.col1,2)=0;
+
+--echo #
+--echo # Test for BKA's variant of Index Condition Pushdown. With BKA,
+--echo # pushed index conditions that refer to preceding tables are 
+--echo # handled in a special way because there's no clear concept of 
+--echo # "current row" for the preceding table(s)
+--echo #
+
+explain
+select * from t0,t3 where t3.col1=t0.a and mod(t3.pk2,2)=t0.a;
+select * from t0,t3 where t3.col1=t0.a and mod(t3.pk2,2)=t0.a;
+
+set optimizer_switch='mrr=off';
+select * from t0,t3 where t3.col1=t0.a and mod(t3.pk2,2)=t0.a;
+set optimizer_switch='mrr=on';
+
+--echo #
+--echo # A query which has RANGE_SEQ_IF::skip_record != nullptr.
+--echo #
+--echo # MultiGet/MRR does not invoke skip_record() as it would not produce
+--echo # much speedup.
+--echo #
+insert into t3 select 10000+a, 10000+a, a, 'duplicate-match' from t1;
+delete from t3 where col1 in (3,5);
+
+--replace_column 10 # 11 #
+explain
+select * from t0 left join t3 on t3.col1=t0.a where t3.pk1 is null;
+select * from t0 left join t3 on t3.col1=t0.a where t3.pk1 is null;
+set optimizer_switch='mrr=off';
+select * from t0 left join t3 on t3.col1=t0.a where t3.pk1 is null;
+set optimizer_switch='mrr=on';
+
+--replace_column 10 # 11 #
+explain
+select * from t0 where t0.a in (select t3.col1 from t3 where char_length(t3.filler)<30);
+select * from t0 where t0.a in (select t3.col1 from t3 where char_length(t3.filler)<30);
+
+drop table t0,t1,t2,t3,t4;
+
+--echo #
+--echo # Multi-keypart testcase
+--echo #
+create table t0(a int);
+insert into t0 values (0),(1),(2),(3),(4),(5),(6),(7),(8),(9);
+
+CREATE TABLE t2 (
+  id bigint(20) unsigned NOT NULL DEFAULT '0',
+  type tinyint(3) unsigned NOT NULL DEFAULT '0',
+  c1 int(10) unsigned NOT NULL DEFAULT '0',
+  c2 int(10) unsigned NOT NULL DEFAULT '0',
+  time_updated int(10) unsigned NOT NULL DEFAULT '0',
+  id2 bigint(20) unsigned NOT NULL DEFAULT '0',
+  c3 int(10) unsigned NOT NULL DEFAULT '0',
+  c_id bigint(20) unsigned NOT NULL DEFAULT '0',
+  a_id bigint(20) unsigned NOT NULL DEFAULT '0',
+  PRIMARY KEY (id,`type`),
+  KEY time_updated (time_updated),
+  KEY id2 (id2),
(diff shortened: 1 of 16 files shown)


mysql-test/suite/rocksdb/t/rocksdb_mrr.test
```diff
@@ -0,0 +1,353 @@
...
+insert into t1 select A.a + B.a* 10 + C.a * 100 from t0 A, t0 B, t0 C;
+
+create table t2 (
+  pk int primary key,
+  col1 int,
+  filler char(32)
+) engine=rocksdb default charset=latin1;
+
+insert into t2 select a,a,a from t1;
+set global rocksdb_force_flush_memtable_now=1;
```
percona-ysorokin, line 19: Please, double check if this is not a debug-only variable. If it is, add `--source include/have_debug.inc`

mysql-test/suite/rocksdb/t/rocksdb_mrr.test
```diff
@@ -0,0 +1,353 @@
...
+analyze table t1;
+--replace_column 10 # 11 #
+explain
+select * from t1 where kp1=1 order by kp2 limit 10;
+# This shouldn't fail an assert:
+select * from t1 where kp1=1 order by kp2 limit 10;
+
+drop table t0, t1;
+set optimizer_switch=@save_optimizer_switch;
```
percona-ysorokin: not really necessary to save/restore a session variable

mysql-test/suite/rocksdb/t/rocksdb_mrr2.test
```diff
@@ -0,0 +1,38 @@
...
+set optimizer_switch='mrr=on,mrr_cost_based=off,batched_key_access=on';
+
+--replace_column 11 #
+explain 
+select pk, col1 from t2,t0 
+where t2.pk=t0.a*90 and md5(blob1)=md5(repeat(t2.pk,100)) and md5(blob2)=md5(repeat(t2.pk,200));
+
+select pk, col1 from t2,t0
+where t2.pk=t0.a*90 and md5(blob1)=md5(repeat(t2.pk,100)) and md5(blob2)=md5(repeat(t2.pk,200));
+set optimizer_switch=@save_optimizer_switch;
```
percona-ysorokin: not really necessary to save/restore a session variable

mysql-test/suite/rocksdb/t/rocksdb_mrr2.test
```diff
@@ -0,0 +1,38 @@
...
+create table t2 (
+  pk int primary key,
+  col1 int,
+  blob1 text,
+  blob2 text
+) engine=rocksdb;
+
+insert into t2 select a,a,repeat(a,100), repeat(a,200) from t1;
+
+set global rocksdb_force_flush_memtable_now=1;
```
percona-ysorokin, line 24: Likewise

mysql-test/suite/rocksdb/t/rocksdb_mrr_debug.test
```diff
@@ -0,0 +1,59 @@
...
+insert into t1 select A.a + B.a* 10 + C.a * 100 from t0 A, t0 B, t0 C;
+
+create table t2 (
+  pk int primary key,
+  col1 int,
+  filler char(32)
+) engine=rocksdb;
+
+insert into t2 select a,a,a from t1;
+set global rocksdb_force_flush_memtable_now=1;
```
percona-ysorokin, line 19: Likewise

sql/sql_optimizer.cc
```diff
@@ -2559,7 +2579,8 @@ void JOIN::adjust_access_methods() {
         // From table scan to index scan, thus filter effect needs no recalc.
       }
     } else if (tab->type() == JT_REF) {
-      if (can_switch_from_ref_to_range(thd, tab, ORDER_NOT_RELEVANT, false)) {
+      const char *switch_reason = can_switch_from_ref_to_range(thd, tab, ORDER_NOT_RELEVANT, false);
+      if (switch_reason) {
```
percona-ysorokin: Avoid implicit conversions
```suggestion
      if (switch_reason != nullptr) {
```

sql/sql_optimizer.cc
```diff
@@ -2472,10 +2474,27 @@ static bool can_switch_from_ref_to_range(THD *thd, JOIN_TAB *tab,
...
+         - used MRR
+         - MRR implementation indicated that it's preferred to use MRR over
+           ref(const).
+        then switch to range access as it will read rows using the MRR interface
+      */
+      if (tab->position()->key->key == tab->quick()->index &&
+          length == tab->quick()->max_used_key_length &&
+          tab->quick()->get_type() == QUICK_SELECT_I::QS_TYPE_RANGE) {
+        auto *qr = static_cast<QUICK_RANGE_SELECT*>(tab->quick());
+        if (!(qr->mrr_flags & HA_MRR_USE_DEFAULT_IMPL) &&
```
percona-ysorokin: Avoid implicit conversions
```suggestion
        if ((qr->mrr_flags & HA_MRR_USE_DEFAULT_IMPL) == 0 &&
```

sql/sql_optimizer.cc
```diff
@@ -2472,10 +2474,27 @@ static bool can_switch_from_ref_to_range(THD *thd, JOIN_TAB *tab,
...
+            (qr->mrr_flags & HA_MRR_CONVERT_REF_TO_RANGE))
```
percona-ysorokin: Avoid implicit conversions
```suggestion
            (qr->mrr_flags & HA_MRR_CONVERT_REF_TO_RANGE) != 0)
```

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -15634,6 +15672,530 @@ void Rdb_compaction_stats::record_end(rocksdb::CompactionJobInfo info) {
...
+  if (keyno == table->s->primary_key) {
+    // We need all ranges to be single-point lookups using full PK values.
+    // (Range scans, like "pk BETWEEN 10 and 20" or restrictions on PK prefix
+    //  cannot be used)
+    bool all_eq_ranges = true;
+    KEY_MULTI_RANGE range{};
+    range_seq_t seq_it;
+    seq_it = seq->init(seq_init_param, n_ranges, *flags);
+    while (!seq->next(seq_it, &range)) {
+      if (!(range.range_flag & UNIQUE_RANGE)) {
```
percona-ysorokin: Avoid implicit conversions
```suggestion
      if ((range.range_flag & UNIQUE_RANGE) ==0) {
```

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -15634,6 +15672,530 @@ void Rdb_compaction_stats::record_end(rocksdb::CompactionJobInfo info) {
...
+    if (all_eq_ranges) {
+      // Indicate that we will use MultiGet MRR
+      *flags &= ~HA_MRR_USE_DEFAULT_IMPL;
+      *flags |= HA_MRR_SUPPORT_SORTED;
+      *bufsz = mrr_get_length_per_rec() * res * 1.1 + 1;
+    }
+  } else {
+    // For scans on secondary keys, we use MultiGet when we read the PK values.
+    // We only need PK values when the scan is non-index-only.
+    if (!(*flags & HA_MRR_INDEX_ONLY)) {
```
percona-ysorokin: Avoid implicit conversions
```suggestion
    if ((*flags & HA_MRR_INDEX_ONLY) == 0) {
```

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -15634,6 +15672,530 @@ void Rdb_compaction_stats::record_end(rocksdb::CompactionJobInfo info) {
...
+      thd->optimizer_switch_flag(OPTIMIZER_SWITCH_MRR) &&
+      !thd->optimizer_switch_flag(OPTIMIZER_SWITCH_MRR_COST_BASED);
+
+  res =
+      handler::multi_range_read_info(keyno, n_ranges, keys, bufsz, flags, cost);
+  if (res || m_lock_rows != RDB_LOCK_NONE || !mrr_enabled) {
+    return res;
+  }
+
+  if (keyno == table->s->primary_key && (*flags & HA_MRR_FULL_EXTENDED_KEYS)) {
```
percona-ysorokin:
```suggestion
  if (keyno == table->s->primary_key && (*flags & HA_MRR_FULL_EXTENDED_KEYS) != 0) {
```

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -15634,6 +15672,530 @@ void Rdb_compaction_stats::record_end(rocksdb::CompactionJobInfo info) {
...
+    *flags &= ~HA_MRR_USE_DEFAULT_IMPL;
+    *flags |= HA_MRR_CONVERT_REF_TO_RANGE;
+    *flags |= HA_MRR_SUPPORT_SORTED;
+  }
+
+  if (keyno != table->s->primary_key && !(*flags & HA_MRR_INDEX_ONLY)) {
```
percona-ysorokin:
```suggestion
  if (keyno != table->s->primary_key && (*flags & HA_MRR_INDEX_ONLY) == 0) {

File: storage/rocksdb/ha_rocksdb.cc [hunk 13]
```diff
@@ -15634,6 +15672,530 @@ void Rdb_compaction_stats::record_end(rocksdb::CompactionJobInfo info) {
...
+    return self->handler::multi_range_read_init(seq, seq_init_param, n_ranges,
+                                                mode, nullptr);
+  }
+
+  int get_next_rowid(uchar *buf, int *size, char **range_ptr) override {
+    if (err) {
+      return err;
+    }
+
+    while (!(err = self->handler::multi_range_read_next(range_ptr))) {
```
percona-ysorokin:
```suggestion
    while ((err = self->handler::multi_range_read_next(range_ptr)) == 0) {
```

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -15634,6 +15672,530 @@ void Rdb_compaction_stats::record_end(rocksdb::CompactionJobInfo info) {
...
+      if (self->mrr_funcs.skip_index_tuple &&
```
percona-ysorokin:
```suggestion
      if (self->mrr_funcs.skip_index_tuple != nullptr &&
```

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -15634,6 +15672,530 @@ void Rdb_compaction_stats::record_end(rocksdb::CompactionJobInfo info) {
...
+          self->mrr_funcs.skip_index_tuple(self->mrr_iter, *range_ptr)) {
+        // BKA's variant of "Index Condition Pushdown" check failed
+        continue;
+      }
+
+      if (self->mrr_funcs.skip_record &&
```
percona-ysorokin:
```suggestion
      if (self->mrr_funcs.skip_record != nullptr &&
```

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -15634,6 +15672,530 @@ void Rdb_compaction_stats::record_end(rocksdb::CompactionJobInfo info) {
...
+};
+
+// Initialize an MRR scan
+int ha_rocksdb::multi_range_read_init(RANGE_SEQ_IF *seq, void *seq_init_param,
+                                      uint n_ranges, uint mode,
+                                      HANDLER_BUFFER *buf) {
+  int res;
+
+  if (!current_thd->optimizer_switch_flag(OPTIMIZER_SWITCH_MRR) ||
+      (mode & HA_MRR_USE_DEFAULT_IMPL) ||
```
percona-ysorokin:
```suggestion
      (mode & HA_MRR_USE_DEFAULT_IMPL) !=0 ||
```

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -15634,6 +15672,530 @@ void Rdb_compaction_stats::record_end(rocksdb::CompactionJobInfo info) {
...
+      (buf->buffer_end - buf->buffer < mrr_get_length_per_rec()) ||
+      (active_index != table->s->primary_key && (mode & HA_MRR_SORTED))) {
```
percona-ysorokin:
```suggestion
      (active_index != table->s->primary_key && (mode & HA_MRR_SORTED) != 0)) {
```

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -15634,6 +15672,530 @@ void Rdb_compaction_stats::record_end(rocksdb::CompactionJobInfo info) {
...
+// returned
+uint ha_rocksdb::mrr_get_length_per_rec() {
+  return sizeof(rocksdb::Slice) + sizeof(rocksdb::Status) +
+         sizeof(rocksdb::PinnableSlice) +
+         sizeof(char *) +  // this for KEY_MULTI_RANGE::ptr
+         m_pk_descr->max_storage_fmt_length();
+}
+
+template <typename T> void align_ptr(char **p) {
+  if (((size_t)p) % alignof(T)) {
```
percona-ysorokin: It looks like there is a bug here we are trying to align pointer to pointer (`p`) instead of pointer (`*p`)
```suggestion
  if ((reinterpret_cast<std::uintptr_t>(*p)) % alignof(T)) {
```

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -15634,6 +15672,530 @@ void Rdb_compaction_stats::record_end(rocksdb::CompactionJobInfo info) {
...
+    *p += alignof(T) - ((size_t)p) % alignof(T);
```
percona-ysorokin: Likewise
```suggestion
    *p += alignof(T) - (reinterpret_cast<std::uintptr_t>(*p)) % alignof(T);
```

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -15634,6 +15672,530 @@ void Rdb_compaction_stats::record_end(rocksdb::CompactionJobInfo info) {
...
+    return HA_ERR_INTERNAL_ERROR;  // error
+  }
+
+  ssize_t elem = 0;
+
+  mrr_n_elements = elem;
+  int key_size;
+  char *range_ptr;
+  int err;
+  while (!(err = mrr_rowid_reader->get_next_rowid((uchar*)buf, &key_size,
```
percona-ysorokin:
```suggestion
  while ((err = mrr_rowid_reader->get_next_rowid((uchar*)buf, &key_size,
                                                  &range_ptr)) == 0) {
```

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -15634,6 +15672,530 @@ void Rdb_compaction_stats::record_end(rocksdb::CompactionJobInfo info) {
...
+
+  Rdb_transaction *tx = get_tx_from_thd(table->in_use);
+  int rc;
+
+  while (true) {
+    while (true) {
+      if (table->in_use->killed) return HA_ERR_QUERY_INTERRUPTED;
+
+      if (mrr_read_index >= mrr_n_elements) {
+        if (mrr_rowid_reader->eof() || !mrr_n_elements) {
```
percona-ysorokin:
```suggestion
       if (mrr_rowid_reader->eof() || !mrr_n_elements !=0) {
```

percona-ysorokin (changes requested):

sql/sql_optimizer.cc
```diff
@@ -2472,10 +2474,27 @@ static bool can_switch_from_ref_to_range(THD *thd, JOIN_TAB *tab,
...
+        If there was a quick select that
+         - was scanning the same range
+         - used MRR
+         - MRR implementation indicated that it's preferred to use MRR over
+           ref(const).
+        then switch to range access as it will read rows using the MRR interface
+      */
+      if (tab->position()->key->key == tab->quick()->index &&
+          length == tab->quick()->max_used_key_length &&
+          tab->quick()->get_type() == QUICK_SELECT_I::QS_TYPE_RANGE) {
```
oleksandr-kachan: Please pay attention to this condition, not sure it would be correct in this form.

percona-ysorokin (approved):
LGTM

hermanlee:
Could you update the description of this PR to indicate which patches from 5.6 are referenced here? Are they:

https://github.com/facebook/mysql-5.6/commit/45110dcda26
https://github.com/facebook/mysql-5.6/commit/be12c8cfb44
https://github.com/facebook/mysql-5.6/commit/d0c98c1e6a5
https://github.com/facebook/mysql-5.6/commit/e99310c1197
https://github.com/facebook/mysql-5.6/commit/888d47893f1
https://github.com/facebook/mysql-5.6/commit/40d6766e57b

Thanks!

Updated: Nevermind, they are referenced, but not in the description of the PR. I'll manually pull them over.

hermanlee:
I'm seeing crashes in a number of tests: main.range_icp main.range_icp_mrr main.range_mrr main.order_by_none main.range_none main.order_by_icp_mrr main.order_by_all

The backtrace for main.range_all shows:

```
21:38:57 UTC - mysqld got signal 11 ;          
Most likely, you have hit a bug, but this error can also be caused by malfunctioning hardware.
Thread pointer: 0x7f0fa2c0b000
Attempting backtrace. You can use the following information to find out
where mysqld died. If you see no messages after this, something went
terribly wrong...             
stack_bottom = 7f0fe0c84f68 thread_stack 0x46000
/data/users/herman/rocks-mysql/8.0/_build-8.0-Debug/runtime_output_directory/mysqld(my_print_stacktrace(unsigned char*, unsigned long)+0x54) [0x64e66f4]
/data/users/herman/rocks-mysql/8.0/_build-8.0-Debug/runtime_output_directory/mysqld(handle_fatal_signal+0x229) [0x5354ff9]
/data/users/herman/rocks-mysql/8.0/_build-8.0-Debug/runtime_output_directory/mysqld() [0x50b3785]
/data/users/herman/rocks-mysql/8.0/_build-8.0-Debug/runtime_output_directory/mysqld() [0x50b10e1]
/data/users/herman/rocks-mysql/8.0/_build-8.0-Debug/runtime_output_directory/mysqld(JOIN::test_skip_sort()+0x426) [0x50ad176]
/data/users/herman/rocks-mysql/8.0/_build-8.0-Debug/runtime_output_directory/mysqld(JOIN::optimize()+0x2a69) [0x50a46e9]
/data/users/herman/rocks-mysql/8.0/_build-8.0-Debug/runtime_output_directory/mysqld(SELECT_LEX::optimize(THD*)+0x193) [0x51681c3]
/data/users/herman/rocks-mysql/8.0/_build-8.0-Debug/runtime_output_directory/mysqld(Sql_cmd_dml::execute_inner(THD*)+0x90) [0x516a430]
/data/users/herman/rocks-mysql/8.0/_build-8.0-Debug/runtime_output_directory/mysqld(Sql_cmd_dml::execute(THD*)+0x59d) [0x5169bfd]
/data/users/herman/rocks-mysql/8.0/_build-8.0-Debug/runtime_output_directory/mysqld(mysql_execute_command(THD*, bool, unsigned long long*)+0x8061) [0x50e6041]
/data/users/herman/rocks-mysql/8.0/_build-8.0-Debug/runtime_output_directory/mysqld(mysql_parse(THD*, Parser_state*, unsigned long long*)+0x69b) [0x50dbc5b]
/data/users/herman/rocks-mysql/8.0/_build-8.0-Debug/runtime_output_directory/mysqld(dispatch_command(THD*, COM_DATA const*, enum_server_command)+0x2043) [0x50d8493]
/data/users/herman/rocks-mysql/8.0/_build-8.0-Debug/runtime_output_directory/mysqld(do_command(THD*)+0x42a) [0x50dadaa]
/data/users/herman/rocks-mysql/8.0/_build-8.0-Debug/runtime_output_directory/mysqld() [0x532d07e]
/data/users/herman/rocks-mysql/8.0/_build-8.0-Debug/runtime_output_directory/mysqld() [0x6d97c6a]

Trying to get some variables.
Some pointers may be invalid and cause the dump to abort.
Query (7f0fa270c028): EXPLAIN SELECT * FROM test_ref WHERE id=3 AND c LIKE 'gh%' ORDER BY c LIMIT 1
```

hermanlee:
I'm seeing a few test failures when this PR is applied on top of the latest tree.

rocksdb.type_text_indexes
rocksdb.type_char_indexes_collation
rocksdb.type_blob_indexes

For example:

```
mysqltest: At line 73: Test assertion failed in assert.inc
In included file ./include/assert.inc: 74
included from ./suite/rocksdb/include/prefix_index_only_query_check.inc: 22
included from ./suite/rocksdb/include/covering_index_query.inc: 25
included from /data/users/herman/rocks-mysql/8.0/mysql-test/suite/rocksdb/t/type_text_indexes.test: 295

The result from queries just before the failure was:
33      yyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyy
34                                        
128     zzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzz
# Eligible for covering index optimization
SELECT id, bigfield FROM t1 FORCE INDEX(fid) WHERE fake_id <= 1032 ORDER BY fake_id;
id      bigfield
0
1       a
8       bbbbbbbb
24      cccccccccccccccccccccccc
30                                    
31      ddddddddddddddddddddddddddddddd
32      xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
include/assert.inc [Eligible for covering index optimization: 7 rocksdb_covered_secondary_key_lookups]
######## Test assertion failed: Eligible for covering index optimization: 7 rocksdb_covered_secondary_key_lookups ########
Dumping debug info:
Assertion text: 'Eligible for covering index optimization: 7 rocksdb_covered_secondary_key_lookups'
Assertion condition: '7 - 1 = 7'
Assertion condition, interpolated: '7 - 1 = 7'
Assertion result: '0'
safe_process[3433523]: Child process: 3433524, exit: 1
```

@oleksandr-kachan Could you take a look into this? Thanks!

oleksandr-kachan:
> I'm seeing a few test failures when this PR is applied on top of the latest tree.
> 
> rocksdb.type_text_indexes
> rocksdb.type_char_indexes_collation
> rocksdb.type_blob_indexes

Dear @hermanlee,
Thank you for letting me know. I'll take a look at this in a few days.

oleksandr-kachan:
Hello @hermanlee,
> rocksdb.type_text_indexes
> rocksdb.type_char_indexes_collation
> rocksdb.type_blob_indexes

As I can see these tests started failing because of the following change which comes from https://github.com/facebook/mysql-5.6/commit/45110dcda26#diff-5a5525d170a985d103b518fa307b17ccR8050-R8077

```
@@ -8447,8 +8473,12 @@ int ha_rocksdb::read_row_from_secondary_key(uchar *const buf,
+  // Due to MRR, now an index-only scan have pushed index condition.
+  // (If it does, we follow non-index only code path here, except that
+  //  we don't fetch the row).
+  bool have_icp = (pushed_idx_cond && pushed_idx_cond_keyno == active_index);
 
-  if (covered_lookup && m_lock_rows == RDB_LOCK_NONE) {
+  if (covered_lookup && m_lock_rows == RDB_LOCK_NONE && !have_icp) {
     pk_size =
     } else {
       rc = kd.unpack_record(table, buf, &rkey, &value,
                            m_converter->get_verify_row_debug_checksums());
       global_stats.covered_secondary_key_lookups.inc();
     }
@@ -8469,7 +8499,9 @@ int ha_rocksdb::read_row_from_secondary_key(uchar *const buf,
       } else {
-        rc = get_row_by_rowid(buf, m_pk_packed_tuple, pk_size);
+        if (!covered_lookup || m_lock_rows != RDB_LOCK_NONE) {
+          rc = get_row_by_rowid(buf, m_pk_packed_tuple, pk_size);
+        }
       }
     }
   }
```


As a result of this change covered_secondary_key_lookups value doesn't match expectations in test.
As I understand this is done intentionally but at the moment I'm not sure if covered_secondary_key_lookups should still be incremented here (in case covered_lookup == true and have_icp == true).

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -8443,7 +8482,9 @@ int ha_rocksdb::read_row_from_secondary_key(uchar *const buf,
       if (pk_size == RDB_INVALID_KEY_LEN) {
         rc = HA_ERR_ROCKSDB_CORRUPT_DATA;
       } else {
-        rc = get_row_by_rowid(buf, m_pk_packed_tuple, pk_size);
+        if (!covered_lookup || m_lock_rows != RDB_LOCK_NONE) {
```
hermanlee, line 8485: @Pushapgl We're trying to resolve this diff with covered index lookups. If we enter this code path, and covered_lookup is true, then it seems like we should also call global_stats.covered_secondary_key_lookups.inc();

This doesn't appear to be the case in 5.6 either. Is this a bug?

hermanlee: Actually, find_icp_matching_index_rec() advances the key until an ICP match is found. Doesn't this mean covered_lookup needs to be recalculated based on the new key?

hermanlee: cc: @spetrunia

spetrunia: > Actually, find_icp_matching_index_rec() advances the key until an ICP match is found. Doesn't this mean covered_lookup needs to be recalculated based on the new key?

Yes, correct. The current code has a bug.
if `have_icp==true`, and we start from a row for which `covered_lookup=true`, and go into `find_icp_matching_index_rec`, which will move to the row which should have `covered_lookup=false`  and return, then we will fail to call get_row_by_rowid(). 

(I'm surprised that ICP is used for such partially-covered keys, though.. I that it wasn't)

spetrunia: ... welll apparently it is used. (and there's nothing wrong with that.. there might be difficulties if the range we are scanning uses the partially-covered column, but it's not the case her)

hermanlee: In the case where `covered_lookup == false` and `have_icp == false`, it seems entering the `else` clause and calling find_icp_matching_index_rec() would be the wrong action. Reading the full row is still needed though.

It seems this function should be re-ordered to account for all of these conditions correctly. Something like:

```
int ha_rocksdb::read_row_from_secondary_key(uchar *const buf,
                                            const Rdb_key_def &kd,
                                            bool move_forward) {
  int rc = 0;
  uint pk_size;

  // Due to MRR, now an index-only scan have pushed index condition.
  // (If it does, we follow non-index only code path here, except that
  //  we don't fetch the row).
  bool have_icp = (pushed_idx_cond && pushed_idx_cond_keyno == active_index);
  if (have_icp) {
    if (kd.m_is_reverse_cf) move_forward = !move_forward;
    rc = find_icp_matching_index_rec(move_forward, buf);
    if (rc) return (rc);
  }

  /* Get the key columns and primary key value */
  const rocksdb::Slice &rkey = m_scan_it->key();
  const rocksdb::Slice &value = m_scan_it->value();

#ifndef DBUG_OFF
  bool save_keyread_only = m_keyread_only;
#endif
  DBUG_EXECUTE_IF("dbug.rocksdb.HA_EXTRA_KEYREAD", { m_keyread_only = true; });

  bool covered_lookup = (m_keyread_only && kd.can_cover_lookup()) ||
                         kd.covers_lookup(&value, &m_lookup_bitmap);

#ifndef DBUG_OFF
  m_keyread_only = save_keyread_only;
#endif

  pk_size =
  } else {
    if (covered_lookup && m_lock_rows == RDB_LOCK_NONE) {
      rc = kd.unpack_record(table, buf, &rkey, &value,
                            m_converter->get_verify_row_debug_checksums());
      global_stats.covered_secondary_key_lookups.inc();
    } else {
      rc = get_row_by_rowid(buf, m_pk_packed_tuple, pk_size);
    }
  }

  if (!rc) {
    m_last_rowkey.copy((const char *)m_pk_packed_tuple, pk_size,
                       &my_charset_bin);
  }

  return rc;
}
```

@spetrunia , For the MRR case, the comment is that if `have_icp == true`, then we want to skip fetching the row. But the original code doesn't check for have_icp before fetching the row. If `have_icp == true`, can we just call kd.unpack_record() and change the conditional from `if (covered_lookup && m_lock_rows == RDB_LOCK_NONE)` to `if ((covered_lookup || have_icp) && m_lock_rows == RDB_LOCK_NONE)`.

hermanlee: I just tested this. Skipping `get_row_by_rowid()` based on `have_icp == true` will return the wrong data.

PR #1140, opened 2020-06-04
PS-7027: MyRocks: Implement optimized support for utf8mb4_bin and other binary collations (5.6)

inikep:
https://jira.percona.com/browse/PS-7027

This patch adds optimized support for `utf8mb4_bin` and all collations based on `my_collation_8bit_bin_handler` (i.e. `latinX_bin`, `cp12XX_bin`, `cp85X_bin`, `armscii8_bin`, `ascii_bin` and many more).

It also introduces the `rocksdb.type_char_secondary_index_only_scan` test that:
1. Finds all collations that support index-only scans with secondary keys using the "Extra" field from "EXPLAIN SELECT"
2. Verifies if the original string is reconstructed correctly from SK (or from PK when index-only scans with SK are not supported)

### DIFF
--- a/mysql-test/suite/rocksdb/t/type_char_secondary_index_only_scan.test
+++ b/mysql-test/suite/rocksdb/t/type_char_secondary_index_only_scan.test
@@ -1,0 +1,50 @@
+# 1. Find all collations that support index-only scans with secondary
+#    keys using the "Extra" field from "EXPLAIN SELECT"
+# 2. Verify if the original string is reconstructed correctly from SK
+#    (or from PK when index-only scans with SK are not supported)
+
+--source include/have_rocksdb.inc
+
+CREATE TABLE collations AS SELECT * FROM information_schema.COLLATION_CHARACTER_SET_APPLICABILITY WHERE COLLATION_NAME NOT IN ("latin2_czech_cs", "cp1250_czech_cs") ORDER BY COLLATION_NAME;
+
+--disable_query_log
+--let $i=1
+--let $charset = query_get_value(select * from collations, CHARACTER_SET_NAME, $i)
+--let $collate = query_get_value(select * from collations, COLLATION_NAME, $i)
+--echo CharSet, Collate, Char Extra, VarChar Extra
+while ($charset != 'No such row')
+{
+  --let table_name=t_$charset
+  if ($charset == "binary")
+  {
+    --eval CREATE TABLE $table_name (id INT NOT NULL AUTO_INCREMENT PRIMARY KEY, vch VARBINARY(64), ch BINARY(64), KEY vch_idx (vch), KEY ch_idx (ch)) ENGINE=rocksdb
+  }
+  if ($charset != "binary")
+  {
+    --eval CREATE TABLE $table_name (id INT NOT NULL AUTO_INCREMENT PRIMARY KEY, vch VARCHAR(64), ch CHAR(64), KEY vch_idx (vch), KEY ch_idx (ch)) ENGINE=rocksdb CHARSET $charset COLLATE $collate
+  }
+  --eval INSERT INTO $table_name (vch, ch) VALUES('VARCHAR', 'char')
+  --let $vch_extra = query_get_value(EXPLAIN SELECT vch FROM $table_name, Extra, 1)
+  --let $ch_extra = query_get_value(EXPLAIN SELECT ch FROM $table_name, Extra, 1)
+  --echo $charset, $collate, $ch_extra, $vch_extra
+  --let $ch_value = query_get_value(SELECT ch FROM $table_name, ch, 1)
+  --let $vch_value = query_get_value(SELECT vch FROM $table_name, vch, 1)
+  if ($ch_value != "char")
+  {
+    --echo ERROR for $charset, $collate: \$ch_value=$ch_value, \$vch_value=$vch_value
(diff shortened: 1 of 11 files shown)


mysql-test/suite/rocksdb/t/type_char_secondary_index_only_scan.test
```diff
@@ -0,0 +1,50 @@
+# 1. Find all collations that support index-only scans with secondary
+#    keys using the "Extra" field from "EXPLAIN SELECT"
+# 2. Verify if the original string is reconstructed correctly from SK
+#    (or from PK when index-only scans with SK are not supported)
```
Pushapgl, line 4: I didn't see accessing the PK when SK is not supported. In order to guarantee the correctness of unpack_info, we should ideally rely of PK rather than SK because SK can have weirdness of covering and non covering.

mysql-test/suite/rocksdb/t/type_char_secondary_index_only_scan.test
```diff
@@ -0,0 +1,50 @@
...
+  if ($charset == "binary")
+  {
+    --eval CREATE TABLE $table_name (id INT NOT NULL AUTO_INCREMENT PRIMARY KEY, vch VARBINARY(64), ch BINARY(64), KEY vch_idx (vch), KEY ch_idx (ch)) ENGINE=rocksdb
+  }
+  if ($charset != "binary")
+  {
+    --eval CREATE TABLE $table_name (id INT NOT NULL AUTO_INCREMENT PRIMARY KEY, vch VARCHAR(64), ch CHAR(64), KEY vch_idx (vch), KEY ch_idx (ch)) ENGINE=rocksdb CHARSET $charset COLLATE $collate
+  }
+  --eval INSERT INTO $table_name (vch, ch) VALUES('VARCHAR', 'char')
+  --let $vch_extra = query_get_value(EXPLAIN SELECT vch FROM $table_name, Extra, 1)
```
Pushapgl, line 27: $table_name has a single record and we are checking if the optimizer has picked index or not. This seems very flaky test case as we can't bet on optimizer to always pick index for signal record.

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -7046,16 +7046,19 @@ int ha_rocksdb::rdb_error_to_mysql(const rocksdb::Status &s,
...
     const my_core::Field *const field) {
   const my_core::enum_field_types type = field->real_type();
   /* Handle [VAR](CHAR|BINARY) or TEXT|BLOB */
   if (type == MYSQL_TYPE_VARCHAR || type == MYSQL_TYPE_STRING ||
       type == MYSQL_TYPE_BLOB) {
-    return RDB_INDEX_COLLATIONS.find(field->charset()) !=
-           RDB_INDEX_COLLATIONS.end();
+    return rdb_is_simple_collation(field->charset()) ||
+           rdb_is_binary_collation(field->charset()) ||
+           RDB_INDEX_COLLATIONS.find(field->charset()) !=
```
Pushapgl, line 7060: Nit - rdb_is_binary_collation seems like a super set of RDB_INDEX_COLLATIONS, isn't thats the case?

storage/rocksdb/rdb_datadic.cc
```diff
@@ -2117,10 +2118,11 @@ int Rdb_key_def::unpack_utf8_str(
   uchar *const dst_end = dst + field->pack_length();
 
   while (src < src_end) {
-    my_wc_t wc = (src[0] << 8) | src[1];
-    src += 2;
+    my_wc_t wc = (bytes == 3) ? (src[0] << 16) | (src[1] << 8) | src[2]
```
Pushapgl, line 2121: Nit - This is repeated at 3 places, should we expose a small helper method around this.

storage/rocksdb/ha_rocksdb.cc
inikep: `rdb_is_binary_collation` doesn't cover `BINARY(N)` that is `my_charset_bin` (included in `RDB_INDEX_COLLATIONS`).

storage/rocksdb/rdb_datadic.cc
inikep: It can be done but template instantiation should create separate methods for `bytes == 3` and `bytes != 3`.

mysql-test/suite/rocksdb/t/type_char_secondary_index_only_scan.test
inikep: When `EXPLAIN SELECT` returns `Using index` it means that SK was used.
When `EXPLAIN SELECT` returns `NULL` we know that PK was used because MyRocks switches to PK if SK can't be used.

mysql-test/suite/rocksdb/t/type_char_secondary_index_only_scan.test
inikep: This test tries to verify if data is reconstructed correctly from:
1. SK using `unpack_info` or mem-comparable form (`Using index`)
2. PK when index-only scans with SK are not supported (`EXPLAIN SELECT` returns `NULL`).

The server is deterministic so it should always return the same results unless optimizer will be changed in the future, but I assume we would like to know about it.

PR #1142, opened 2020-06-29
PS-7027 MyRocks: Implement optimized support for collations other than latin1/utf8 (8.0)

inikep:
1. Fix `EXPLAIN SELECT` that always showed `Using index` for RocksDB SKs in 8.0.
    
2. Add `rocksdb.type_char_index_ver_bump_verify` to check backward compatibility of index-only scans with SK after index ver bumping.

3. Fix support of NO_PAD collations i.e. "utf8mb4*_0900_*"

4. Add `rocksdb.type_varchar_packing` to validate padding of `CHAR` and `VARCHAR` collations. Only multi-level collations (`utf8mb4_*_0900_as_cs`, `utf8mb4_0900_as_ci`, `latin2_czech_cs` and `cp1250_czech_cs`) are not verified.

5. Add optimized support for `utf8mb4_bin` and all collations based on `my_collation_8bit_bin_handler` (i.e. `latinX_bin`, `cp12XX_bin`, `cp85X_bin`, `armscii8_bin`, `ascii_bin` and many more). Index-only scans with SK are supported by decoding the memcomparable form back into the original form.

6. Introduce the `rocksdb.type_char_secondary_index_only_scan` test that:
    a) Finds all collations that support index-only scans with secondary keys using the "Extra" field from "EXPLAIN SELECT"
    b) Verifies if the original string is reconstructed correctly from SK (or from PK when index-only scans with SK are not supported)

7. Turn on support for all remaining collations by storing entire copy of original field in the value. The downside is that extra storage is used for the index.

8. Add backward compatibility (with `SECONDARY_FORMAT_VERSION_UPDATE5`) after introducing a support for index-only scans with SK for all previously-unsupported collations by storing entire copy of original field in `unpack_info`.


sql/table.cc
```diff
@@ -751,7 +751,9 @@ void setup_key_part_field(TABLE_SHARE *share, handler *handler_file,
   } else if (part_of_key_not_extended) {
     field->part_of_prefixkey.set_bit(key_n);
   }
-  if ((handler_file->index_flags(key_n, key_part_n, 0) & HA_KEYREAD_ONLY) &&
+  /* call index_flags() only after all key parts were processed */
+  if ((key_part_n == keyinfo->actual_key_parts - 1) &&
```
hermanlee: I'm seeing some test failures that look like the wrong result based on this change.

```
CURRENT_TEST: rocksdb.type_varchar
--- /mnt/btrfs/trunk-git-mysql-111611707-1596472507/mysql-test/suite/rocksdb/r/type_varchar.result	2020-08-03 19:35:41.696722657 +0300
+++ /mnt/btrfs/trunk-git-mysql-111611707-1596472507/_build-8.0-Release/mysql-test/var/16/log/type_varchar.reject	2020-08-03 20:20:57.945493924 +0300
@@ -444,28 +444,24 @@
 explain
 select col1, hex(col1) from t1 force index(col1);
 id	select_type	table	partitions	type	possible_keys	key	key_len	ref	rows	filtered	Extra
-1	SIMPLE	t1	NULL	index	NULL	col1	131	NULL	#	#	Using index
+1	SIMPLE	t1	NULL	ALL	NULL	NULL	NULL	NULL	#	#	NULL
```

I thought there is a comment in ha_rocksdb::index_flags on the function being called by the SQL layer before the tables are necessarily fully setup. However, it seems like changing the code here leads to other failures.

Do you have the test case for 8.0 Sk always showing using index when it should not be? Is there a better fix that can be made inside of ha_rocksdb::index_flags() instead?

inikep: To check that 8.0 Sk always shows `using index` please just import: 
```
	new file:   mysql-test/suite/rocksdb/r/type_char_secondary_index_only_scan.result
	new file:   mysql-test/suite/rocksdb/t/type_char_secondary_index_only_scan.test
```
from https://github.com/facebook/mysql-5.6/commit/fb4549660c

inikep: I will work on moving a fix from `sql/table.cc` to `ha_rocksdb::index_flags`.
BTW, do you run MTR tests with ` --character-set-server` and `--collation-server`?
I see a lot of issues in `rocksdb.type_*` tests:
```
CURRENT_TEST: rocksdb.type_varchar
--- /data/mysql-server/fb-8.0.17/mysql-test/suite/rocksdb/r/type_varchar.result	2020-07-29 16:22:29.948380928 +0300
+++ /data/mysql-server/fb-8.0.17-deb-clang9/mysql-test/var/log/type_varchar.reject	2020-08-10 15:26:19.743306454 +0300
@@ -16,16 +16,22 @@
 v65000 VARCHAR(65000) ,
 PRIMARY KEY (v64)
 ) ENGINE=rocksdb;
+Warnings:
+Note	1246	Converting column 'v65000' from VARCHAR to TEXT
 SHOW COLUMNS IN t1;
 Field	Type	Null	Key	Default	Extra
 v0	varchar(0)	YES		NULL	
 v1	varchar(1)	YES		NULL	
 v64	varchar(64)	NO	PRI	NULL	
-v65000	varchar(65000)	YES		NULL	
-CREATE TABLE t2 (v VARCHAR(65532), PRIMARY KEY (v(255))) ENGINE=rocksdb;
+v65000	mediumtext	YES		NULL	
+CREATE TABLE t2 (v VARCHAR(65532), PRIMARY KEY (v(255))) ENGINE=rocksdb CHARSET utf8 COLLATE utf8_general_ci;
+Warnings:
+Warning	3719	'utf8' is currently an alias for the character set UTF8MB3, but will be an alias for UTF8MB4 in a future release. Please consider using UTF8MB4 in order to be unambiguous.
+Warning	3778	'utf8_general_ci' is a collation of the deprecated character set UTF8MB3. Please consider using UTF8MB4 with an appropriate collation instead.
+Note	1246	Converting column 'v' from VARCHAR to TEXT
 SHOW COLUMNS IN t2;
 Field	Type	Null	Key	Default	Extra
-v	varchar(65532)	NO	PRI	NULL	
+v	mediumtext	NO	PRI	NULL	
```

hermanlee: It's not clear to me if the following query plan using filesort is problematic for `select id from t3 order by vch;`, vs the current solution (use index for ordering and fetch the primary key on each row when needed).

```
create table t3 (id int NOT NULL AUTO_INCREMENT PRIMARY KEY, vch VARCHAR(64), ch CHAR(64), KEY vch_idx(vch), key ch_idx(ch)) ENGINE=rocksdb CHARSET utf8mb4 collate=utf8mb4_croatian_ci;

mysql> mysql> explain select id from t3 order by vch;
+----+-------------+-------+------------+------+---------------+------+---------+------+------+----------+----------------+
| id | select_type | table | partitions | type | possible_keys | key  | key_len | ref  | rows | filtered | Extra          |
+----+-------------+-------+------------+------+---------------+------+---------+------+------+----------+----------------+
|  1 | SIMPLE      | t3    | NULL       | ALL  | NULL          | NULL | NULL    | NULL |   15 |   100.00 | Using filesort |
+----+-------------+-------+------------+------+---------------+------+---------+------+------+----------+----------------+
1 row in set, 1 warning (0.00 sec)
```

However, bd907766e basically addresses the issue in the general case and provides the fix needed so that Using index is accurate. Given this, is a targeted fix in ha_rocksdb.cc still needed?

Since this change in sql/table.cc is causing other potential issues, I'm considering dropping it.

And yes, we do have different settings for the rocksdb suite to maintain the backward compatibility with 5.6 test results. 3976b20b5c introduced the settings in mysql-test/suite/rocksdb/my.cnf.

inikep:
I updated this PR:
1. Rebased to the current trunk.
2. Removed changes from `sql/table.cc`.

I tested `rocksdb.type_varchar` and your example and they seems to work fine:
```
create table t3 (id int NOT NULL AUTO_INCREMENT PRIMARY KEY, vch VARCHAR(64), ch CHAR(64), KEY vch_idx(vch), key ch_idx(ch)) ENGINE=rocksdb CHARSET utf8mb4 collate=utf8mb4_croatian_ci;
explain select id from t3 order by vch;
id	select_type	table	partitions	type	possible_keys	key	key_len	ref	rows	filtered	Extra
1	SIMPLE	t3	NULL	index	NULL	vch_idx	259	NULL	10	100.00	Using index
Warnings:
Note	1003	/* select#1 */ select `test`.`t3`.`id` AS `id` from `test`.`t3` order by `test`.`t3`.`vch`
```

storage/rocksdb/rdb_datadic.cc
```diff
@@ -2964,34 +2967,56 @@ bool Rdb_key_def::is_variable_length_field(const enum_field_types type) {
...
  */
-static int unpack_charset(
-    const CHARSET_INFO *cset,  // character set information
-    const uchar *src,          // source data to unpack
-    uint src_len,              // length of source data
-    uchar *dst,                // destination of unpacked data
-    uint dst_len,              // length of destination data
-    uint *used_bytes)          // output number of bytes used
+template <const int bytes>
+int unpack_charset(const CHARSET_INFO *cset,  // character set information
```
hermanlee: It seems like this template is unused at the moment since it's meant for no_pad 2/3 byte characters. Was the intention to keep unpack_binary_varlength as a standalone function, rather than templatize it like unpack_binary_or_utf8_varlength_space_pad? Is this function being kept to support possibly utf8mb4_0900_bin collations in the future?

inikep: I didn't notice that `unpack_charset` is now unused. I removed it.

PR #1144, opened 2020-07-27
inplace drop partition

luqun:
This change is to enable inplace drop partition support for myrocks.  Major code change is to iterator all partitions and find which partition need to be dropped, call SE(myrocks) API  to drop that table with constructed partition name.

This doesn't solve DDL crash safe issue.

also steal np_inplace_vs_copy testcase from George


hermanlee:
cc: @georgelorchpercona @satya-bodapati

sql/partitioning/partition_base.cc
```diff
@@ -4046,6 +4049,31 @@ bool Partition_base::commit_inplace_alter_table(
   if (ha_alter_info->alter_info->flags == Alter_info::ALTER_PARTITION)
     DBUG_RETURN(false);
 
+  std::string full_table_path = std::string("./") + altered_table->s->db.str +
+                                std::string("/") +
+                                altered_table->s->table_name.str;
```
satya-bodapati: this looks problematic. Better use tablename_to_filename()

sql/partitioning/partition_base.cc
```diff
@@ -4046,6 +4049,31 @@ bool Partition_base::commit_inplace_alter_table(
...
+      partition_state state = part_elem->part_state;
+      switch (state) {
+        case PART_TO_BE_DROPPED:
+          char name[FN_REFLEN];
+          // TODO(luqun): add subpartition support
+          create_partition_name(name, full_table_path.c_str(),
+                                part_elem->partition_name, false);
+          (*file)->ha_external_lock(get_thd(), F_UNLCK);
+          (*file)->ha_delete_table(name, old_table_def);
+          break;
```
satya-bodapati: how do we handle if one partition couldn't be deleted/dropped? I guess ALTER guarantees no concurrent DMLS.

sql/partitioning/partition_base.cc
```diff
@@ -4046,6 +4049,31 @@ bool Partition_base::commit_inplace_alter_table(
...
+    while ((part_elem = part_it++) != nullptr) {
+      partition_state state = part_elem->part_state;
+      switch (state) {
+        case PART_TO_BE_DROPPED:
+          char name[FN_REFLEN];
+          // TODO(luqun): add subpartition support
+          create_partition_name(name, full_table_path.c_str(),
+                                part_elem->partition_name, false);
+          (*file)->ha_external_lock(get_thd(), F_UNLCK);
+          (*file)->ha_delete_table(name, old_table_def);
```
satya-bodapati: Please check if concurrent SELECTs can be done.  Does the delete_table() implementation in myrocks gurantee to wait until pending IO is completed?

satya-bodapati (changes requested):

sql/partitioning/partition_base.cc
```diff
@@ -4046,6 +4049,31 @@ bool Partition_base::commit_inplace_alter_table(
...
+    handler **file = m_file;
+    partition_info *part_info = ha_alter_info->modified_part_info;
+    List_iterator_fast<partition_element> part_it(part_info->partitions);
+    partition_element *part_elem;
+    while ((part_elem = part_it++) != nullptr) {
+      partition_state state = part_elem->part_state;
+      switch (state) {
+        case PART_TO_BE_DROPPED:
+          char name[FN_REFLEN];
+          // TODO(luqun): add subpartition support
```
satya-bodapati: Time to add the support for sub partition as well?

sql/partitioning/partition_base.cc
```diff
@@ -4046,6 +4049,31 @@ bool Partition_base::commit_inplace_alter_table(
...
+          create_partition_name(name, full_table_path.c_str(),
+                                part_elem->partition_name, false);
+          (*file)->ha_external_lock(get_thd(), F_UNLCK);
```
satya-bodapati: why is the unlock necessary here?

satya-bodapati (changes requested):

sql/partitioning/partition_base.cc
luqun: Yes, SQL layer will acquire exclusive lock and clean table cache before commit

https://github.com/facebook/mysql-5.6/blob/5dd1adebbcc81a8db2c8c30eb20c63d484ee68de/sql/sql_table.cc#L12648

sql/partitioning/partition_base.cc
luqun: handler::ha_delete_table requires table/partition is unlocked..
https://github.com/facebook/mysql-5.6/blob/fb-mysql-8.0.17/sql/handler.cc#L4962

also table/partition is locked in 
https://github.com/facebook/mysql-5.6/blob/fb-mysql-8.0.17/sql/sql_table.cc#L12516

sql/partitioning/partition_base.cc
luqun: Good point.. Use table->s->normalized_path.str instead.

sql/partitioning/partition_base.cc
luqun: Done

sql/partitioning/partition_base.cc
luqun: myrocks is "append only" database and there is no background threads to service IO request. AFAIK, it doesn't need to wait for pending IO.

luqun: > Please check if concurrent SELECTs can be done. 

there are two small time windows drop partition will acquire/hold exclusive table lock, During these two time windows, SELECT will wait for acquire table lock if it doesn't hold table lock yet.

luqun:
@georgelorchpercona @satya-bodapati Thanks a lot for your comments, would you take a look again?

george-lorch:
Sure @luqun I will pull it, take a look, and test it again.

PR #1145, opened 2020-07-28
inplace add partition

luqun:
This change is to enable inplace add partition(range/list) support for myrocks. Major code change is to iterator all partitions and find which partition need to be created, call SE(myrocks) API to create that table with constructed partition name. 

This doesn't solve DDL crash safe issue.

These code are prototypes and need be refined..


luqun:
cc: @georgelorchpercona @satya-bodapati

agniutkarsh (approved):

PR #1158, opened 2021-02-17
Fix mysql range query optimizer bug in key_or

yizhang82:
For range query case with multiple OR, optimizer rely on tree_or and key_or to 'merge' the OR sub-tree together.
And in the case of having multiple ranges that are the same, key_or would see that both ranges key1 and key2 are exactly the same, release key2->next_key_part, and then OR key1->next_key_part and key2->next_key_part= NULL (since it is released) with key_or. However, because NULL sub-tree are treat as TRUE (same as is_always), the entirety of key1->next_key_part gets dropped, so you end up with just the first key.

It makes sense for NULL sub-tree to be treat as TRUE for key_or, because for example (A > 1 AND NULL) OR (A > 1 AND B > 1 AND NULL) should be merged as (A > 1 AND NULL), so it makes sense for NULL to be treated as TRUE so that key_or(NULL, key) should come back as NULL. So the right way to fix this is to stop handling current range and simply move to the next range (without falling through to the remaining logic).

Because of this bug the range query plan end up using much less keys and end up being more expensive, so usually a ref plan gets picked with less keys, and the query becomes much more expensive.

A small repro:

```
CREATE TABLE test_or (
  a bigint unsigned NOT NULL,
  b bigint unsigned NOT NULL,
  c bigint unsigned NOT NULL,
  d VARCHAR(10), PRIMARY KEY (a, b, c))
```

Before:
```
mysql> explain select * from test_or FORCE INDEX (PRIMARY) where (a=1 and b=1 and c=1) OR (a=1 and B=1 and c=1) OR (a=1 and b=2 and c=2);
+----+-------------+---------+------------+------+---------------+---------+---------+-------+------+----------+-------------+
| id | select_type | table   | partitions | type | possible_keys | key     | key_len | ref   | rows | filtered | Extra       |
+----+-------------+---------+------------+------+---------------+---------+---------+-------+------+----------+-------------+
|  1 | SIMPLE      | test_or | NULL       | ref  | PRIMARY       | PRIMARY | 8       | const |    1 |    33.33 | Using where |
+----+-------------+---------+------------+------+---------------+---------+---------+-------+------+----------+-------------+
1 row in set, 1 warning (0.00 sec)
```

After:
```
mysql> explain select * from test_or FORCE INDEX (PRIMARY) where (a=1 and b=1 and c=1) OR (a=1 and B=1 and c=1) OR (a=1 and b=2 and c=2);
+----+-------------+---------+------------+-------+---------------+---------+---------+------+------+----------+-------------+
| id | select_type | table   | partitions | type  | possible_keys | key     | key_len | ref  | rows | filtered | Extra       |
+----+-------------+---------+------------+-------+---------------+---------+---------+------+------+----------+-------------+
|  1 | SIMPLE      | test_or | NULL       | range | PRIMARY       | PRIMARY | 24      | NULL |    2 |   100.00 | Using where |
+----+-------------+---------+------------+-------+---------------+---------+---------+------+------+----------+-------------+
1 row in set, 1 warning (0.00 sec)
```

### DIFF
--- a/mysql-test/include/range.inc
+++ b/mysql-test/include/range.inc
@@ -2841,3 +2841,25 @@ eval EXPLAIN $query2;
 eval $query2;
 
 DROP TABLE test_ref;
+
+--echo #
+--echo # BUG: Range optimizer loses keys when ORing identical ranges
+--echo #
+
+CREATE TABLE test_or (
+  a bigint unsigned NOT NULL,
+  b bigint unsigned NOT NULL,
+  c bigint unsigned NOT NULL,
+  d VARCHAR(10), PRIMARY KEY (a, b, c));
+
+INSERT INTO test_or VALUES (1, 1, 1, 'a'), (2, 2, 2, 'b'), (3, 3, 3, 'c');
+
+# This gives us range plan
+EXPLAIN SELECT * FROM test_or FORCE INDEX (PRIMARY) WHERE
+(a=1 and b=1 and c=1) OR (a=1 and b=2 and c=2);
+
+# Bug: this falls back to a ref plan
+EXPLAIN SELECT * FROM test_or FORCE INDEX (PRIMARY) WHERE
+(a=1 and b=1 and c=1) OR (a=1 and b=1 and c=1) OR (a=1 and b=2 and c=2);
+
+DROP TABLE test_or;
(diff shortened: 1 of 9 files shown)


yizhang82:
@spetrunia We recently found this bug in 8.0 optimizer and it would be great if you can take a quick look at the fix. I've filed upstream bug https://bugs.mysql.com/bug.php?id=102634.

spetrunia:
Ok, looking

spetrunia:
Looks good to me

spetrunia (approved):

yizhang82:
@spetrunia thanks!

PR #1165, opened 2021-05-17
Range Locking - rebased on latest fb-mysql-8.0.20

spetrunia:
This includes:
1. Iterator patches, re-based on top of the latest fb-mysql-8.0.20 tree 
2. Range Locking patch

### DIFF
--- a/storage/rocksdb/ha_rocksdb.cc
+++ b/storage/rocksdb/ha_rocksdb.cc
@@ -12507,7 +12507,7 @@ int ha_rocksdb::external_lock(THD *const thd, int lock_type) {
       }
     }
     tx->m_n_mysql_tables_in_use++;
-    rocksdb_register_tx(rocksdb_hton, thd, tx, (lock_type == F_WRLCK));
+    rocksdb_register_tx(rocksdb_hton, thd, tx, (m_lock_rows != RDB_LOCK_NONE));
     tx->io_perf_start(&m_io_perf);
 
     m_use_range_locking= false;
@@ -12542,7 +12542,7 @@ int ha_rocksdb::start_stmt(THD *const thd,
 
   Rdb_transaction *const tx = get_or_create_tx(thd);
   read_thd_vars(thd);
-  rocksdb_register_tx(ht, thd, tx, (lock_type == F_WRLCK));
+  rocksdb_register_tx(ht, thd, tx, (m_lock_rows != RDB_LOCK_NONE));
   tx->io_perf_start(&m_io_perf);
 
   DBUG_RETURN(HA_EXIT_SUCCESS);
--- a/storage/rocksdb/ha_rocksdb.cc
+++ b/storage/rocksdb/ha_rocksdb.cc
@@ -9590,40 +9590,29 @@ int ha_rocksdb::set_range_lock(Rdb_transaction *tx,
     Figure out the right endpoint.
   */
 
-  if (find_flag == HA_READ_KEY_EXACT) {
+  if (find_flag == HA_READ_KEY_EXACT ||
+      find_flag == HA_READ_PREFIX_LAST) {
     if (slice.size() == Rdb_key_def::INDEX_NUMBER_SIZE) {
       // This is a full table/index scan
+      // (in case of HA_READ_PREFIX_LAST, a reverse-ordered one)
       start_has_inf_suffix= false;
       big_range = true;
     } else {
       /*
-        This is "key_part= const" interval. We need to lock this range:
-        (lookup_value, -inf) < key < (lookup_value, +inf)
+        HA_READ_KEY_EXACT: 
+          This is "key_part= const" interval. We need to lock this range:
+          (lookup_value, -inf) < key < (lookup_value, +inf)
+        HA_READ_PREFIX_LAST:
+          We get here for queries like:
+
+            select * from t1 where pk1=const order by pk1 desc for update
+
+          assuming this uses an index on (pk1, ...).
+          We get end_key=nullptr.
       */
       start_has_inf_suffix= false;
       end_has_inf_suffix= true;
       end_slice= slice;
-    }
-  }
-  else if (find_flag == HA_READ_PREFIX_LAST) {
-    if (slice.size() == Rdb_key_def::INDEX_NUMBER_SIZE) {
-      /* Reverse-ordered full index scan */
-      start_has_inf_suffix= true;
-      big_range = true;
-    } else {
-      /*
-         We get here for queries like:
-
-           select * from t1 where pk1=const order by pk1 desc for update
-
-         assuming this uses an index on (pk1, ...)
-         We get end_key=nullptr.
-
-         The range to lock is the same as with HA_READ_KEY_EXACT above.
-      */
-      end_slice= slice;
-      start_has_inf_suffix= false;
-      end_has_inf_suffix= true;
     }
   }
   else if (find_flag == HA_READ_PREFIX_LAST_OR_PREV) {
@@ -9689,8 +9678,11 @@ int ha_rocksdb::set_range_lock(Rdb_transaction *tx,
       start_has_inf_suffix= false;
     else if (find_flag == HA_READ_AFTER_KEY)
       start_has_inf_suffix= true;
-    else
+    else {
+      // Unknown type of range, shouldn't happen
       DBUG_ASSERT(0);
+      big_range = true;
+    }
 
     // Known end range bounds: HA_READ_AFTER_KEY, HA_READ_BEFORE_KEY
     if (end_key->flag == HA_READ_AFTER_KEY) {
@@ -9699,8 +9691,11 @@ int ha_rocksdb::set_range_lock(Rdb_transaction *tx,
     } else if (end_key->flag == HA_READ_BEFORE_KEY) {
       // this is "key_part < const", non-inclusive.
       end_has_inf_suffix= false;
-    } else
+    } else {
+      // Unknown type of range, shouldn't happen
       DBUG_ASSERT(0);
+      big_range = true;
+    }
 
     uchar pack_buffer[MAX_KEY_LENGTH];
     uint end_slice_size= kd.pack_index_tuple(table, pack_buffer, end_slice_buf,
--- a/storage/rocksdb/ha_rocksdb.cc
+++ b/storage/rocksdb/ha_rocksdb.cc
@@ -7668,6 +7668,7 @@ ha_rocksdb::ha_rocksdb(my_core::handlerton *const hton,
       m_sk_packed_tuple_old(nullptr),
       m_pack_buffer(nullptr),
       m_lock_rows(RDB_LOCK_NONE),
+      m_use_range_locking(false),
       m_keyread_only(false),
       m_iteration_only(false),
       m_insert_with_update(false),
@@ -7990,6 +7991,7 @@ int ha_rocksdb::open(const char *const name,
   }
 
   m_lock_rows = RDB_LOCK_NONE;
+  m_use_range_locking = false;
   m_locked_row_action = THR_WAIT;
   m_key_descr_arr = m_tbl_def->m_key_descr_arr;
 
--- a/storage/rocksdb/ha_rocksdb.cc
+++ b/storage/rocksdb/ha_rocksdb.cc
@@ -9611,7 +9611,7 @@ int ha_rocksdb::set_range_lock(Rdb_transaction *tx,
   rocksdb::Slice slice(slice_arg);
   *use_locking_iterator= false;
 
-  if (m_lock_rows == RDB_LOCK_NONE || !rocksdb_use_range_locking) {
+  if (!m_use_range_locking) {
     return 0;
   }
   bool big_range= false;
@@ -10322,7 +10322,7 @@ int ha_rocksdb::index_next_with_direction_intern(uchar *const buf,
       table->m_status = 0;
       rc = 0;
     } else if (active_index == table->s->primary_key) {
-      if (m_lock_rows != RDB_LOCK_NONE && !rocksdb_use_range_locking) {
+      if (m_lock_rows != RDB_LOCK_NONE && !m_use_range_locking) {
         DEBUG_SYNC(ha_thd(), "rocksdb_concurrent_delete");
         /* We need to put a lock and re-read */
         bool skip_row = false;
@@ -11424,7 +11424,7 @@ int ha_rocksdb::update_write_sk(const TABLE *const table_arg,
         reinterpret_cast<const char *>(m_sk_packed_tuple_old), old_packed_size);
 
     /* Range locking: lock the index tuple being deleted */
-    if (rocksdb_use_range_locking) {
+    if (m_use_range_locking) {
       auto s= row_info.tx->lock_singlepoint_range(kd.get_cf(), old_key_slice);
       if (!s.ok()) {
         return (row_info.tx->set_status_error(table->in_use, s, kd,
@@ -11478,7 +11478,7 @@ int ha_rocksdb::update_write_sk(const TABLE *const table_arg,
     rc = bulk_load_key(row_info.tx, kd, new_key_slice, new_value_slice, true);
   } else {
     /* Range locking: lock the index tuple being inserted */
-    if (rocksdb_use_range_locking) {
+    if (m_use_range_locking) {
       auto s= row_info.tx->lock_singlepoint_range(kd.get_cf(), new_key_slice);
       if (!s.ok()) {
         return (row_info.tx->set_status_error(table->in_use, s, kd,
@@ -11921,7 +11921,7 @@ int ha_rocksdb::delete_row(const uchar *const buf) {
         For point locking, Deleting on secondary key doesn't need any locks.
         Range locking must get a lock.
       */
-      if (rocksdb_use_range_locking) {
+      if (m_use_range_locking) {
         auto s= tx->lock_singlepoint_range(kd.get_cf(), secondary_key_slice);
         if (!s.ok()) {
           DBUG_RETURN(tx->set_status_error(table->in_use, s, kd, m_tbl_def,
@@ -12504,6 +12504,13 @@ int ha_rocksdb::external_lock(THD *const thd, int lock_type) {
     tx->m_n_mysql_tables_in_use++;
     rocksdb_register_tx(rocksdb_hton, thd, tx, (lock_type == F_WRLCK));
     tx->io_perf_start(&m_io_perf);
+
+    m_use_range_locking= false;
+    if (rocksdb_use_range_locking &&
+        m_lock_rows != RDB_LOCK_NONE &&
+        my_core::thd_tx_isolation(thd) >= ISO_REPEATABLE_READ) {
+      m_use_range_locking= true;
+    }
   }
 
   DBUG_RETURN(res);
--- a/storage/rocksdb/ha_rocksdb.cc
+++ b/storage/rocksdb/ha_rocksdb.cc
@@ -104,6 +104,9 @@
 #include "./ObjectFactory.h"
 #endif
 
+#include "./rdb_locking_iter.h"
+
+
 // Internal MySQL APIs not exposed in any header.
 extern "C" {
 /**
@@ -655,6 +658,10 @@ static void rocksdb_set_delayed_write_rate(THD *thd, struct SYS_VAR *var,
 static void rocksdb_set_max_latest_deadlocks(THD *thd, struct SYS_VAR *var,
                                              void *var_ptr, const void *save);
 
+static void rocksdb_set_max_lock_memory(THD *thd,
+                                        struct SYS_VAR *var,
+                                        void *var_ptr, const void *save);
+
 static void rdb_set_collation_exception_list(const char *exception_list);
 static void rocksdb_set_collation_exception_list(THD *thd, struct SYS_VAR *var,
                                                  void *var_ptr,
@@ -807,6 +814,16 @@ static unsigned long long  // NOLINT(runtime/int)
     rocksdb_select_bypass_multiget_min = 0;
 static bool rocksdb_skip_locks_if_skip_unique_check = false;
 static bool rocksdb_alter_column_default_inplace = false;
+
+// Range Locking: how much memory can be used used for the lock data structure
+//  (which hold the locks acquired by all clients).
+static ulonglong rocksdb_max_lock_memory;
+
+static bool rocksdb_use_range_locking = 0;
+static bool rocksdb_use_range_lock_manager_as_point = 0;
+std::shared_ptr<rocksdb::RangeLockManagerHandle> range_lock_mgr;
+
+std::shared_ptr<rocksdb::RangeLockManagerHandle> range_lock_mgr_used_as_point;
 
 std::atomic<uint64_t> rocksdb_row_lock_deadlocks(0);
 std::atomic<uint64_t> rocksdb_row_lock_wait_timeouts(0);
@@ -1507,6 +1524,13 @@ static MYSQL_SYSVAR_UINT(max_latest_deadlocks, rocksdb_max_latest_deadlocks,
                          "deadlocks to store",
                          nullptr, rocksdb_set_max_latest_deadlocks,
                          rocksdb::kInitialMaxDeadlocks, 0, UINT32_MAX, 0);
+
+static MYSQL_SYSVAR_ULONGLONG(max_lock_memory, rocksdb_max_lock_memory,
+                              PLUGIN_VAR_RQCMDARG,
+                              "Range-locking mode: Maximum amount of memory "
+                              "that locks from all transactions can use at a time",
+                              nullptr, rocksdb_set_max_lock_memory,
+                              /*initial*/1073741824, 0, UINT64_MAX, 0);
 
 static MYSQL_SYSVAR_ENUM(
     info_log_level, rocksdb_info_log_level, PLUGIN_VAR_RQCMDARG,
@@ -2364,6 +2388,19 @@ static MYSQL_SYSVAR_BOOL(table_stats_use_table_scan,
                          rocksdb_update_table_stats_use_table_scan,
                          rocksdb_table_stats_use_table_scan);
 
+static MYSQL_SYSVAR_BOOL(use_range_locking, rocksdb_use_range_locking,
+                         PLUGIN_VAR_OPCMDARG | PLUGIN_VAR_READONLY,
+                         "Use Range Locking",
+                         nullptr, nullptr,
+                         rocksdb_use_range_locking);
+
+static MYSQL_SYSVAR_BOOL(use_range_lock_manager_as_point,
+                         rocksdb_use_range_lock_manager_as_point,
+                         PLUGIN_VAR_OPCMDARG | PLUGIN_VAR_READONLY,
+                         "Use Range Lock Manager as point",
+                         nullptr, nullptr,
+                         rocksdb_use_range_lock_manager_as_point);
+
 static MYSQL_SYSVAR_BOOL(
     large_prefix, rocksdb_large_prefix, PLUGIN_VAR_RQCMDARG,
     "Support large index prefix length of 3072 bytes. If off, the maximum "
@@ -2680,7 +2717,9 @@ static struct SYS_VAR *rocksdb_system_variables[] = {
     MYSQL_SYSVAR(manual_compaction_threads),
     MYSQL_SYSVAR(manual_compaction_bottommost_level),
     MYSQL_SYSVAR(rollback_on_timeout),
-
+    MYSQL_SYSVAR(use_range_locking),
+    MYSQL_SYSVAR(use_range_lock_manager_as_point),
+    MYSQL_SYSVAR(max_lock_memory),
     MYSQL_SYSVAR(enable_insert_with_update_caching),
     MYSQL_SYSVAR(trace_block_cache_access),
     MYSQL_SYSVAR(trace_queries),
@@ -2966,8 +3005,41 @@ class Rdb_transaction {
   virtual rocksdb::Status do_pop_savepoint() = 0;
   virtual void do_rollback_to_savepoint() = 0;
 
+ private:
+  /*
+    If true, the current statement should not use a snapshot for reading.
+    Note that in a multi-statement transaction, the snapshot may have been
+    allocated by another statement.
+  */
+  bool m_stmt_ignores_snapshot = false;
+
+  /* Snapshot-ignore mode will put away m_reads_opts.snapshot here: */
+  const rocksdb::Snapshot *m_saved_snapshot= nullptr;
+
  public:
+
+  void start_ignore_snapshot() {
+    // note: this may be called several times for the same statement
+    if (!m_stmt_ignores_snapshot) {
+      m_saved_snapshot = m_read_opts.snapshot;
+      m_read_opts.snapshot = nullptr;
+      m_stmt_ignores_snapshot= true;
+      if (!m_snapshot_timestamp)
+        rdb->GetEnv()->GetCurrentTime(&m_snapshot_timestamp);
+    }
+  }
+
+  void end_ignore_snapshot_if_needed() {
+    if (m_stmt_ignores_snapshot) {
+      m_stmt_ignores_snapshot = false;
+      m_read_opts.snapshot = m_saved_snapshot;
+      m_saved_snapshot = nullptr;
+    }
+  }
+  bool in_snapshot_ignore_mode() const { return m_stmt_ignores_snapshot; }
+
   rocksdb::ReadOptions m_read_opts;
+
   const char *m_mysql_log_file_name;
   my_off_t m_mysql_log_offset;
   const char *m_mysql_gtid;
@@ -3148,6 +3220,19 @@ class Rdb_transaction {
   virtual void release_lock(const Rdb_key_def &key_descr,
                             const std::string &rowkey, bool force = false) = 0;
 
+  virtual
+  rocksdb::Status lock_range(rocksdb::ColumnFamilyHandle *const cf,
+                             const rocksdb::Endpoint &start,
+                             const rocksdb::Endpoint &end) = 0;
+
+  rocksdb::Status lock_singlepoint_range(rocksdb::ColumnFamilyHandle *const cf,
+                                         const rocksdb::Slice &point) {
+    // Normally, one needs to "flip" the endpoint type for reverse-ordered CFs.
+    // But here we are locking just one point so this is not necessary.
+    rocksdb::Endpoint endp(point, false);
+    return lock_range(cf, endp, endp);
+  }
+
   virtual bool prepare() = 0;
 
   bool commit_or_rollback() {
@@ -3206,10 +3291,17 @@ class Rdb_transaction {
     m_is_delayed_snapshot = false;
   }
 
+  void locking_iter_created() {
+    if (!m_snapshot_timestamp)
+      rdb->GetEnv()->GetCurrentTime(&m_snapshot_timestamp);
+  }
+
   virtual void acquire_snapshot(bool acquire_now) = 0;
   virtual void release_snapshot() = 0;
 
-  bool has_snapshot() const { return m_read_opts.snapshot != nullptr; }
+  bool has_snapshot() const {
+    return m_read_opts.snapshot != nullptr || m_saved_snapshot;
+  }
 
  private:
   // The Rdb_sst_info structures we are currently loading.  In a partitioned
@@ -3763,7 +3855,9 @@ class Rdb_transaction {
 
   virtual rocksdb::Iterator *get_iterator(
       const rocksdb::ReadOptions &options,
-      rocksdb::ColumnFamilyHandle *column_family) = 0;
+      rocksdb::ColumnFamilyHandle *column_family,
+      bool is_rev_cf,
+      bool use_locking_iterator=false) = 0;
 
   virtual void multi_get(rocksdb::ColumnFamilyHandle *const column_family,
                          const size_t num_keys, const rocksdb::Slice *keys,
@@ -3772,10 +3866,12 @@ class Rdb_transaction {
                          const bool sorted_input) const = 0;
 
   rocksdb::Iterator *get_iterator(
-      rocksdb::ColumnFamilyHandle *const column_family, bool skip_bloom_filter,
+      rocksdb::ColumnFamilyHandle *const column_family, bool is_rev_cf,
+      bool skip_bloom_filter,
       const rocksdb::Slice &eq_cond_lower_bound,
       const rocksdb::Slice &eq_cond_upper_bound, bool read_current = false,
-      bool create_snapshot = true) {
+      bool create_snapshot = true,
+      bool use_locking_iterator=false) {
     // Make sure we are not doing both read_current (which implies we don't
     // want a snapshot) and create_snapshot which makes sure we create
     // a snapshot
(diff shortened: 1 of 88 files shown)


storage/rocksdb/rdb_locking_iter.h
```diff
@@ -0,0 +1,190 @@
...
+  template <bool forward> void Scan(const rocksdb::Slice& target,
+                                    bool call_next) {
+    if (!iter_->Valid()) {
+      status_ = iter_->status();
+      valid_ = false;
+      return;
+    }
+
+    while (1) {
+      /*
```
hermanlee: Could you thd->killed in this loop and terminate/return HA_ERR_QUERY_INTERRUPTED if set?

storage/rocksdb/rdb_locking_iter.h
```diff
@@ -0,0 +1,190 @@
...
+//    lock_iter->Seek('abc');
+//    lock_iter->Valid()==true && lock_iter->key() == 'bcd';
+//
+//   After the above, the returned record 'bcd' is locked by transaction trx.
+//   Also, the range between ['abc'..'bcd'] is empty and locked by trx.
+//
+//    lock_iter->Next();
+//    lock_iter->Valid()==true && lock_iter->key() == 'efg'
+//
+//   Now, the range ['bcd'.. 'efg'] (bounds incluive) is also locked, and there are no
```
hermanlee: nit: inclusive

storage/rocksdb/rdb_locking_iter.h
```diff
@@ -0,0 +1,190 @@
...
+//   records between 'bcd'  and 'efg'.
+//
+class LockingIterator : public rocksdb::Iterator {
+
+  rocksdb::Transaction *txn_;
```
hermanlee: Could these member names be prefixed with m_?

hermanlee (commented):
Just a few initial comments. I will continue to read through this.

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -14738,6 +14641,36 @@ static int show_rocksdb_stall_vars(THD *thd MY_ATTRIBUTE((unused)),
...
+static longlong rocksdb_locktree_escalation_count=1234;
+static longlong rocksdb_locktree_current_lock_memory=0;
+
+static SHOW_VAR rocksdb_locktree_status_variables[] = {
+    DEF_STATUS_VAR_FUNC("escalation_count",
+                        &rocksdb_locktree_escalation_count, SHOW_LONGLONG),
+    DEF_STATUS_VAR_FUNC("current_lock_memory",
+                        &rocksdb_locktree_current_lock_memory, SHOW_LONGLONG),
+    // end of the array marker
+    {NullS, NullS, SHOW_LONG}};
```
hermanlee: Should  {NullS, NullS, SHOW_LONG, SHOW_SCOPE_GLOBAL}

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -14738,6 +14641,36 @@ static int show_rocksdb_stall_vars(THD *thd MY_ATTRIBUTE((unused)),
...
+
+static SHOW_VAR rocksdb_empty_status_variables[] = {
+    {NullS, NullS, SHOW_LONG}};
```
hermanlee: Should {NullS, NullS, SHOW_LONG, SHOW_SCOPE_GLOBAL}

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -9338,140 +9253,297 @@ int ha_rocksdb::index_read_map(uchar *const buf, const uchar *const key,
...
+  bool start_has_inf_suffix = false, end_has_inf_suffix = false;
+  rocksdb::Slice slice(slice_arg);
+  *use_locking_iterator= false;
 
-    while (1) {
-      int rc = rocksdb_skip_expired_records(kd, m_scan_it, !move_forward);
-      if (rc != HA_EXIT_SUCCESS) {
-        return rc;
-      }
+  if (m_lock_rows == RDB_LOCK_NONE || !rocksdb_use_range_locking) {
```
hermanlee: For read-committed isolation level, we should probably skip range locks? It seems InnoDB read-committed isolation level doesn't hold range locks (except in short instances like inserting a new row).

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -9338,140 +9253,297 @@ int ha_rocksdb::index_read_map(uchar *const buf, const uchar *const key,
...
+      /*
+        This is "key_part= const" interval. We need to lock this range:
+        (lookup_value, -inf) < key < (lookup_value, +inf)
+      */
+      start_has_inf_suffix= false;
+      end_has_inf_suffix= true;
+      end_slice= slice;
+    }
+  }
+  else if (find_flag == HA_READ_PREFIX_LAST) {
```
hermanlee: It seems this case can be combined with HA_READ_KEY_EXACT since both are setting the range lock at [(key,0) (key, +inf)]

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -9338,140 +9253,297 @@ int ha_rocksdb::index_read_map(uchar *const buf, const uchar *const key,
...
+    if (slice.size() == Rdb_key_def::INDEX_ID_SIZE) {
+      /* Reverse-ordered full index scan */
+      start_has_inf_suffix= true;
```
hermanlee: Even though start_has_inf_suffix is not used in the case of big_range = true, it's not clear why start_has_inf_suffix is true here? Isn't the range lock still [(INDEX_ID, 0), (INDEX_ID, +inf)], same as read_key_exact?

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -9338,140 +9253,297 @@ int ha_rocksdb::index_read_map(uchar *const buf, const uchar *const key,
...
+    if (end_key) {
+      uchar pack_buffer[MAX_KEY_LENGTH];
+      uint end_slice_size=
+          kd.pack_index_tuple(table, pack_buffer, end_slice_buf,
+                              end_key->key, end_key->keypart_map);
+
+      slice= rocksdb::Slice(reinterpret_cast<char *>(end_slice_buf),
+                            end_slice_size);
+
+      end_has_inf_suffix= false;
```
hermanlee: It seems these two lines aren't needed, but maybe a comment that end_key doesn't not need inf_suffix because we're looking at < end_key?

Or include start_has_inf_suffix = false for completeness?

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -9338,140 +9253,297 @@ int ha_rocksdb::index_read_map(uchar *const buf, const uchar *const key,
...
   }
-  return HA_EXIT_SUCCESS;
+  else if (end_key) {
+    // Known start range bounds: HA_READ_KEY_OR_NEXT, HA_READ_AFTER_KEY
+    if (find_flag == HA_READ_KEY_OR_NEXT)
+      start_has_inf_suffix= false;
+    else if (find_flag == HA_READ_AFTER_KEY)
+      start_has_inf_suffix= true;
+    else
+      DBUG_ASSERT(0);
```
hermanlee, line 9713: If we hit this case, should this just set big_range = true?

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -9338,140 +9253,297 @@ int ha_rocksdb::index_read_map(uchar *const buf, const uchar *const key,
...
+    *use_locking_iterator= true;
+    return 0;
+  }
+
+  rocksdb::Endpoint start_endp;
+  rocksdb::Endpoint end_endp;
+
+  if (kd.m_is_reverse_cf) {
+    // Flip the endpoints
+    start_endp =rocksdb::Endpoint(end_slice, !end_has_inf_suffix);
```
hermanlee, line 9768: It's not clear why the suffixes are inverted here?

If the forward cf range to be locked is (pk=10, 0) -> (pk=12, +inf), then isn't the reverse cf range to be locked
(pk=12, +inf) -> (pk=10, 0)? It seems (pk=12, 0) -> (pk=10, +inf) would not be right?

storage/rocksdb/rdb_iterator.cc
```diff
@@ -0,0 +1,974 @@
...
+          m_thd, *m_kd, eq_cond,
+          std::max(eq_cond_len, (uint)Rdb_key_def::INDEX_ID_SIZE),
+          m_scan_it_lower_bound, m_scan_it_upper_bound,
+          &m_scan_it_lower_bound_slice, &m_scan_it_upper_bound_slice)) {
+    skip_bloom = false;
+  }
+
+  // Save the value of m_use_locking_iter because release_scan_iterator()
+  // will set it to false.
+  bool use_locking_iter= m_use_locking_iter;
```
hermanlee: It doesn't look like m_use_locking_iter is set back to use_locking_iter after release_scan_iterator() is called. 

In index_read_intern(), for read-committed, we could call setup_scan_iterator and release_iterator multiple times. If we ever did that for repeatable-read, then it seems m_use_locking_iter, then while the first couple of times through the loop might grab a locking iterator, the third and later times through the loop would not allocate a locking iterator.

storage/rocksdb/rdb_locking_iter.h
```diff
@@ -0,0 +1,190 @@
...
+      return;
+    }
+
+    while (1) {
+      /*
+        note: the underlying iterator checks iterator bounds, so we don't need
+        to check them here
+      */
+      DEBUG_SYNC(my_core::thd_get_current_thd(), "rocksdb.locking_iter_scan");
+      auto end_key = iter_->key();
```
hermanlee: Would it might make sense to store the range locked so far. If this key falls within the range, we won't have to perform another GetRangeLock(). Thinking ahead, if we wanted to optimize this by scan ahead a few keys and locking the whole range first, we can reduce the number of GetRangeLock() calls.

spetrunia: > If this key falls within the range, we won't have to perform another GetRangeLock().

Are you talking about the case where another transaction is inserting records into the range we're scanning? Unless I am missing something, this is the only scenario where the current code will make more GetRangeLock() calls than necessary.

I agree that if we scan ahead several values, we'll need to remember what range we've locked.

hermanlee: Yes, I'm thinking of the case where our snapshot gives us a next_key and as we attempt to acquire a range lock up to next_key, a concurrent transaction inserts and commits multiple keys before our next_key. Once we lock up to next_key, we refresh the snapshot and discover the next few committed keys is already in the range lock up to next_key.

storage/rocksdb/rdb_iterator.cc
spetrunia: Addressed

storage/rocksdb/ha_rocksdb.cc
spetrunia: Right. It should be false, but also it doesn't matter because it is not used.

storage/rocksdb/ha_rocksdb.cc
spetrunia: The reason suffixes are inverted is as follows:

Let's consider a forward-ordered CF. The endpoint values would compare as follows (here LOWEST=0, HIGHEST=1, missing inf_suffix means any inf_suffix value):
 
```
 key=a, inf_suffix=LOWEST
 key=ab
 key=az
 key=a, inf_suffix=HIGHEST

 key=b, inf_suffix=LOWEST
 key=ba
 key=bz
 key=b, inf_suffix=HIGHEST
 ...
```
Putting the same set of keys into a reverse CF:

```
 key=b, inf_suffix=HIGHEST
 key=bz
 key=ba
 key=b, inf_suffix=LOWEST

 key=a, inf_suffix=HIGHEST
 key=az
 key=ab
 key=a, inf_suffix=LOWEST.
```
Note that the flag values have swapped their meanings, HIGHEST comes before LOWEST.

However, the endpoint comparison function in RocksDB doesn't "know" if the CF is reverse-ordered or not (it just makes use of the Key Comparator).

As for inf_suffix value, `Endpoint(key=$VAL, inf_suffix=0)` always compares before any endpoint with key=$VAL, and `Endpoint(key=$VAL, inf_suffix=1)` always compares after any endpoint with key=$VAL.

This is why one needs to flip the inf_suffix values.

storage/rocksdb/ha_rocksdb.cc
spetrunia: Removed the lines, added a comment.

storage/rocksdb/ha_rocksdb.cc
spetrunia: ... added a comment about this.

storage/rocksdb/ha_rocksdb.cc
spetrunia: Yes, done.

storage/rocksdb/rdb_locking_iter.h
spetrunia: Ok, done.

hermanlee (commented):
I'm seeing a different behavior from InnoDB where the gap from the last key to supremum is not being acquired.

On InnoDB:

TXN 1:
create table t1 (pk int primary key) engine=innodb;
insert into t1 values (1), (10), (100);
begin;
select * from t1 for update;

TXN 2:
begin;
select * from t1;
insert into t1 values (150); # times out on lock

On MyRocks:
insert into t1 values (150); # succeeds

hermanlee (commented):
Another difference with InnoDB:

```
TXN 1:
create table t1 (pk int primary key) engine=rocksdb;
insert into t1 values (1), (10), (100);
begin;
select * from t1;

TXN 2:
begin;
select * from t1;
insert into t1 values (90);
commit;

TXN 1:
select * from t1 lock in share mode; /* InnoDB returns 4 rows, but MyRocks only returns 3 rows: 90 is missing. */
```

Using `select * from t1 for update`, MyRocks correctly returns 4 rows instead of 3.

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -4287,17 +4400,35 @@ class Rdb_transaction_impl : public Rdb_transaction {
...
-  void start_stmt() override {
+  void start_stmt(bool is_dml_statement) override {
+
+    if (rocksdb_use_range_locking && is_dml_statement) {
+      /*
+        In Range Locking mode, RocksDB does not do "key tracking".
+        Use InnoDB-like concurrency mode: make the DML statements always read
+        the latest data (instead of using transaction's snapshot).
+        This "downgrades" the transaction isolation to READ-COMMITTED on the
+        master, but in return the actions can be replayed on the slave.
```
hermanlee: For terminology, please use `primary` instead of `master` and `secondary` instead of `slave`.

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -6667,6 +6882,15 @@ static int rocksdb_init_internal(void *const p) {
     DBUG_RETURN(HA_EXIT_FAILURE);
   }
 
+  if (range_lock_mgr)
+  {
+    range_lock_mgr->SetMaxLockMemory(rocksdb_max_lock_memory);
+    sql_print_information("RocksDB: USING NEW RANGE LOCKING");
```
hermanlee: Can probably remove `NEW`. It's just range locking vs. point locking.

storage/rocksdb/rdb_locking_iter.h
```diff
@@ -0,0 +1,222 @@
...
+#include "./ha_rocksdb.h"
+
+namespace myrocks {
+
+//////////////////////////////////////////////////////////////////////////////
+// Locking iterator
+//////////////////////////////////////////////////////////////////////////////
+
+//
+// LockingIterator is an iterator that locks the rows before returning, as well
```
hermanlee, line 14: I don't think this actually works with non-materialized partial indexes. In the case where partial index is not materialized, the rows are read from the primary key, sorted in the secondary key order and then returned directly by the partial_index iterator. It would not make Seek()/Next() calls on the secondary index down to rocksdb, and so the range locks in the non-materialized partial index case would not be acquired through this iterator.

I'm getting an assertion when trying to test this though:

```
create table t1 (a int, b int, c int, d int, e int default 0  primary key(a, b, c), key (a, b, d) comment 'partial_group_keyparts=2,partial_group_threshold=10000') engine=rocksdb;
insert into t1 (a,b,c,d) values (1, 1, 100, 1000), (1,1,200, 2000), (1, 1, 300, 3000), (1, 1, 400, 4000);
update t1 set d = d + 1 where a = 1 and b = 1 and d > 2000; 
```

```
(gdb) bt
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
#1  0x00007ffff7e124ed in __GI_abort () at abort.c:79
#2  0x00007ffff7e224d3 in __assert_fail_base (fmt=0x7ffff7f9a088 "%s%s%s:%u: %s%sAssertion `%s' failed.\n%n", 
    assertion=assertion@entry=0x3eda1fc "!key_descr.is_primary_key() || (key_descr.is_primary_key() && m_row_lock_count > 0)", 
    file=file@entry=0x3cdbdc3 "/home/herman/rocks-mysql/8.0/storage/rocksdb/ha_rocksdb.cc", line=line@entry=4040, 
    function=function@entry=0x3a0cb52 "virtual void myrocks::Rdb_transaction_impl::release_lock(const myrocks::Rdb_key_def &, const std::string &, bool)") at assert.c:92
#3  0x00007ffff7e22543 in __GI___assert_fail (
    assertion=0x3eda1fc "!key_descr.is_primary_key() || (key_descr.is_primary_key() && m_row_lock_count > 0)", 
    file=0x3cdbdc3 "/home/herman/rocks-mysql/8.0/storage/rocksdb/ha_rocksdb.cc", line=4040, 
    function=0x3a0cb52 "virtual void myrocks::Rdb_transaction_impl::release_lock(const myrocks::Rdb_key_def &, const std::string &, bool)")
    at assert.c:101
#4  0x0000000007b54dfb in myrocks::Rdb_transaction_impl::release_lock (this=0x7fffbc84aa80, key_descr=..., 
    rowkey="\000\000\001\004\200\000\000\001\200\000\000\001\200\000", <incomplete sequence \310>, force=false)
    at /home/herman/rocks-mysql/8.0/storage/rocksdb/ha_rocksdb.cc:4039
#5  0x0000000007b0337c in myrocks::ha_rocksdb::unlock_row (this=0x7fffbc883428) at /home/herman/rocks-mysql/8.0/storage/rocksdb/ha_rocksdb.cc:10463
#6  0x0000000005cdd1a0 in Sql_cmd_update::update_single_table (this=0x7fffbc80faf0, thd=0x7fffbc80b000)
    at /home/herman/rocks-mysql/8.0/sql/sql_update.cc:809
#7  0x0000000005ce0c47 in Sql_cmd_update::execute_inner (this=0x7fffbc80faf0, thd=0x7fffbc80b000)
    at /home/herman/rocks-mysql/8.0/sql/sql_update.cc:1614
#8  0x0000000005bf58be in Sql_cmd_dml::execute (this=0x7fffbc80faf0, thd=0x7fffbc80b000) at /home/herman/rocks-mysql/8.0/sql/sql_select.cc:738
#9  0x0000000005b51f1e in mysql_execute_command (thd=0x7fffbc80b000, first_level=true, last_timer=0x7ffff53b39b0)
    at /home/herman/rocks-mysql/8.0/sql/sql_parse.cc:4204
#10 0x0000000005b4b38d in mysql_parse (thd=0x7fffbc80b000, parser_state=0x7ffff53b4560, last_timer=0x7ffff53b39b0)
    at /home/herman/rocks-mysql/8.0/sql/sql_parse.cc:6202
#11 0x0000000005b46cf9 in dispatch_command (thd=0x7fffbc80b000, com_data=0x7ffff53b4d98, command=COM_QUERY)
    at /home/herman/rocks-mysql/8.0/sql/sql_parse.cc:2226
#12 0x0000000005b49ee1 in do_command (thd=0x7fffbc80b000) at /home/herman/rocks-mysql/8.0/sql/sql_parse.cc:1560
#13 0x0000000005dcaa21 in handle_connection (arg=0x7fffcea982e0)
    at /home/herman/rocks-mysql/8.0/sql/conn_handler/connection_handler_per_thread.cc:330
#14 0x00000000079f00fb in pfs_spawn_thread (arg=0x7ffff748a620) at /home/herman/rocks-mysql/8.0/storage/perfschema/pfs.cc:2884
#15 0x00007ffff7d5d20c in start_thread (arg=0x7ffff53b6700) at pthread_create.c:479
#16 0x00007ffff7f0581f in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:95
```

However, performance with partial indexes might be very poor because we sort the primary keys, set the range lock, and then we need to re-read/re-sort the primary keys. We'd want to lock based on the rows returned by the partial index iterator.

Alternatively, we could skip supporting range locks for partial indexes and use only point locks, but going down this road means the compatibility matrix among different myrocks features like ttl, range locking, partial indexes, etc. gets complicated quickly.

cc: @lth

hermanlee: This assertion happens even with a regular SK (not partial index) in this table schema.

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -10850,6 +11331,15 @@ int ha_rocksdb::update_write_sk(const TABLE *const table_arg,
     old_key_slice = rocksdb::Slice(
         reinterpret_cast<const char *>(m_sk_packed_tuple_old), old_packed_size);
 
+    /* Range locking: lock the index tuple being deleted */
+    if (m_use_range_locking) {
```
hermanlee: It seems like this lock should go into check_and_lock_sk() instead?

hermanlee: Also, it seems like this check should be using  `rocksdb_use_range_locking` instead? If the range lock manager is being used, then a singlepoint lock is needed on each secondary key regardless of the current transaction's isolation level.

For example:

```
TXN 1:

create table i1 (pk int, a int, b int, primary key (pk), key(a)) engine=innodb;
insert into i1 values (1, 100, 1000), (2, 200, 2000), (3, 300, 3000);
set transaction isolation level repeatable read;
begin;
update i1 set b = b + 1 where a > 200;

TXN 2:

set transaction isolation level read committed;
begin;
insert into i1 values (4, 150, 1500); /* Success */
insert into i1 values (5, 250, 1500); /* Blocks on InnoDB, but succeeds on MyRocks. */
```

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -11900,8 +12411,15 @@ int ha_rocksdb::external_lock(THD *const thd, int lock_type) {
       }
     }
     tx->m_n_mysql_tables_in_use++;
-    rocksdb_register_tx(rocksdb_hton, thd, tx);
+    rocksdb_register_tx(rocksdb_hton, thd, tx, (lock_type == F_WRLCK));
```
hermanlee: This seems like it should be `lock_type == F_WRLCK || lock_type == F_RDLCK` so that `select lock in share mode` can also read the latest data?

storage/rocksdb/rdb_i_s.cc
```diff
@@ -1771,31 +1771,63 @@ static int rdb_i_s_lock_info_fill_table(
...
-      tables->table->field[RDB_LOCKS_FIELD::MODE]->store(
-          key_lock_info.exclusive ? "X" : "S", 1, system_charset_info);
-
-      /* Tell MySQL about this row in the virtual table */
-      ret = static_cast<int>(
-          my_core::schema_table_store_record(thd, tables->table));
-
-      if (ret != 0) {
-        break;
+  if (range_lock_mgr) {
```
hermanlee, line 1828: Given how similar the range lock and non-range lock sections are, would it be better to refactor this code? The code populating the tables->table->field could be added to separate helper function that takes a std::string for the key_hexstr and char for lock_mode.

storage/rocksdb/rdb_locking_iter.h
```diff
@@ -0,0 +1,222 @@
...
+          m_status = rocksdb::Status::OK();
+          m_valid= true;
+          break;
+        } else {
+          // We've got a key but it is outside the range we've locked.
+          // Re-try the lock-and-read step.
+          continue;
+        }
+      } else {
+        // There's no row (within the iterator bounds perhaps). Exit now.
```
hermanlee: It seems like we may need range lock between the target slice and the end range of the iterator bounds.

storage/rocksdb/rdb_locking_iter.h
hermanlee: Some further thoughts. I think we can either range lock the PK or range lock the SK here. Range locking the PK would still be correct since any insert/updates/scans would still need to lock the PK. We may end up locking more than needed, but I think it simplifies the implementation.

Since the PK and SK share the same prefixes and the PK is read to find the prefixes needed for the SK, we can range lock the prefixes on the PK in get_prefix_from_start(). m_pk_iterator.seek() is used to find the next prefixes, so we need to range lock the start_key to the returned prefix. We can range lock from endpoint (start_key, 0) to the endpoint (returned_prefix, +inf) so that we don't need to individually lock each gap between rows of the returned_prefix.

We would need to also implement the read/lock/re-read logic using the m_pk_iterator here. Given how the partial index iterator uses m_pk_iterator, it's not clear to me pushing m_iter_should_use_locking to the m_pk_iterator is sufficient because the range locks need to cover the whole range from one prefix to the next.

lth: Currently, partial indexes will call always Seek() on the secondary key range before reading from the PK, and that theoretically should acquire the correct range locks, and partial indexes should just work without much additional changes.

What we'll end up is that the locking PK iterator will ensure that no new prefixes show up in between consecutive prefixes, and the locking SK iterator will ensure that no new keys can be added within a prefix. 

That said, I think there are a few problems:
1. I don't think we're acquiring range locks if the iterator is invalid (in general, it doesn't seem like we acquire range locks up to the iterator end bound).
2. We also need to make sure we're acquiring point locks on the SK even if they're not materialized (it doesn't seem like that is done in this diff though).
3. In the partial index seek code, we always expect that we'd be able to read some rows out after getting the next prefix from the PK. However, with iterator refreshing, that might not be true anymore, so we probably need to add a while loop there.

There also may be some opportunity for optimization, since partial indexes currently uses a point lock on the prefix key to stand in as a range lock over the prefix (we do something similar for unique keys), so we probably don't have to lock twice. Also, it might be cleaner to lock only on PK instead as described above.

hermanlee: Yes, Rdb_iterator_partial::seek() does always call Rdb_iterator_base::seek() to lock the range. We should get the correct range locking out of the partial index iterator's base iterator. The main downside is the ranges might be very large if the SK doesn't have any materialized prefixes, but that's a performance/concurrency issue, and not a correctness one.

I'd think we'd also need to set m_iterator_pk.m_iter_should_use_locking based on the partial index iterator's setting, which I think is not currently happening. And yes, I think locking only on the PK might be cleaner. We might be able to get away with not having to acquire SK point locks if we lock the whole prefix range in the PK before doing anything.

storage/rocksdb/rdb_i_s.cc
spetrunia: Done

storage/rocksdb/rdb_locking_iter.h
spetrunia: Yes, done.

storage/rocksdb/rdb_locking_iter.h
spetrunia: > This assertion happens even with a regular SK (not partial index) in this table schema.

Confirm.  I've checked the cause:  The code attempts to count `m_row_lock_count`, and it counts range lock acquisition as one lock.  For rows that didn't match the WHERE, `unlock_row` decrements the count. If there is more than one row in the range, the counter can go below zero and fire an assert.

storage/rocksdb/rdb_locking_iter.h
spetrunia: Wait, the fix fix crashes with `rocksdb_enable_iterate_bounds=off` . Will fix

spetrunia: Now done.

storage/rocksdb/ha_rocksdb.cc
spetrunia: > Also, it seems like this check should be using rocksdb_use_range_locking instead?


Agree. done.

storage/rocksdb/ha_rocksdb.cc
spetrunia: Agree that `select ... lock in share mode` should read the latest data.
However, `F_RDLCK` is passed for both SELECTs and `SELECT ... LOCK IN SHARE MODE`. 
Changed the check to analyze `m_lock_rows`.

storage/rocksdb/rdb_locking_iter.h
spetrunia: Ok finally committed 9f29d708d47db4aaf4e015697dd088d2ee408b44  (pushed into this PR, too)  with a solution for this.

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -4346,17 +4460,35 @@ class Rdb_transaction_impl : public Rdb_transaction {
...
-  void start_stmt() override {
+  void start_stmt(bool is_dml_statement) override {
+
+    if (rocksdb_use_range_locking && is_dml_statement) {
+      /*
+        In Range Locking mode, RocksDB does not do "key tracking".
+        Use InnoDB-like concurrency mode: make the DML statements always read
+        the latest data (instead of using transaction's snapshot).
+        This "downgrades" the transaction isolation to READ-COMMITTED on the
+        primary, but in return the actions can be replayed on the replica.
```
hermanlee, line 4477: The first half of the sentence makes sense, but the second half: "but in return the actions can be replayed on the replica" seems to involve replication and it's not clear how replication to secondaries is affected by this concurrency mode?

storage/rocksdb/rdb_locking_iter.cc
```diff
@@ -0,0 +1,286 @@
...
+        m_status = rocksdb::Status::OK();
+        m_valid= true;
+        break;
+      } else {
+        // We've got a key but it is outside the range we've locked.
+        // Re-try the lock-and-read step.
+        continue;
+      }
+    } else {
+      // Some error
```
hermanlee: Seems like this comment is inaccurate? If we reached the end of the range, it's not really an error?

hermanlee (commented):
The following doesn't look right:

```
/* TXN 1 */
mysql> create table t1 (pk int not null auto_increment, a int, b int, primary key (pk), unique key (a)) engine=rocksdb;
Query OK, 0 rows affected (0.05 sec)
                                         
mysql> begin;
                                                                                   
mysql> insert into t1 (a, b) values (1, 1), (2, 2);                               
Query OK, 2 rows affected (0.01 sec)                                               
Records: 2  Duplicates: 0  Warnings: 0                                             
                                                                                   
mysql> select * from information_schema.rocksdb_locks;                            
+------------------+----------------+--------------------+------+
| COLUMN_FAMILY_ID | TRANSACTION_ID | KEY                | MODE |
+------------------+----------------+--------------------+------+
|                0 |             17 | 000001080180000002 | X    |
|                0 |             17 | 000001080180000001 | X    |
|                0 |             17 | 0000010780000002   | X    |
|                0 |             17 | 0000010780000001   | X    |
+------------------+----------------+--------------------+------+
4 rows in set (0.01 sec)

/* TXN 2 */
mysql> begin;

mysql> insert into t1 (a, b) values (10, 10), (20, 20);
ERROR 1205 (HY000): Lock wait timeout exceeded; try restarting transaction: Timeout on index: test.t1.PRIMARY

/* TXN 1 */
mysql> select * from information_schema.rocksdb_locks;
+------------------+----------------+-------------------------------------+------+
| COLUMN_FAMILY_ID | TRANSACTION_ID | KEY                                 | MODE |
+------------------+----------------+-------------------------------------+------+
|                0 |             17 | 0000010780000001-000001080180000002 | X    |
+------------------+----------------+-------------------------------------+------+
1 row in set (0.01 sec)
```

The range lock seems to span multiple indexes.

Another example:

```
mysql> create table t6 (pk int not null auto_increment, a int, b int, c int, primary key (pk), unique key (a), key (b), unique key (c)) engine=rocksdb;
Query OK, 0 rows affected (0.03 sec)
                                          
mysql> begin;

mysql> insert into t6 (a, b, c) values (1, 1, 1), (2, 2, 2);                                                                                                             
Query OK, 2 rows affected (0.00 sec)                                                                                                                                     
Records: 2  Duplicates: 0  Warnings: 0    
                                          
mysql> select * from information_schema.rocksdb_locks;                                                                                                                   
+------------------+----------------+---------------------------------------------+------+                                                                               
| COLUMN_FAMILY_ID | TRANSACTION_ID | KEY                                         | MODE |                                                                               
+------------------+----------------+---------------------------------------------+------+                                                                               
|                0 |             21 | 00000113018000000280000002                  | X    |                                                                               
|                0 |             21 | 0000011080000001-00000113018000000180000001 | X    |                                                                               
+------------------+----------------+---------------------------------------------+------+                                                                               
2 rows in set (0.00 sec)                  
                                                                                                                                                                         
mysql> select index_name, hex(index_number) from information_schema.rocksdb_ddl where table_name = 't6';                                                                 
+------------+-------------------+
| index_name | hex(index_number) |  
+------------+-------------------+        
| PRIMARY    | 110               |
| a          | 111               |  
| c          | 112               |        
| b          | 113               |                                                                                                                                       
+------------+-------------------+  
4 rows in set (0.00 sec)

File: storage/rocksdb/rdb_locking_iter.cc (hunk 22 above)
Comment (spetrunia) [reply storage/rocksdb/rdb_locking_iter.cc]: Agree... "Some error" here included EoF which is not really an error.

**Comment by spetrunia** [comment]
https://github.com/facebook/mysql-5.6/pull/1165#pullrequestreview-769443923
> The following doesn't look right: 
> ...

I can't reproduce this behavior... This is what I'm getting:

```
mysql> begin;

mysql> insert into t1 (a, b) values (1, 1), (2, 2);
Query OK, 2 rows affected (0.01 sec)
Records: 2  Duplicates: 0  Warnings: 0

mysql> select * from information_schema.rocksdb_locks;
+------------------+----------------+--------------------+------+
| COLUMN_FAMILY_ID | TRANSACTION_ID | KEY                | MODE |
+------------------+----------------+--------------------+------+
|                0 |              2 | 0000016e0180000002 | X    |
|                0 |              2 | 0000016d80000002   | X    |
|                0 |              2 | 0000016e0180000001 | X    |
|                0 |              2 | 0000016d80000001   | X    |
+------------------+----------------+--------------------+------+
4 rows in set (0.00 sec)
```
so far the same. But TXN 2 will succeed to get the lock:

```
/* TXN 2 */

mysql> begin;

mysql> insert into t1 (a, b) values (10, 10), (20, 20);
Query OK, 2 rows affected (0.00 sec)
Records: 2  Duplicates: 0  Warnings: 0

mysql> select * from information_schema.rocksdb_locks;
+------------------+----------------+--------------------+------+
| COLUMN_FAMILY_ID | TRANSACTION_ID | KEY                | MODE |
+------------------+----------------+--------------------+------+
|                0 |              3 | 0000016e0180000014 | X    |
|                0 |              3 | 0000016e018000000a | X    |
|                0 |              2 | 0000016e0180000002 | X    |
|                0 |              2 | 0000016e0180000001 | X    |
|                0 |              3 | 0000016d80000004   | X    |
|                0 |              3 | 0000016d80000003   | X    |
|                0 |              2 | 0000016d80000002   | X    |
|                0 |              2 | 0000016d80000001   | X    |
+------------------+----------------+--------------------+------+
```
..which looks correct to me

**Comment by spetrunia** [comment]
> Another example:
> 
> ```
> mysql> create table t6 (pk int not null auto_increment, a int, b int, c int, primary key (pk), unique key (a), key (b), unique key (c)) engine=rocksdb;
> Query OK, 0 rows affected (0.03 sec)
>                                           
> mysql> begin;
> Query OK, 0 rows affected (0.00 sec)
> 
> mysql> insert into t6 (a, b, c) values (1, 1, 1), (2, 2, 2);                                                                                                             
> Query OK, 2 rows affected (0.00 sec)                                                                                                                                     
> Records: 2  Duplicates: 0  Warnings: 0    
>                                           
> mysql> select * from information_schema.rocksdb_locks;                                                                                                                   
> +------------------+----------------+---------------------------------------------+------+                                                                               
> | COLUMN_FAMILY_ID | TRANSACTION_ID | KEY                                         | MODE |                                                                               
> +------------------+----------------+---------------------------------------------+------+                                                                               
> |                0 |             21 | 00000113018000000280000002                  | X    |                                                                               
> |                0 |             21 | 0000011080000001-00000113018000000180000001 | X    |                                                                               
> +------------------+----------------+---------------------------------------------+------+                                                                               
> 2 rows in set (0.00 sec)                  
> ```

For me, I get this:
```
mysql> begin;

mysql> insert into t6 (a, b, c) values (1, 1, 1), (2, 2, 2);
Query OK, 2 rows affected (0.00 sec)
Records: 2  Duplicates: 0  Warnings: 0

mysql> 
mysql> select * from information_schema.rocksdb_locks;
+------------------+----------------+----------------------------+------+
| COLUMN_FAMILY_ID | TRANSACTION_ID | KEY                        | MODE |
+------------------+----------------+----------------------------+------+
|                0 |              1 | 00000172018000000280000002 | X    |
|                0 |              1 | 000001710180000002         | X    |
|                0 |              1 | 000001700180000002         | X    |
|                0 |              1 | 0000016f80000002           | X    |
|                0 |              1 | 00000172018000000180000001 | X    |
|                0 |              1 | 000001710180000001         | X    |
|                0 |              1 | 000001700180000001         | X    |
|                0 |              1 | 0000016f80000001           | X    |
+------------------+----------------+----------------------------+------+
8 rows in set (0.00 sec)

mysql>  select index_name, hex(index_number) from information_schema.rocksdb_ddl where table_name = 't6';
+------------+-------------------+
| index_name | hex(index_number) |
+------------+-------------------+
| PRIMARY    | 16F               |
| a          | 170               |
| c          | 171               |
| b          | 172               |
+------------+-------------------+
4 rows in set (0.01 sec)
```
which looks ok at first glance

**Comment by spetrunia** [comment]
Double checking if I am using the right tree... it is https://github.com/spetrunia/mysql-5.6/commits/fb-mysql-8.0.20-range-locking-may 
```
commit 9f29d708d47db4aaf4e015697dd088d2ee408b44 (HEAD -> fb-mysql-8.0.20-range-locking-may, origin/fb-mysql-8.0.20-range-locking-may)
Author: Sergei Petrunia <psergey@askmonty.org>
Date:   Mon Sep 13 20:23:12 2021 +0300

    Range Locking: Make partial index support LockingIterator.

```

storage/rocksdb/ha_rocksdb.cc
spetrunia: Agree, it doesn't sound clear. 

IIRC I had this in mind when writing this: 

With statement-based replication and point locking, one can get the slave 
out of sync. Using gap(or range) locking allows to avoid that. But when 
one uses Gap (or range) locking, the locking reads must read the latest 
committed data.

I guess this should either be stated in full, or not at all...

hermanlee:
I ran my tests with the range locking change based on top of: https://github.com/facebook/mysql-5.6/commit/b22013dba3a9f0d242004f71bcac4427c7249abb.

In your lock outputs, none of the locks look like they were escalated and merged together. What triggers that process?

spetrunia:
lock escalation is triggered by lock tree memory exceeding the limit.  The current lock memory usage should be shown in `rocksdb_current_lock_memory`, and  `rocksdb_max_lock_memory` is where one sets the limit.

hermanlee:
Ah, my setup also inherited these settings from an existing test:

-rocksdb_use_range_locking=1 --rocksdb_max_lock_memory=1024

laurynas-biveinis:
Should this one be closed because of https://github.com/facebook/mysql-5.6/pull/1185 ?

PR #1173, opened 2021-10-07
Fixed PS-7827 (partition table created with RocksDB engine crashes on a select query)

i-rinat:
https://jira.percona.com/browse/PS-7827
https://jira.percona.com/browse/PS-7815

Problem:
Code in `Rdb_key_def::unpack_record()` and `Rdb_key_def::get_data_start_ptr()`
in MyRocks casts `handler*` pointers back to `ha_rocksdb*` assuming that
they do point to `ha_rocksdb` instances. However, when partitioning is in
use, `rocksdb_create_handler()` creates instances of `ha_rockspart`, which
cannot be casted to `ha_rocksdb`. Trying to do so results in memory
corruptions, and sometimes crashes.

Fix:
Extract buffer management parts of `ha_rocksdb` into a separate class, add
it as a mixin to both `ha_rocksdb` and `ha_rockspart`, and then recover from
`handler*` pointers with `dynamic_cast` at run time. RTTI is already enabled
and used in MyRocks, so no build flags change is necessary.

### DIFF
--- a/mysql-test/suite/rocksdb/r/bug_ps7827.result
+++ b/mysql-test/suite/rocksdb/r/bug_ps7827.result
@@ -1,0 +1,1 @@
+bug_ps7827.test
--- a/mysql-test/suite/rocksdb/t/bug_ps7827-master.opt
+++ b/mysql-test/suite/rocksdb/t/bug_ps7827-master.opt
@@ -1,0 +1,1 @@
+--default-storage-engine=rocksdb
--- a/mysql-test/suite/rocksdb/t/bug_ps7827.test
+++ b/mysql-test/suite/rocksdb/t/bug_ps7827.test
@@ -1,0 +1,52 @@
+--source include/have_debug.inc
+--source include/have_rocksdb.inc
+
+# The following statements were triggering memory corruption and subsequent
+# server crash. The defect was caused by dereferencing a pointer to a RocksDB
+# partition handler as a generic RocksDB handler which caused possible
+# writes to other allocations memory areas.
+#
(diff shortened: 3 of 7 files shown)


i-rinat:
Only branches `fb-mysql-8.0.17` and `fb-mysql-8.0.20` seems to be affected by the issue. Others either have no RocksDB, or have no partitioning support for RocksDB engine. This pull request is against `fb-mysql-8.0.17` (at the time of writing). Should I make another pull request for `fb-mysql-8.0.20` too?

If there is some preferred naming scheme that I'm not following, feel free to notify me, so I can make required changes. Or just make changes without notifying. My primary interest is to have the issue fixed.

hermanlee:
Thanks for filing this issue and providing a test case. 8.0.20 is the latest codebase, but we can port this over to that branch.

cc: @yizhang82

i-rinat:
Here is the patch adapted for the `fb-mysql-8.0.20` branch: https://github.com/facebook/mysql-5.6/compare/fb-mysql-8.0.20...i-rinat:ri-PS-7827-rocksdb-ub-dereferences-8.0.20.

i-rinat:
There was input from Percona's side; so I've changed patch a bit according to their comments. Changes are minor: a bit larger comment in the test; a couple of `assert()`'s in the code where pointers from `dynamic_cast<>` are expected to never be `nullptr`. Branch with the patch adaptation for `fb-mysql-8.0.20` was also updated similarly.

inikep:

i-rinat:
0f79c6669e767006f4ef427e2f1d361694df1600

PR #1175, opened 2021-11-02
Fixed PS-7882 (Assertion `m_output_buf->m_curr_offset <= m_output_buf->m_total_size' failed)

i-rinat:
https://jira.percona.com/browse/PS-7882

Problem:
There was a missing size check in `Rdb_index_merge::add()` just after `merge_buf_write()` emptied the unsorted buffer. With extremely small merge buffer sizes, moderately big data could trigger a buffer overflow. Shortly after, a failed assertion terminates the server (debug builds only).

Fix:
Check if the unsorted buffer size is enough to fit a key-value pair right after an unsorted buffer is flushed and cleared.


inikep:

i-rinat:
92f53de5fa9b59657e4b195d4d1855e36814299e

PR #1176, opened 2021-11-02
Fix return type of oveloaded global_value_ptr() method

i-rinat:
Class `sys_var` which is a base for both `Sys_var_gtid_purged_for_tailing` and `Sys_var_gtid_set`, declares return value of `global_value_ptr()` as `const uchar *`, so each overload with exactly same arguments must return the same type.

### DIFF
--- a/sql/sys_vars.h
+++ b/sql/sys_vars.h
@@ -2242,7 +2242,7 @@ class Sys_var_gtid_set : public sys_var {
     global_sid_lock->unlock();
     return (uchar *)buf;
   }
-  uchar *global_value_ptr(THD *thd, LEX_STRING *base) {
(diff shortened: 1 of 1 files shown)


i-rinat:
This error prevents debug builds on at least GCC 9.3.0. I believe that it's actually a warning, but build scripts extensively set `-Werror`, so this warning is treated as an error. And it cannot be muffled by specific `-Wno-error=...` as other warnings.

inikep:

i-rinat:
Yes, it seems that fb-mysql-8.0.23 has this issue fixed.
`Sys_var_gtid_set::global_value_ptr` is still `uchar *` and not expected `const uchar *`. But since that piece of code is hidden behind `HAVE_GTID_NEXT_LIST` feature macro which is not enabled, it doesn't really matter.

PR #1179, opened 2022-01-21
Add clangd version >= 11 index to .gitignore

laurynas-biveinis:
Tweak the existing  comment too, as the index is used for more than just
completion.


laurynas-biveinis:

PR #1181, opened 2022-01-27
Add missing #include <array> directives

laurynas-biveinis:
in MyRocks and patches elsewhere. Their absence is benign on Linux build, but it
produces actual compilation errors on XCode toolchain.

If you plan to squash the non-MyRocks bits with the feature patches, then:
- control_events.h patch goes with e5b0781e4cf9b0b6ac41a515c1f578c19b861715
- sql_admission_control.h with dbbe068c39679bf8d57979626fadde687b1e51e6
- pfs_struct.h with 45019198d5f7350ba1d311f292c888e29a91c9d3


laurynas-biveinis:

PR #1182, opened 2022-02-03
Remove a misleading comment from commit and rollback by XID handlerto…

laurynas-biveinis:
…n methods

The "do nothing" commit originates from
346c587763e0540eac30a633eddd42e0e8e09663, which added the methods as no-op
stubs. They were later properly implemented in
d59142529e587c363eb845baf69f2a938fab96c6, but the comment survived and is
actively misleading.


laurynas-biveinis:

PR #1183, opened 2022-02-03
Fix LLVM codegen for struct st_filenum_pos atomic operations

laurynas-biveinis:
LLVM has an issue where atomic operations on a struct with 32-bit fields are
compiled using libatomic library calls instead of direct assembly, as if the
whole struct were 32-bit aligned, i.e. its objects could cross machine word
boundary: https://bugs.llvm.org/show_bug.cgi?id=45055.

Workaround this issue by aligning the first 32-bit field at 64 bits.

This allows not linking mysys with libatomic.

Squash with 894aea72d40277eaf5636bf37b95e7206c8ff2aa

### DIFF
--- a/mysys/CMakeLists.txt
+++ b/mysys/CMakeLists.txt
@@ -203,10 +203,6 @@ ADD_CONVENIENCE_LIBRARY(mysys ${MYSYS_SOURCES}
   ${SSL_LIBRARIES}
   )
 
-IF(MY_COMPILER_IS_CLANG)
-  TARGET_LINK_LIBRARIES(mysys atomic)
-ENDIF()
-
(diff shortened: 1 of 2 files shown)


laurynas-biveinis:

mpercy:
> Now a part of #1184 

Hi, how should this PR be tested? Does it need to be tested in conjunction with #1184 or can it simply be tested standalone?

laurynas-biveinis:
@mpercy - standalone. PR #1184 contained unrelated changes due to organizational, not technical, reasons

abhinav04sharma:
There is an existing MTR test (mysql-test/suite/rpl_gtid/t/rpl_wait_for_semi_sync_ack.test) which can be used for smoke testing.

PR #1184, opened 2022-02-04
A single PR that has all the previous PRs in separate commits

laurynas-biveinis:
@hermanlee, as discussed yesterday, this is a single PR that has all the previous PRs in separate commits.


laurynas-biveinis:
@hermanlee, https://github.com/facebook/mysql-5.6/pull/1183 reopened and removed from the current PR

PR #1190, opened 2022-04-12
Cleanup: add static to ha_rocksdb.cc declarations

laurynas-biveinis:
Add a few const at the same time, and replace some const with constexpr. Remove
several unused declarations.


hermanlee:
We're seeing a compiler error with release builds:

error: unused function 'rdb_dbug_set_ttl_read_filter_ts' [-Werror,-Wunused-function]
static int rdb_dbug_set_ttl_read_filter_ts();

laurynas-biveinis:
@hermanlee fixed

PR #1212, opened 2022-08-15
Add refresh_iter to release pinned MemTable and SST

rockeet:
During long running trx, there is a rocksdb::Version object is referenced by iterator, thus the **MemTable**s and **SST**s referenced by **Version** can not be released even the **MemTable**s are flushed or **SST**s are compacted.

This PR is intended for releasing MemTable and SST objects held by
rocksdb::Version object which referenced by old rocksdb::Iterator, newly
created Iterator may reference a newer rocksdb::Version object, The data
view of these 2 iterators are identical.

### DIFF
--- a/storage/rocksdb/ha_rocksdb.cc
+++ b/storage/rocksdb/ha_rocksdb.cc
@@ -4061,6 +4061,32 @@ class Rdb_transaction {
     return get_iterator(options, column_family, table_type);
   }
 
+  rocksdb::Iterator *refresh_iterator(
+      const rocksdb::Snapshot* snapshot,
+      rocksdb::ColumnFamilyHandle *const column_family, bool skip_bloom_filter,
+      const rocksdb::Slice &eq_cond_lower_bound,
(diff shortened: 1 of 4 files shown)


yoshinorim:
I read the diff and I think it is better to submit as a RocksDB PR (https://github.com/facebook/rocksdb) as a read option or a wrapper of Iterator, rather than MyRocks specific functions. Would you mind filing an issue in RocksDB (https://github.com/facebook/rocksdb/issues) first and see feedbacks? Snapshot has to be carried over for consistent reads, and your diff looks to take into account that already.

rockeet:
> I read the diff and I think it is better to submit as a RocksDB PR (https://github.com/facebook/rocksdb) as a read option or a wrapper of Iterator, rather than MyRocks specific functions. Would you mind filing an issue in RocksDB (https://github.com/facebook/rocksdb/issues) first and see feedbacks? Snapshot has to be carried over for consistent reads, and your diff looks to take into account that already.

Thank you, I have filed such a RocksDB [issue](https://github.com/facebook/rocksdb/issues/10536)!

PR #1213, opened 2022-08-15
properties_collector: (un)materialize: add missing m_name

rockeet:
This PR bump stats version to `INDEX_STATS_VERSION_WITH_NAME = 3` for added `m_name`.


yoshinorim:
Could you elaborate more about why it is necessary to store index names in SST files? Index name is defined in MySQL data dictionary, so MyRocks can extract index names, map between internal index ids and index names when opening tables.

rockeet:
We have an visualization tool which can view structures inside an SST file, by this tool we found **name** is missing, then we added name in (un)materialize.

image

PR #1218, opened 2022-08-28
set m_read_opts.ignore_range_deletions properly

rockeet:
ignore_range_deletions should be set to true when range del is not enabled


Pushapgl:
whats the motivation of this change?

rockeet:
> whats the motivation of this change?

if dont `ignore_range_deletions`, DB reads will try to construct RangeDelAggregator which waste amount of CPU time without any useful work when `rocksdb_enable_delete_range_for_drop_index` was set to false.

PR #1220, opened 2022-08-31
Disable MyRocks I/O watchdog under macOS

laurynas-biveinis:
macOS does not have the timer_create APIs. It should be possible to port the
feature to dispatch event source APIs, but such effort is hard to justify.


laurynas-biveinis:
Will rebase to fix the conflict

laurynas-biveinis:
Ready for respin

PR #1237, opened 2022-09-05
Disable INFORMATION_SCHEMA.SOCKET_DIAG_SLAVES table on macOS

laurynas-biveinis:
Because macOS does not have the required functionality.

Squash with accb01cc760345cba3f01776c997780935b6546b

### DIFF
--- a/mysql-test/include/mysqlshow.inc
+++ b/mysql-test/include/mysqlshow.inc
@@ -1,6 +1,7 @@
 # Test lists tables in Information_schema, and InnoDB adds some
 # Don't test when thread_pool active
 --source include/not_threadpool.inc
+--source include/not_mac_os.inc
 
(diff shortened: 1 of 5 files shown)


yoshinorim:
This makes the socket_diag_slaves no longer tested from dd_is_compatibility_ci, is_tables_is etc. How about adding 	
"--source include/not_mac_os.inc" on these files and forking another test for mac?

laurynas-biveinis:
@yoshinorim that makes sense. Since macOS is not going to be a first class platform, and any changes go through Linux testing eventually anyway, I am thinking of just adding `--source include/not_mac_os.inc`, and leaving the forking of the test to indeterminate future

laurynas-biveinis:
Ready for respin

PR #1247, opened 2022-10-12
Refactor sql_print_error + abort pair into a new helper fatal_error

laurynas-biveinis:
Removes some repetition from the call sites.

The motivation for this changes comes from clone, where startup time fixup adds many such print-error-and-abort instances.

### DIFF
--- a/storage/rocksdb/rdb_utils.h
+++ b/storage/rocksdb/rdb_utils.h
@@ -219,6 +219,17 @@ inline int purge_all_jemalloc_arenas() {
 #endif
 }
 
+template <typename... Params>
+// format attribute would be great here too, but it's unavailable with the
+// variadic templates
+MY_ATTRIBUTE((cold, noreturn,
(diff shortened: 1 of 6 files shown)


storage/rocksdb/rdb_utils.h
```diff
@@ -219,6 +219,16 @@ inline int purge_all_jemalloc_arenas() {
 #endif
 }
 
+template <typename... Params>
+// format attribute would be great here too, but it's unavailable with the
+// variadic templates
+MY_ATTRIBUTE((cold, noreturn, noinline)) void fatal_error(const char *fmt,
```
hermanlee: Can you rename this to rdb_fatal_error()?

laurynas-biveinis: Done

(the reason I did not add `rdb_` prefix before was that I thought namespace `myrocks::` did its job)

PR #1248, opened 2022-10-13
Add a MyRocks testcase for non-default rocksdb-wal-dir value

laurynas-biveinis:
No such test currently exists.

### DIFF
--- a/mysql-test/suite/rocksdb/t/wal_dir.test
+++ b/mysql-test/suite/rocksdb/t/wal_dir.test
@@ -1,0 +1,30 @@
+#
+# Test MyRocks with non-default rocksdb-wal-dir setting
+#
+--source include/have_rocksdb.inc
+
+--let WAL_DIR=$MYSQL_TMP_DIR/rdb_wal
+
+call mtr.add_suppression("dir NotFound");
+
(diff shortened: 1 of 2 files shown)


laurynas-biveinis:
This testcase happens to be a regression testcase for clone

mysql-test/suite/rocksdb/t/wal_dir.test
```diff
@@ -0,0 +1,26 @@
...
+--let $restart_parameters=restart:--rocksdb-wal-dir=$WAL_DIR
+--source include/restart_mysqld.inc
+
+CREATE TABLE t1 (a INT PRIMARY KEY, b VARCHAR(50)) ENGINE=ROCKSDB;
+
+INSERT INTO t1 VALUES (1, "foo");
+INSERT INTO t1 VALUES (2, "bar");
+
+SELECT * FROM t1;
+
```
hermanlee, line 20: Overall, this looks good. Could you also add in

```
SET GLOBAL rocksdb_force_flush_memtable_now=1;

SELECT * FROM t1;
```

So that we add a little more WAL related functionality.

laurynas-biveinis: Done

PR #1251, opened 2022-10-27
Extract a helper myrocks::file_in

laurynas-biveinis:
This avoids temporary strings and memory re-allocations.

At the same time replace some uses of "/" with '/', where possible.

### DIFF
--- a/storage/rocksdb/rdb_utils.h
+++ b/storage/rocksdb/rdb_utils.h
@@ -289,6 +289,9 @@ std::string rdb_hexdump(const char *data, const std::size_t data_len,
                         const std::size_t maxsize = 0)
     MY_ATTRIBUTE((__nonnull__));
 
+// Return dir + '/' + file
+std::string rdb_concat_paths(const std::string &dir, const std::string &file);
+
 /*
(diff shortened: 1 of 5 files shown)


storage/rocksdb/rdb_utils.h
```diff
@@ -289,6 +289,9 @@ std::string rdb_hexdump(const char *data, const std::size_t data_len,
                         const std::size_t maxsize = 0)
     MY_ATTRIBUTE((__nonnull__));
 
+// Return dir + '/' + file
+std::string file_in(const std::string &dir, const std::string &file);
```
hermanlee: Maybe have a more descriptive name that is consistent with the other helper functions?  Maybe rdb_concat_filepaths?

laurynas-biveinis: Renamed to `rdb_concat_paths` (let me know if you prefer `rdb_concat_filepaths` or `rdb_concat_file_paths` instead)

(the reason I've been avoiding the `rdb_` prefix was that I thought that the `myrocks::` namespace did the same job)

PR #1255, opened 2022-11-17
MyRocks to support both dynamic and static plugins

yoshinorim:
Summary:
Prior to this diff, fb-mysql MyRocks only supported static plugins, and used sql_print_* APIs for logging, which was not available for dynamic plugins.

This diff allows to create either static or dynamic plugins, depending on -DMYSQL_DYNAMIC_PLUGIN. In addition to that, adding another option to skip creating MyRocks plugin by -DWITHOUT_ROCKSDB_STORAGE_ENGINE.

Without -DMYSQL_DYNAMIC_PLUGIN, it creates static plugin, basically the same behavior as before.
With -DMYSQL_DYNAMIC_PLUGIN, it creates dynamic plugin, and using plugin logging functions (LogPluginErr). Linking with other libraries such as jemalloc is adjusted accordingly.

Test Plan: mtr with dynamic builds

### DIFF
--- a/mysql-test/suite/rocksdb/r/truncate_failures.result
+++ b/mysql-test/suite/rocksdb/r/truncate_failures.result
@@ -89,7 +89,7 @@ pk
 SET DEBUG = "+d,rocksdb_truncate_failure_crash";
 TRUNCATE TABLE t1_crash;
 ERROR HY000: Lost connection to MySQL server during query
- [MY-000000
+1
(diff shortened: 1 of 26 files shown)


laurynas-biveinis:
Please close, done in https://github.com/facebook/mysql-5.6/commit/2a347f6520235a6c278f505ea456bf18c5dba9a7

PR #1256, opened 2022-11-21
Rocksdb valgrind

laurynas-biveinis:


laurynas-biveinis:
Need to re-evaluate after https://github.com/facebook/mysql-5.6/commit/aa84ea9a948157011eab5d444b3a98e23d928107

cmake/compiler_features.cmake
```diff
@@ -271,4 +271,12 @@ MACRO(ROCKSDB_SET_DEFINTIONS)
 	if(HAVE_AUXV_GETAUXVAL AND NOT ROCKSDB_DISABLE_AUXV_GETAUXVAL)
 	  add_definitions(-DROCKSDB_AUXV_GETAUXVAL_PRESENT)
 	endif()
+
+        if(WITH_UBSAN)
+          add_definitions(-DROCKSDB_UBSAN_RUN)
```
lth, line 276: Do you have more context on how these flags are consumed?

laurynas-biveinis: `ROCKSDB_UBSAN_RUN` appears to suppress UBSan warnings for misaligned pointers for the sites where RocksDB does that intentionally. 
`ROCKSDB_VALGRIND_RUN` makes RocksDB to release more of allocated memory on shutdown.

PR #1260, opened 2023-01-16
Fix -Wformat-security warning for variadic templates

laurynas-biveinis:
If -Wformat-security is enabled,
rdb_fatal_error(... "not a printf format string")
gives
rdb_utils.h:254:27: error: format string is not a string literal (potentially insecure) [-Werror,-Wformat-security]
  myrocks_sql_print_error(fmt, std::forward<Params>(params)...);

Fix by introducing an rdb_fatal_error overload for messages without format strings, i.e. the ones which could serve as a parameter for a "%s" format string.

At the same time at a format function attribute for rdb_fatal_error for supported compilers, which at the moment are clang >= 15.

### DIFF
--- a/storage/rocksdb/CMakeLists.txt
+++ b/storage/rocksdb/CMakeLists.txt
@@ -181,9 +181,6 @@ SET(ROCKSDB_SOURCES
   ${ROCKSDB_LIB_SOURCES}
 )
 
-# To avoid compiler error on rdb_fatal_error() in rdb_util.h
-add_compile_flags(${ROCKSDB_SOURCES} COMPILE_FLAGS "-Wno-format-security")
-
 IF(WITH_FB_TSAN OR ROCKSDB_DYNAMIC_PLUGIN)
   SET(PIC_EXT "_pic")
 ELSE()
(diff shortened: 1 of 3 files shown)


laurynas-biveinis:
The LLVM >= 15 branch is not tested because MySQL 8.0.28 requires many fixes to compile with it elsewhere. Still, I believe it's a good thing to keep to get the format string mismatch warnings once the fork rebases on a sufficiently new MySQL version.

yoshinorim:
Thanks for the patch. With your fix, couldn't you remove these two lines from storage/rocksdb/CMakeLists.txt?

```
# To avoid compiler error on rdb_fatal_error() in rdb_util.h
add_compile_flags(${ROCKSDB_SOURCES} COMPILE_FLAGS "-Wno-format-security")
```

laurynas-biveinis:
> Thanks for the patch. With your fix, couldn't you remove these two lines from storage/rocksdb/CMakeLists.txt?
> 
> ```
> # To avoid compiler error on rdb_fatal_error() in rdb_util.h
> add_compile_flags(${ROCKSDB_SOURCES} COMPILE_FLAGS "-Wno-format-security")
> ```

Done - forgot the `git add` in the inital push :)

PR #1267, opened 2023-01-29
Fix rdb_i_s_sst_props_fields_info

rockeet:
In `rdb_i_s_sst_props_fields_info` definition, flag `MY_I_S_MAYBE_NULL` is missing for fields `COMPRESSION_ALGO`, `FILTER_POLICY`, `COMPRESSION_OPTIONS`, causing the `set_null` does not take effect, these 3 fields will keep last SST's value.

This PR fixed this issue.


hermanlee:
Thank you for the PR to fix this issue!

PR #1275, opened 2023-03-01
Commit sst in dedicated threads

rockeet:
Execute `Rdb_sst_info::commit_sst_file` in dedicated threads, this improves performance:
1. `Rdb_sst_file_ordered::commit` may use stack to reverse input data, this is time consuming
2. `m_sst_file_writer->Finish` may be consuming, at least it need to call `fsync`

Execute `Rdb_sst_info::commit_sst_file` in dedicated threads increase parallelization with mimimal code changes.

### DIFF
--- a/storage/rocksdb/rdb_sst_info.cc
+++ b/storage/rocksdb/rdb_sst_info.cc
@@ -483,21 +483,17 @@ int Rdb_sst_info::finish(Rdb_sst_commit_info *commit_info,
     return ret;
   }
 
-  auto join_commiting_threads = [this]() {
-    for (auto& thr : m_commiting_threads) {
-      thr.join();
-    }
-    m_commiting_threads.clear();
-  };
-  join_commiting_threads();
-
   m_print_client_error = print_client_error;
 
   if (m_curr_size > 0) {
     // Close out any existing files
     close_curr_sst_file();
-    join_commiting_threads();
   }
+
+  for (auto& thr : m_commiting_threads) {
+    thr.join();
+  }
+  m_commiting_threads.clear();
 
   // This checks out the list of files so that the caller can collect/group
   // them and ingest them all in one go, and any racing calls to commit
--- a/storage/rocksdb/rdb_sst_info.cc
+++ b/storage/rocksdb/rdb_sst_info.cc
@@ -355,6 +355,7 @@ Rdb_sst_info::Rdb_sst_info(rocksdb::DB *const db, const std::string &tablename,
 
 Rdb_sst_info::~Rdb_sst_info() {
   assert(m_sst_file == nullptr);
+  SHIP_ASSERT(m_commiting_threads.empty());
 
   for (const auto &sst_file : m_committed_files) {
     // In case something went wrong attempt to delete the temporary file.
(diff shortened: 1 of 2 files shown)


storage/rocksdb/rdb_sst_info.cc
```diff
@@ -392,13 +393,23 @@ int Rdb_sst_info::open_new_sst_file() {
 }
 
 void Rdb_sst_info::commit_sst_file(Rdb_sst_file_ordered *sst_file) {
+  m_commiting_threads_mutex.lock();
```
bladepan, line 396: consider std::lock_guard<std::mutex>? we use `RDB_MUTEX_LOCK_CHECK` macros elsewhere in this file

storage/rocksdb/rdb_sst_info.cc
```diff
@@ -479,6 +490,11 @@ int Rdb_sst_info::finish(Rdb_sst_commit_info *commit_info,
     close_curr_sst_file();
   }
 
+  for (auto& thr : m_commiting_threads) {
```
bladepan, line 493: do we need to grab a log here too?

bladepan:
> Execute `Rdb_sst_info::commit_sst_file` in dedicated threads, this improves performance:
> 
>     1. `Rdb_sst_file_ordered::commit` may use stack to reverse input data, this is time consuming
> 
>     2. `m_sst_file_writer->Finish` may be consuming, at least it need to call `fsync`
> 
> 
> Execute `Rdb_sst_info::commit_sst_file` in dedicated threads increase parallelization with mimimal code changes.

1. the keys are already ordered, why it would reverse input data?
2. the `fsync` happens when writing sst file to disk? 

could you please provide some rough calculations about the time savings?

rockeet:
> > Execute `Rdb_sst_info::commit_sst_file` in dedicated threads, this improves performance:
> > ```
> > 1. `Rdb_sst_file_ordered::commit` may use stack to reverse input data, this is time consuming
> > 
> > 2. `m_sst_file_writer->Finish` may be consuming, at least it need to call `fsync`
> > ```
> > Execute `Rdb_sst_info::commit_sst_file` in dedicated threads increase parallelization with mimimal code changes.
> 
> 1. the keys are already ordered, why it would reverse input data?
> 2. the `fsync` happens when writing sst file to disk?
> 
> could you please provide some rough calculations about the time savings?

1. In `Rdb_sst_file_ordered::commit()`, if `m_use_stack` is true, it will write kv from stack.
2. In `Rdb_sst_file_ordered::Rdb_sst_file::commit()`, it calls `m_sst_file_writer->Finish()`, in which will write file and call fsync.

In our private branch, when `m_use_stack` is true(reverse cf), `create index` time reduces 30+%.

storage/rocksdb/rdb_sst_info.cc
bladepan: >     1. In `Rdb_sst_file_ordered::commit()`, if `m_use_stack` is true, it will write kv from stack.
> 
>     2. In `Rdb_sst_file_ordered::Rdb_sst_file::commit()`, it calls `m_sst_file_writer->Finish()`, in which will write file and call fsync.
> 
> 
> In our private branch, when `m_use_stack` is true(reverse cf), `create index` time reduces 30+%.

Rdb_index_merge and Rdb_sst_file_ordered both use cf's comparator, when Rdb_index_merge write to Rdb_sst_file_ordered, the keys should be in cf's increasing order, m_use_stack should be false in this case.

bladepan:
>     1. In `Rdb_sst_file_ordered::commit()`, if `m_use_stack` is true, it will write kv from stack.
> 
>     2. In `Rdb_sst_file_ordered::Rdb_sst_file::commit()`, it calls `m_sst_file_writer->Finish()`, in which will write file and call fsync.
> 
> 
> In our private branch, when `m_use_stack` is true(reverse cf), `create index` time reduces 30+%.

Rdb_index_merge and Rdb_sst_file_ordered both use cf's comparator, when Rdb_index_merge write to Rdb_sst_file_ordered, the keys should be in cf's increasing order, m_use_stack should be false in this case.

rockeet:
> > ```
> > 1. In `Rdb_sst_file_ordered::commit()`, if `m_use_stack` is true, it will write kv from stack.
> > 
> > 2. In `Rdb_sst_file_ordered::Rdb_sst_file::commit()`, it calls `m_sst_file_writer->Finish()`, in which will write file and call fsync.
> > ```

> > In our private branch, when `m_use_stack` is true(reverse cf), `create index` time reduces 30+%.
> 
> Rdb_index_merge and Rdb_sst_file_ordered both use cf's comparator, when Rdb_index_merge write to Rdb_sst_file_ordered, the keys should be in cf's increasing order, m_use_stack should be false in this case.

We removed `merge_record::m_comparator` since it is identical in each `std::set` element, and using `Slice::operator<` to avoid `rocksdb::Comparator` virtual function call, thus it is reversed with regard to rev cf and makes `m_use_stack` being true.

PR #1008, opened 2019-03-29
FB8-221: Change SET GLOBAL [super_]read_only to not acquire the global read lock

dutow:
Summary:

Jira ticket: https://jira.percona.com/browse/FB8-221

Reference commit: https://github.com/facebook/mysql-5.6/commit/d8657c3d
Reference commit: https://github.com/facebook/mysql-5.6/commit/a3f808d
Reference commit: https://github.com/facebook/mysql-5.6/commit/f2dfbd0

-------- a3f808d --------

Change SET GLOBAL [super_]read_only to not acquire the global read lock

Currently calling SET GLOBAL [super_]read_only acquires the global read lock before setting the flag and this will block on a long running DML statement.  We think it is safe in our environment to not acquire this lock.

This behavior can be changed back to the original by SET @global.legacy_global_read_lock_mode=1

PR #1276, opened 2023-03-04
Make blackhole engine transactional

alanliang:
Summary: 

The main motivation for this diff is that we want to be able to set super_read_only and read_only=1 very quickly. Normally, when there are ongoing transactions happening, setting the read only vars can time out. This is problematic for promotions in both semi-sync and raft.

The timeout is caused by long running transactions holding on to the read lock. In 2016, we introduced a sysvar legacy_global_read_lock_mode in MySQL 5.6 ([Change SET GLOBAL [super_]read_only to not acquire the global read lock](https://github.com/facebook/mysql-5.6/pull/1008/files)) to allow us to skip acquiring the read lock when turning on read only variables. It would skip the read lock if legacy_global_read_lock_mode=0  This would mean that turning read only would be instant and any ongoing transactions would eventually die at commit.

However, it was deemed not safe to skip the read lock because DDL statements could still be completed despite read only vars being turned on. Instead, the team decided to implement the strategy of killing of conflicting connections ([Add kill_conflicting_connections option for admin user](https://github.com/facebook/mysql-5.6/commit/f2ed7c3e4ce)) after read only is turned on to mitigate the long running transactions problem. It is only a mitigation because after killing the transactions, we still have to wait for the rollback before the read lock is released and this can be slow and still result in a timeout.

In MySQL 8.0, transactional DDL was supported. This is great because DDL statements would be treated the same way as DML statements. So if read-only was enabled and there were an ongoing DDL statement, it will fail at commit. This means that we can safely utilize legacy_global_read_lock_mode=0 and skip the read lock when setting read only vars.
However, the catch is that in MySQL 8.0, DML's in non-transactional engines like Blackhole engine, do not fail when legacy_global_read_lock_mode=0 and read-only is turned on. Specifically, when read-only is turned on, in-flight DMLs do not get blocked at commit time and are able to go through. This means that DML statements can be lost since they are seen as committed, despite read-only being on.

To address this, we can make Blackhole engine transactional. 

Test Plan:

- Added tests

Reviewers: 

Subscribers: 

Tasks: 

Tags: 

Differential Revision: https://phabricator.intern.facebook.com/D43785777


laurynas-biveinis (commented):
The core changes in `external_lock` look fine. I am not sure if that covers all the code paths, I'd suggest to add MTR tests and make code changes if needed (I am not saying any changes will be needed, just that these are some things to check):
- `START TRANSACTION WITH CONSISTENT SNAPSHOT` - does `blackhole_hton` need `start_consistent_snapshot` which would also call `trans_register_ha`?
- Blackhole DML in `START TRANSACTION WITH SHARED INNODB|ROCKSDB SNAPSHOT` transactions (`blackhole_hton::start_shared_snapshot`)?
- Blackhole DML under `LOCK TABLES` (`ha_blackhole::start_stmt`). Source code comments say that `external_lock` will not be called in this case.

Interestingly this work should require removing `HA_NO_TRANSACTIONS` from `ha_blackhole::table_flags` but it's already absent there.

alanliang:
> * `START TRANSACTION WITH CONSISTENT SNAPSHOT` - does `blackhole_hton` need `start_consistent_snapshot` which would also call `trans_register_ha`?

In all places where start_consistent_snapshot() is called, there's a check to see whether the function pointer is defined first. For blackhole, which doesn't store data, it doesn't make sense to implement consistent snapshots. `START TRANSACTION WITH CONSISTENT SNAPSHOT` will act as a normal transaction. It just won't create a snapshot for blackhole engine.

> * Blackhole DML in `START TRANSACTION WITH SHARED INNODB|ROCKSDB SNAPSHOT` transactions (`blackhole_hton::start_shared_snapshot`)?

Similar to previous, there's a check for the function pointer before it's called. Again, it doesn't make sense to implement shared snapshots. I added tests to make sure it fails properly.

> * Blackhole DML under `LOCK TABLES` (`ha_blackhole::start_stmt`). Source code comments say that `external_lock` will not be called in this case.

This is a good catch. I added ha_blackhole::start_stmt and added tests. Before adding this, commit while read-only lock is set would just work. With ha_blackhole::start_stmt implemented, commit will now fail.

> Interestingly this work should require removing `HA_NO_TRANSACTIONS` from `ha_blackhole::table_flags` but it's already absent there.

I noticed that too! However, when I issue a `SHOW ENGINES`, it correctly says that BLACKHOLE now supports transactions.

mysql-test/t/blackhole.test
```diff
@@ -41,3 +41,43 @@ SELECT MATCH (a) AGAINST ('abc') AS score FROM t;
 # But this index scan failed. Should succeed now.
 SELECT 1 FROM t WHERE MATCH (a) AGAINST ('abc');
 DROP TABLE t;
+
+--echo # Test Transaction support
+connect(con1,localhost,root,,);
```
laurynas-biveinis: Looks like an unused connection

laurynas-biveinis (commented):
LGTM, with one minor nit

alanliang:
I recently added a plugin sysvar so that we can enable/disable it. It will default to be disabled for safety.

mysql-test/suite/funcs_1/r/is_engines_blackhole.result
```diff
@@ -3,6 +3,6 @@ WHERE ENGINE = 'BLACKHOLE';
 ENGINE	BLACKHOLE
 SUPPORT	YES
 COMMENT	/dev/null storage engine (anything you write to it disappears)
-TRANSACTIONS	NO
+TRANSACTIONS	YES
```
laurynas-biveinis, line 6: Is this with the default blackhole_transactions=OFF?

storage/blackhole/ha_blackhole.cc
```diff
@@ -363,6 +394,16 @@ static int blackhole_fini(void *) {
 struct st_mysql_storage_engine blackhole_storage_engine = {
     MYSQL_HANDLERTON_INTERFACE_VERSION};
 
+static MYSQL_SYSVAR_BOOL(transactions, blackhole_transactions,
+                         PLUGIN_VAR_OPCMDARG,
+                         "Enables transaction support for Blackhole engine",
+                         /*check=*/nullptr,
+                         /*update=*/nullptr,
+                         /*default=*/blackhole_transactions);
```
laurynas-biveinis, line 402: While this should just take value from the variable (`false`, but even that requires knowing that statics are zero-initialized), the sysvar definition looks circular superficially.
```suggestion
                         /*default=*/false);
```

storage/blackhole/ha_blackhole.cc
```diff
@@ -40,6 +40,7 @@ using std::string;
 using std::unique_ptr;
 
 static PSI_memory_key bh_key_memory_blackhole_share;
+static bool blackhole_transactions;
```
laurynas-biveinis, line 43: For clarity
```suggestion
static bool blackhole_transactions = false;
```

mysql-test/t/blackhole.test
```diff
@@ -41,3 +41,38 @@ SELECT MATCH (a) AGAINST ('abc') AS score FROM t;
 # But this index scan failed. Should succeed now.
 SELECT 1 FROM t WHERE MATCH (a) AGAINST ('abc');
 DROP TABLE t;
+
```
laurynas-biveinis, line 44: This only tests non-transactional blackhole: please set the plugin var first.

Consider converting this snippet to an include file, and testing both false and true for the sysvar.

mysql-test/t/global_read_lock.inc
```diff
@@ -1,8 +1,14 @@
+--echo # Test setting read only with long running transactions engine=$engine,
```
laurynas-biveinis, line 1: Does this need blackhole_transactions=ON?

storage/blackhole/ha_blackhole.cc
alanliang: Makes sense. I initially tried to avoid having a default set in multiple places.

storage/blackhole/ha_blackhole.cc
alanliang: Yes, perhaps it is OK as it does technically have transaction support but not enabled. However, I can also see the other side where it can be confusing.

mysql-test/t/global_read_lock.inc
alanliang: Yes, perhaps it is OK as it does technically have transaction support but not enabled. However, I can also see the other side where it can be confusing.

mysql-test/t/blackhole.test
alanliang: It looks like even before implementing transactions, BEGIN TRANSACTION and COMMIT were allowed and it would just no-op. So with blackhole_transactions on or off, they both yielded the same test results. 

Do you know of a way to actually ensure commit does something?

I tried to use SHOW BINLOG EVENTS via --source include/show_binlog_events.inc but whether blackhole_transactions is on or off, it produces the same binlog events. What's curious is that, even with:

START TRANSACTION;
... 
ROLLBACK;

they produce the same binlog events. So it seems like rollback works even without transaction support?

laurynas-biveinis: > Do you know of a way to actually ensure commit does something?

Check the binlog comment below. If that does not help, maybe the `Handler_commit` status variable will be bumped more? If that doesn't help, I would still consider testing both variable settings even without user-visible changes between them: it would test that the server does not assert etc.

> I tried to use SHOW BINLOG EVENTS via --source include/show_binlog_events.inc but whether blackhole_transactions is on or off, it produces the same binlog events. What's curious is that, even with:
> 
> START TRANSACTION; ... ROLLBACK;
> 
> they produce the same binlog events. So it seems like rollback works even without transaction support?

Could you please re-check whether that's the case with transactional tables? AFAIK, binlog has different behavior for transactions that mix-in non-transactional SE changes:
1) If all SEs in a transaction are transactional, then the transaction gets binlogged on COMMIT, and on ROLLBACK nothing gets binlogged
2) If there is at least one non-transcational SE in a transaction, then the transaction gets binlogged (immediately after each statement? not sure) always, and on ROLLBACK the ROLLBACK statement itself gets binlogged. This way replicas can try to reproduce what happened on the primary.

The behavior you described matches 2) but with transactional Blackhole 1) must also happen.

alanliang: What I find interesting is that without my changes to the blackhole engine, start transaction and rollback seems to work, as in, nothing gets binlogged due to the rollback. Looking at the blackhole documentation, this statement seems to hold true in my tests:

"The BLACKHOLE engine is transaction-aware, in the sense that committed transactions are written to the binary log and rolled-back transactions are not."
https://dev.mysql.com/doc/refman/8.0/en/blackhole-storage-engine.html

I tested another non-transactional engine such as Memory and found that the rollback statement does nothing. So, there's something in blackhole implementation that is making it "transaction aware", even before my changes.

However, the most important thing we add with my changes is that blackhole transactions actually block on commit, which it didn't before.

alanliang: I'll leave the tests as-is as it is still valuable since ensures we don't crash with and without transactions turned on.

alanliang:
Already committed: https://github.com/facebook/mysql-5.6/commit/d1d23e7c7ca2deb6012cd12545993fb29ea01bfc

PR #1277, opened 2023-03-07
Add support for DSCP (Differentiated Services Field) to MySQL Clone Plugin for remote clone process

akshaysuryawanshi:
MySQL Clone Plugin allows for faster data copy using multi-threaded process. Setting DSCP fields on sockets that are created using randomly selected ports is quite difficult and will always be delayed due to fetching of the socket information from MySQL processlist or other methods. With clone_dscp_on_socket MySQL Clone plugin variable we are able to set the DSCP value on the socket as soon as they are created for remote transfers.

### DIFF
--- a/plugin/clone/src/clone_plugin.cc
+++ b/plugin/clone/src/clone_plugin.cc
@@ -85,6 +85,9 @@ static char *clone_ssl_ca;
 /** Clone system variable: timeout for clone restart after n/w failure */
 uint clone_restart_timeout;
 
+/** Clone system variable: DSCP value to set on socket */
+static uint clone_dscp_on_socket;
+
 /** Key for registering clone allocations with performance schema */
 PSI_memory_key clone_mem_key;
 
@@ -517,6 +520,40 @@ static int plugin_clone_remote_client(THD *thd, const char *remote_host,
 @param[in]	socket	network socket to remote client
 @return error code */
 static int plugin_clone_remote_server(THD *thd, MYSQL_SOCKET socket) {
+  int dscp_val = clone_dscp_on_socket;
+
+  if (dscp_val != 0) {
+    int tos = dscp_val << 2;
+    uint16_t test_family;
+    socklen_t len = sizeof(test_family);
+    int res =
+        mysql_socket_getsockopt(socket, SOL_SOCKET, SO_DOMAIN,
+                                reinterpret_cast<void *>(&test_family), &len);
+
+    if (res != 0) {
+      // NO_LINT_DEBUG
+      LogPluginErr(
+          WARNING_LEVEL, ER_CLONE_DONOR,
+          "Failed to get socket domain while adjusting DSCP_QOS on donor");
+    } else if (test_family == AF_INET6) {
+      res = mysql_socket_setsockopt(socket, IPPROTO_IPV6, IPV6_TCLASS, &tos,
+                                    sizeof(tos));
+    } else if (test_family == AF_INET) {
+      res = mysql_socket_setsockopt(socket, IPPROTO_IP, IP_TOS, &tos,
+                                    sizeof(tos));
+    } else {
+      // NO_LINT_DEBUG
+      LogPluginErr(
+          WARNING_LEVEL, ER_CLONE_DONOR,
+          "Failed to get socket family while adjusting DSCP_QOS on donor");
+    }
+
+    if (res != 0) {
+      // NO_LINT_DEBUG
+      LogPluginErr(WARNING_LEVEL, ER_CLONE_DONOR,
+                   "Failed to set TOS/TCLASS on donor");
+    }
+  }
   myclone::Server clone_inst(thd, socket);
 
   auto err = clone_inst.clone();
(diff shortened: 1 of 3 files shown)


mysql-test/suite/sys_vars/t/clone_dscp_on_socket_basic.test
```diff
@@ -0,0 +1,125 @@
+##################### dscp_on_socket_basic.test ###############################
```
laurynas-biveinis: Plugin sys var tests do no go to the `sys_var` suite, usually there are just functional tests added as needed in the specific suites (`clone` and/or `rocksdb_clone` here). In this case a functional test is not possible if I understand correctly, so no need for this test. You can check by running MTR `sys_vars.all_vars` test and if it fails, checking its failure output for any mentions of this sysvar.

In the case this test is needed, it looks like one of the verbose templates for a sysvar test has been used here. It is OK to leave it as-is, but in general these tests can be brief and with fewer comments

plugin/clone/include/clone.h
```diff
@@ -41,6 +41,9 @@ Clone Plugin: Common objects and interfaces
 #include "mysql/components/services/log_builtins.h"
 #include "violite.h"
 
+#define DEFAULT(X) X
```
laurynas-biveinis: These two are only used in the .cc file, so should be declared there. But for these particular macros, no need to have them at all I believe, i.e. the DEFAULT one already has a comment next to it at the use place.

plugin/clone/include/clone.h
```diff
@@ -111,6 +114,9 @@ const uint CLONE_MIN_BLOCK = 1024 * 1024;
 /** Minimum network packet. Safe margin for meta information */
 const uint CLONE_MIN_NET_BLOCK = 2 * CLONE_MIN_BLOCK;
 
+/** Clone system variable: DSCP value to set on socket */
+extern uint clone_dscp_on_socket;
```
laurynas-biveinis: Only used in the .cc file, no need for this declaration

plugin/clone/src/clone_plugin.cc
```diff
@@ -85,6 +85,9 @@ static char *clone_ssl_ca;
 /** Clone system variable: timeout for clone restart after n/w failure */
 uint clone_restart_timeout;
 
+/** Clone system variable: DSCP value to set on socket */
+uint clone_dscp_on_socket;
```
laurynas-biveinis:
```suggestion
static uint clone_dscp_on_socket;
```

plugin/clone/src/clone_plugin.cc
```diff
@@ -517,6 +520,36 @@ static int plugin_clone_remote_client(THD *thd, const char *remote_host,
 @param[in]	socket	network socket to remote client
 @return error code */
 static int plugin_clone_remote_server(THD *thd, MYSQL_SOCKET socket) {
+  int dscp_val = clone_dscp_on_socket;
+
+  bool set_dscp = false;
+  if (dscp_val != 0) {
+    set_dscp = true;
+  }
```
laurynas-biveinis: `const auto set_dscp = (dscp_val != 0);`
or just check directly below
`if (dscp_val != 0) {`

plugin/clone/src/clone_plugin.cc
```diff
@@ -517,6 +520,36 @@ static int plugin_clone_remote_client(THD *thd, const char *remote_host,
...
+
+  if (set_dscp) {
+    int tos = dscp_val << 2;
+    uint16_t test_family;
+    socklen_t len = sizeof(test_family);
+    int res = mysql_socket_getsockopt(socket, SOL_SOCKET, SO_DOMAIN, reinterpret_cast<void *>(&test_family), &len);
+
+    if (res != 0) {
+    // NO_LINT_DEBUG
```
laurynas-biveinis: Indentation

plugin/clone/src/clone_plugin.cc
```diff
@@ -517,6 +520,36 @@ static int plugin_clone_remote_client(THD *thd, const char *remote_host,
...
+      my_message(ER_CLONE_DONOR, "Failed to get socket domain while adjusting DSCP_QOS on donor", MYF(0));
```
laurynas-biveinis: I believe all these `my_message` calls should be something else, e.g.  `my_error` followed by `LogPluginErr(ERROR_LEVEL, ER_CLONE_CLIENT_TRACE, ... )` - I'd suggest to simulate a failure here and see what happens, both what is returned to the client and what is printed to the error log.

plugin/clone/src/clone_plugin.cc
```diff
@@ -517,6 +520,31 @@ static int plugin_clone_remote_client(THD *thd, const char *remote_host,
...
+      res = mysql_socket_setsockopt(socket, IPPROTO_IPV6, IPV6_TCLASS, &tos, sizeof(tos));
+    } else if (test_family == AF_INET) {
+      res = mysql_socket_setsockopt(socket, IPPROTO_IP, IP_TOS, &tos, sizeof(tos));
+    } else {
+      // NO_LINT_DEBUG
+      LogPluginErr(WARNING_LEVEL, ER_CLONE_DONOR, "Failed to get socket family while adjusting DSCP_QOS on donor");
+    }
+
+    if (res != 0) {
+    // NO_LINT_DEBUG
```
laurynas-biveinis: Indentation (you can also use `clang-format` version 10)

laurynas-biveinis:
LGTM

laurynas-biveinis:
This PR should be closed

PR #1279, opened 2023-03-09
Add dd default engine

sunshine-Chun:
Summary:
Add a read only global sys variable default_dd_storage_engine
Test Plan:
MTR test
Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/mysql-test/r/mysqld--help-notwin.result
+++ b/mysql-test/r/mysqld--help-notwin.result
@@ -2461,6 +2461,8 @@ The following options may be given as the first argument:
  WriteOptions::ignore_missing_column_families for RocksDB
  --rocksdb-write-policy=name 
  DBOptions::write_policy for RocksDB
+ --rocksdb-wsenv-oncall=name 
+ Oncall for RocksDB WSEnv
  --rocksdb-wsenv-path=name 
  Path for RocksDB WSEnv
  --rocksdb-wsenv-tenant=name 
@@ -3733,6 +3735,7 @@ rocksdb-write-batch-max-bytes 0
 rocksdb-write-disable-wal FALSE
 rocksdb-write-ignore-missing-column-families FALSE
 rocksdb-write-policy write_committed
+rocksdb-wsenv-oncall 
 rocksdb-wsenv-path 
 rocksdb-wsenv-tenant 
 rpl-read-size 8192
--- a/mysql-test/suite/rocksdb_sys_vars/r/rocksdb_wsenv_oncall_basic.result
+++ b/mysql-test/suite/rocksdb_sys_vars/r/rocksdb_wsenv_oncall_basic.result
@@ -1,0 +1,7 @@
+SET @start_global_value = @@global.ROCKSDB_WSENV_ONCALL;
+SELECT @start_global_value;
+@start_global_value
+
+"Trying to set variable @@global.ROCKSDB_WSENV_ONCALL to 444. It should fail because it is readonly."
+SET @@global.ROCKSDB_WSENV_ONCALL   = 444;
+ERROR HY000: Variable 'rocksdb_wsenv_oncall' is a read only variable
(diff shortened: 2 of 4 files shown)


sql/mysqld.cc
```diff
@@ -7669,6 +7671,9 @@ static int init_server_components() {
   if (initialize_storage_engine(default_tmp_storage_engine, " temp",
                                 &global_system_variables.temp_table_plugin))
     unireg_abort(MYSQLD_ABORT_EXIT);
+  if (initialize_storage_engine(default_dd_storage_engine, " data dictionary",
```
laurynas-biveinis: Unfortunately this will not work, and at this stage it is too early to tell the correct fix. The default DD engine is special in that it has to be loaded significantly earlier in the startup process and its load method is diferent (`initialize_storage_engine` calls `ha_resolve_by_name` - which needs `mysql.plugin` table already accessible).

For example, in my experimental branch I load the plugin in a different way: https://github.com/laurynas-biveinis/mysql-5.6/commit/5838003fce1b77e06c1ea9b3657c031bb35c10ba:
- in CMakeLists.txt, the plugin has to be marked as `MANDATORY`, which makes it statically-linked too
- then it has to be handled in `plugin_register_builtin_and_init_core_se` in `sql_plugin.cc`.

But at this time I still cannot tell whether the above is sufficient too. I think we simply must revisit this at a later point.

sql/sys_vars.cc
```diff
@@ -6071,6 +6071,13 @@ static Sys_var_plugin Sys_default_tmp_storage_engine(
     MYSQL_STORAGE_ENGINE_PLUGIN, DEFAULT(&default_tmp_storage_engine),
     NO_MUTEX_GUARD, NOT_IN_BINLOG, ON_CHECK(check_storage_engine));
 
+static Sys_var_plugin Sys_default_dd_storage_engine(
```
laurynas-biveinis: Since this is different from other "default SE" sysvars in that each allowed value is handled specifically, I'd consider making it enum, not string.

sql/mysqld.cc
sunshine-Chun: I removed this initialization and changed the sys var to enum. I hit some issue updating this PR. So end up with closing this one and creating a new PR.

sunshine-Chun:
Will create a new PR

sunshine-Chun:
Issue fixed. Back to this pr

sql/sys_vars.cc
```diff
@@ -6071,6 +6071,14 @@ static Sys_var_plugin Sys_default_tmp_storage_engine(
     MYSQL_STORAGE_ENGINE_PLUGIN, DEFAULT(&default_tmp_storage_engine),
     NO_MUTEX_GUARD, NOT_IN_BINLOG, ON_CHECK(check_storage_engine));
 
+static const char *dd_storage_engine[] = {"InnoDB", "RocksDB", NullS};
+static Sys_var_enum Sys_default_dd_storage_engine(
+    "default_dd_storage_engine",
+    "The default storage engine for data dictionary",
+    READ_ONLY GLOBAL_VAR(default_dd_storage_engine), NO_CMD_LINE,
+    dd_storage_engine, DEFAULT(DEFAULT_DD_INNODB), NO_MUTEX_GUARD,
+    NOT_IN_BINLOG, ON_CHECK(0), ON_UPDATE(0));
```
```suggestion
    NOT_IN_BINLOG, ON_CHECK(nullptr), ON_UPDATE(nullptr));
```

sql/sys_vars.cc
```diff
@@ -6071,6 +6071,14 @@ static Sys_var_plugin Sys_default_tmp_storage_engine(
     MYSQL_STORAGE_ENGINE_PLUGIN, DEFAULT(&default_tmp_storage_engine),
     NO_MUTEX_GUARD, NOT_IN_BINLOG, ON_CHECK(check_storage_engine));
 
+static const char *dd_storage_engine[] = {"InnoDB", "RocksDB", NullS};
+static Sys_var_enum Sys_default_dd_storage_engine(
+    "default_dd_storage_engine",
+    "The default storage engine for data dictionary",
+    READ_ONLY GLOBAL_VAR(default_dd_storage_engine), NO_CMD_LINE,
```
```suggestion
    READ_ONLY GLOBAL_VAR(default_dd_storage_engine), CMD_LINE(REQUIRED_ARG),
```

laurynas-biveinis (commented):
Please check `main.mysqld--help-notwin` test, I think it will need a re-record

Otherwise this PR looks about as complete as it could be at this point

PR #1282, opened 2023-03-13
Create rocksdb_datadir if not exist

sunshine-Chun:
Summary: When rebuilding a db, mysqld fails because it can not find rocksdb_datadir in move_wals_to_target_dir(). rocksdb_datadir has not been created at the time, but move_wals_to_target_dir() assumes it has. 

Add a check about rocksdb_datadir, if it doesn't exist, create the path.

The reason why we never hit this before is, when we test the clone, the db has already been built and the rocksdb_datadir has been created. 

Test Plan:

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/mysql-test/suite/rocksdb/t/check_datadir_existence.test
+++ b/mysql-test/suite/rocksdb/t/check_datadir_existence.test
@@ -1,0 +1,32 @@
+--source include/have_rocksdb.inc
+# Make sure mysqld will not crash when creating a new db instance,
+# if rocksdb_wal_dir and rocksdb_datadir both exist and are different
+
+let BASEDIR= `select @@basedir`;
+let DDIR=$MYSQL_TMP_DIR/installdb_test;
+let WDIR=$MYSQL_TMP_DIR/waltest;
+let MYSQLD_LOG=$MYSQL_TMP_DIR/server.log;
+let extra_args=--no-defaults --basedir=$BASEDIR;
+
+--echo # shut server down
+--source include/shutdown_mysqld.inc
+--echo # Server is down
+
+--echo #
+--echo # Try --initialize
+--echo #
+
+--echo # Run the server with --initialize
+--exec $MYSQLD $extra_args --initialize --datadir=$DDIR --rocksdb_wal_dir=$WDIR --log-error-verbosity=1 > $MYSQLD_LOG 2>&1
+
+--echo # Remove data dir and log file.
+--remove_file $MYSQLD_LOG
+--force-rmdir $DDIR
+--error 1,0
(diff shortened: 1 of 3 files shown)


storage/rocksdb/ha_rocksdb.cc
```diff
@@ -7193,6 +7193,10 @@ static void move_wals_to_target_dir() {
         (my_errno() != EEXIST))
       rdb_fatal_error("Failed to create %s", rocksdb_wal_dir);
 
+    if ((my_mkdir(rocksdb_datadir, S_IRWXU, MYF(0)) == -1) &&
```
laurynas-biveinis, line 7196: What does "rebuilding a db" mean here? Specifically, is a testcase possible?

luqun: I think it means When it try to bootstrap a new mysql instance.
you can repro it when rocksdb_wal_dir and rocksdb_data_dir point to different directory, and then try to bootstrap it.

sunshine-Chun: Test case added.

mysql-test/suite/rocksdb/t/check_datadir_existence.test
```diff
@@ -0,0 +1,35 @@
...
+let DDIR=$MYSQL_TMP_DIR/installdb_test;
+let WDIR=$MYSQL_TMP_DIR/waltest;
+let MYSQLD_LOG=$MYSQL_TMP_DIR/server.log;
+let extra_args=--no-defaults --basedir=$BASEDIR;
+
+--echo # shut server down
+--exec echo "wait" > $MYSQLTEST_VARDIR/tmp/mysqld.1.expect
+--shutdown_server
+--source include/wait_until_disconnected.inc
+--echo # Server is down
```
laurynas-biveinis: Lines 11-15 should be replaced with
`--source include/shutdown_mysqld.inc`

mysql-test/suite/rocksdb/t/check_datadir_existence.test
```diff
@@ -0,0 +1,35 @@
...
+--force-rmdir $DDIR
+--force-rmdir $WDIR
+
+--echo #
+--echo # Cleanup
+--echo #
+--echo # Restarting the server
+--exec echo "restart" > $MYSQLTEST_VARDIR/tmp/mysqld.1.expect
+--enable_reconnect
+--source include/wait_until_connected_again.inc
```
laurynas-biveinis: Lines 28-35 should be replaced with
`--source include/start_mysqld.inc`

PR #1283, opened 2023-03-14
Do not try to access plugin name for unloaded plugins

laurynas-biveinis:
If early plugin initialization fails, the plugin is unloaded with its entry remaining in plugin_array with PLUGIN_IS_FREED. The unloading of plugin will call dlclose, potentially unmapping its memory, including the plugin name constant. This may cause a crash in plugin_init_initialize_and_reap in the code which delays the raft plugin initialization by comparing its name. The crash is seen on macOS, MTR test perfschema.table_plugin_early_load, on a sanitizer-less build (enabling sanitizers disables the dlclose call).

Fix by moving plugin name check after the plugin state check, which excludes the PLUGIN_IS_FREED entries.

Squash with 0cf9341ab0a9751a8895dd073fc6ae9f3aa0bd89

### DIFF
--- a/sql/sql_plugin.cc
+++ b/sql/sql_plugin.cc
@@ -1441,9 +1441,9 @@ static bool plugin_init_initialize_and_reap() {
   for (st_plugin_int **it = plugin_array->begin(); it != plugin_array->end();
        ++it) {
     plugin_ptr = *it;
-    // Delay initialze raft plugin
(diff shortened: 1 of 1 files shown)


sunshine-Chun:
Is this change data dictionary related?

laurynas-biveinis:
It is a pre-existing regression that is visible on one of the DD-related testcases

PR #1284, opened 2023-03-16
Early load rocksdb plugin if ddse is rocksdb

sunshine-Chun:
Summary: 
Load rocksdb plugin in early stage if ddse is rocksdb

Test Plan: MTR

Reviewers:

Subscribers:

Tasks:

Tags:


mysql-test/suite/rocksdb/t/early_load_rocksdb_plugin.test
```diff
@@ -0,0 +1,34 @@
+--source include/have_rocksdb.inc
+
+# Set the default ddse to Rocksdb
+# Stimulate bootstrap by starting mysqld with --initialize
```
laurynas-biveinis:
```suggestion
# Simulate bootstrap by starting mysqld with --initialize
```

But this is not a "simulation" - this is an actual bootstrap on some instance

mysql-test/suite/rocksdb/t/early_load_rocksdb_plugin.test
```diff
@@ -0,0 +1,34 @@
...
+--remove_file $MYSQLD_LOG
+--force-rmdir $DDIR
+
+--echo #
+--echo # Cleanup
+--echo #
+--echo # Restarting the server
+--source include/start_mysqld.inc
+
+SELECT count(*) FROM INFORMATION_SCHEMA.PLUGINS where PLUGIN_NAME like '%ROCKSDB%';
```
laurynas-biveinis: Redundant?

sql/sql_plugin.cc
```diff
@@ -1577,20 +1577,23 @@ bool plugin_register_builtin_and_init_core_se(int *argc, char **argv) {
       if (register_builtin(plugin, &tmp, &plugin_ptr)) goto err_unlock;
 
       /*
-        Only initialize daemon_keyring_proxy, MyISAM, InnoDB and CSV at this
-        stage. Note that when the --help option is supplied,
-        daemon_keyring_proxy and InnoDB are not initialized because the plugin
-        table will not be read anyway, as indicated by the flag set when the
-        plugin_init() function is called.
+        Only initialize daemon_keyring_proxy, MyISAM, InnoDB, CSV and ROCKSDB
+        (only when defaul dd storage engine is Rocksdb) at this stage. Note that
```
laurynas-biveinis: Inconsistent "ROCKSDB/Rocksdb" case

sql/sql_plugin.cc
```diff
@@ -1577,20 +1577,23 @@ bool plugin_register_builtin_and_init_core_se(int *argc, char **argv) {
...
           &my_charset_latin1, plugin->name, "daemon_keyring_proxy_plugin");
       bool is_myisam =
           !my_strcasecmp(&my_charset_latin1, plugin->name, "MyISAM");
       bool is_innodb =
           !my_strcasecmp(&my_charset_latin1, plugin->name, "InnoDB");
+      bool is_rocksdb =
+          !my_strcasecmp(&my_charset_latin1, plugin->name, "ROCKSDB");
       if ((!is_daemon_keyring_proxy || is_help_or_validate_option()) &&
           !is_myisam && (!is_innodb || is_help_or_validate_option()) &&
+          (!is_rocksdb || default_dd_storage_engine != DEFAULT_DD_ROCKSDB) &&
```
laurynas-biveinis: Not sure what should we do about this at this stage - we will need to implement cross-grade between different DD engines. Might be OK for now though

sql/sql_plugin.cc
```diff
@@ -3237,8 +3240,9 @@ static int construct_options(MEM_ROOT *mem_root, st_plugin_int *tmp,
       (char *)mem_root->Alloc(plugin_name_len + plugin_dash.length + 1);
   strxmov(plugin_name_with_prefix_ptr, plugin_dash.str, plugin_name_ptr, NullS);
 
-  if (tmp->load_option != PLUGIN_FORCE &&
-      tmp->load_option != PLUGIN_FORCE_PLUS_PERMANENT) {
+  if ((tmp->load_option != PLUGIN_FORCE &&
+       tmp->load_option != PLUGIN_FORCE_PLUS_PERMANENT) ||
+      (strcmp(plugin_name, "ROCKSDB") == 0)) {
```
laurynas-biveinis: Why is this needed?

storage/rocksdb/rdb_datadic.cc
```diff
@@ -4342,6 +4342,12 @@ int Rdb_validate_tbls::add_table(Rdb_tbl_def *tdef) {
 */
 bool Rdb_validate_tbls::validate(void) {
   THD *const thd = my_core::thd_get_current_thd();
+
+  if(!thd) {
```
laurynas-biveinis: Formatting (`clang-format`)

storage/rocksdb/CMakeLists.txt
```diff
@@ -266,7 +266,7 @@ IF (ROCKSDB_DYNAMIC_PLUGIN)
   MYSQL_ADD_PLUGIN(rocksdb ${ROCKSDB_SOURCES} STORAGE_ENGINE DEFAULT
     MODULE_ONLY LINK_LIBRARIES ${rocksdb_static_libs})
 ELSE()
-  MYSQL_ADD_PLUGIN(rocksdb_se ${ROCKSDB_SOURCES} STORAGE_ENGINE DEFAULT
+  MYSQL_ADD_PLUGIN(rocksdb_se ${ROCKSDB_SOURCES} STORAGE_ENGINE MANDATORY
```
laurynas-biveinis: The ability to build RocksDB as a dynamic plugin remains - which is incompatible with native DD and is likely to be broken by the `sql_plugin.cc` changes. As a minimum, please test this PR with `-DWITH_ROCKSDB_DYNAMIC_PLUGIN=ON` (not sure if that's the exact CMake option name), and then please either make any changes to make it work, either remove the dynamic plugin option. But the latter may break some user scripts for a feature that does not exist yet.

sql/sql_plugin.cc
luqun: yes. I think this PR only enable load rocksdb plugin before DD initialize..

mysql-test/suite/rocksdb/t/early_load_rocksdb_plugin.test
```diff
@@ -0,0 +1,34 @@
...
+--source include/shutdown_mysqld.inc
+--echo # Server is down
+
+--echo #
+--echo # Try --initialize
+--echo #
+
+--echo # Run the server with --initialize
+--exec $MYSQLD $extra_args --initialize --datadir=$DDIR --log-error-verbosity=1 > $MYSQLD_LOG 2>&1
+
```
luqun: could you check "RocksDB:Init column families" or similar string in MYSQLD_LOG?

storage/rocksdb/CMakeLists.txt
sunshine-Chun: In the diff tool, we have test set for both dynamicRocksdb and static rocksdb. From the test result, most of them passed. There are few test case failure. But I don't think it's related to the change. @luqun Could you please help to verify it's not related to this diff's change?

sql/sql_plugin.cc
sunshine-Chun: This is needed when I run setup command in dev server to build a database. Without this change, the setup script will fail.

laurynas-biveinis: What is the command that the setup script executes and what error is it getting?

sunshine-Chun: It's about handling the command options. When executing mysqld command in the setup script, we pass --rocksdb as a command parameter. After changing Rocksdb as mandatory, this option is considered loose anymore. And will throw an error about unknown option and fail the mysqld. The change here helps to add the option to option list to pass the option check.

mysql-test/suite/rocksdb/t/early_load_rocksdb_plugin.test
```diff
@@ -0,0 +1,66 @@
...
+--echo #
+--echo # Try --initialize
+--echo #
+
+--echo # Run the server with --initialize
+--exec $MYSQLD $extra_args --initialize --default_dd_storage_engine=RocksDB --datadir=$DDIR --log-error-verbosity=3 > $MYSQLD_LOG 2>&1
+
+--let OUTF = $MYSQLD_LOG
+
+perl;
```
laurynas-biveinis: Please look into `assert_grep.inc`, specifically its `assert_only_after` parameter

sql/sql_plugin.cc
laurynas-biveinis: I don't like the special casing here, but what has to be done has to be done. Perhaps the scripts should no longer pass `--rocksdb` (as they are not passing `--innodb` currently, right?), and then later this could be removed.

sql/sql_plugin.cc
```diff
@@ -3237,8 +3240,12 @@ static int construct_options(MEM_ROOT *mem_root, st_plugin_int *tmp,
       (char *)mem_root->Alloc(plugin_name_len + plugin_dash.length + 1);
   strxmov(plugin_name_with_prefix_ptr, plugin_dash.str, plugin_name_ptr, NullS);
 
-  if (tmp->load_option != PLUGIN_FORCE &&
-      tmp->load_option != PLUGIN_FORCE_PLUS_PERMANENT) {
+  /* After changing static ROCKSDB plugin as mandatory, the --rocksdb option is
+   * not considered as loose. Add the option in the option list to avoid unknown
```
laurynas-biveinis: Not a common MySQL comment style, the common ones are `//` and `/* */` without leading asterix in the intermediate lines

storage/rocksdb/rdb_datadic.cc
```diff
@@ -4342,6 +4342,15 @@ int Rdb_validate_tbls::add_table(Rdb_tbl_def *tdef) {
 */
 bool Rdb_validate_tbls::validate(void) {
   THD *const thd = my_core::thd_get_current_thd();
+
+  /* When ddse is ROCKSDB, we are not able to validate the Rocksdb tables since
+   * we load rocksdb plugin before initializing dd. So skip validation here. */
```
laurynas-biveinis: Comment style

sql/sql_plugin.cc
luqun: @sunshine-Chun , if we revert this change, and if MTR passes, we can upgrade script to remove that --rocksdb instead.

sunshine-Chun: Sounds good. Let me remove this line and see if MTR test passes

storage/rocksdb/CMakeLists.txt
luqun: I will take a look this part to see whether it works with rocksdb dynamic plugin..

sql/sql_plugin.cc
sunshine-Chun: Looks like the MTR passed. Will remove this line.

PR #1285, opened 2023-03-20
Load rocksdb se plugin if default dd se is rocksdb

sunshine-Chun:
Early load rocksdb plugin if ddse is rocksdb

Summary:

Test Plan: MTR

Reviewers:

Subscribers:

Tasks:

Tags:


laurynas-biveinis (commented):
LGTM (conditionally on the dynamic plugin configuration not broken)

luqun:
> LGTM (conditionally on the dynamic plugin configuration not broken)

maybe we can add a check ddse is initialized  after plugin_register_builtin_and_init_core_se, if not, failed

sql/sql_plugin.cc
```diff
@@ -1618,6 +1625,9 @@ bool plugin_register_builtin_and_init_core_se(int *argc, char **argv) {
   assert(global_system_variables.table_plugin);
   assert(global_system_variables.temp_table_plugin);
 
+  /* if ddse is rocksdb, rocksdb plugin should be loaded */
+  if (default_dd_storage_engine == DEFAULT_DD_ROCKSDB) assert(rocksdb_loaded);
```
laurynas-biveinis: On release build the logic becomes unactionable, thus
```suggestion
assert(rocksdb_loaded || default_dd_storage_engine != DEFAULT_DD_ROCKSDB);
```

PR #1286, opened 2023-03-20
Experimental change to solve performance overshooting issue

sunshine-Chun:
Summary:
 
This is a experimental change to solve the performance overshooting issue. The full change need to change the number of function parameters for several functions. Will apply the changes after we reach an agreement to make this change.

Test Plan:

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/plugin/clone/src/clone_client.cc
+++ b/plugin/clone/src/clone_client.cc
@@ -228,6 +228,8 @@ uint64_t Client_Stat::task_target(uint64_t target_speed, uint64_t current_speed,
   not set yet, start by assuming all active thread. */
   auto active_tasks =
       (current_target == 0) ? num_tasks : (current_speed / current_target);
+  
+  active_tasks = num_tasks;
 
   /* Keep the value within current boundary. */
   if (active_tasks == 0) {


plugin/clone/src/clone_client.cc
```diff
@@ -228,6 +228,8 @@ uint64_t Client_Stat::task_target(uint64_t target_speed, uint64_t current_speed,
   not set yet, start by assuming all active thread. */
   auto active_tasks =
       (current_target == 0) ? num_tasks : (current_speed / current_target);
+  
+  active_tasks = num_tasks;
```
sunshine-Chun, line 232: The issue here is the calculation of active  tasks is not correct. Let's say,  the current_target is 700MB/S, and the current_speed is 900MB/S. Then the active_tasks will be 1. 

According to the calculation for task_target below, the task target will be 700MB/s.
`  auto task_target = target_speed / active_tasks;
`
The task_target will be used as the current_target of next call of this task_target() function. Hence, it will go over the above calculation again and the task target speed will not be updated at all.

Since the max speed capacity for each thread is around 450MB/s, it will never throttle the thread. (Because it's always smaller than the target task speed 700MB/s).

For the change I made, I assume all threads created are active.

laurynas-biveinis: Why InnoDB clone does not exhibit the issue?

PR #1287, opened 2023-04-01
[MyRocksDD] Disable DESC for mysql.password_history when creating sys…

sunshine-Chun:
…tem table

Summary: Myrocks doesn't support DESC. Disable DESC when creating rocksdb system tables.

Test Plan: Don't break current MTR

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/scripts/mysql_system_tables.sql
+++ b/scripts/mysql_system_tables.sql
@@ -26,6 +26,7 @@
 
 set @have_innodb= (select count(engine) from information_schema.engines where engine='INNODB' and support != 'NO');
 set @is_mysql_encrypted = (select ENCRYPTION from information_schema.INNODB_TABLESPACES where NAME='mysql');
+set @ddse= (select @@default_dd_storage_engine);
 
 -- Tables below are NOT treated as DD tables by MySQL server yet.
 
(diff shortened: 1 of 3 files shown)


scripts/mysql_system_tables.sql
```diff
@@ -26,6 +26,7 @@
 
 set @have_innodb= (select count(engine) from information_schema.engines where engine='INNODB' and support != 'NO');
 set @is_mysql_encrypted = (select ENCRYPTION from information_schema.INNODB_TABLESPACES where NAME='mysql');
+set @ddse= (select @@default_dd_storage_engine);
```
laurynas-biveinis, line 29: ddse == ROCKSDB && is_mysql_encrypted == true combination should be diagnosed and rejected?

sunshine-Chun: would you mind elaborating this? I didn't quite follow.

laurynas-biveinis: This same script allows creating an instance with mysql tablespace in InnoDB being encrypted, which MyRocks would not support. I believe we should return an error if the DD engine is MyRocks and mysql tablespace encryption is requested.

sunshine-Chun:
closed.

PR #1292, opened 2023-04-07
Index read fixes

laurynas-biveinis:
Minor index access cleanups for MyRocks:
- `ha_rocksdb::index_init`: remove duplicate `active_index` assignment. Move
  `m_need_build_decoder` assignment past the error exit so that taking it does not
  result in any state changes.
- `ha_rocksdb::index_next_same`: add an assert for `active_index` value to prevent
  out-of-bound accesses in the next line.


hermanlee:
Let me backport the first two patches directly, and then we can rebase this PR so that it contains only the MyRocks changes.

laurynas-biveinis:
Right, I forgot about the squashing

laurynas-biveinis:
Rebased, ready for review

PR #1294, opened 2023-04-17
Add multiple checks for clone

sunshine-Chun:
Summary: 

This diff add two checks:
1. mysqld version check will change from major version check to strict version check. Need to change the config name to version_comment to achieve this.
2. Check the client and donor have the same default storage engine plugin.

Test Plan: Test by running clone

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/plugin/clone/include/clone_hton.h
+++ b/plugin/clone/include/clone_hton.h
@@ -75,6 +75,9 @@ struct Locator {
       /* Should not lock plugin for auxiliary threads */
       assert(thd != nullptr);
 
+      DBUG_EXECUTE_IF("client_has_more_se", if (db_type == DB_TYPE_ROCKSDB)
+                                                db_type = DB_TYPE_UNKNOWN;);
+
       m_hton = ha_resolve_by_legacy_type(thd, db_type);
 
     } else {
--- a/plugin/clone/src/clone_hton.cc
+++ b/plugin/clone/src/clone_hton.cc
@@ -27,6 +27,7 @@ Clone Plugin: Interface with SE handlerton
 */
 
 #include "plugin/clone/include/clone_hton.h"
+#include "mysqld_error.h"
 
 /* Namespace for all clone data types */
 namespace myclone {
@@ -105,6 +106,15 @@ static bool run_hton_clone_begin(THD *thd, plugin_ref plugin, void *arg) {
   return (false);
 }
 
+static bool run_collect_expected_storage_vector(THD *, plugin_ref plugin,
+                                                void *arg) {
+  auto clone_arg = static_cast<std::vector<legacy_db_type> *>(arg);
+  auto hton = plugin_data<handlerton *>(plugin);
+  if (hton->clone_interface.clone_begin != nullptr)
+    clone_arg->push_back(hton->db_type);
+  return (false);
+}
+
 // Make InnoDB first in the storage locator and task vectors - to drive the
 // cross-engine synchronization & to wait for reconnects
 static void make_innodb_first(Storage_Vector &clone_loc_vec,
@@ -124,6 +134,31 @@ static void make_innodb_first(Storage_Vector &clone_loc_vec,
     }
   }
   assert(clone_loc_vec[0].m_hton->db_type == DB_TYPE_INNODB);
+}
+
+static int validate_storage_engine(THD *thd, Storage_Vector &clone_loc_vec) {
+  std::vector<legacy_db_type> expected_storage_vec;
+  expected_storage_vec.reserve(clone_loc_vec.capacity());
+
+  std::unordered_set<legacy_db_type> storage_vec;
+
+  std::transform(
+      clone_loc_vec.begin(), clone_loc_vec.end(),
+      std::inserter(storage_vec, storage_vec.end()),
+      [](const myclone::Locator &loc) { return loc.m_hton->db_type; });
+
+  plugin_foreach(thd, run_collect_expected_storage_vector,
+                 MYSQL_STORAGE_ENGINE_PLUGIN, &expected_storage_vec);
+
+  for (const auto &iter : expected_storage_vec) {
+    if (storage_vec.find(iter) == storage_vec.end()) {
+      my_error(ER_CLONE_PROTOCOL, MYF(0),
+               "storage engine plugin set mismatch between donor and client");
+      return (ER_CLONE_PROTOCOL);
+    }
+  }
+
+  return (0);
 }
 
 int hton_clone_begin(THD *thd, Storage_Vector &clone_loc_vec,
@@ -152,6 +187,9 @@ int hton_clone_begin(THD *thd, Storage_Vector &clone_loc_vec,
 
   assert(clone_loc_vec[0].m_hton->db_type == DB_TYPE_INNODB);
 
+  auto err = validate_storage_engine(thd, clone_loc_vec);
+  if (err != 0) return (err);
+
   for (auto &loc_iter : clone_loc_vec) {
     uint32_t task_id = 0;
 
@@ -170,7 +208,7 @@ int hton_clone_begin(THD *thd, Storage_Vector &clone_loc_vec,
       assert(flags[HA_CLONE_RESTART]);
     }
 #endif
-    auto err = loc_iter.m_hton->clone_interface.clone_begin(
+    err = loc_iter.m_hton->clone_interface.clone_begin(
         loc_iter.m_hton, thd, loc_iter.m_loc, loc_iter.m_loc_len, task_id,
         clone_type, clone_mode);
 
@@ -247,6 +285,10 @@ static bool run_hton_clone_apply_begin(THD *thd, plugin_ref plugin, void *arg) {
     uint32_t task_id = 0;
 
     assert(clone_arg->m_mode == HA_CLONE_MODE_VERSION);
+
+    DBUG_EXECUTE_IF(
+        "client_has_less_se",
+        if (hton->db_type == DB_TYPE_ROCKSDB) { return (false); });
 
     clone_arg->m_err = hton->clone_interface.clone_apply_begin(
         hton, thd, loc.m_loc, loc.m_loc_len, task_id, clone_arg->m_mode,
--- a/plugin/clone/src/clone_server.cc
+++ b/plugin/clone/src/clone_server.cc
@@ -381,6 +381,12 @@ int Server::deserialize_init_buffer(const uchar *init_buf, size_t init_len) {
       goto err_end;
     }
 
+    if (!loc.m_hton) {
+      my_error(ER_CLONE_PROTOCOL, MYF(0),
+               "storage engine plugin set mismatch between donor and client");
+      return (ER_CLONE_PROTOCOL);
+    }
+
     m_storage_vec.push_back(loc);
 
     init_len -= serialized_length;
--- a/mysql-test/suite/rocksdb_clone/r/remote_se_check.result
+++ b/mysql-test/suite/rocksdb_clone/r/remote_se_check.result
@@ -1,0 +1,27 @@
+INSTALL PLUGIN clone SONAME 'CLONE_PLUGIN';
+SET global debug="+d,client_has_less_se";
+SET GLOBAL clone_autotune_concurrency = OFF;
+SET GLOBAL clone_max_concurrency = 8;
+SET GLOBAL clone_valid_donor_list = 'HOST:PORT';
+CLONE INSTANCE FROM USER@HOST:PORT IDENTIFIED BY '' DATA DIRECTORY = 'CLONE_DATADIR';
+ERROR 08S01: Got an error reading communication packets
+select ID, STATE, ERROR_NO from performance_schema.clone_status;
+ID	STATE	ERROR_NO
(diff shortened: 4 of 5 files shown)


plugin/clone/src/clone_server.cc
```diff
@@ -380,8 +380,10 @@ int Server::deserialize_init_buffer(const uchar *init_buf, size_t init_len) {
     if (init_len < serialized_length) {
       goto err_end;
     }
-
-    m_storage_vec.push_back(loc);
+    
+    if (loc.m_hton != nullptr) {
```
sunshine-Chun: When client is rocksdb and donor is innodb. The client sends rocksdb locator to donor but the donor can not serialize it with valid m_hton. It will cause nullptr error later, so I only add loc to storage_vector when m_hton is not null. Looking for better solution.

sql/server_component/clone_protocol_service.cc
```diff
@@ -307,12 +307,8 @@ DEFINE_METHOD(int, mysql_clone_validate_configs,
     because user can no way clone from the current donor. */
     if (config_name.compare("version_compile_os") == 0) {
       critical_error = ER_CLONE_OS;
-    } else if (config_name.compare("version") == 0) {
-      /* we want to allow to add some suffix to the version and still match
-      i.e. 8.0.25 should be the same as 8.0.25-debug */
-      if (compare_prefix_version(config_val, donor_val)) {
-        continue;
-      }
+    } else if (config_name.compare("version_comment") == 0) {
```
sunshine-Chun: since we change the config name here. Do we need to bump the protocol version?

plugin/clone/src/clone_server.cc
sunshine-Chun: The clone will actually fail with ```ERROR 3863 (HY000): Clone received unexpected response from Donor : Wrong Clone RPC response length for COM_RES_LOCS.``` if I remove the loc here.

plugin/clone/src/clone_client.cc
```diff
@@ -1036,6 +1036,9 @@ bool Client::plugin_is_installed(std::string &plugin_name) {
 
 int Client::validate_remote_params() {
   int last_error = 0;
+  bool donar_rocksdb_installed = false;
+
+  std::string ROCKSDB = "ROCKSDB";
```
laurynas-biveinis:
```suggestion
  static constexpr char ROCKSDB[] = "ROCKSDB";
```

This creates a new string every time this method is called. `static` would avoid that, but `constexpr` would move it to compile-time. As C++17 does not yet have a constexpr std::string, a char array would do.

plugin/clone/src/clone_client.cc
```diff
@@ -1056,10 +1059,18 @@ int Client::validate_remote_params() {
     auto &plugin_name = plugin.first;
     auto &so_name = plugin.second;
 
+    if (plugin_name.compare(ROCKSDB) == 0) donar_rocksdb_installed = true;
+
     if (plugin_is_installed(plugin_name)) {
       continue;
     }
 
+    /* Donor installs ROCKSDB and client doesn't. Fail the clone. */
```
laurynas-biveinis: The commit message / PR description needs updating, as it might be misleading: the code checks for plugins, the commit message talks about default storage engine, which is an entirely different thing (a server sysvar).

plugin/clone/src/clone_client.cc
```diff
@@ -1072,6 +1083,13 @@ int Client::validate_remote_params() {
     last_error = ER_CLONE_PLUGIN_MATCH;
   }
 
+  /* Donor plugin is not there in donor but installed in recipient. Fail the
+   * clone. */
```
laurynas-biveinis: Not the MySQL comment style, either `//`, either `/* */` without leading stars in the middle lines

plugin/clone/src/clone_client.cc
```diff
@@ -1036,6 +1036,9 @@ bool Client::plugin_is_installed(std::string &plugin_name) {
 
 int Client::validate_remote_params() {
   int last_error = 0;
+  bool donar_rocksdb_installed = false;
```
laurynas-biveinis:
```suggestion
  bool donor_rocksdb_installed = false;
```

plugin/clone/src/clone_server.cc
laurynas-biveinis: Can you just do `my_error(ER_CLONE_PROTOCOL, MYF(0), "SE mismatch")` when `loc.m_hton = nullptr`?

sql/server_component/clone_protocol_service.cc
laurynas-biveinis: I don't think it has any practical difference whether we do or we do not, as the Oracle <-> FB clone will never be supported, and the server version check is already stricter.

plugin/clone/include/clone_server.h
```diff
@@ -29,6 +29,7 @@ Clone Plugin: Server interface
 #ifndef CLONE_SERVER_H
 #define CLONE_SERVER_H
 
+#include <cstdint>
```
laurynas-biveinis: Why?

plugin/clone/src/clone_hton.cc
```diff
@@ -105,6 +106,14 @@ static bool run_hton_clone_begin(THD *thd, plugin_ref plugin, void *arg) {
   return (false);
 }
 
+static bool run_collect_expected_storage_vector([[maybe_unused]] THD *thd,
```
laurynas-biveinis:
```suggestion
static bool run_collect_expected_storage_vector(THD *,
```

plugin/clone/src/clone_hton.cc
```diff
@@ -126,11 +135,31 @@ static void make_innodb_first(Storage_Vector &clone_loc_vec,
...
 
+int validate_storage_engine(THD *thd, Storage_Vector &clone_loc_vec) {
+  std::vector<legacy_db_type> expected_storage_vec;
+  expected_storage_vec.reserve(clone_loc_vec.capacity());
+
+  plugin_foreach(thd, run_collect_expected_storage_vector,
+                 MYSQL_STORAGE_ENGINE_PLUGIN, &expected_storage_vec);
+
+  if (expected_storage_vec.size() != clone_loc_vec.size()) {
+    my_error(ER_CLONE_PROTOCOL, MYF(0), "SE mismatch.");
```
laurynas-biveinis: Maybe
```suggestion
    my_error(ER_CLONE_PROTOCOL, MYF(0), "storage engine plugin set mismatch between donor and client");
```

plugin/clone/src/clone_hton.cc
```diff
@@ -126,11 +135,31 @@ static void make_innodb_first(Storage_Vector &clone_loc_vec,
...
 }
 
+int validate_storage_engine(THD *thd, Storage_Vector &clone_loc_vec) {
+  std::vector<legacy_db_type> expected_storage_vec;
+  expected_storage_vec.reserve(clone_loc_vec.capacity());
+
+  plugin_foreach(thd, run_collect_expected_storage_vector,
+                 MYSQL_STORAGE_ENGINE_PLUGIN, &expected_storage_vec);
+
+  if (expected_storage_vec.size() != clone_loc_vec.size()) {
```
laurynas-biveinis: Is it possible to check not only size but that the actual plugins match?

plugin/clone/src/clone_server.cc
```diff
@@ -381,6 +381,11 @@ int Server::deserialize_init_buffer(const uchar *init_buf, size_t init_len) {
       goto err_end;
     }
 
+    if (!loc.m_hton) {
+      my_error(ER_CLONE_PROTOCOL, MYF(0), "SE mismatch");
```
laurynas-biveinis: Likewise

plugin/clone/src/clone_hton.cc
```diff
@@ -105,6 +106,14 @@ static bool run_hton_clone_begin(THD *thd, plugin_ref plugin, void *arg) {
   return (false);
 }
 
+static bool run_collect_expected_storage_vector(THD *, plugin_ref plugin,
+                                                void *arg) {
+  auto clone_arg = static_cast<std::unordered_set<legacy_db_type> *>(arg);
+  auto hton = plugin_data<handlerton *>(plugin);
```
hermanlee, line 112: I think we should only collect InnoDB and MyRocks plugins, or something that supports transactions (though there is code for blackhole to be transactional). It is possible for other storage engine types like TempTable might be disabled on the donor and enabled on the client. While this engine won't affect the clone process, it does prevent the clone plugin from running since this check would fail.

hermanlee (commented):
Can you add a test case? Feel free to use the mtr debug utility to inject different responses if creating a test case that enable different storage plugins is difficult.

plugin/clone/src/clone_hton.cc
laurynas-biveinis: We could collect clone-supporting plugins, i.e. whose handlertons have `clone_interface.clone_begin != nullptr`

plugin/clone/src/clone_hton.cc
```diff
@@ -105,6 +106,15 @@ static bool run_hton_clone_begin(THD *thd, plugin_ref plugin, void *arg) {
...
+  if (hton->db_type == DB_TYPE_INNODB || hton->db_type == DB_TYPE_ROCKSDB)
```
laurynas-biveinis: We could collect clone-supporting plugins, i.e. whose handlertons have `clone_interface.clone_begin != nullptr`. This resolves the issue of other SEs and avoids hardcoding InnoDB/MyRocks here.

plugin/clone/src/clone_hton.cc
```diff
@@ -126,11 +136,41 @@ static void make_innodb_first(Storage_Vector &clone_loc_vec,
   assert(clone_loc_vec[0].m_hton->db_type == DB_TYPE_INNODB);
 }
 
+int validate_storage_engine(THD *thd, Storage_Vector &clone_loc_vec) {
```
laurynas-biveinis:
```suggestion
static int validate_storage_engine(THD *thd, Storage_Vector &clone_loc_vec) {
```

plugin/clone/src/clone_hton.cc
```diff
@@ -126,11 +136,41 @@ static void make_innodb_first(Storage_Vector &clone_loc_vec,
...
+  plugin_foreach(thd, run_collect_expected_storage_vector,
+                 MYSQL_STORAGE_ENGINE_PLUGIN, &expected_storage_vec);
+
+  if (expected_storage_vec.size() != clone_loc_vec.size()) {
+    my_error(ER_CLONE_PROTOCOL, MYF(0),
+             "storage engine plugin set mismatch between donor and client");
+    return (ER_CLONE_PROTOCOL);
+  }
+
+  for (auto &loc_iter : clone_loc_vec) {
```
laurynas-biveinis:
```suggestion
  for (const auto &loc_iter : clone_loc_vec) {
```

plugin/clone/src/clone_hton.cc
```diff
@@ -126,6 +136,30 @@ static void make_innodb_first(Storage_Vector &clone_loc_vec,
...
 
+static int validate_storage_engine(THD *thd, Storage_Vector &clone_loc_vec) {
+  std::vector<legacy_db_type> expected_storage_vec;
+  expected_storage_vec.reserve(clone_loc_vec.capacity());
+
+  std::unordered_set<legacy_db_type> storage_vec;
+
+  std::transform(clone_loc_vec.begin(), clone_loc_vec.end(),
+                 std::inserter(storage_vec, storage_vec.end()),
+                 [](myclone::Locator loc) { return loc.m_hton->db_type; });
```
laurynas-biveinis:
```suggestion
                 [](const myclone::Locator &loc) { return loc.m_hton->db_type; });
```

plugin/clone/include/clone_hton.h
```diff
@@ -77,6 +77,10 @@ struct Locator {
 
       m_hton = ha_resolve_by_legacy_type(thd, db_type);
 
+      DBUG_EXECUTE_IF(
+          "client_has_more_se",
+          if (m_hton->db_type == DB_TYPE_ROCKSDB) m_hton = nullptr;);
```
hermanlee: We might get a little more coverage by converting db_type to DB_TYPE_UNKNOWN and moving this before the ha_resolve_by_legacy_type() call.

plugin/clone/include/clone_hton.h
```diff
@@ -75,6 +75,10 @@ struct Locator {
       /* Should not lock plugin for auxiliary threads */
       assert(thd != nullptr);
 
+      DBUG_EXECUTE_IF(
+          "client_has_more_se",
+          if (m_hton->db_type == DB_TYPE_ROCKSDB) db_type = DB_TYPE_UNKNOWN;);
```
hermanlee: m_hton is going to be nullptr here. I think the check should be `db_type == DB_TYPE_ROCKSDB`

laurynas-biveinis:
@sunshine-Chun, please log a bug with upstream like the ones listed in https://bugs.mysql.com/bug.php?id=109926, consider also sending your patch

PR #1296, opened 2023-04-24
Add set/get server version api

sunshine-Chun:
Summary: This diff added support for set/get server version api in rocksdb

Test Plan: Test it locally

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/storage/rocksdb/ha_rocksdb.cc
+++ b/storage/rocksdb/ha_rocksdb.cc
@@ -7434,6 +7434,8 @@ static int rocksdb_init_internal(void *const p) {
   rocksdb_hton->clone_interface.clone_apply_end = rocksdb_clone_apply_end;
 
   rocksdb_hton->dict_register_dd_table_id = rocksdb_dict_register_dd_table_id;
+  rocksdb_hton->dict_get_server_version = rocksdb_dict_get_server_version;
+  rocksdb_hton->dict_set_server_version = rocksdb_dict_set_server_version;
 
   rocksdb_hton->flags = HTON_SUPPORTS_EXTENDED_KEYS | HTON_CAN_RECREATE;
 
--- a/storage/rocksdb/rdb_datadic.cc
+++ b/storage/rocksdb/rdb_datadic.cc
@@ -5311,6 +5311,11 @@ bool Rdb_dict_manager::init(rocksdb::TransactionDB *const rdb_dict,
       rocksdb::Slice(reinterpret_cast<char *>(m_key_buf_max_index_id),
                      Rdb_key_def::INDEX_NUMBER_SIZE);
 
+  rdb_netbuf_store_index(m_key_buf_server_version, Rdb_key_def::SERVER_VERSION);
+  m_key_slice_server_version =
+      rocksdb::Slice(reinterpret_cast<char *>(m_key_buf_server_version),
+                     Rdb_key_def::INDEX_NUMBER_SIZE);
+
   resume_drop_indexes();
   rollback_ongoing_index_creation();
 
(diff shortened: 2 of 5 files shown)


storage/rocksdb/rdb_datadic.h
```diff
@@ -1725,6 +1730,10 @@ class Rdb_dict_manager : public Ensure_initialized {
   bool update_max_index_id(rocksdb::WriteBatch *const batch,
                            const uint32_t index_id,
                            bool is_dd_tbl = false) const;
+
+  bool get_server_version(uint *const version);
```
luqun: nit: also const

storage/rocksdb/rdb_datadic.cc
```diff
@@ -6033,6 +6038,40 @@ bool Rdb_dict_manager::update_max_index_id(rocksdb::WriteBatch *const batch,
...
+  }
+
+  return true;
+}
+
+bool Rdb_dict_manager::set_server_version() const {
+  const std::unique_ptr<rocksdb::WriteBatch> wb = begin();
+  rocksdb::WriteBatch *const batch = wb.get();
+
+  Rdb_buf_writer<Rdb_key_def::VERSION_SIZE + Rdb_key_def::INDEX_NUMBER_SIZE>
```
luqun: please don't use Rdb_key_def::INDEX_NUMBER_SIZE for value writer, although the value is same. you can define a new variable or ..

storage/rocksdb/rdb_datadic.cc
```diff
@@ -6033,6 +6038,40 @@ bool Rdb_dict_manager::update_max_index_id(rocksdb::WriteBatch *const batch,
   return false;
 }
 
+bool Rdb_dict_manager::get_server_version(uint *const ser_version) const {
+  std::string value;
+
+  const rocksdb::Status status = get_value(m_key_slice_server_version, &value);
+  if (status.ok()) {
+    const uchar *const val = (const uchar *)value.c_str();
```
laurynas-biveinis: C++ style cast

storage/rocksdb/rdb_datadic.cc
```diff
@@ -6033,6 +6038,40 @@ bool Rdb_dict_manager::update_max_index_id(rocksdb::WriteBatch *const batch,
...
+      return false;
+    }
+  }
+
+  return true;
+}
+
+bool Rdb_dict_manager::set_server_version() const {
+  const std::unique_ptr<rocksdb::WriteBatch> wb = begin();
+  rocksdb::WriteBatch *const batch = wb.get();
```
laurynas-biveinis: Why do we mirror a smart pointer in a raw pointer? IMHO removing the raw one, renaming `wb` to `batch`, and using `batch.get()` inline as needed would result in as clean or cleaner code

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -3257,6 +3257,25 @@ std::string dump_ingest_external_file_options(
   return s.str();
 }
 
+/**
```
laurynas-biveinis: Please move to that new source file in the other PR

sunshine-Chun:
dict_manager is a static variable declared in ha_rocksdb.cc. The get/set server version call rely on this variable. However this variable is not accessible in rdb_native_dd. we can either make it global or we can wrap it in ha_rocksdb. (The current implementation). LMK your thoughts.

storage/rocksdb/rdb_native_dd.h
```diff
@@ -16,6 +16,7 @@
 #pragma once
 
 /* C++ standard header files */
+#include <sys/types.h>
```
laurynas-biveinis, line 19: Is this for `uint`? If so, forward-declare it instead

laurynas-biveinis:
> dict_manager is a static variable declared in ha_rocksdb.cc. The get/set server version call rely on this variable. However this variable is not accessible in rdb_native_dd. we can either make it global or we can wrap it in ha_rocksdb. (The current implementation). LMK your thoughts.

Wrapping LGTM

storage/rocksdb/rdb_native_dd.h
```diff
@@ -21,13 +21,17 @@
 /* MySQL header files */
 #include "sql/dd/object_id.h"
 
+typedef unsigned int uint;
```
laurynas-biveinis: I am sorry, I didn't realize you cannot forward declare a simple `typedef`. While repeating a `typedef` is correct, it's probably better to include the header in this case.

PR #1297, opened 2023-04-24
Add ddst_dict_init api support for rocksdb dd

sunshine-Chun:
Summary: Add ddst_dict_init api support for rocksdb dd

Test Plan: Test locally, not crush....

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/storage/rocksdb/ha_rocksdb.cc
+++ b/storage/rocksdb/ha_rocksdb.cc
@@ -7511,6 +7511,7 @@ static int rocksdb_init_internal(void *const p) {
   rocksdb_hton->dict_get_server_version = rocksdb_dict_get_server_version;
   rocksdb_hton->dict_set_server_version = rocksdb_dict_set_server_version;
   rocksdb_hton->is_supported_system_table = rocksdb_is_supported_system_table;
+  rocksdb_hton->ddse_dict_init = rocksdb_ddse_dict_init;
 
   rocksdb_hton->flags = HTON_SUPPORTS_EXTENDED_KEYS | HTON_CAN_RECREATE;
 
@@ -19170,7 +19171,6 @@ bool ha_rocksdb::get_se_private_data(dd::Table *, bool reset) {
   if (reset) {
     native_dd::clear_dd_table_ids();
   }
-
   return false;
 }
 
--- a/storage/rocksdb/rdb_native_dd.h
+++ b/storage/rocksdb/rdb_native_dd.h
@@ -21,6 +21,7 @@
 
 /* MySQL header files */
 #include "sql/dd/object_id.h"
+#include "sql/handler.h"
 
 namespace dd {
 class Table;
@@ -28,10 +29,15 @@ class Table;
 
 namespace myrocks {
 
+constexpr const char rocksdb_dd_space_name[] = "mysql";
+
 void rocksdb_dict_register_dd_table_id(dd::Object_id dd_table_id);
 bool rocksdb_dict_get_server_version(uint *version);
 bool rocksdb_dict_set_server_version();
 bool rocksdb_is_supported_system_table(const char *, const char *, bool);
+bool rocksdb_ddse_dict_init(dict_init_mode_t dict_init_mode, uint version,
+                            List<const dd::Object_table> *tables,
+                            List<const Plugin_tablespace> *tablespaces);
 
 class native_dd {
  private:
--- a/storage/rocksdb/rdb_native_dd.cc
+++ b/storage/rocksdb/rdb_native_dd.cc
@@ -25,6 +25,7 @@
 #include "ha_rocksdb.h"
 #include "storage/rocksdb/ha_rocksdb_proto.h"
 #include "storage/rocksdb/rdb_datadic.h"
+#include "sql/plugin_table.h"
 
 namespace myrocks {
 std::unordered_set<dd::Object_id> native_dd::s_dd_table_ids = {};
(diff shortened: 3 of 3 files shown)


storage/rocksdb/ha_rocksdb.cc
```diff
@@ -19020,6 +19028,22 @@ static bool parse_fault_injection_params(
   return false;
 }
 
+static bool rocksdb_ddse_dict_init(
+    dict_init_mode_t dict_init_mode, [[maybe_unused]] uint version,
+    [[maybe_unused]] List<const dd::Object_table> *tables,
+    List<const Plugin_tablespace> *tablespaces) {
+  assert(tables && tables->is_empty());
```
laurynas-biveinis: Split asserts on `&&`

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -19020,6 +19028,22 @@ static bool parse_fault_injection_params(
...
+  assert(tablespaces && tablespaces->is_empty());
+
+  assert(dict_init_mode == DICT_INIT_CREATE_FILES ||
+         dict_init_mode == DICT_INIT_CHECK_FILES);
+
+  static Plugin_tablespace dd_space("mysql", "", "", "", rocksdb_hton_name);
```
laurynas-biveinis: This `"mysql"` can go to a `constexpr`, we will definitely reference it many more times. Maybe a server layer already has such constant

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -19020,6 +19028,22 @@ static bool parse_fault_injection_params(
...
+  tablespaces->push_back(&dd_space);
+
+  return false;
```
laurynas-biveinis, line 19174: We don't want this code to actually run yet, consider adding `assert(false)` and returning an error

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -19020,6 +19028,22 @@ static bool parse_fault_injection_params(
   return false;
 }
 
+static bool rocksdb_ddse_dict_init(
+    dict_init_mode_t dict_init_mode, [[maybe_unused]] uint version,
```
laurynas-biveinis: Drop the argument names for fully-unused parameters, use [[maybe_unused]] for the debug build-only parameters only.

storage/rocksdb/rdb_global.h
```diff
@@ -357,6 +357,7 @@ static_assert(HA_ERR_ROCKSDB_FIRST > HA_ERR_LAST,
 #define HA_ERR_ROCKSDB_LAST HA_ERR_ROCKSDB_TMP_TABLE_COMMIT_FAILED
 
 const char *const rocksdb_hton_name = "ROCKSDB";
+const char *const rocksdb_dd_space_name = "mysql";
```
laurynas-biveinis: constexpr in `rdb_native_dd.h`

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -19020,6 +19028,28 @@ static bool parse_fault_injection_params(
   return false;
 }
 
+static bool rocksdb_ddse_dict_init(
```
laurynas-biveinis: Move to `rdb_native_dd.cc`?

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -19020,6 +19028,28 @@ static bool parse_fault_injection_params(
...
+  assert(tables->is_empty());
+  assert(tablespaces);
+  assert(tablespaces->is_empty());
+
+  assert(dict_init_mode == DICT_INIT_CREATE_FILES ||
+         dict_init_mode == DICT_INIT_CHECK_FILES);
+
+  static Plugin_tablespace dd_space(rocksdb_dd_space_name, "", "", "",
+                                    rocksdb_hton_name);
+  tablespaces->push_back(&dd_space);
```
laurynas-biveinis: On a second thought, tablespaces are purely InnoDB concept. So nothing should be added here?

luqun: I think DD expect mysql namespace always exists.. Let me find related code in a while..

laurynas-biveinis: The current DD layer assumes it does, but I am patching it out in my prototype.
I am concerned about pretending something exists when MyRocks does not even support the concept of it.

storage/rocksdb/rdb_native_dd.h
```diff
@@ -20,15 +20,22 @@
...
 #include "sql/dd/object_id.h"
+#include "sql/handler.h"
 
 namespace dd {
 class Table;
 }
 
 namespace myrocks {
 
+constexpr const char *rocksdb_dd_space_name = "mysql";
```
laurynas-biveinis:
```suggestion
constexpr const char rocksdb_dd_space_name[] = "mysql";
```

The pointer version works, but its sizeof is 8, which is very annoying to find

sunshine-Chun:
Do we still need this diff. Sounds like we are not adding the. dd_space in the ddst_dict_init function now.

laurynas-biveinis:
As with the rest of the current diffs, it's hard to say, as the code is too incomplete to run. IMHO we can hold until it becomes clear.

luqun:
@laurynas-biveinis , according to previous conversation, we may still need this mysql tablespace. Currently https://github.com/facebook/mysql-5.6/pull/1319 will fail if there is testcase using rocksdb DD.

laurynas-biveinis:
@sunshine-Chun , @luqun, It is true that we'll need the mysql tablespace. But, in the current prototype it is InnoDB that sets it up, thus at this point I am not comfortable to committing to RocksDB creating it. We could quickly merge a PR for `ddse_dict_init`, which would be mostly empty, and leave the tablespace for later.

laurynas-biveinis:
@sunshine-Chun , please close this PR

PR #1299, opened 2023-04-24
Add rocksdb_is_supported_system_table api support for rocksdb dd

sunshine-Chun:
Summary:

Test Plan:

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/storage/rocksdb/ha_rocksdb.cc
+++ b/storage/rocksdb/ha_rocksdb.cc
@@ -7436,6 +7436,7 @@ static int rocksdb_init_internal(void *const p) {
   rocksdb_hton->dict_register_dd_table_id = rocksdb_dict_register_dd_table_id;
   rocksdb_hton->dict_get_server_version = rocksdb_dict_get_server_version;
   rocksdb_hton->dict_set_server_version = rocksdb_dict_set_server_version;
+  rocksdb_hton->is_supported_system_table = rocksdb_is_supported_system_table;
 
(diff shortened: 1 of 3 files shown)


storage/rocksdb/ha_rocksdb.cc
```diff
@@ -3162,6 +3162,11 @@ static int rocksdb_compact_column_family(
   return HA_EXIT_SUCCESS;
 }
 
+static bool rocksdb_is_supported_system_table(const char *, const char *,
+                                              bool is_sql_layer_system_table) {
+  return is_sql_layer_system_table;
```
laurynas-biveinis: I don't think this is safe: this allows `ALTER TABLE` for completely unsupported tables, and `ALTER TABLE` back to InnoDB would not be guaranteed. If we want to have the handlerton method anyway this early, then let's start with `return false` and enable the tables individually or in groups as the relevant work is done.

sunshine-Chun:
Can we merge this api to the code? PR https://github.com/facebook/mysql-5.6/pull/1287 is blocked without this api implemented. Without this api, the `early_load_rocksdb_plugin.test` will fail on the "mysql.password_history" because of the system table check inside of the engine check. I tried to return true for that table, it can pass the test. I will set a debug execute in the pr cited above to pass the mtr test.

laurynas-biveinis:
LGTM

sunshine-Chun:
PR merged, close here.

PR #1302, opened 2023-05-01
Bug #34638573 Compile MySQL with clang 15

laurynas-biveinis:
Clang 15 comes with a few new warnings:

variable 'cnt' set but not used
-Wunused-but-set-variable

argument 'ptr' of type 'const LinearSectionPtr[3]' with mismatched bound -Warray-parameter

out-of-line definition of constexpr static data member is redundant in C++17 and is deprecated
-Wdeprecated

unqualified call to 'std::move'
-Wunqualified-std-cast-call

builtin __has_trivial_copy is deprecated; use __is_trivially_copyable instead -Wdeprecated-builtins

Fix these, to get a clean build also with Clang 15.

Change-Id: I315b7b77e2bf3a6bacf091aeb8887d0902d3f950


sql/auth/sql_auth_cache.cc
```diff
@@ -2489,8 +2488,7 @@ bool acl_reload(THD *thd, bool mdl_locked) {
   old_acl_users = acl_users;
   old_acl_dbs = acl_dbs;
   old_acl_proxy_users = acl_proxy_users;
-  old_acl_restrictions = move(acl_restrictions);
-  old_acl_fast_lookup = acl_fast_lookup;
```
hermanlee, line 2493: Is this removal intended? This causes a memory leak because old_acl_fast_lookup is never set, so the old acl_fast_lookup is not deleted.

laurynas-biveinis: I made an error while cherry-picking. https://github.com/facebook/mysql-5.6/pull/1307

PR #1307, opened 2023-05-09
Restore an erroneously-removed assignment

laurynas-biveinis:
A regression from 78c1954822958d96a0ef527eeb3a0b8744be520d


laurynas-biveinis:
41d4cc7b17b74fd2d1b832f53f0233f24c5f473c

Issue #1295, opened 2023-04-20
force_flush_memtable_and_lzero_now always compacts to physical L1


mdcallag:
force_flush_memtable_and_lzero_now always compacts to the physical L1, which might not be in use when dynamic leveled is enabled. In that case it should compact to the logical L1, which might be the level named L2 or L3.

The buggy code, written by me, is here for [5.6.35](https://github.com/facebook/mysql-5.6/blob/2881af8d76226e3df563a3bf4439cd2451cae413/storage/rocksdb/ha_rocksdb.cc#L483) and [8.0.28](https://github.com/facebook/mysql-5.6/blob/5fdafcc1379c2f2f3de87bc26dc892885f04e28a/storage/rocksdb/ha_rocksdb.cc#L607).

Instead of "1" the argument to CompactFiles can be determined via a call to getProperty using the "rocksdb.base-level" property, or kBaseLevel. [See here](https://github.com/facebook/rocksdb/blob/main/db/internal_stats.cc#L397).

mdcallag:
AFAIK the fix has been merged for FB MySQL 8.0.32

Issue #1200, opened 2022-07-27
rocksdb_force_flush_memtable_and_lzero_now has a race


mdcallag:
The race is that there is no pause between flushing memtables (rocksdb_flush_all_memtables) and requesting compaction from L0 to L1. So if flushing memtables triggers some L0->L1 compaction then the attempt at a manual flush can fail. Code [is here](https://github.com/facebook/mysql-5.6/blob/fb-mysql-8.0.23/storage/rocksdb/ha_rocksdb.cc#L529).

While [this code](https://github.com/facebook/mysql-5.6/blob/fb-mysql-8.0.23/storage/rocksdb/ha_rocksdb.cc#L573) is there to catch a race, or maybe this race, it didn't based on the debug printfs I added because s.InvalidArgument() was not true. The diff for the debug printfs [is here](https://gist.github.com/mdcallag/b8a7668bb81ea6e60b8c1bfb50818498).

The workaround is to do this in my SQL scripts. First I flush the memtable, then guess at how long it might take for flush + compaction to finish, then request L0->L1 compaction:
```
set global rocksdb_force_flush_memtable_now=1;
select sleep(30);
set global rocksdb_force_flush_memtable_and_lzero_now=1;
```

I added printfs to the current code to see what happens. This is without the workaround:
```
2022-07-27T13:16:18.821162-07:00 12 [Note] [MY-010926] [Server] Access denied for user 'UNKNOWN_MYSQL_USER'@'localhost' (using password: NO)
2022-07-27T13:17:25.426564-07:00 18 [Note] [MY-000000] [Server] RocksDB: Manual memtable and L0 flush.
2022-07-27T13:17:25.705741-07:00 18 [Note] [MY-000000] [Server] RocksDB: ff for a CF default
2022-07-27T13:17:25.705788-07:00 18 [Note] [MY-000000] [Server] ... attempt
2022-07-27T13:17:25.705852-07:00 18 [Note] [MY-000000] [Server] ... calling CompactFiles
2022-07-27T13:17:25.706149-07:00 18 [Note] [MY-000000] [Server] ... not OK: Operation aborted: Necessary compaction input file /000041.sst is currently being compacted.
2022-07-27T13:17:25.706163-07:00 18 [Note] [MY-000000] [Server] RocksDB: ff for a CF __system__
2022-07-27T13:17:25.706169-07:00 18 [Note] [MY-000000] [Server] ... attempt
[...]

PR #1309, opened 2023-05-11
Fix issues 1200 and 1295, add rocksdb_compact_lzero_now global variable

mdcallag:
This fixes issue 1295 and has a workaround for issue 1200. https://github.com/facebook/mysql-5.6/issues/1200
https://github.com/facebook/mysql-5.6/issues/1295

The fix for issue 1295 is to get the value of the base level from RocksDB rather than assuming that L0 compacts into L1 because L1 might not be the base level when dynamic leveled compaction is used -- in that case the base level can be Ln where n > 1 because L1 isn't needed yet.

There are two workarounds for issue 1200. The hacky one is to sleep for 1 second between requesting memtable flush and L0 -> base_level compaction.

The alternative is to add a new global variable, rocksdb_compact_lzero_now, that when set will request L0 -> base_level compaction. This allows a client to first set rocksdb_force_flush_memtable_now, wait for that to finish, then set rocksdb_compact_lzero_now.

Other changes:
* rocksdb_force_flush_memtable_now default value changed to OFF to match what is done for rocksdb_force_flush_memtable_and_lzero_now and rocksdb_compact_lzero_now
* confirm that the value to which these variables are set can be parsed, that wasn't done for all of them
* don't raise an error when these are set to OFF, that will be a no-op

Note: these variables are triggers, an action is taken when this is done: set global var = ON | true | 1

But these variables always show the value 0 (OFF, false). Their value doesn't change.


mdcallag:
I ran the rocksdb, rocksdb_rpl and rocksdb_sys_vars suites

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -643,6 +659,60 @@ static int rocksdb_force_flush_memtable_and_lzero_now(
...
+    return HA_EXIT_SUCCESS;
+  }
+
+  // NO_LINT_DEBUG
+  LogPluginErrMsg(INFORMATION_LEVEL, ER_LOG_PRINTF_MSG,
+                  "RocksDB: Manual memtable and L0 flush.");
+  rocksdb_flush_all_memtables();
+
+  // Try to avoid https://github.com/facebook/mysql-5.6/issues/1200
+  my_sleep(1000000);
```
bladepan, line 711: shall we wait for the rocksdb fix https://github.com/facebook/rocksdb/issues/10428?

mdcallag: I see a PR to add a function to wait for compaction to finish. I don't see one for requesting L0->L1 compaction.
https://github.com/facebook/rocksdb/pull/11436

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -643,6 +659,60 @@ static int rocksdb_force_flush_memtable_and_lzero_now(
...
+  } else if (!parsed_value) {
+    // Setting to OFF is a no-op and this supports mtr tests
+    return HA_EXIT_SUCCESS;
+  }
+
+  return rocksdb_compact_lzero();
+}
+
+static void rocksdb_force_flush_memtable_and_lzero_now_stub(
+    THD *const thd MY_ATTRIBUTE((__unused__)),
```
luqun, line 686: nit: could you omit "thd MY_ATTRIBUTE((__unused__))"? if you don't specify parameter name, c++ treat as unused parameters.

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -643,6 +659,60 @@ static int rocksdb_force_flush_memtable_and_lzero_now(
...
+    // Setting to OFF is a no-op and this supports mtr tests
+    return HA_EXIT_SUCCESS;
+  }
+
+  // NO_LINT_DEBUG
+  LogPluginErrMsg(INFORMATION_LEVEL, ER_LOG_PRINTF_MSG,
+                  "RocksDB: Manual memtable and L0 flush.");
+  rocksdb_flush_all_memtables();
+
+  // Try to avoid https://github.com/facebook/mysql-5.6/issues/1200
```
luqun, line 710: If two threads both try to set rocksdb_compact_lzero=1 or rocksdb_force_flush_memtable_and_lzero_now=1, will they hit similar race?

mdcallag: Note that by race I mean a performance issue, not a correctness issue:
1) request memtable flush
2) immediately request L0 -> L1 compaction
3) wait for both to finish

If L0 -> L1 starts before memtable flush finishes then there might still be SSTs in the L0.

The "hack", sleep for 1 second in rocksdb_force_flush_memtable_and_lzero now between memtable flush and L0->L1 compaction, reduces the window for the race, but doesn't prevent it. There are other ways for the race to occur and I am reluctant to try and prevent them as any solution requires usage of a robust solution for waiting for compaction/flush to finish. Such a robust solution doesn't exist today. Hopefully it will in a few weeks or months. But even then I am still reluctant to try too hard to make this perfect.

It is more than good enough if I fix the bug WRT to getting the base level for L0->L1 compaction

storage/rocksdb/ha_rocksdb.cc
mdcallag: This is the style used by existing code. I prefer to match it.

mdcallag:
AFAIK the fix has been merged

PR #1310, opened 2023-05-12
Add different compression choice

sunshine-Chun:
Summary: when using clone to copy data in innodb instance and turn compression on. The speed is about 10 times slower than with compression off. The default compression algorithm used is zlib. Try to use other compression algorithm to tune the performance. From testing, zstd is twice faster than zlib

Test Plan:

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/plugin/clone/include/clone.h
+++ b/plugin/clone/include/clone.h
@@ -90,6 +90,10 @@ extern uint clone_max_io_bandwidth;
 /** Clone system variable: If network compression is enabled */
 extern bool clone_enable_compression;
 
+extern ulong clone_compression_algorithm;
+
+extern uint clone_zstd_compression_level;
+
 /** Clone system variable: SSL private key */
 extern char *clone_client_ssl_private_key;
 
--- a/plugin/clone/src/clone_client.cc
+++ b/plugin/clone/src/clone_client.cc
@@ -896,6 +896,13 @@ int Client::connect_remote(bool is_restart, bool use_aux) {
   mysql_clone_ssl_context ssl_context;
 
   ssl_context.m_enable_compression = clone_enable_compression;
+
+  if (ssl_context.m_enable_compression) {
+    ssl_context.m_compression_algorithm =
+        mysql_compression_lib_names[clone_compression_algorithm];
+    ssl_context.m_compression_level = clone_zstd_compression_level;
+  }
+
   ssl_context.m_server_extn =
       ssl_context.m_enable_compression ? &m_conn_server_extn : nullptr;
   ssl_context.m_ssl_mode = m_share->m_ssl_mode;
--- a/plugin/clone/src/clone_plugin.cc
+++ b/plugin/clone/src/clone_plugin.cc
@@ -70,6 +70,10 @@ uint clone_max_io_bandwidth;
 /** Clone system variable: If network compression is enabled */
 bool clone_enable_compression;
 
+ulong clone_compression_algorithm;
+
+uint clone_zstd_compression_level;
+
 /** Clone system variable: valid list of donor addresses. */
 static char *clone_valid_donor_list;
 
@@ -636,6 +640,26 @@ static MYSQL_SYSVAR_BOOL(enable_compression, clone_enable_compression,
                          "If compression is done at network", nullptr, nullptr,
                          false); /* Disable compression by default */
 
+enum enum_clone_compression_algorithm { ZLIB = 0, ZSTD };
+
+static TYPELIB clone_compression_algorithms_typelib = {
+    array_elements(mysql_compression_lib_names) - 1,
+    "clone_compression_algorithms_typelib", mysql_compression_lib_names,
+    nullptr};
+
+static MYSQL_SYSVAR_ENUM(compression_algorithm, clone_compression_algorithm,
+                         PLUGIN_VAR_OPCMDARG | PLUGIN_VAR_MEMALLOC,
+                         "compression algorithm used in clone", nullptr,
+                         nullptr, enum_clone_compression_algorithm::ZSTD,
+                         &clone_compression_algorithms_typelib);
+
+static MYSQL_SYSVAR_UINT(zstd_compression_level, clone_zstd_compression_level,
+                         PLUGIN_VAR_NOCMDARG, "zstd compression level", nullptr,
+                         nullptr, 3, /* Default */
+                         1,          /* Minimum */
+                         10,         /* Maximum */
+                         1);
+
 /** List of valid donor addresses allowed to clone from. */
 static MYSQL_SYSVAR_STR(valid_donor_list, clone_valid_donor_list,
                         PLUGIN_VAR_OPCMDARG | PLUGIN_VAR_MEMALLOC,
@@ -705,6 +729,8 @@ static SYS_VAR *clone_system_variables[] = {
     MYSQL_SYSVAR(ssl_cert),
     MYSQL_SYSVAR(ssl_ca),
     MYSQL_SYSVAR(donor_timeout_after_network_failure),
+    MYSQL_SYSVAR(compression_algorithm),
+    MYSQL_SYSVAR(zstd_compression_level),
     nullptr};
 
 /** Declare clone plugin */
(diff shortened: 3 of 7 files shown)


sunshine-Chun:
Need to fix test.

plugin/clone/src/clone_plugin.cc
```diff
@@ -636,6 +640,18 @@ static MYSQL_SYSVAR_BOOL(enable_compression, clone_enable_compression,
                          "If compression is done at network", nullptr, nullptr,
                          false); /* Disable compression by default */
 
+static MYSQL_SYSVAR_STR(compression_algorithm, clone_compression_algorithm,
```
laurynas-biveinis: Shoud be enum sysvar

plugin/clone/include/clone_client.h
```diff
@@ -51,6 +51,10 @@ using Time_Msec = std::chrono::milliseconds;
 using Time_Sec = std::chrono::seconds;
 using Time_Min = std::chrono::minutes;
 
+static const char *clone_compression_algorithms[] = {
+    "dummy",       "UNCOMPRESSED", "ZLIB",    "ZSTD",
```
laurynas-biveinis: Why "dummy" and "INVALID"?

plugin/clone/include/clone_client.h
```diff
@@ -51,6 +51,10 @@ using Time_Msec = std::chrono::milliseconds;
 using Time_Sec = std::chrono::seconds;
 using Time_Min = std::chrono::minutes;
 
+static const char *clone_compression_algorithms[] = {
```
laurynas-biveinis: If it's only used in a single source file, it should be declared there instead of a header. If it is used in multiple source files, it should not be `static`

plugin/clone/include/clone.h
```diff
@@ -90,6 +90,10 @@ extern uint clone_max_io_bandwidth;
 /** Clone system variable: If network compression is enabled */
 extern bool clone_enable_compression;
 
+extern unsigned long clone_compression_algorithm;
```
laurynas-biveinis:
```suggestion
extern ulong clone_compression_algorithm;
```
for consistency with `uint`

plugin/clone/include/clone_client.h
```diff
@@ -51,6 +51,8 @@ using Time_Msec = std::chrono::milliseconds;
 using Time_Sec = std::chrono::seconds;
 using Time_Min = std::chrono::minutes;
 
+extern const char *clone_compression_algorithms[3];
```
laurynas-biveinis: This (and the corresponding declaration in `clone_client.cc`) are not specific to the client, but general to both sides. IMHO should go to `clone_plugin.h`/`.cc`

plugin/clone/src/clone_client.cc
```diff
@@ -26,7 +26,10 @@ Clone Plugin: Client implementation
 
 */
 #include <inttypes.h>
+#include <string>
```
laurynas-biveinis: Redundant?

plugin/clone/src/clone_client.cc
```diff
@@ -26,7 +26,10 @@ Clone Plugin: Client implementation
...
 
+#include "my_compress.h"
```
laurynas-biveinis: Redundant?

plugin/clone/src/clone_plugin.cc
```diff
@@ -70,6 +70,10 @@ uint clone_max_io_bandwidth;
 /** Clone system variable: If network compression is enabled */
 bool clone_enable_compression;
 
+unsigned long clone_compression_algorithm;
```
laurynas-biveinis:
```suggestion
ulong clone_compression_algorithm;
```
for symmetry with `uint`

plugin/clone/include/clone_client.h
sunshine-Chun: Will this introduce circular dependency between clone_plugin.h and clone_client.h? The clone_plugin.h already include clone_client.h. With this change, the clone_client will need to include clone_plugin.h as well.

sql/server_component/clone_protocol_service.cc
```diff
@@ -394,6 +394,11 @@ DEFINE_METHOD(MYSQL *, mysql_clone_connect,
   /* Enable compression. */
   if (ssl_ctx->m_enable_compression) {
     mysql_options(mysql, MYSQL_OPT_COMPRESS, nullptr);
+    if (ssl_ctx->m_compression_algorithm)
+      mysql_options(mysql, MYSQL_OPT_COMPRESSION_ALGORITHMS,
+                    ssl_ctx->m_compression_algorithm);
+    mysql_options(mysql, MYSQL_OPT_ZSTD_COMPRESSION_LEVEL,
+                  &(ssl_ctx->m_compression_level));
```
laurynas-biveinis:
```suggestion
                  &ssl_ctx->m_compression_level);
```

plugin/clone/include/clone_client.h
laurynas-biveinis: I am very sorry, I meant to say `clone.h`, unless it results in a circular dependency.

include/mysql/plugin_clone.h
```diff
@@ -69,6 +69,8 @@ struct Mysql_clone {
   int (*clone_server)(THD *thd, MYSQL_SOCKET socket);
 };
 
+extern const char *clone_compression_algorithms[3];
```
laurynas-biveinis: See the other comment about `clone.h`.
Also 
```suggestion
extern const char *clone_compression_algorithms[];
```
(the 3 is dropped by the language)

sunshine-Chun: I am afraid we will hit compiler error if we don't specify the array size here. The clone_compression_algorithm array will be used in array_elements. It requires a fixed size for the  parameter.

```
template <class T, size_t N>
constexpr size_t array_elements(T (&)[N]) noexcept {
  return N;
}
```

laurynas-biveinis (commented):
LGTM

sunshine-Chun:
Looks like we have memory leakage from valgrind test. Need to debug a bit. Will update here once find anything.
```
  at 0xA5A6795: malloc (vg_replace_malloc.c:381)
   by 0x90917AD: ZSTD_customMalloc (allocations.h:30)
   by 0x90917AD: ZSTD_createDCtx_internal (zstd_decompress.c:296)
   by 0x90917AD: ZSTD_createDCtx (zstd_decompress.c:312)
by 0x7700BEA: zstd_uncompress(mysql_zstd_compress_context*, unsigned char*, unsigned long, unsigned long*) [clone .__uniq.287700427806220246369010965997275247218] (my_compress.cc:232)
   by 0x77009DE: my_uncompress(mysql_compress_context*, unsigned char*, unsigned long, unsigned long*) (my_compress.cc:699)
   by 0x64D4473: net_read_compressed_packet(NET*, unsigned long&) [clone .__uniq.40191296941203724010080517394248553221] (net_serv.cc:2259)
   by 0x64D4364: my_net_read(NET*) (net_serv.cc:2297)
   by 0x85CBC4D: mysql_clone_get_response(THD*, MYSQL*, bool, unsigned int, unsigned char**, unsigned long*, unsigned long*) (clone_protocol_service.cc:545)

```

sunshine-Chun:
Memory leakage fixed.

sql/server_component/clone_protocol_service.cc
```diff
@@ -539,6 +545,13 @@ DEFINE_METHOD(int, mysql_clone_get_response,
   *net_length = 0;
   *length = my_net_read(net);
 
+  /** The ZSTD object has been updated to include the newly created decompressor
```
laurynas-biveinis: This is not a Doxygen comment:
```suggestion
  /* The ZSTD object has been updated to include the newly created decompressor
```

sql/server_component/clone_protocol_service.cc
```diff
@@ -539,6 +545,13 @@ DEFINE_METHOD(int, mysql_clone_get_response,
...
+   * context. Include the change in the net extension so that the resource can
```
laurynas-biveinis: This and the next line: the comment style with leading stars is not used in MySQL:
```suggestion
   context. Include the change in the net extension so that the resource can
```

laurynas-biveinis (commented):
LGTM with minor comment formatting issue fixed

PR #1311, opened 2023-05-18
Cleanup non-null pointer args, increase const correctness

laurynas-biveinis:
Various minor cleanups

All done or noticed in the native DD prototype work, are independent from it,
and will help to reduce its diff.

- `ha_rocksdb::create_key_defs` had non-null attribute for the 2nd and 3rd
  argument. It seems that the attribute non-null arg indexes got desynced,
  because the 3rd argument is a reference, and by usage the 1st and the 2nd args
  are pointers that cannot be `nullptr`. Instead of fixing the attribute indexes,
  convert pointers to references, which cannot be null by design.
- Do the same for `ha_rocksdb::create_cfs`. At the same time make `tbl_def_arg`
  `const`, which previously was not, and make the function return `bool`, as it only
  returns two values for success and error. Convert it from
  `DBUG_ENTER_FUNC`/`DBUG_RETURN` to a newer (and lighter) `DBUG_TRACE`.
- `ha_rocksdb::create_table` made `private`, pointer args converted to reference
  args.
- Added missing error checking for RocksDB `WriteBatch::SingleDelete` calls
- `ha_rocksdb::has_hidden_pk` made `static`, args changed from pointers to
  references, removed redundant `nonnull` attribute.
- `ha_rocksdb::pk_index` args changed from pointers to references, removed
  redundant `nonnull` attribute.
- `ha_rocksdb::set_last_rowkey` argument was marked unused even though it was used
- `ha_rocksdb::check_and_lock_sk`, `ha_rocksdb::check_uniqueness_and_lock`,
  `ha_rocksdb::update_write_row`, `ha_rocksdb::reset`,
  `ha_rocksdb::calc_updated_indexes`: added asserts for relationships between
  arguments and `table->record[0]`/`table->record[1]`
- `ha_rocksdb::adjust_handler_stats_sst_and_memtable`: reduce scope of two local
  vars
- `Rdb_key_def`: introduce invalid index ID constant based on
  `dd::INVALID_OBJECT_ID`. Check it on construction and in the index ID getter.
  Replace direct field accesses with the getter call.
- `Rdb_key_def::table_has_hidden_pk` arg changed from pointer to reference.
- Replace `warn_unused_result` attribute with the standard `[[nodiscard]]` for all
  the touched signatures.

### DIFF
--- a/storage/rocksdb/ha_rocksdb.cc
+++ b/storage/rocksdb/ha_rocksdb.cc
@@ -5304,8 +5304,7 @@ class Rdb_writebatch_impl : public Rdb_transaction {
           "Not supported for intrinsic tmp tables");
     }
     ++m_write_count[table_type];
-    m_batch->SingleDelete(column_family, key);
-    return rocksdb::Status::OK();
+    return m_batch->SingleDelete(column_family, key);
   }
 
   bool has_modifications() const override {
@@ -8513,7 +8512,7 @@ exit:
 
 /* Get PK value from m_tbl_def->m_hidden_pk_info. */
 longlong ha_rocksdb::update_hidden_pk_val() {
-  assert(has_hidden_pk(table));
+  assert(has_hidden_pk(*table));
   const longlong new_val = m_tbl_def->m_hidden_pk_val++;
   return new_val;
 }
@@ -8521,7 +8520,7 @@ longlong ha_rocksdb::update_hidden_pk_val() {
 /* Get the id of the hidden pk id from m_last_rowkey */
 int ha_rocksdb::read_hidden_pk_id_from_rowkey(longlong *const hidden_pk_id) {
   assert(table != nullptr);
-  assert(has_hidden_pk(table));
+  assert(has_hidden_pk(*table));
 
   rocksdb::Slice rowkey_slice(m_last_rowkey.ptr(), m_last_rowkey.length());
 
@@ -8826,7 +8825,7 @@ int ha_rocksdb::alloc_key_buffers(const TABLE *const table_arg,
   uint max_packed_sk_len = 0;
   uint pack_key_len = 0;
 
-  m_pk_descr = kd_arr[pk_index(table_arg, tbl_def_arg)];
+  m_pk_descr = kd_arr[pk_index(*table_arg, *tbl_def_arg)];
 
   // move this into get_table_handler() ??
   m_pk_descr->setup(table_arg, tbl_def_arg);
@@ -8993,7 +8992,7 @@ int ha_rocksdb::open(const char *const name,
   }
 
   /* Load hidden pk only once on first use. */
-  if (has_hidden_pk(table) && m_tbl_def->m_hidden_pk_val == 0 &&
+  if (has_hidden_pk(*table) && m_tbl_def->m_hidden_pk_val == 0 &&
       (err = load_hidden_pk_value()) != HA_EXIT_SUCCESS) {
     rdb_open_tables.release_table_handler(m_table_handler);
     free_key_buffers();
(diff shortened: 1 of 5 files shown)


storage/rocksdb/rdb_datadic.h
```diff
@@ -810,6 +816,9 @@ class Rdb_key_def {
   }
 #endif  // NDEBUG
 
+  static constexpr auto INVALID_INDEX_NUMBER =
+      static_cast<std::uint32_t>(dd::INVALID_OBJECT_ID);
```
luqun, line 820: maybe just use -1 instead of INVALID_OBJECT_ID.. since it looks like INDEX_NUMBER and INVALID_OBJECT_ID are two different number.

storage/rocksdb/rdb_datadic.cc
```diff
@@ -954,7 +955,7 @@ uint Rdb_key_def::get_primary_key_tuple(const Rdb_key_def &pk_descr,
   assert(m_pk_key_parts);
 
   /* Put the PK number */
-  rdb_netbuf_store_index(buf, pk_descr.m_index_number);
```
luqun, line 957: nit: could you also change m_index_number as private field?

storage/rocksdb/rdb_datadic.h
laurynas-biveinis: I can, but these numbers will come from the same number space in the DDSE, just with 64 -> 32 truncation. It was `INVALID_OBJECT_ID` that was the motivation for this change in the prototype: to catch uninitialized key definitions being created, which was one of the bugs I had. Given this, do you still want me to use -1?

storage/rocksdb/rdb_datadic.cc
laurynas-biveinis: It seems it is already a private field

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -12456,16 +12463,24 @@ int ha_rocksdb::update_write_sk(const TABLE *const table_arg,
       rc = check_partial_index_prefix(table_arg, kd, row_info.tx,
                                       row_info.old_data);
       if (!rc) {
-        row_info.tx->get_indexed_write_batch(m_tbl_def->get_table_type())
-            ->SingleDelete(kd.get_cf(), old_key_slice);
+        const auto s =
+            row_info.tx->get_indexed_write_batch(m_tbl_def->get_table_type())
+                ->SingleDelete(kd.get_cf(), old_key_slice);
+        if (!s.ok()) {
+          return rdb_error_to_mysql(s);
```
luqun: nit: in ha_rocksdb::delete_row(), if SD or delete fail, it will call tx->set_status_error() instead of rdb_error_to_mysql(). Although these two function are similar, maybe we should have same behavior?

storage/rocksdb/rdb_datadic.h
luqun: For "to catch uninitialized key definitions being created, which was one of the bugs I had.", do you need to assign m_index_number with default value INVALID_INDEX_NUMBER if key isn't initialized?

saw INVALID_OBJECT_ID is also used in innodb, guess its value won't be changed later.. so we should be okay.

laurynas-biveinis: MyRocks does not seem to create "uninitialized" keys ATM, all the object instances are valid and represent valid indexes. The `INVALID_OBJECT_ID` comes from the persistent ID space, so changing it would be a big deal with upgrade code being needed etc.

luqun: okay.. then current change looks good me..

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -9933,15 +9932,15 @@ static void rdb_gen_normalized_tablename(const std::string *db,
 */
 int ha_rocksdb::create_table(const std::string &table_name,
                              const std::string &actual_user_table_name,
-                             const TABLE *table_arg,
+                             const TABLE &table_arg,
                              ulonglong auto_increment_value,
-                             dd::Table *table_def MY_ATTRIBUTE((__unused__)),
+                             const dd::Table &table_def [[maybe_unused]],
```
luqun: This change will cause issue.. due to https://github.com/facebook/mysql-5.6/blob/fb-mysql-8.0.28/sql/sql_tmp_table.cc#L2279 will pass table_def as nullptr..

laurynas-biveinis: Good catch. This can only happen with optimizer temp tables, which in future should be allowed to materialize in MyRocks too

laurynas-biveinis:
Updated, ready for review

luqun:
still waiting for test result and will update when it is done.

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -12456,16 +12463,25 @@ int ha_rocksdb::update_write_sk(const TABLE *const table_arg,
...
       }
     } else {
       // Unconditionally issue SD if rocksdb_partial_index_blind_delete.
-      row_info.tx->get_indexed_write_batch(m_tbl_def->get_table_type())
-          ->SingleDelete(kd.get_cf(), old_key_slice);
+      const auto s =
+          row_info.tx->get_indexed_write_batch(m_tbl_def->get_table_type())
+              ->SingleDelete(kd.get_cf(), old_key_slice);
+      if (!s.ok()) {
+        return rdb_error_to_mysql(s);
```
luqun, line 12483: nit: maybe also change this from rdb_error_to_mysql to row_info.tx->set_status_error().

PR #1312, opened 2023-05-29
Move default PATH setting from CMake to get_rocksdb_files.sh

laurynas-biveinis:
In c3f1703231d7fec34de1479b8913a06cd9c4cd79, get_rocksdb_files.sh invocation got PATH setting. The public discussion is incomplete but b739eac1760c2451200246c5d7fe3233787053b8 suggests that there is a problem with CentOS 9 and make 4.3 that undefined PATH does not get a default fallback, thus a PATH value is provided for the invocation.

The provided PATH value is however incompatible with building on macOS. Fix by setting the fallback PATH value in the get_rocksdb_files.sh script itself.

### DIFF
--- a/storage/rocksdb/CMakeLists.txt
+++ b/storage/rocksdb/CMakeLists.txt
@@ -52,7 +52,7 @@ ELSE()
   # get a list of rocksdb library source files
   # run with env -i to avoid passing variables
   EXECUTE_PROCESS(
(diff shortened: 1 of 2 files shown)


laurynas-biveinis:
I tried reproducing the original issue with CentOS 9, GNU Make 4.3 on 8cc47c6b, but couldn't. Both current HEAD and ths commit build there fine for me too. Thus all I can say is that this change fixes the issue for me, and I hope it doesn't break in your environment.

PR #1313, opened 2023-05-30
Explicitly specify char signedness for the underlying enum type

laurynas-biveinis:
A plain 'char' without 'signed' or 'unsigned' may default to either one, depending on the platform. On x86_64 Linux it is signed, while on ARM Linux it is unsigned, resulting in a build error. Fix by explicitly marking it signed.

Squash with 295ba21bc021c0758ac4f6bb7c35b54ac5093012

### DIFF
--- a/client/mysqlbinlog.cc
+++ b/client/mysqlbinlog.cc
@@ -792,7 +792,7 @@ static const std::string binlog_server_finish_err_msg =
     "The binlog server has finished sending all available binlogs from the "
     "HDFS and has no more binlogs to send.";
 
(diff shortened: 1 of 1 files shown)


luqun:
could we just use default enum type int instead? I mean remove ": char "?

laurynas-biveinis:
> could we just use default enum type int instead? I mean remove ": char "?

Indeed it seems the users of the enum do not depend on the underlying type, I have update the PR to remove it.

laurynas-biveinis:
The PR has been merged with the new code but old commit message?

PR #1314, opened 2023-06-02
MyRocks: implement attachable transaction support

laurynas-biveinis:
The storage engine requirement for supporting attachable transaction is being
able to handle vanished THD::ha_data field in the middle of transaction. The
server layer swaps its existing data with a nullptr before calling the SE, and
expects the SE to allocate a new transaction object and to invalidate any cached
info depending on the previous one. Once the server is done with the attachable
transaction, it swaps back the original SE data pointer.

MyRocks already happens to work correctly in this case, so the main change is to
advertise the attachable transaction support by adding
HA_ATTACHABLE_TRX_COMPATIBLE to ha_rocksdb::table_flags return value.

Since the attachable transactions are
AC-NL-RO-RC (AutoCommit-NonLocking-ReadOnly-ReadCommitted) at this point, the
rest of code patch introduces invariants asserting these properties for
Rdb_transcation and child class objects and minor changes such as not setting
lock timeouts as no locks can be taken. Specifically, such transaction is
indicated by m_max_row_locks == 0. This is decided in set_params which now
handles both regular and attachable transactions.

Make m_ddl_transaction a private field with getters and setters so that the
invariants could be checked.

Factor out is_autocommit and get_ha_data_or_null helper functions.

At the same time make cleanups in the touched code:
- make field modified_tables private (trivially)
- remove unused write-only field m_mysql_gtid.
- simplify can_acquire_snapshot_without_conflicts, make it `[[nodiscard]]`.
- make ingest_bulk_load_files take const reference argument.
- remove redundant "virtual" keyword from
  Rdb_transaction_impl::is_writebatch_trx.
- remove unused method Rdb_transaction_impl::has_snapshot

For testing, convert the existing InnoDB attachable_trx.test to an include file
that is shared by both engine tests.

At a later point the attachable transactions may become read-write (if i.e.
mysql.gtid_executed is converted to MyRocks) and read-uncommitted, if this
isolation level will be implemented for the DDSE. In this case the invariant
will have to be split.

Reference:
- Server attachable transaction support:
  https://dev.mysql.com/worklog/task/?id=8003
- InnoDB attachable transaction support:
  https://dev.mysql.com/worklog/task/?id=7828
- The initial InnoDB implementation: 5e9eedb9123d0199f6c42c3e5f0a8cd24260c07c


laurynas-biveinis:
Please review only the last commit of the three, the first two will disappear once their branch is merged

mysql-test/suite/innodb/t/attachable_trx.test
```diff
@@ -1,154 +1,3 @@
 --source include/have_debug.inc
```
luqun, line 1: nit: do you still need this? since mysql-test/include/attachable_trx.inc already contains this..

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -3295,6 +3295,10 @@ static bool rdb_is_ttl_read_filtering_enabled();
 static int rdb_dbug_set_ttl_read_filter_ts();
 #endif
 
+[[nodiscard]] static bool is_autocommit(const THD &thd) {
```
luqun, line 3431: could we cache the value for this(to avoid recalculation)? such as add a member variable(such as m_is_autocommit) into Rdb_transaction?

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -5206,30 +5290,55 @@ class Rdb_writebatch_impl : public Rdb_transaction {
...
   /* Implementations of do_*savepoint based on rocksdB::WriteBatch savepoints */
-  void do_set_savepoint() override { m_batch->SetSavePoint(); }
+  void do_set_savepoint() override {
+    assert(!is_ac_nl_ro_rc_transaction());
+
+    m_batch->SetSavePoint();
+  }
+
   rocksdb::Status do_pop_savepoint() override {
+    assert(!is_ac_nl_ro_rc_transaction());
```
luqun, line 5433: Looks like there are a lot of these asserts(easy to missing one)? maybe we can do reverse assert? such as assert( if it is attachable trx, then it should be ac_nl_ro_rc in myrocks)?

assert_IFF(thd->is_attachable_ro_transaction_active(), is_ac_nl_ro_rc_transaction());

mysql-test/suite/innodb/t/attachable_trx.test
laurynas-biveinis: It does its test run time effect there, but MTR framework also greps the testcases for it to classify them, i.e. https://github.com/facebook/mysql-5.6/blob/18381192a55a018c0fb8a838659019f2d1f1b3e4/mysql-test/lib/mtr_cases.pm#L1534

storage/rocksdb/ha_rocksdb.cc
laurynas-biveinis: We could, but I am not sure it's worth it: the savings would consist of a single AND operation, and maybe of some pointer loads. In all the cases the THD is readily available (no cases of getting it through `current_thd`).

storage/rocksdb/ha_rocksdb.cc
laurynas-biveinis: These asserts serve to mark the code paths that are not allowed to execute under attachable transactions. Not sure how this check is possible to implement with inverted (and there would be many "inverted" locations too). 

The missed assert locations are indeed a problem, but not critical one. Since the asserts are on all levels on callstacks, they would likely to trigger on different levels, allowing to improve the code.

luqun:
@laurynas-biveinis , could you rebase this PR? it will easy to do testing after rebase.

laurynas-biveinis:
> @laurynas-biveinis , could you rebase this PR? it will easy to do testing after rebase.

Will do

laurynas-biveinis:
@luqun Rebased

laurynas-biveinis:
Found a minor issue after rebasing the whole prototype branch

laurynas-biveinis:
Ready for review. It is independent from https://github.com/facebook/mysql-5.6/pull/1323, although trivial text conflicts might emerge between the two.

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -5417,11 +5534,15 @@ class Rdb_writebatch_impl : public Rdb_transaction {
...
                              const bool assume_tracked
                                  MY_ATTRIBUTE((unused))) override {
+    assert(!is_ac_nl_ro_rc_transaction());
+
     if (table_type == TABLE_TYPE::INTRINSIC_TMP) {
       assert(false);
       return rocksdb::Status::NotSupported(
           "Not supported for intrinsic tmp tables");
     }
+    assert(!is_ac_nl_ro_rc_transaction());
```
luqun: nit: line 5544 duplicate with line 5537?

storage/rocksdb/ha_rocksdb.cc
luqun: yeah, previous I was thinking that add a variable m_dd_trx whose value is from thd_tx_is_dd_trx API, if m_dd_trx is true, then it must use attachable trx(AC-NL-RO-RC).. anyway, since you added all(maybe almost all) API with your current assert and testing looks good so far, maybe keep as it is..

laurynas-biveinis:
Updated & rebased

PR #1316, opened 2023-06-05
Refactor data dictionary transaction isolation setting

laurynas-biveinis:
InnoDB uses READ UNCOMMITTED, which is not supported with MyRocks. Instead of
hardcoding READ UNCOMMITTED at several locations, introduce a helper function
that returns the desired DD transaction isolation level, based on the default DD
engine.

No functional changes if the default DD engine is InnoDB.

### DIFF
--- a/sql/dd/impl/tables/dd_properties.cc
+++ b/sql/dd/impl/tables/dd_properties.cc
@@ -34,6 +34,7 @@
 #include "mysql/udf_registration_types.h"
 #include "mysqld_error.h"
 #include "sql/auth/sql_security_ctx.h"
+#include "sql/dd/dd_utility.h"
 #include "sql/dd/dd_version.h"  // dd::DD_VERSION
 #include "sql/dd/impl/raw/raw_table.h"
 #include "sql/dd/impl/transaction_impl.h"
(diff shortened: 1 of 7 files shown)


laurynas-biveinis:
There are two commits but only the second one is for review. The first one will be removed once the prerequisite branch is merged.

storage/rocksdb/get_rocksdb_files.sh
```diff
@@ -1,4 +1,8 @@
 #!/bin/bash
+if [ -z "$PATH" ]; then
+    export PATH="/sbin:/usr/sbin:/bin:/usr/bin"
```
luqun, line 3: unrelated changes?

laurynas-biveinis: This branch contains a commit that fixes the build for me. I'll remove it once it is merged, until then, please review 3dae3ef

sql/dd/impl/tables/dd_properties.cc
```diff
@@ -117,12 +118,7 @@ bool DD_properties::init_cached_properties(THD *thd) {
   // Early exit in case the properties are already initialized.
   if (!m_properties.empty()) return false;
 
-  /*
-    Start a DD transaction to get the properties. Please note that we
-    must do this read using isolation level ISO_READ_UNCOMMITTED
```
luqun, line 122: could we keep these comments? maybe useful for myrock atomic DDL

laurynas-biveinis: They are consolidated in dd_utility.h

PR #1319, opened 2023-06-12
Refactor the acquisition of DDSE handlerton

laurynas-biveinis:
Make it support both InnoDB and MyRocks DDSEs by introducing a helper function
dd::get_dd_engine that returns either based on the default_dd_storage_engine
sysvar setting. To have correct diagnostics in one instance at least, introduce
a get_dd_engine_name utility function too.

No functional changes if default_dd_storage_engine == InnoDB.


luqun:
still work on update one MTR(rocksdb.early_load_rocksdb_plugin) according to this PR.

laurynas-biveinis:
> still work on update one MTR(rocksdb.early_load_rocksdb_plugin) according to this PR.

@luqun, I am sorry, I forgot that we already have an MTR test for DDSE == RocksDB. Do you need me to do anything at this point?

PR #1321, opened 2023-06-13
implement handler API table_exists_in_engine for myrocks

luqun:
Summary:
During Data Dictioanry(DD) upgrade or restart, mysql should decide which SE(Storage Engine) contains actual DD tables if multiple SEs can be DDSE. 

Currently MyRocks still have its own private data dictionary, which contains table name->table_def mapping.  also these private data dictionary is initialized during plugin initialization. Thus implements handler API  table_exists_in_engine to check whether a table exists in myrocks internal data dictionary.

 

Test Plan:

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/sql/dd/impl/bootstrap/bootstrap_ctx.h
+++ b/sql/dd/impl/bootstrap/bootstrap_ctx.h
@@ -30,7 +30,8 @@
 #include "mysql_version.h"                // MYSQL_VERSION_ID
 #include "sql/dd/dd_version.h"            // DD_VERSION
 #include "sql/dd/info_schema/metadata.h"  // IS_DD_VERSION
-#include "sql/mysqld.h"                   // opt_initialize
+#include "sql/handler.h"
+#include "sql/mysqld.h"  // opt_initialize
 
 class THD;
 
@@ -100,7 +101,12 @@ class DD_bootstrap_ctx {
 
   uint m_did_I_S_upgrade_from = 0;
   uint m_actual_I_S_version = 0;
-  uint m_actual_dd_engine = 0;
+  // actual DDSE holds dd tables
+  // its values is indicted by which SE hold mysql.dd_properties table
+  //  Druing boostrap time,  its value is equal to default_dd_storage_engine
+  //  During upgrade time, its value maybe not equal to
+  //  default_dd_storage_engine
+  legacy_db_type m_actual_dd_engine = DB_TYPE_UNKNOWN;
 
  public:
   DD_bootstrap_ctx() = default;
@@ -124,7 +130,8 @@ class DD_bootstrap_ctx {
     m_actual_I_S_version = actual_I_S_version;
   }
 
-  void set_actual_dd_engine(uint actual_dd_engine) {
+  void set_actual_dd_engine(enum legacy_db_type actual_dd_engine) {
+    assert(m_actual_dd_engine == DB_TYPE_UNKNOWN);
     m_actual_dd_engine = actual_dd_engine;
   }
 
@@ -132,7 +139,10 @@ class DD_bootstrap_ctx {
 
   uint get_actual_I_S_version() const { return m_actual_I_S_version; }
 
-  uint get_actual_dd_engine() const { return m_actual_dd_engine; }
+  uint get_actual_dd_engine() const {
+    assert(m_actual_dd_engine != DB_TYPE_UNKNOWN);
+    return m_actual_dd_engine;
+  }
 
   void set_dd_upgrade_done() {
     assert(m_did_dd_upgrade_from == 0);
--- a/sql/dd/impl/bootstrap/bootstrap_ctx.h
+++ b/sql/dd/impl/bootstrap/bootstrap_ctx.h
@@ -139,7 +139,7 @@ class DD_bootstrap_ctx {
 
   uint get_actual_I_S_version() const { return m_actual_I_S_version; }
 
-  uint get_actual_dd_engine() const {
+  legacy_db_type get_actual_dd_engine() const {
     assert(m_actual_dd_engine != DB_TYPE_UNKNOWN);
     return m_actual_dd_engine;
   }
--- a/sql/dd/impl/bootstrap/bootstrap_ctx.h
+++ b/sql/dd/impl/bootstrap/bootstrap_ctx.h
@@ -100,6 +100,7 @@ class DD_bootstrap_ctx {
 
   uint m_did_I_S_upgrade_from = 0;
   uint m_actual_I_S_version = 0;
+  uint m_actual_dd_engine = 0;
 
  public:
   DD_bootstrap_ctx() = default;
@@ -123,9 +124,15 @@ class DD_bootstrap_ctx {
     m_actual_I_S_version = actual_I_S_version;
   }
 
+  void set_actual_dd_engine(uint actual_dd_engine) {
+    m_actual_dd_engine = actual_dd_engine;
+  }
+
   uint get_actual_dd_version() const { return m_actual_dd_version; }
 
   uint get_actual_I_S_version() const { return m_actual_I_S_version; }
+
+  uint get_actual_dd_engine() const { return m_actual_dd_engine; }
 
   void set_dd_upgrade_done() {
     assert(m_did_dd_upgrade_from == 0);
--- a/sql/dd/impl/bootstrap/bootstrapper.cc
+++ b/sql/dd/impl/bootstrap/bootstrapper.cc
@@ -1084,15 +1084,11 @@ bool initialize_dd_properties(THD *thd) {
           rocksdb_hton, thd, MYSQL_SCHEMA_NAME.str,
           dd::tables::DD_properties::instance().name().c_str());
     }
-    if (error == HA_ERR_TABLE_EXIST) {
-      // actual ddse maybe rocksdb
-      bootstrap::DD_bootstrap_ctx::instance().set_actual_dd_engine(
-          DB_TYPE_ROCKSDB);
-    } else {
-      // actual ddse maybe innodb
-      bootstrap::DD_bootstrap_ctx::instance().set_actual_dd_engine(
-          DB_TYPE_INNODB);
-    }
+
+    // HA_ERR_TABLE_EXIST means actual ddse maybe rocksdb
+    // otherwise, actual ddse maybe innodb
+    bootstrap::DD_bootstrap_ctx::instance().set_actual_dd_engine(
+        error == HA_ERR_TABLE_EXIST ? DB_TYPE_ROCKSDB : DB_TYPE_INNODB);
   }
 
   // Create the dd_properties table.
--- a/sql/dd/impl/bootstrap/bootstrapper.cc
+++ b/sql/dd/impl/bootstrap/bootstrapper.cc
@@ -1073,6 +1073,27 @@ bool create_dd_schema(THD *thd) {
 }
 
 bool initialize_dd_properties(THD *thd) {
+  if (!opt_initialize) {
+    // Find out which ddse contains dd_properties table during restart
+    // Try rocksdb first
+    handlerton *rocksdb_ddse = ha_resolve_by_legacy_type(thd, DB_TYPE_ROCKSDB);
+    int error = HA_ERR_NO_SUCH_TABLE;
+    if (rocksdb_ddse != nullptr || rocksdb_ddse->table_exists_in_engine) {
+      error = rocksdb_ddse->table_exists_in_engine(
+          rocksdb_ddse, thd, MYSQL_SCHEMA_NAME.str,
+          dd::tables::DD_properties::instance().name().c_str());
+    }
+    if (error == HA_ERR_TABLE_EXIST) {
+      // actual ddse maybe rocksdb
+      bootstrap::DD_bootstrap_ctx::instance().set_actual_dd_engine(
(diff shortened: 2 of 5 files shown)


storage/rocksdb/ha_rocksdb.cc
```diff
@@ -7168,6 +7168,28 @@ static bool rocksdb_get_table_statistics(
...
+  NOTE: Don't call any DD API, due to this function maybe called during
+  DD initializing.
+
+  @return HA_ERR_TABLE_EXIST - if table exist in ROCKSDB
+  @return HA_ERR_NO_SUCH_TABLE - if table isn't exists in ROCKSDB.
+*/
+static int rocksdb_table_exists_in_engine(handlerton *, THD *,
+                                          const char *db_name,
+                                          const char *table_name) {
+  std::string fullname = db_name;
```
laurynas-biveinis: Copy-patste with nosql_access.c:`select_exec::run` (around line 2140), `rocksdb_get_table_statistics`, and `rdb_normalize_tablename`. Can you extract a helper? If you do, start with an empty string and do `reserve` first for the exact length

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -7168,6 +7168,28 @@ static bool rocksdb_get_table_statistics(
...
+  fullname.append(".");
+  fullname.append(table_name);
+
+  auto tbl_def = ddl_manager.find(fullname);
```
```suggestion
  const auto * const tbl_def = ddl_manager.find(fullname);
```

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -7168,6 +7168,28 @@ static bool rocksdb_get_table_statistics(
...
+  if (!tbl_def) {
```
laurynas-biveinis: Consider single return with a ternary operator

sql/dd/impl/bootstrap/bootstrapper.cc
```diff
@@ -1072,6 +1072,27 @@ bool create_dd_schema(THD *thd) {
 }
 
 bool initialize_dd_properties(THD *thd) {
+  if (!opt_initialize) {
+    // Find out which ddse contains dd_properties table during restart
+    // Try rocksdb first
+    handlerton *rocksdb_ddse = ha_resolve_by_legacy_type(thd, DB_TYPE_ROCKSDB);
+    int error = HA_ERR_NO_SUCH_TABLE;
+    if (rocksdb_ddse != nullptr || rocksdb_ddse->table_exists_in_engine) {
```
laurynas-biveinis: Important suggestion:
```suggestion
    if (rocksdb_ddse != nullptr && rocksdb_ddse->table_exists_in_engine) {
```

Less important suggestion (both subexpressions should use the same way for null comparison):
```suggestion
    if (rocksdb_ddse != nullptr && rocksdb_ddse->table_exists_in_engine != nullptr) {
```

sql/dd/impl/bootstrap/bootstrapper.cc
```diff
@@ -1072,6 +1072,27 @@ bool create_dd_schema(THD *thd) {
 }
 
 bool initialize_dd_properties(THD *thd) {
+  if (!opt_initialize) {
+    // Find out which ddse contains dd_properties table during restart
```
laurynas-biveinis: There can only be one DDSE, so the comment is not correct. I'd just say SE
```suggestion
    // Find out which SE contains dd_properties table during restart
```

sql/dd/impl/bootstrap/bootstrapper.cc
```diff
@@ -1072,6 +1072,27 @@ bool create_dd_schema(THD *thd) {
...
+    // Try rocksdb first
+    handlerton *rocksdb_ddse = ha_resolve_by_legacy_type(thd, DB_TYPE_ROCKSDB);
```
laurynas-biveinis: Likewise
```suggestion
    handlerton *rocksdb_hton = ha_resolve_by_legacy_type(thd, DB_TYPE_ROCKSDB);
```

sql/dd/impl/bootstrap/bootstrap_ctx.h
```diff
@@ -100,6 +100,7 @@ class DD_bootstrap_ctx {
 
   uint m_did_I_S_upgrade_from = 0;
   uint m_actual_I_S_version = 0;
+  uint m_actual_dd_engine = 0;
```
laurynas-biveinis: Here and elsewhere use the actual type:
```suggestion
  enum legacy_db_type m_actual_dd_engine = DB_TYPE_UNKNOWN;
```

laurynas-biveinis: This needs a comment explaining what is the "actual DDSE". If I understand correctly, it is the one that at the moment happens to contain the DD tables, as indicated by the presence of `mysql.dd_properties`. It is assumed (again, if I understand correctly), that moving any single table between SEs is atomic, and so `dd_properties` will be present in exactly one SE at all times. The actual DDSE may be different from `default_dd_storage_engine` setting. There may be more subtlety still.

sql/dd/impl/bootstrap/bootstrapper.cc
```diff
@@ -1072,6 +1072,27 @@ bool create_dd_schema(THD *thd) {
...
+    int error = HA_ERR_NO_SUCH_TABLE;
+    if (rocksdb_ddse != nullptr || rocksdb_ddse->table_exists_in_engine) {
+      error = rocksdb_ddse->table_exists_in_engine(
+          rocksdb_ddse, thd, MYSQL_SCHEMA_NAME.str,
+          dd::tables::DD_properties::instance().name().c_str());
+    }
+    if (error == HA_ERR_TABLE_EXIST) {
+      // actual ddse maybe rocksdb
```
laurynas-biveinis: Regarding the atomicity of moving the table between engines, add `assert(!innodb_hton->table_exists_in_engine(the same table))`?

sql/dd/impl/bootstrap/bootstrapper.cc
```diff
@@ -1072,6 +1072,27 @@ bool create_dd_schema(THD *thd) {
...
+      bootstrap::DD_bootstrap_ctx::instance().set_actual_dd_engine(
+          enum_dd_default_engine::DEFAULT_DD_ROCKSDB);
+    } else {
+      // actual ddse maybe innodb
+      bootstrap::DD_bootstrap_ctx::instance().set_actual_dd_engine(
```
laurynas-biveinis: Likewise an opposite assert here

sql/dd/impl/bootstrap/bootstrap_ctx.h
```diff
@@ -123,10 +124,16 @@ class DD_bootstrap_ctx {
     m_actual_I_S_version = actual_I_S_version;
   }
 
+  void set_actual_dd_engine(uint actual_dd_engine) {
+    m_actual_dd_engine = actual_dd_engine;
```
laurynas-biveinis, line 135: This can be called at most once, consider adding `assert(m_actual_dd_engine == DB_TYPE_UNKNOWN);`

sql/dd/impl/bootstrap/bootstrap_ctx.h
```diff
@@ -123,10 +124,16 @@ class DD_bootstrap_ctx {
...
+  }
+
   uint get_actual_dd_version() const { return m_actual_dd_version; }
 
   uint get_actual_I_S_version() const { return m_actual_I_S_version; }
 
+  uint get_actual_dd_engine() const { return m_actual_dd_engine; }
```
laurynas-biveinis: This should never be called if `set_actual_dd_engine` was not called, consider adding `assert(m_actual_dd_engine != DB_TYPE_UNKNOWN);`

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -7168,6 +7168,28 @@ static bool rocksdb_get_table_statistics(
   return false;
 }
 
+/**
+  Check if a table exists in ROCKSDB during DD upgrade/restart
+  NOTE: Don't call any DD API, due to this function maybe called during
```
laurynas-biveinis: Clarify that this is server DD API, (not MyRocks DD API)

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -7168,6 +7168,28 @@ static bool rocksdb_get_table_statistics(
...
+  DD initializing.
+
+  @return HA_ERR_TABLE_EXIST - if table exist in ROCKSDB
```
```suggestion
  @return HA_ERR_TABLE_EXIST - if table exists in ROCKSDB
```

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -7168,6 +7168,28 @@ static bool rocksdb_get_table_statistics(
...
+  @return HA_ERR_NO_SUCH_TABLE - if table isn't exists in ROCKSDB.
```
```suggestion
  @return HA_ERR_NO_SUCH_TABLE - if table doesn't exist in ROCKSDB.
```

storage/rocksdb/ha_rocksdb.cc
luqun: add a helper function, but use reserve requires to use strlen to add these length together? maybe doesn't worth it..

sql/dd/impl/bootstrap/bootstrapper.cc
luqun: innodb doesn't have table_exists_in_engine(since it doesn't contains its own/private DD). maybe assert mysql.idb exists?

storage/rocksdb/ha_rocksdb.cc
laurynas-biveinis: Saving heap reallocations pays off more than `strlen`, but it's a minor comment, up to you to decide

sql/dd/impl/bootstrap/bootstrapper.cc
laurynas-biveinis: In this case the current code is fine

sql/dd/impl/bootstrap/bootstrapper.cc
laurynas-biveinis: The current code OK as is

storage/rocksdb/rdb_utils.h
```diff
@@ -473,4 +473,13 @@ class Ensure_initialized {
   return fn.compare(fn_len - ext_len, ext_len, extension) == 0;
 }
 
+// return <db_name>.<table_name>
+inline std::string get_full_tablename(const char *const db_name,
+                                      const char *const table_name) {
+  std::string fullname = db_name;
+  fullname.append(".");
```
laurynas-biveinis: Nit: since this is a single character, a single-character version should be more efficient:
```suggestion
  fullname += '.';
```

luqun: Good point.. just realized if it is value type, it will perform better

 https://github.com/llvm-mirror/libcxx/blob/master/include/string#L985

storage/rocksdb/ha_rocksdb.cc
luqun: updated with strlen.. For non-partition table, I can simple reserve 2 * NAME_CHAR_LEN + 1. For partition table, maybe reserve more.. anyway, use strlen instead.

sql/dd/impl/bootstrap/bootstrap_ctx.h
```diff
@@ -123,10 +130,20 @@ class DD_bootstrap_ctx {
...
+  void set_actual_dd_engine(enum legacy_db_type actual_dd_engine) {
+    assert(m_actual_dd_engine == DB_TYPE_UNKNOWN);
+    m_actual_dd_engine = actual_dd_engine;
+  }
+
   uint get_actual_dd_version() const { return m_actual_dd_version; }
 
   uint get_actual_I_S_version() const { return m_actual_I_S_version; }
 
+  uint get_actual_dd_engine() const {
```
laurynas-biveinis:
```suggestion
  legacy_db_type get_actual_dd_engine() const {
```

sql/dd/impl/bootstrap/bootstrapper.cc
```diff
@@ -1073,6 +1073,28 @@ bool create_dd_schema(THD *thd) {
...
+    // Try rocksdb first
+    handlerton *rocksdb_hton = ha_resolve_by_legacy_type(thd, DB_TYPE_ROCKSDB);
+    int error = HA_ERR_NO_SUCH_TABLE;
+    if (rocksdb_hton != nullptr &&
+        rocksdb_hton->table_exists_in_engine != nullptr) {
+      error = rocksdb_hton->table_exists_in_engine(
+          rocksdb_hton, thd, MYSQL_SCHEMA_NAME.str,
+          dd::tables::DD_properties::instance().name().c_str());
+    }
+    if (error == HA_ERR_TABLE_EXIST) {
```
laurynas-biveinis: Consider replacing with a ternary operator

storage/rocksdb/nosql_access.cc
```diff
@@ -14,6 +14,7 @@
    along with this program; if not, write to the Free Software
    Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA */
 
+#include "rdb_utils.h"
```
laurynas-biveinis: Move to the block starting at line 46/47

laurynas-biveinis (commented):
LGTM

PR #1322, opened 2023-06-15
Set rocksdb_large_prefix to ON and deprecate this system variable

laurynas-biveinis:
This variable was introduced to mirror innodb_large_prefix, which was deprecated in 5.7 and removed in 8.0, and no longer serves any real purpose, nor there is anything to gain with its 'OFF' setting, which is also incompatible with the data dictionary schema. Thus 1) change the default to 'ON'; 2) deprecate it.

Squash with 749f1dc5ef0327e55616441b399000510db9a54d

### DIFF
--- a/storage/rocksdb/ha_rocksdb.cc
+++ b/storage/rocksdb/ha_rocksdb.cc
@@ -62,6 +62,7 @@
 #include "sql/sql_partition.h"
 #include "sql/sql_table.h"
 #include "sql/sql_thd_internal_api.h"
+#include "sql/strfunc.h"
 
 /* RocksDB includes */
 #include "env/composite_env_wrapper.h"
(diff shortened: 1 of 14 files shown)


storage/rocksdb/ha_rocksdb.cc
```diff
@@ -1414,6 +1414,18 @@ static constexpr ulonglong RDB_DEFAULT_MAX_COMPACTION_HISTORY = 64;
...
+static constexpr char large_prefix_deprecated_msg[] =
+    "using rocksdb_large_prefix is deprecated and it will be removed in a "
+    "future release";
+
+static void rocksdb_large_prefix_update(THD *thd, SYS_VAR *, void *var_ptr,
+                                        const void *save) {
+  push_warning(thd, Sql_condition::SL_WARNING, HA_ERR_WRONG_COMMAND,
+               large_prefix_deprecated_msg);
+
+  *static_cast<bool*>(var_ptr) = *static_cast<const bool*>(save);
```
luqun: could we add a check that if DDSE is rocksdb, user can't turn off rocksdb_large_prefix

luqun:
could you also record main.mysqld--help-notwin with your change?

laurynas-biveinis:
Both comments addressed, ready for review again

PR #1323, opened 2023-06-15
Do not use THR_LOCK for MyRocks

laurynas-biveinis:
Since 5.7, all the THR_LOCK uses cases for InnoDB are fully covered by MDL locking and thus InnoDB removed THR_LOCK-based locking. The same MDL lock coverage applies to MyRocks as well, thus remove THR_LOCK-based locking there too.
- Remove fields Rdb_table_handler::m_thr_lock and ha_rocksdb::m_db_lock
- Add HA_NO_READ_LOCAL_LOCK to ha_rocksdb::table_flags to indicate that the SE does not have any specific support for LOCK TABLE READ LOCAL
- New overridden method ha_rocksdb::lock_count that always returns zero
- Remove all code in ha_rocksdb::store_lock that created THR_LOCK objects. Leave only the code that decides on the internal RocksDB locking.

References:
- https://dev.mysql.com/worklog/task/?id=6671
- https://dev.mysql.com/doc/relnotes/mysql/5.7/en/news-5-7-5.html
- 5c6171e2536b3274044354471e1ace9c878b8aa4

### DIFF
--- a/storage/rocksdb/ha_rocksdb.cc
+++ b/storage/rocksdb/ha_rocksdb.cc
@@ -8608,7 +8608,6 @@ Rdb_table_handler *Rdb_open_tables_map::get_table_handler(
 
     m_table_map.emplace(table_name_str, table_handler);
 
-    thr_lock_init(&table_handler->m_thr_lock);
     table_handler->m_io_perf_read.init();
     table_handler->m_io_perf_write.init();
   }
@@ -8867,7 +8866,6 @@ void Rdb_open_tables_map::release_table_handler(
     const auto ret MY_ATTRIBUTE((__unused__)) =
         m_table_map.erase(std::string(table_handler->m_table_name));
     assert(ret == 1);  // the hash entry must actually be found and deleted
-    my_core::thr_lock_delete(&table_handler->m_thr_lock);
     my_free(table_handler);
   }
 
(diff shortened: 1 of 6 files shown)


storage/rocksdb/ha_rocksdb.cc
```diff
@@ -8364,8 +8364,6 @@ Rdb_table_handler *Rdb_open_tables_map::get_table_handler(
 
     m_table_map.emplace(table_name_str, table_handler);
 
-    thr_lock_init(&table_handler->m_thr_lock);
-    table_handler->m_io_perf_read.init();
```
luqun, line 8612: accidentally remove "table_handler->m_io_perf_read.init();"?

laurynas-biveinis:
Rebased and retested with no changes

luqun:
Please rebase this PR.

laurynas-biveinis:
@luqun , rebased

laurynas-biveinis:
This has a minor effect for DD: by making MyRocks behave like InnoDB WRT locking, at least one more existing testcase can be run as-is under both DDSEs (main.status)

luqun:
> This has a minor effect for DD: by making MyRocks behave like InnoDB WRT locking, at least one more existing testcase can be run as-is under both DDSEs (main.status)


Currently, do you try to run main test suite under myrocks dd and find out main.status need this change?

luqun:
T

laurynas-biveinis:
> > This has a minor effect for DD: by making MyRocks behave like InnoDB WRT locking, at least one more existing testcase can be run as-is under both DDSEs (main.status)
> 
> Currently, do you try to run main test suite under myrocks dd and find out main.status need this change?

Yes, but that was not the motivation to do it. If this PR did not exist, I'd just skip the testcase

PR #1326, opened 2023-06-20
Fix a too-strict assertion for --help if default_dd_storage_engine=Ro…

laurynas-biveinis:
…cksDB


luqun:
LGTM, thanks for fixing.

PR #1328, opened 2023-06-22
Handle log archiver archiving past the clone SE sync point

laurynas-biveinis:
Under write workload, it is possible for the InnoDB log archiver to archive log data past the clone SE synchronization point LSN, with the following thread interleaving:

1) Clone SE synchronization executes the performance_schema.LOG_STATUS query 
2) User query performs a write to InnoDB, advancing the LSN 
3) InnoDB log archiver archives the log written in the previous step
4) Clone SE synchronization asks the InnoDB log archiver to stop archiving at the clone LSN, which is less than the already-archived LSN.

It was assumed that this scenario cannot happen, thus code hits various asserts, and on the release build may continue archiving indefinitely.

Since the log archiver already has support for providing less log data than archived, fix by setting the log archiver stop LSN correctly to smaller of SE synchronization and already-archived LSN.

Add a DEBUG_SYNC-using testcase.

### DIFF
--- a/storage/innobase/arch/arch0log.cc
+++ b/storage/innobase/arch/arch0log.cc
@@ -439,9 +439,11 @@ int Arch_Log_Sys::stop(Arch_Group *group, lsn_t &stop_lsn, byte *log_blk,
                        uint32_t &blk_len) {
   int err = 0;
   blk_len = 0;
-  stop_lsn = m_archived_lsn.load();
+
   auto se_sync_stop_lsn = get_stop_lsn_master_thread();
   ut_ad(se_sync_stop_lsn != LSN_MAX || log_blk == nullptr);
+
+  stop_lsn = std::min(m_archived_lsn.load(), se_sync_stop_lsn);
   ut_ad(stop_lsn <= se_sync_stop_lsn);
 
   if (log_blk != nullptr) {
@@ -595,7 +597,6 @@ Arch_State Arch_Log_Sys::check_set_state(bool is_abort, lsn_t *archived_lsn,
       if (*archived_lsn != LSN_MAX) {
         /* Update system archived LSN from input */
         ut_ad(*archived_lsn >= m_archived_lsn.load());
-        ut_ad(*archived_lsn <= get_stop_lsn_any_thread());
         m_archived_lsn.store(*archived_lsn);
       } else {
         /* If input is not initialized,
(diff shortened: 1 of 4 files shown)


storage/innobase/arch/arch0log.cc
```diff
@@ -439,9 +439,11 @@ int Arch_Log_Sys::stop(Arch_Group *group, lsn_t &stop_lsn, byte *log_blk,
                        uint32_t &blk_len) {
   int err = 0;
   blk_len = 0;
-  stop_lsn = m_archived_lsn.load();
+
   auto se_sync_stop_lsn = get_stop_lsn_master_thread();
   ut_ad(se_sync_stop_lsn != LSN_MAX || log_blk == nullptr);
+
+  stop_lsn = std::min(m_archived_lsn.load(), se_sync_stop_lsn);
```
sunshine-Chun, line 446: I wonder why change the stop_lsn value will fix the issue. The stop_lsn will be changed to se_sync_stop_lsn anyway after running wait_archive_complete() function. And it is not used in wait_archived_complete() function. Looks like the log archiver is still running in the same way and the m_archiver_lsn will be the same as what it is before the change.

I tried to revert the stop_lsn change and comment out the debug assert `ut_ad(stop_lsn <= se_sync_stop_lsn);`. Looks like the above test can still pass.

laurynas-biveinis: The important part is that this too-large `stop_lsn` value can be observed globally, in this case by log archiver threads, and do the wrapping-around calculation for the bug before it is fixed up.

To confirm, you can replace both the fixing `std::min` and the debug assert with
```
  stop_lsn = m_archived_lsn.load();
  if (stop_lsn > se_sync_stop_lsn) {
    sleep(2);
  }
```

Then (non-deterministically; might take a few runs) you will get
```
Assertion failure: arch0log.cc:604:*archived_lsn <= get_stop_lsn_any_thread() thread 0x17101b000
```
in `Arch_log_sys::check_set_state`. And if you comment that one out, then
```
Assertion failure: arch0log.cc:623:stop_lsn >= local_archived_lsn thread 0x16ce73000
```
right before the calculation which subtracts the two and is the smoking gun for the bug.

BTW the testcase for the bug has a minor issue that it only synchronizes the user threads, but not the log archiver threads. I am not sure if it's worth fixing though.

sunshine-Chun: Would you mind explaining my following confusion? 
The stop_lsn in the diff is `lsn_t m_end_lsn`. I didn't see the log archiver uses the `m_end_lsn` to do calculation. Instead, the stop_lsn in log archiver is got from `atomic_lsn_t m_stop_lsn`. I wonder how `m_end_lsn`  affects log archiver?

laurynas-biveinis: You are right, I was able to break my "fix" with:
```diff
--- a/storage/innobase/arch/arch0log.cc
+++ b/storage/innobase/arch/arch0log.cc
@@ -443,7 +443,10 @@ int Arch_Log_Sys::stop(Arch_Group *group, lsn_t &stop_lsn, byte *log_blk,
   auto se_sync_stop_lsn = get_stop_lsn_master_thread();
   ut_ad(se_sync_stop_lsn != LSN_MAX || log_blk == nullptr);
 
-  stop_lsn = std::min(m_archived_lsn.load(), se_sync_stop_lsn);
+  const auto archived_lsn = m_archived_lsn.load();
+  stop_lsn = std::min(archived_lsn, se_sync_stop_lsn);
+  if (archived_lsn > se_sync_stop_lsn)
+    sleep(2);
   ut_ad(stop_lsn <= se_sync_stop_lsn);
 
   if (log_blk != nullptr) {
```

which is functionally the same as the fix (writes to `stop_lsn` (aka `m_end_lsn`) ASAP), but sleeps if already archived too much. Continuing debugging.

laurynas-biveinis: Updated the fix, rebased & pushed. The fix has been updated to handle the archived LSN > stop LSN in `Arch_log_sys::check_set_state` by archiving 0 instead of u64 maxint LSN delta

PR #1334, opened 2023-06-28
Enable engine specific DD tables

luqun:
During ddse upgrade/downgrade, these DDSE_PRIVATE and  DDSE_PROTECTED dd tables(such as innodb_table_stats , innodb_index_stats and  innodb_ddl_log) should stay at its original SE. 

Changes:
- add an overload constructor Object_table_impl() to allow pass engine specific tables
- For innodb DD tables, always add tablespaces (not matter whether default_dd_storage_engine)

The change shouldn't break anything if current DDSE is innodb.

### DIFF
--- a/sql/dd/impl/types/object_table_impl.h
+++ b/sql/dd/impl/types/object_table_impl.h
@@ -85,6 +85,7 @@ class Object_table_impl : virtual public Object_table {
     common options.
   */
   Object_table_impl();
+  Object_table_impl(const String_type &engine);
 
   const String_type &name() const override {
     return m_target_def.get_table_name();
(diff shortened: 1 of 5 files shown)


luqun:
CC: @laurynas-biveinis

sql/dd/impl/types/object_table_impl.h
```diff
@@ -85,6 +85,7 @@ class Object_table_impl : virtual public Object_table {
     common options.
   */
   Object_table_impl();
+  Object_table_impl(const String_type &engine);
```
laurynas-biveinis:
```suggestion
  explicit Object_table_impl(const String_type &engine);
```

laurynas-biveinis (commented):
LGTM with a minor nit

PR #1337, opened 2023-07-14
Remove extra assert in Rdb_writebatch_impl class

luqun:
Rdb_writebatch_impl is used in replication if rocksdb_skip_trx_api = 1. thus it should have same behavior as Rdb_transaction_impl. 

After MyRocks DD, these Rdb_writebatch_impl member function will be used to access DD table through attachable transaction which break these asserts.

Remove these asserts and keep same assert as Rdb_transaction_impl member functions


laurynas-biveinis (commented):
LGTM

luqun:
committed

PR #1329, opened 2023-06-22
Always assume existence of persisted autoincrement values for MyRocks…

laurynas-biveinis:
… table

The non-persistent autoincrement values upgrade to persistent ones was implemented in 2017, in 5.6. Thus no 8.0 instance should encounter non-persistent values, ever. If this assumption is incorrect, the upgrade is still possible by going through the latest 5.6 release first.

Thus remove myrocks_autoinc_upgrade debug code injection point, simplify the surrounding code, remove the tests that use it.

PR #1338, opened 2023-07-18
update rocksdb suite MTR for rocksdb DD

luqun:
Try to update all MTR in rocksdb suites to works on rocksdb DD.

Details:

Code changes:
- update most of these debug hook to be more specific, such as check table name, check thread, column family to execute these hook in a correct time when running under rocksdb DD; without these check, most likely these hook will be executed for DD tables.

MTR changes:
- rocksdb large prefix: add a new large_prefix.test which is excluded from rocksdb DD, due to when rocksdb DD is enabled, rocksdb_large_prefix must be on.
-  rocksdb stats variables value change a lot when running under rocksdb DD, exclude rocksdb.bloomfilter, rocksdb.bloomfilter2., rocksdb.bloomfilter3, rocksdb.bloomfilter_bulk_load, rocksdb.read_only_tx from rocksdb DD combination
- rocksdb doesn't support SERIALIZABLE ISOLATION, exclude rocksdb.issue314 from rocksdb DD combination

- When restart mysqld with options, also pass default DDSE value to avoid accidentally change DD table from rocksdb->innodb or innodb->rocksdb


mysql-test/suite/rocksdb/t/unsupported_tx_isolations.test
```diff
@@ -22,4 +22,6 @@ insert into t1 values (2,1);
 select * from t1 where id=1;
 rollback;
 
+# rocksdb dd doesn't support serializable
+SET session transaction isolation level read committed;
```
laurynas-biveinis: Inconsistent case (it is inconsistent in the whole file, but good to keep it consistent at least at line level)

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -10260,8 +10260,9 @@ int ha_rocksdb::create_table(const std::string &table_name,
   DBUG_ENTER_FUNC();
 
   int err;
-  bool is_dd_tbl = dd::get_dictionary()->is_dd_table_name(
-      table_arg.s->db.str, table_arg.s->table_name.str);
+  bool is_dd_tbl = false;
+
+  if (ha_thd()->is_dd_system_thread()) is_dd_tbl = true;
```
laurynas-biveinis: Consider using the ternary operator. Could you maybe also tell what made you do this change?

mysql-test/suite/rocksdb/t/validate_datadic.test
```diff
@@ -1,5 +1,12 @@
 --source include/have_rocksdb.inc
 
+# skip validate_datadic test for rocksdb DD
+# With rocksdb DD, table metadata lives in rocksdb
+# update mysql.ibd file won't get these error message
+if (`select count(*) = 1 from performance_schema.global_variables where variable_name = 'default_dd_storage_engine' and variable_value = 'RocksDB';`) {
```
laurynas-biveinis: Extract an include file i.e. `include/have_innodb_dd.inc` or similar, use here and in `large_prefix.test`

laurynas-biveinis (commented):
How are you testing these changes?

luqun:
> How are you testing these changes?

in mysql-test/suite/rocksdb/combinations, add following lines..so these testcase will run after upgrade DD from innodb to rocksdb

  [rocksdb_dd]
  default_dd_storage_engine = 1

storage/rocksdb/ha_rocksdb.cc
luqun: I will exclude these line from this PR. Anyway, let me explain a little bit first, When DD upgrade occurs(either innodb->rocksdb or different version), we will 
1. create a temporarily db, 
2. create all of these DD table inside the db, 
3. "rename" temporarily db to mysql db after rename existing mysql db to another name.

These line change is "required" for 2nd step, otherwise, these DD table will created in default column family instead of __system__ column family. INNODB has similar issue when DD upgrade since it want to create these DD table in temporarily db under mysql tablespace..

luqun:
I should mention that the main purpose of this PR is to avoid duplicate effort.. @laurynas-biveinis and I both need to update some MTR for rocksdb DD..

laurynas-biveinis:
> I should mention that the main purpose of this PR is to avoid duplicate effort.. @laurynas-biveinis and I both need to update some MTR for rocksdb DD..

My DD changes focus on collecting new tests, as the ideal of always running all the tests with both DDSEs settings is unattainable. For example, https://github.com/laurynas-biveinis/mysql-5.6/commit/9f081c91e302ff0418e23d43b42eb3e8154d0c35: rocksdb.ddse, rocksdb_rpl.ddse_rpl

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -3499,11 +3499,13 @@ class Rdb_transaction {
     @param wb
    */
   rocksdb::Status merge_auto_incr_map(rocksdb::WriteBatchBase *const wb) {
-    DBUG_EXECUTE_IF("myrocks_autoinc_upgrade", return rocksdb::Status::OK(););
 
     // Iterate through the merge map merging all keys into data dictionary.
     rocksdb::Status s;
     for (auto &it : m_auto_incr_map) {
+      DBUG_EXECUTE_IF(
```
laurynas-biveinis: This and similar changes below should be covered by https://github.com/facebook/mysql-5.6/pull/1329

mysql-test/suite/rocksdb/combinations
```diff
@@ -2,3 +2,6 @@
 
 [drop_table_with_delete_range]
 rocksdb_enable_delete_range_for_drop_index = ON
+
+[rocksdb_dd]
+default_dd_storage_engine = 1
```
laurynas-biveinis: If this bit will be merged eventually, use the string name instead of 1

mysql-test/suite/rocksdb/t/allow_to_start_after_corruption.test
```diff
@@ -60,7 +62,7 @@ select * from t1;
 # remove flag to ignore corruption
 --let $_mysqld_option=--log-error=$LOG
 --error 0
---exec $MYSQLD_CMD $_mysqld_option
+--exec $MYSQLD_CMD $_mysqld_option --default_dd_storage_engine=$ddse
```
laurynas-biveinis, line 65: Here and elsewhere, why it's needed? Does it use the other DDSE otherwise?

mysql-test/include/not_rocksdb_dd.inc
```diff
@@ -0,0 +1,3 @@
+if (`select count(*) = 1 from performance_schema.global_variables where variable_name = 'default_dd_storage_engine' and variable_value = 'RocksDB';`) {
```
laurynas-biveinis: @luqun, I think it makes sense to split this one file to its own separate PR and merge ASAP. We can already start annotating tests with it.

(I'd also consider replacing negative with positive i.e. `have_innodb_ddse` in the filename and same in the check itself)

laurynas-biveinis: It looks like it also makes sense to have `have_innodb_ddse_unless_row_binlog.inc`, to address testcases that try to replicate administrative statements in row format, thus require MyRocks to replicate in row. I'll work on this some more.

luqun: > @luqun, I think it makes sense to split this one file to its own separate PR and merge ASAP. We can already start annotating tests with it.
> 
> (I'd also consider replacing negative with positive i.e. `have_innodb_ddse` in the filename and same in the check itself)

how about https://github.com/facebook/mysql-5.6/pull/1344? also Let me update this PR

mysql-test/suite/rocksdb/t/allow_to_start_after_corruption.test
luqun: yes. if it isn't specified, it will use INNODB(since this is default DDSE). During restart, it found out the actual ddse is rocksdb, then try to convert from ROCKSDB DDSE to INNODB DDSE, which isn't supported yet.

mysql-test/suite/rocksdb/t/bloomfilter.inc
```diff
@@ -1,4 +1,7 @@
 --source include/have_rocksdb.inc
+# with rocksdb DD, these status variable values also inclue DD tables. etc
```
laurynas-biveinis, line 2: Here and below: marking a MyRocks test with a skip-if-MyRocks-DDSE is a bit of a red flag: it means the test will be effectively removed once the DDSE is switched over. If we are not OK with dropping this test coverage in general, we need to find a way to maintain it.

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -4764,7 +4764,9 @@ class Rdb_transaction {
 
   void remove_from_global_trx_list(void) {
     DBUG_EXECUTE_IF("rocksdb_trx_list_crash", {
-      THD *thd = new THD();
+      if (!m_thd->is_system_thread() &&
+          !m_thd->is_attachable_transaction_active()) {
+        THD *thd = new THD();
```
laurynas-biveinis, line 4769: Indentation

laurynas-biveinis (commented):
@luqun I am sorry, I didn't realize there will not be another PR with the pushed bits, thus I have some late review comments

Issue #1336, opened 2023-07-10
Allow users to manually set the number of block cache shards


mdcallag:
While RocksDB allows for num_shard_bits to be set for the block cache, MyRocks does not expose this.
This is a feature request for MyRocks to expose an option that allows me to set it.

There can be serious perf problems when the block cache is too small. With intra-L0 compaction there can be large SSTs in the L0, large SSTs imply large filter blocks and reading a large object into a small cache shard can evict everything from that shard.

A blog post on this [is here](http://smalldatum.blogspot.com/2023/07/one-source-write-stalls-with-rocksdb.html).

mdcallag:
Fixed

PR #1339, opened 2023-07-19
Add rocksdb_block_cache_numshardbits for issue 1336

mdcallag:
This fixes https://github.com/facebook/mysql-5.6/issues/1336

This adds the my.cnf options: rocksdb_block_cache_numshardbits

This option can be set so that RocksDB to fix the number of block cache shards.

The default value is -1 to match existing behavior. When -1 RocksDB code will determine the number of block cache shards as min(6, rocksdb_block_cache_size / min_shard_size) and today min_shard_size is 512K for LRU and 32M for Hyper.

The math above frequently results in a block cache with too many small shards when rocksdb_block_cache_size is not too big (a few GB is not too big) and there will be perf problems that are hard to debug in such a case.

### DIFF
--- a/storage/rocksdb/ha_rocksdb.cc
+++ b/storage/rocksdb/ha_rocksdb.cc
@@ -920,7 +920,7 @@ static long long rocksdb_compaction_sequential_deletes_file_size = 0l;
 static uint32_t rocksdb_validate_tables = 1;
 char *rocksdb_datadir;
 static uint32_t rocksdb_max_bottom_pri_background_compactions = 0;
-static int32_t rocksdb_block_cache_numshardbits = -1;
+static int rocksdb_block_cache_numshardbits = -1;
 static uint32_t rocksdb_table_stats_sampling_pct;
 static uint32_t rocksdb_table_stats_recalc_threshold_pct = 10;
 static unsigned long long rocksdb_table_stats_recalc_threshold_count = 100ul;
@@ -2056,9 +2056,9 @@ static MYSQL_SYSVAR_INT(table_cache_numshardbits,
 static MYSQL_SYSVAR_INT(block_cache_numshardbits,
                         rocksdb_block_cache_numshardbits,
                         PLUGIN_VAR_RQCMDARG | PLUGIN_VAR_READONLY,
-                        "Block cache numshardbits for RocksDB",
-                        nullptr, nullptr,
-                       /* default */ -1, /* min */ -1, /* max */ 8, 0);
+                        "Block cache numshardbits for RocksDB", nullptr,
+                        nullptr,
+                        /* default */ -1, /* min */ -1, /* max */ 8, 0);
 
 static MYSQL_SYSVAR_UINT64_T(wal_ttl_seconds,
                              rocksdb_db_options->WAL_ttl_seconds,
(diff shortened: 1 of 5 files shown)


storage/rocksdb/ha_rocksdb.cc
```diff
@@ -920,6 +920,7 @@ static long long rocksdb_compaction_sequential_deletes_file_size = 0l;
 static uint32_t rocksdb_validate_tables = 1;
 char *rocksdb_datadir;
 static uint32_t rocksdb_max_bottom_pri_background_compactions = 0;
+static int32_t rocksdb_block_cache_numshardbits = -1;
```
```suggestion
static int rocksdb_block_cache_numshardbits = -1;
```
`MYSQL_SYSVAR_INT`, `HyperClockCacheOptions`, `NewLRUCache` all expect an `int` here

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -2052,6 +2053,13 @@ static MYSQL_SYSVAR_INT(table_cache_numshardbits,
                         // fails to create a cache and returns a nullptr
                         /* min */ 0, /* max */ 19, 0);
 
+static MYSQL_SYSVAR_INT(block_cache_numshardbits,
+                        rocksdb_block_cache_numshardbits,
+                        PLUGIN_VAR_RQCMDARG | PLUGIN_VAR_READONLY,
+                        "Block cache numshardbits for RocksDB",
+                        nullptr, nullptr,
+                       /* default */ -1, /* min */ -1, /* max */ 8, 0);
```
```suggestion
                        /* default */ -1, /* min */ -1, /* max */ 8, 0);
```

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -7952,13 +7961,13 @@ static int rocksdb_init_internal(void *const p) {
         rocksdb_use_hyper_clock_cache
             ? rocksdb::HyperClockCacheOptions(
                   rocksdb_block_cache_size, rocksdb_tbl_options->block_size,
-                  -1
-                  /* num_shard_bits */,
+                  rocksdb_block_cache_numshardbits /* num_shard_bits */,
```
laurynas-biveinis: The variable name already tells it's numshardbits, making the comment redundant
```suggestion
                  rocksdb_block_cache_numshardbits,
```
This might also need a clang-format run to reflow (which may touch the rest of the file, so might be a good idea to limit it to this region)

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -7952,13 +7961,13 @@ static int rocksdb_init_internal(void *const p) {
...
                   false /* strict_capacity_limit */, memory_allocator)
                   .MakeSharedCache()
 
             : rocksdb::NewLRUCache(
-                  rocksdb_block_cache_size, -1 /*num_shard_bits*/,
+                  rocksdb_block_cache_size,
+                  rocksdb_block_cache_numshardbits /*num_shard_bits*/,
```
```suggestion
                  rocksdb_block_cache_numshardbits,
```

laurynas-biveinis (commented):
It is likely `main.mysqld--help-notwin` test needs a re-record

storage/rocksdb/ha_rocksdb.cc
mdcallag: Updated the PR to fix this

storage/rocksdb/ha_rocksdb.cc
mdcallag: Updated the PR to fix this

storage/rocksdb/ha_rocksdb.cc
mdcallag: Updated the PR to fix this including running clang-format (phabricator also warned about that)

storage/rocksdb/ha_rocksdb.cc
mdcallag: updated the PR to fix this

mdcallag:
> It is likely `main.mysqld--help-notwin` test needs a re-record

Done

PR #1341, opened 2023-08-09
dont bulk load for table metadata change

luqun:
Summary:

When customer issue a DDL statement, it will change a table's metadata(whose data is stored in DD) and a tables' raw data.

- during table metadata change, DD will enable OPTION_DD_UPDATE_CONTEXT flag(https://github.com/facebook/mysql-5.6/blob/fb-mysql-8.0.28/sql/dd/impl/transaction_impl.cc#L169).

- during table raw data change, OPTION_DD_UPDATE_CONTEXT flag isn't enabled.

if current DD is myrocks and customer try to run a DDL with bulk loading enabled,

- before this change, myrocks write table metadata using bulk loading when DD save new schema(such as dd::Table_impl::store_children). Later, DD will try to read new schema(such as  dd::Table_impl::restore_children), then it fail to find new schema due myrocks bulk load doesn't support read your own data.

- after this change, table metadata won't use bulk loading(its change data is small) and current trx can read its own data.

Test Plan:

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/storage/rocksdb/ha_rocksdb.cc
+++ b/storage/rocksdb/ha_rocksdb.cc
@@ -12548,17 +12548,8 @@ int ha_rocksdb::update_write_pk(const Rdb_key_def &kd,
     get_ha_data(ha_thd())->total_tmp_table_size += row_size;
   }
 
-  /*
-    DDL trx update table metadata(DD table) and raw data,
-    For table metadata(thd_is_dd_update_stmt()==True), always skip bulk loading
-    due to DD will read/restore latest metadata after updating metadata
-    For raw data(thd_is_dd_update_stmt()==False), use bulk-load if requested
-  */
-  bool is_dd_operation = default_dd_storage_engine == DEFAULT_DD_ROCKSDB &&
-                         thd_is_dd_update_stmt(ha_thd());
-
   if (rocksdb_enable_bulk_load_api && THDVAR(table->in_use, bulk_load) &&
-      !hidden_pk && !is_dd_operation) {
+      !hidden_pk && !is_dd_operation()) {
     /*
       Write the primary key directly to an SST file using an SstFileWriter
      */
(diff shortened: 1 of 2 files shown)


storage/rocksdb/ha_rocksdb.cc
```diff
@@ -12565,8 +12565,17 @@ int ha_rocksdb::update_write_pk(const Rdb_key_def &kd,
...
 
+  /*
+    DDL trx update table metadata(DD table) and raw data,
+    For table metadata(thd_is_dd_update_stmt()==True), always skip bulk loading
+    due to DD will read/restore latest metadata after updating metadata
+    For raw data(thd_is_dd_update_stmt()==False), use bulk-load if requested
+  */
+  bool is_dd_operation = default_dd_storage_engine == DEFAULT_DD_ROCKSDB &&
+                         thd_is_dd_update_stmt(ha_thd());
+
```
laurynas-biveinis: Can you make this a helper function so that the comment is present only once?

storage/rocksdb/ha_rocksdb.h
```diff
@@ -1000,6 +1000,8 @@ class ha_rocksdb : public my_core::handler, public blob_buffer {
   void build_decoder();
   void check_build_decoder();
 
+  inline bool is_dd_operation();
```
laurynas-biveinis: Add `const`, suggesting to use "update" instead of "operation" to indicate that it involves a write: 
```suggestion
  inline bool is_dd_update() const;
```

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -16413,6 +16413,19 @@ inline bool ha_rocksdb::is_instant(const Alter_inplace_info *ha_alter_info) {
...
 
+/*
+  DDL trx update both table metadata(DD table) and raw data,
+  - For table metadata(thd_is_dd_update_stmt()==True), always skip bulk loading
+  due to DD will read/restore latest metadata after updating metadata and bulk
+  loading doesn't suppose read before commit
+  - For raw data(thd_is_dd_update_stmt()==False), use bulk-load if requested
+  @return True if DDSE is rocksdb and it is updating table metadata
+*/
+inline bool ha_rocksdb::is_dd_update() {
```
```suggestion
inline bool ha_rocksdb::is_dd_update() const {
```
But, `inline` in source file will not be inlined unless LTO is used. Please either remove `inline`, or move to the header file

luqun: remove inline..

laurynas-biveinis: Please add `const` too :)

luqun:
@laurynas-biveinis , I will commit this PR today if it looks okay to you.

laurynas-biveinis:
> @laurynas-biveinis , I will commit this PR today if it looks okay to you.

Please add `const` to the method definition (now only declaration has it), and then go ahead 👍

PR #1342, opened 2023-08-09
Reimplement InnoDB data dictionary INFORMATION_SCHEMA views through s…

laurynas-biveinis:
…erver layer

Currently, the DD-related INFORMATION_SCHEMA views in InnoDB use the low level InnoDB APIs directly to access the underlying DD tables. That stops working when those tables are moved to a different storage engine. To keep these views working, re-implement them using the server layer engine-agnostic APIs. Leave the old implementation enabled when InnoDB is the data dictionary storage engine.

To test the implementation without a viable alternate DDSE yet, the new implementation was enabled unconditionally and all the test regressions fixed and collected to a new testsuite rocksdb_dd_innodb. Going forward it will collect other InnoDB-specific tests when it is not the DDSE.


mysql-test/suite/rocksdb_dd_innodb/t/i_s_innodb_tables.test
```diff
@@ -0,0 +1 @@
+--source ../../innodb/t/i_s_innodb_tables.test
```
luqun, line 1: maybe I miss something... do you also need to config default_dd_storage_engine to rocksdb for this test suite?

storage/innobase/include/dict0dd.h
```diff
@@ -1352,5 +1353,15 @@ the dictionary.
...
 bool dd_is_table_in_encrypted_tablespace(const dict_table_t *table);
 
+/** Check if InnoDB must access MySQL data dictionary through the server layer
+APIs. If InnoDB is the data dictionary storage engine, it may access the data
+dictionary through lower level direct InnoDB interface. But if it is not the
+DDSE, the tables will be in another storage engine, thus the server layer
+interface must be used.
+@return whether server layer interface must be used to access the DD */
+[[nodiscard]] inline bool dd_access_through_server() noexcept {
+  return (default_dd_storage_engine != DEFAULT_DD_INNODB);
```
luqun, line 1363: how about add a new variable? with that variable enabled(it will always access through server), maybe we can reuse more innodb MTRs?

storage/innobase/handler/i_s.cc
```diff
@@ -6873,6 +7106,96 @@ static int i_s_innodb_tablespaces_fill_table(THD *thd, TABLE_LIST *tables,
...
+                 tablespace.name().length() + 1);
+
+          const auto &properties = tablespace.options();
+
+          uint32_t flags;
+          if (se_private_data.get(dd_space_key_strings[DD_SPACE_FLAGS],
+                                  &flags)) {
+            assert(false);
+            mem_heap_free(heap);
+            return 1;
```
luqun, line 7136: return 0 to minic old innodb behavior? https://github.com/facebook/mysql-5.6/blob/fb-mysql-8.0.28/storage/innobase/dict/dict0dd.cc#L5453

mysql-test/suite/rocksdb_dd_innodb/t/i_s_innodb_tables.test
laurynas-biveinis: Yes, it will be set once it is viable (the rest of the feature gets merged). Now these tests were developed with the `dd_access_through_server()` hardcoded to return `true`

storage/innobase/include/dict0dd.h
laurynas-biveinis: Possible, but I view adding new system variables as last resort, because MySQL already has the problem of too many sysvars

storage/innobase/handler/i_s.cc
laurynas-biveinis: The old InnoDB behavior is different in different callstack levels: the lower `dd_process_dd_tablespaces_rec` function that you linked to indeed returns false on error, true on success, but here we are in `i_s_innodb_tablespaces_fill_table`, which returns 0 on success, 1 on error.

Actually I see that to fully mimic the old behavior the error at this point has to be silently eaten

storage/innobase/include/dict0dd.h
laurynas-biveinis: I have also considered making the new method mandatory for InnoDB DDSE too, but that might be quite risky

laurynas-biveinis: As discussed elsewhere, let's keep as is.

PR #1343, opened 2023-08-22
Fix MYSQL_BIN_LOG::update_prev_gtid_and_opid declaration

laurynas-biveinis:
It was declared with long and defined with int64_t, causing

fb-mysql/sql/binlog.cc:6305:21: error: out-of-line definition of 'update_prev_gtid_and_opid' does not match any declaration in 'MYSQL_BIN_LOG' void MYSQL_BIN_LOG::update_prev_gtid_and_opid(Gtid_set *prev_gtid,

on macOS, where long != int64_t.

Squash with 10f0fe819b417767857146c0a6c5fa152f8e0238

### DIFF
--- a/sql/binlog.h
+++ b/sql/binlog.h
@@ -586,8 +586,8 @@ class MYSQL_BIN_LOG : public TC_LOG {
 
   inline uint get_sync_period() { return *sync_period_ptr; }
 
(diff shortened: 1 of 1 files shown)


laurynas-biveinis:
ping @li-chi

li-chi (approved):

li-chi:
Nice catch

PR #1344, opened 2023-08-31
add have_innodb_ddse.inc file

luqun:
This include file is used to exclude some testcase which requires innodb ddse


laurynas-biveinis:
LGTM

PR #1345, opened 2023-09-01
[DD] add DD_ENGINE property into dd_properties table

luqun:
Summary:
With multiple DDSE support, DD tables can be stored in different DDSE. 

During upgrade, it is possible that some of DD table is upgrade to target engine but others DD tables isn't. To skip check all of these tables SE, store DD_ENGINE value into dd_properties and the value is stored/updated at the end of upgrade.
-  if DD_ENGINE value is same as default ddse, upgrade has been finished. 
-  if DD_ENGINE value isn't same as default ddse, continue upgrade from start.

Changes:
- Add more diagnostic info, such as DDSE changes into mysqld log
- is_dd_upgrade() will return true if either version upgrade or ddse upgrade(change)
- DD_ENGINE value is stored at the end of upgrade, similar to mysqld_version

### DIFF
--- a/share/messages_to_error_log.txt
+++ b/share/messages_to_error_log.txt
@@ -11596,6 +11596,12 @@ ER_IB_SLOWRM_CREATE_DIR
 ER_IB_SLOWRM_MOVE_FILE
   eng "%s"
 
+ER_DDSE_CHANGE
+  eng "MySQL DDSE is changing from '%s' to '%s'."
+
+ER_DD_ENGINE_NOT_FOUND
+  eng "The DD_ENGINE setting for the data dictionary was not found. Starting the server using DD_ENGINE= '%u'."
+
 # DO NOT add server-to-client messages here;
 # they go in messages_to_clients.txt
 # in the same directory as this file.
--- a/share/messages_to_error_log.txt
+++ b/share/messages_to_error_log.txt
@@ -11602,6 +11602,9 @@ ER_DDSE_CHANGE
 ER_DD_ENGINE_NOT_FOUND
   eng "The DD_ENGINE setting for the data dictionary was not found. Starting the server using DD_ENGINE= '%u'."
 
+ER_UNSUPPORTED_DD_ENGINE
+  eng "The DD_ENGINE value '%u' found in dd_properties isn't supported DDSE."
+
 # DO NOT add server-to-client messages here;
 # they go in messages_to_clients.txt
 # in the same directory as this file.
--- a/share/messages_to_error_log.txt
+++ b/share/messages_to_error_log.txt
@@ -11600,10 +11600,10 @@ ER_DDSE_CHANGE
   eng "MySQL DDSE is changing from '%s' to '%s'."
 
 ER_DD_ENGINE_NOT_FOUND
-  eng "The DD_ENGINE setting for the data dictionary was not found. Starting the server using DD_ENGINE= '%u'."
+  eng "The DD_ENGINE property for the data dictionary was not found. Starting the server using DD_ENGINE='%s'."
 
 ER_UNSUPPORTED_DD_ENGINE
-  eng "The DD_ENGINE value '%u' found in dd_properties isn't supported DDSE."
+  eng "The DD_ENGINE value '%u' found in dd_properties isn't supported DDSE. Currently supported DDSE are '%s'"
 
 # DO NOT add server-to-client messages here;
 # they go in messages_to_clients.txt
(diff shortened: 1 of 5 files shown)


sql/dd/impl/bootstrap/bootstrapper.cc
```diff
@@ -1211,6 +1213,17 @@ bool initialize_dd_properties(THD *thd) {
       LogErr(ERROR_LEVEL, ER_LCTN_CHANGED, lower_case_table_names, actual_lctn);
       return true;
     }
+
+    // read actual dd engine from dd_properties
+    uint actual_dd_engine = DB_TYPE_INNODB;
```
laurynas-biveinis: Maybe initialize to something != INNODB, != ROCKSDB to catch corrupted reads?

sql/dd/impl/bootstrap/bootstrapper.cc
```diff
@@ -1211,6 +1213,17 @@ bool initialize_dd_properties(THD *thd) {
...
+    exists = false;
+    if (dd::tables::DD_properties::instance().get(thd, "DD_ENGINE",
+                                                  &actual_dd_engine, &exists) ||
+        !exists) {
+      LogErr(WARNING_LEVEL, ER_DD_ENGINE_NOT_FOUND, actual_dd_engine);
+    }
+    bootstrap::DD_bootstrap_ctx::instance().set_actual_dd_engine(
+        (legacy_db_type)actual_dd_engine);
```
```suggestion
        static_cast<legacy_db_type>(actual_dd_engine);
```

laurynas-biveinis (commented):
What would happen in the following case:
- A DDSE change is initiated and completes half-way
- The instance is killed
- The user resets the DDSE to the old pre-upgrade value in my.cnf, and attempts to start the instance again

sql/dd/impl/bootstrap/bootstrapper.cc
luqun: add another check for actual_dd_engine value for corrupted read.. 

another way is that initialize to unknown first, then if the property doesn't exist, initialize to innodb?

laurynas-biveinis: > add another check for actual_dd_engine value for corrupted read..
> 
> another way is that initialize to unknown first, then if the property doesn't exist, initialize to innodb?

I'd prefer to fail early on such corruption instead of trying to overcome it silently, but my operations background is weak, hence no strong opinion

share/messages_to_error_log.txt
```diff
@@ -11596,6 +11596,15 @@ ER_IB_SLOWRM_CREATE_DIR
 ER_IB_SLOWRM_MOVE_FILE
   eng "%s"
 
+ER_DDSE_CHANGE
+  eng "MySQL DDSE is changing from '%s' to '%s'."
+
+ER_DD_ENGINE_NOT_FOUND
+  eng "The DD_ENGINE setting for the data dictionary was not found. Starting the server using DD_ENGINE= '%u'."
```
laurynas-biveinis:
```suggestion
  eng "The DD_ENGINE setting for the data dictionary was not found. Starting the server using DD_ENGINE='%s'."
```

and a corresponing change in the code. IMHO the engines must be identified by their names, not by enum values

share/messages_to_error_log.txt
```diff
@@ -11596,6 +11596,15 @@ ER_IB_SLOWRM_CREATE_DIR
...
+
+ER_UNSUPPORTED_DD_ENGINE
+  eng "The DD_ENGINE value '%u' found in dd_properties isn't supported DDSE."
```
laurynas-biveinis: Similar problem but IMHO it wouldn't be a good idea to translate all the enum values neither. Either point the user to the source code with enum values, either add something like "supported DDSEs are InnoDB (value 1) and MyRocks (value 2)"

sql/dd/impl/bootstrap/bootstrapper.cc
laurynas-biveinis: > another way is that initialize to unknown first, then if the property doesn't exist, initialize to innodb?

+1

share/messages_to_error_log.txt
luqun: Good point. use name is better than enum. fix both

sql/dd/impl/bootstrap/bootstrapper.cc
```diff
@@ -1211,6 +1213,35 @@ bool initialize_dd_properties(THD *thd) {
       LogErr(ERROR_LEVEL, ER_LCTN_CHANGED, lower_case_table_names, actual_lctn);
       return true;
     }
+
+    /*
+      Read actual dd engine from dd_properties: reject restarting if found
+      unsupported DD_ENGINE; use INNODB if couldn't find DD_ENGINE, since
+      existing INNODB DDSE instance doesn't contains DD_ENGINE property
```
laurynas-biveinis, line 1220:
```suggestion
      existing INNODB DDSE instance doesn't contain DD_ENGINE property
```

laurynas-biveinis (commented):
Could you explain how the scenario in my first comment works?

luqun:
> Could you explain how the scenario in my first comment works?

I will add this information into https://github.com/facebook/mysql-5.6/pull/1354

luqun:
The PR has been committed

PR #1353, opened 2023-09-04
[DD] use actual DDSE during ddse upgrade scenario

luqun:
Summary:

- During boostrap, default_dd_storage_engine and actual DDSE is same;
- During DDSE change(innodb->rocksdb or rocksdb->innodb), default_dd_storage_engine and actual DDSE aren't same. default_dd_storage_engine is the target(or expected) DDSE, actual DDSE is existing DDSE and was fetched from dd_properties tables.

The change:
- use actual DDSE during upgrade(! opt_initialize) scenario.
- always call INNODB handler API for innobase_dict_recover and innobase_dict_init unless INNODB is disabled. This is required for DDSE change --- need to access data in INNODB
- call dict_set_server_version to store server version info
- etc

### DIFF
--- a/sql/dd/dd_table.cc
+++ b/sql/dd/dd_table.cc
@@ -2339,8 +2339,7 @@ static std::unique_ptr<dd::Table> create_dd_system_table(
   } else {
     assert(file->ht->db_type == DB_TYPE_INNODB ||
            file->ht->db_type == DB_TYPE_ROCKSDB);
-    if (file->ht == nullptr || file->ht->dict_register_dd_table_id == nullptr)
-      return nullptr;
+    if (file->ht->dict_register_dd_table_id == nullptr) return nullptr;
     file->ht->dict_register_dd_table_id(tab_obj->se_private_id());
   }
   return tab_obj;
--- a/sql/dd/dd_table.cc
+++ b/sql/dd/dd_table.cc
@@ -2332,10 +2332,17 @@ static std::unique_ptr<dd::Table> create_dd_system_table(
   }
 
   // Register the se private id with the DDSE.
-  handlerton *ddse = get_dd_engine(thd);
-  if (ddse->dict_register_dd_table_id == nullptr) return nullptr;
-  ddse->dict_register_dd_table_id(tab_obj->se_private_id());
-
+  if (opt_initialize) {
+    handlerton *ddse = get_dd_engine(thd);
+    if (ddse->dict_register_dd_table_id == nullptr) return nullptr;
+    ddse->dict_register_dd_table_id(tab_obj->se_private_id());
+  } else {
+    assert(file->ht->db_type == DB_TYPE_INNODB ||
+           file->ht->db_type == DB_TYPE_ROCKSDB);
+    if (file->ht == nullptr || file->ht->dict_register_dd_table_id == nullptr)
+      return nullptr;
+    file->ht->dict_register_dd_table_id(tab_obj->se_private_id());
+  }
   return tab_obj;
 }
 
--- a/sql/dd/impl/bootstrap/bootstrap_ctx.h
+++ b/sql/dd/impl/bootstrap/bootstrap_ctx.h
@@ -140,7 +140,7 @@ class DD_bootstrap_ctx {
 
   uint get_actual_I_S_version() const { return m_actual_I_S_version; }
 
-  legacy_db_type get_actual_dd_engine() const {
+  [[nodiscard]] legacy_db_type get_actual_dd_engine() const {
     assert(m_actual_dd_engine != DB_TYPE_UNKNOWN);
     return m_actual_dd_engine;
   }
--- a/sql/dd/impl/bootstrap/bootstrap_ctx.h
+++ b/sql/dd/impl/bootstrap/bootstrap_ctx.h
@@ -140,7 +140,7 @@ class DD_bootstrap_ctx {
 
   uint get_actual_I_S_version() const { return m_actual_I_S_version; }
 
-  uint get_actual_dd_engine() const {
+  legacy_db_type get_actual_dd_engine() const {
     assert(m_actual_dd_engine != DB_TYPE_UNKNOWN);
     return m_actual_dd_engine;
   }
--- a/sql/dd/impl/bootstrap/bootstrapper.cc
+++ b/sql/dd/impl/bootstrap/bootstrapper.cc
@@ -82,11 +82,30 @@ namespace {
 // Initialize recovery in the DDSE.
 bool DDSE_dict_recover(THD *thd, dict_recovery_mode_t dict_recovery_mode,
                        uint version) {
-  handlerton *ddse = get_dd_engine(thd);
+  handlerton *ddse = nullptr;
+
+  if (opt_initialize) {
+    ddse = get_dd_engine(thd);
+  } else {
+    ddse = ha_resolve_by_legacy_type(
+        thd, bootstrap::DD_bootstrap_ctx::instance().get_actual_dd_engine());
+  }
+
   if (ddse->dict_recover == nullptr) return true;
-
   bool error = ddse->dict_recover(dict_recovery_mode, version);
+  if (error) return error;
 
+  /*
+    Always call innodb handler API unless disabled
+    innobase_dict_recover will initialze innodb services, which is required
+    for DDSE dd change
+  */
+  if (ddse->db_type != DB_TYPE_INNODB) {
+    ddse = ha_resolve_by_legacy_type(thd, DB_TYPE_INNODB);
+    bool disabled = ha_is_storage_engine_disabled(ddse);
+    if (!disabled && ddse->dict_recover == nullptr) return true;
+    if (!disabled) error = ddse->dict_recover(dict_recovery_mode, version);
+  }
   /*
     Commit when tablespaces have been initialized, since in that
     case, tablespace meta data is added.
@@ -725,7 +744,13 @@ namespace bootstrap {
   predefined tables and tablespaces.
 */
 bool DDSE_dict_init(THD *thd, dict_init_mode_t dict_init_mode, uint version) {
-  handlerton *ddse = get_dd_engine(thd);
+  handlerton *ddse = nullptr;
+  if (opt_initialize) {
+    ddse = get_dd_engine(thd);
+  } else {
+    ddse = ha_resolve_by_legacy_type(
+        thd, bootstrap::DD_bootstrap_ctx::instance().get_actual_dd_engine());
+  }
 
   /*
     The lists with element wrappers are mem root allocated. The wrapped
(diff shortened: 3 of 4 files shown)


laurynas-biveinis:
https://github.com/facebook/mysql-5.6/pull/1357 duplicated some of the changes, will review to reconcile both PRs (and likely drop mine)

sql/dd/dd_table.cc
```diff
@@ -2332,10 +2332,17 @@ static std::unique_ptr<dd::Table> create_dd_system_table(
...
-  ddse->dict_register_dd_table_id(tab_obj->se_private_id());
-
+  if (opt_initialize) {
+    handlerton *ddse = get_dd_engine(thd);
+    if (ddse->dict_register_dd_table_id == nullptr) return nullptr;
+    ddse->dict_register_dd_table_id(tab_obj->se_private_id());
+  } else {
+    assert(file->ht->db_type == DB_TYPE_INNODB ||
+           file->ht->db_type == DB_TYPE_ROCKSDB);
+    if (file->ht == nullptr || file->ht->dict_register_dd_table_id == nullptr)
```
laurynas-biveinis: `file->ht == nullptr` check comes after unconditional dereference in the assert above, either the assert needs fixing, either (more likely IMHO) the first condition has to be dropped

sql/dd/impl/bootstrap/bootstrap_ctx.h
```diff
@@ -140,7 +140,7 @@ class DD_bootstrap_ctx {
 
   uint get_actual_I_S_version() const { return m_actual_I_S_version; }
 
-  uint get_actual_dd_engine() const {
+  legacy_db_type get_actual_dd_engine() const {
```
laurynas-biveinis: Might also `[[nodiscard]]` at the same time :)

sql/dd/impl/bootstrap/bootstrapper.cc
```diff
@@ -82,11 +82,30 @@ namespace {
...
-  handlerton *ddse = get_dd_engine(thd);
-  if (ddse->dict_recover == nullptr) return true;
+  handlerton *ddse = nullptr;
 
+  if (opt_initialize) {
+    ddse = get_dd_engine(thd);
+  } else {
+    ddse = ha_resolve_by_legacy_type(
+        thd, bootstrap::DD_bootstrap_ctx::instance().get_actual_dd_engine());
+  }
```
laurynas-biveinis, line 90: `DDSE_dict_recover`, `DDSE_dict_init`, and `sync_meta_data` use the same code, please extract a helper that returns `ddse`

sql/dd/impl/bootstrap/bootstrapper.cc
```diff
@@ -82,11 +82,30 @@ namespace {
...
+
+  if (ddse->dict_recover == nullptr) return true;
   bool error = ddse->dict_recover(dict_recovery_mode, version);
+  if (error) return error;
 
+  /*
+    Always call innodb handler API unless disabled
```
laurynas-biveinis, line 103: the "unless disabled" part looks incorrect, until full conversion of system tables. InnoDB needs to be initialized always, because the system tables are still in it.
Same comment for `DDSE_dict_init` below

sql/dd/impl/bootstrap/bootstrapper.cc
```diff
@@ -739,6 +764,22 @@ bool DDSE_dict_init(THD *thd, dict_init_mode_t dict_init_mode, uint version) {
                            &ddse_tablespaces))
     return true;
 
+  /*
+    Always call innodb handler API unless disabled
+    innobase_dict_init will initialze these innodb specific tables
```
laurynas-biveinis: "initialize"

sql/dd/impl/bootstrap/bootstrapper.cc
```diff
@@ -1815,13 +1864,16 @@ bool update_versions(THD *thd, bool is_dd_upgrade_57) {
 
   /*
     Update the server version number in the bootstrap ctx and the
-    DD tablespace header if we have been doing a server upgrade.
+    DD tablespace header if we have been doing a server upgrade
+    or dd upgrade.
```
```suggestion
    or DDSE change.
```

sql/dd/impl/bootstrap/bootstrapper.cc
```diff
@@ -82,11 +82,30 @@ namespace {
...
+  }
+
+  if (ddse->dict_recover == nullptr) return true;
   bool error = ddse->dict_recover(dict_recovery_mode, version);
+  if (error) return error;
 
+  /*
+    Always call innodb handler API unless disabled
+    innobase_dict_recover will initialze innodb services, which is required
+    for DDSE dd change
```
laurynas-biveinis, line 105: I think code from https://github.com/facebook/mysql-5.6/pull/1357/files#diff-018eea8366c7c269ff17becee828ebbd4d3e029b47a766d199689119732e8f8bR89 has to be added there for the error case

sql/dd/impl/bootstrap/bootstrapper.cc
```diff
@@ -739,6 +764,22 @@ bool DDSE_dict_init(THD *thd, dict_init_mode_t dict_init_mode, uint version) {
                            &ddse_tablespaces))
     return true;
```
laurynas-biveinis, line 765: Add diagnostics here: https://github.com/facebook/mysql-5.6/pull/1357/files#diff-018eea8366c7c269ff17becee828ebbd4d3e029b47a766d199689119732e8f8bR749

sql/dd/impl/bootstrap/bootstrapper.cc
```diff
@@ -739,6 +764,22 @@ bool DDSE_dict_init(THD *thd, dict_init_mode_t dict_init_mode, uint version) {
...
+  */
+  if (ddse->db_type != DB_TYPE_INNODB) {
+    handlerton *innodb_ddse = ha_resolve_by_legacy_type(thd, DB_TYPE_INNODB);
+    bool innodb_disabled = ha_is_storage_engine_disabled(innodb_ddse);
+    if (!innodb_disabled) {
+      assert(!ddse_tablespaces.size());
+      if (innodb_ddse->ddse_dict_init == nullptr ||
+          innodb_ddse->ddse_dict_init(dict_init_mode, version, &ddse_tables,
+                                      &ddse_tablespaces))
+        return true;
```
laurynas-biveinis, line 783: Add diagnostics here: https://github.com/facebook/mysql-5.6/pull/1357/files#diff-018eea8366c7c269ff17becee828ebbd4d3e029b47a766d199689119732e8f8bR760

laurynas-biveinis (commented):
Perhaps you could take the last chunk from my PR: https://github.com/facebook/mysql-5.6/pull/1357/files#diff-018eea8366c7c269ff17becee828ebbd4d3e029b47a766d199689119732e8f8bR924, in which case my PR would be fully consumed by this one

sql/dd/impl/bootstrap/bootstrapper.cc
luqun: Yes. We could only disable innodb after all tables has been move to rocksdb(including these system tables). that disabled check is helpful for future upgrade all tables to rocksdb development, such as easy to know which table still need innodb after we disable innodb..

laurynas-biveinis:
@luqun , is this ready for review?

luqun:
> @luqun , is this ready for review?

Yes. it is. :)

sql/dd/impl/bootstrap/bootstrapper.cc
```diff
@@ -79,13 +79,35 @@ using namespace dd;
 
 namespace {
 
+[[nodiscard]] handlerton *calculate_dd_engine(THD *thd) {
```
laurynas-biveinis, line 84: Maybe a comment explaining the need in upgrade scenario?

sql/dd/impl/bootstrap/bootstrapper.cc
```diff
@@ -79,13 +79,35 @@ using namespace dd;
...
+  handlerton *ddse = calculate_dd_engine(thd);
   if (ddse->dict_recover == nullptr) return true;
-
   bool error = ddse->dict_recover(dict_recovery_mode, version);
+  if (error && dict_recovery_mode == DICT_RECOVERY_INITIALIZE_TABLESPACES)
+    return dd::end_transaction(thd, error);
+
+  /*
+    Always call innodb handler API unless disabled
+    innobase_dict_recover will initialze innodb services, which is required
```
```suggestion
    innobase_dict_recover will initialize innodb services, which is required
```

sql/dd/impl/bootstrap/bootstrapper.cc
```diff
@@ -736,8 +757,31 @@ bool DDSE_dict_init(THD *thd, dict_init_mode_t dict_init_mode, uint version) {
...
+           "Calling DDSE handlerton ddse_dict_init failed");
     return true;
+  }
+
+  /*
+    Always call innodb handler API unless disabled
+    innobase_dict_init will initialize these innodb specific tables
+  */
+  if (ddse->db_type != DB_TYPE_INNODB) {
+    handlerton *innodb_ddse = ha_resolve_by_legacy_type(thd, DB_TYPE_INNODB);
```
laurynas-biveinis: InnoDB is not the DDSE, thus might be clearer to rename the local var:

```suggestion
    handlerton *innodb_se = ha_resolve_by_legacy_type(thd, DB_TYPE_INNODB);
```

sql/dd/impl/bootstrap/bootstrapper.cc
```diff
@@ -736,8 +757,31 @@ bool DDSE_dict_init(THD *thd, dict_init_mode_t dict_init_mode, uint version) {
...
+    bool innodb_disabled = ha_is_storage_engine_disabled(innodb_ddse);
+    if (!innodb_disabled) {
+      assert(!ddse_tablespaces.size());
```
```suggestion
      assert(ddse_tablespaces.empty());
```

sql/dd/impl/bootstrap/bootstrapper.cc
```diff
@@ -736,8 +757,31 @@ bool DDSE_dict_init(THD *thd, dict_init_mode_t dict_init_mode, uint version) {
...
+      if (innodb_ddse->ddse_dict_init == nullptr ||
+          innodb_ddse->ddse_dict_init(dict_init_mode, version, &ddse_tables,
+                                      &ddse_tablespaces)) {
+        LogErr(
+            ERROR_LEVEL, ER_LOG_PRINTF_MSG,
+            "Calling non-Default DDSE InnoDB handlerton ddse_dict_init failed");
```
laurynas-biveinis: I am not sure about the wording here: the DDSE is singleton, and it is not InnoDB here, and this is not about upgrade scenario. So InnoDB here is not "a DDSE, but non-default one", IMHO I'd call it just a SE
```suggestion
            "Calling non-default DDSE InnoDB handlerton ddse_dict_init failed");
```

laurynas-biveinis (commented):
The bit from PR https://github.com/facebook/mysql-5.6/pull/1357/files#diff-018eea8366c7c269ff17becee828ebbd4d3e029b47a766d199689119732e8f8bR924 does not seem to be here?

sql/dd/impl/bootstrap/bootstrapper.cc
luqun: just copy original words from your PR

sql/dd/impl/bootstrap/bootstrapper.cc
```diff
@@ -79,13 +79,37 @@ using namespace dd;
 
 namespace {
 
+// This method is required for upgrade sceanrio
+// during upgrade, there are two dd engine: actual engine and target engine
```
laurynas-biveinis, line 83:
```suggestion
// during upgrade, there are two dd engines: current engine and target engine
```

laurynas-biveinis (commented):
LGTM with a minor comment change

sql/dd/impl/bootstrap/bootstrapper.cc
luqun: Let me add both.. since in code(upstream), it also use actual..yeah, but 'current' maybe easy to understand.

PR #1355, opened 2023-09-05
Cleanup pointers in method signatures called from ha_rocksdb::create_…

laurynas-biveinis:
…key_defs

Refactoring with no behavior changes:
- Whenever a pointer argument must be non-null, replace it with a lvalue or rvalue reference. Drop any asserts testing non-nullness, and nonnull attributes.
- At the same time drop any redundant const qualifiers from scalar parameters
- At the same time replace any warn_unused_result attributes with nodiscard, and add some extra ones.

### DIFF
--- a/storage/rocksdb/ha_rocksdb.cc
+++ b/storage/rocksdb/ha_rocksdb.cc
@@ -9121,16 +9121,16 @@ int ha_rocksdb::convert_record_from_storage_format(
                                            : rc;
 }
 
-int ha_rocksdb::alloc_key_buffers(const TABLE *const table_arg,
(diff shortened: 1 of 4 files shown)


luqun:
most change looks okay. for these non-const pointer to reference change, previously linter may complain "Is this a non-const reference? If so, make const or use a pointer".. with google c++ style change, now it allows non-const reference as parameter(https://github.com/google/styleguide/commit/7a7a2f510efe7d7fc), maybe our codebase can also use latest google c++ style change..

luqun:
@laurynas-biveinis , Current PR has been pushed.

PR #1356, opened 2023-09-05
Remove an incorrect assert from MYSQL_BIN_LOG::prepare

laurynas-biveinis:
It fires on MyRocks DDSE setup, whenever binlog is enabled globally, but turned off at the moment in one thread.

This is fixed in upstream 8.0.32:  https://github.com/laurynas-biveinis/mysql-5.6/commit/c1401ada9b17ecddbbf71706e886f162cff1a37a

### DIFF
--- a/sql/binlog.cc
+++ b/sql/binlog.cc
@@ -10401,12 +10401,6 @@ int MYSQL_BIN_LOG::prepare(THD *thd, bool all) {
   DBUG_TRACE;
 
   assert(opt_bin_log);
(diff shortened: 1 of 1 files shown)


luqun:
Could you add commit hash in upstream 8.0.32?

laurynas-biveinis:
> Could you add commit hash in upstream 8.0.32?

Sure, even though it's missing from the PR description, I have added it already to the commit message by a force push

PR #1357, opened 2023-09-05
Always call InnoDB dict_recover and ddse_dict_init handlerton methods

laurynas-biveinis:
InnoDB performs some of its initialization actions in dict_recover and ddse_dict_init, which cannot be skipped even when the DDSE is different. Thus call them in those cases too.

At the same time add one more diagnostic and split an if statement for easier debugging.

No functional changes when InnoDB is the DDSE.


laurynas-biveinis:
I need to reconcile with https://github.com/facebook/mysql-5.6/pull/1353

laurynas-biveinis:
Subsumed by #1353

PR #1359, opened 2023-09-12
Run clang-format on storage/rocksdb

laurynas-biveinis:
Since it touched one of the lines with these debugging sync points, use this opportunity to fix "stimulate/simulate" typo too.

### DIFF
--- a/mysql-test/suite/rocksdb/r/error_abort_warning_on_corrupt_data.result
+++ b/mysql-test/suite/rocksdb/r/error_abort_warning_on_corrupt_data.result
@@ -1,4 +1,4 @@
-SET SESSION debug="+d, stimulate_corrupt_data_read";
+SET SESSION debug="+d, simulate_corrupt_data_read";
 CREATE DATABASE a;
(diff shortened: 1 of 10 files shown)


laurynas-biveinis:
It is less than ideal to make a formatting-only PR, but tools complained all the time and
```
  void remove_from_global_trx_list(void) {
    DBUG_EXECUTE_IF("rocksdb_trx_list_crash", {
      if (!m_thd->is_system_thread() &&
          !m_thd->is_attachable_transaction_active()) {
        THD *thd = new THD();
      thd->thread_stack = reinterpret_cast<char *>(&(thd));
```
crossed the threshold for me

luqun:
I had thought we had done this before. let's test it in CI

laurynas-biveinis:
This is clang-format 16. Let me know if you are using a different version

luqun:
> This is clang-format 16. Let me know if you are using a different version

we use an old version and it doesn't complain your change. so we should be good to go.

PR #1361, opened 2023-09-15
Fix and skip some MTR tests for MyRocks DDSE

laurynas-biveinis:
- Make some MTR tests to run only under InnoDB DDSE:
  - main.commit_1innodb, main.partition_locking due to counter values being bumped differently between InnoDB and MyRocks DSSEs.
  - main.rename due to multiple operations in a single SQL statement depending on READ UNCOMMITTED support available for DD transactions.
  - main.unsafe_binlog_innodb, binlog.binlog_stm_unsafe_read_acl_tables_in_dml_ddl due to SERIALIZABLE transaction isolation level not being supported.
  - innodb.create_tablespace due to --innodb-read-only not blocking DD updates when it's in another engine.
  - innodb.log_ddl_must_flush due to it testing specifically the InnoDB DDSE code path.
  - innodb.sdi because of different mysql tablespace SDI dump when DDSE is different.
- Reset the transaction isolation level at the end of binlog.binlog_mix_unsafe_read_acl_tables_in_dml_ddl so that the cleanup DDL statements can execute.
- Fix the option syntax loose_rocksdb to loose-rocksdb for rpl_gtid.rpl_wait_for_semi_sync_ack test so that the "loose" prefix actually works, when MyRocks is compiled in statically.
- Do not print storage engine system vars in innodb_gis.geometry.
- Make InnoDB read only mode tests accept ER_INNODB_READ_ONLY error to in addition to ER_READ_ONLY_MODE. The former is returned by the storage engine at a later point, whereas the latter triggers earlier when InnoDB is the DDSE. The tests are: innodb.innodb_force_recovery, innodb.innodb_read_only-2, innodb_undo.undo_settings.


luqun:
how do you verify these change? run whole suites under rocksdb dd ?

laurynas-biveinis:
> how do you verify these change? run whole suites under rocksdb dd ?

Yes

PR #1367, opened 2023-09-19
Make InnoDB work when it's not the DDSE

laurynas-biveinis:
No functional changes when InnoDB is the DDSE. When InnoDB is not the DDSE:
- ha_innobase::get_se_private_data: make sure the InnoDB-specific DD tables (mysql.innodb_dynamic_metadata, mysql.innodb_table_stats, mysql.innodb_index_stats, & mysql.innodb_ddl_log) get the same hardcoded table, index, & root page IDs as they do when InnoDB is the DDSE.
- innobase_ddse_dict_init: register with the DD layer and create a placeholder DD table mysql.dd_properties_placeholder, whose purpose is to reserve the table, index, and root page IDs for the mysql.dd_properties table that is created when InnoDB is the DDSE.
- Make dict_build_table_def give out the hardcoded table IDs for the InnoDB DD tables.
- Add extra asserts to btr_create to ensure that the hardcoded InnoDB DD tables are created with the right IDs.
- Make purge truncate undo tablespaces unconditionally instead of when something purgeable exists. With DDSE in another engine, nothing purgable exists on e.g. dropping an undo tablespace, and without the truncate the undo tablespace state will not change to the one allowing the drop to complete.
- innodb_drop_tablespace, innodb_drop_undo_tablespace: register InnoDB DDL transactions with 2PC because the DDSE participates too.
- Expand MTR suite rocksdb_dd_innodb with InnoDB-specific tests that regressed during development with MyRocks DDSE.

### DIFF
--- a/storage/innobase/handler/ha_innodb.cc
+++ b/storage/innobase/handler/ha_innodb.cc
@@ -5930,7 +5930,8 @@ static int innobase_commit(handlerton *hton, /*!< in: InnoDB handlerton */
     }
 
     /* If SE needs to persist GTID we must have a transaction. */
-    if (thd->se_persists_gtid_explicit()) {
+    if (thd->se_persists_gtid_explicit() &&
+        (innobase_is_ddse() || !trx_state_eq(trx, TRX_STATE_PREPARED))) {
       trx_start_if_not_started(trx, true);
     }
 
@@ -12922,6 +12923,18 @@ static bool innobase_ddse_dict_init(
     return true;
   }
 
+  if (!innobase_is_ddse()) {
+    dd::Object_table *dd_properties_placeholder =
+        dd::Object_table::create_object_table("INNODB");
+    dd_properties_placeholder->set_hidden(true);
+    dd::Object_table_definition *def =
+        dd_properties_placeholder->target_table_definition();
+    def->set_table_name("dd_properties_placeholder");
+    def->add_field(0, "dummy", "dummy INT UNSIGNED NOT NULL");
+    def->add_index(0, "dummy_pk", "PRIMARY KEY (dummy)");
+    tables->push_back(dd_properties_placeholder);
+  }
+
   /* Instantiate table defs only if we are successful so far. */
   dd::Object_table *innodb_dynamic_metadata =
       dd::Object_table::create_object_table("INNODB");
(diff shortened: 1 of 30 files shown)


laurynas-biveinis:
@luqun as discussed, this is the `mysql.dd_properties_placeholder` patch.

While the PR is not complete in that InnoDB is not fully debugged yet, it is safe to merge

laurynas-biveinis:
Added support for the missing `innodb_page_size` values (8k, 32k, & 64k), and rebased

laurynas-biveinis:
Added a check to `innobase_commit` to avoid starting an already-prepared transaction if GTIDs need to be persisted explicitly, added a regression testcase for this change `rocksdb_dd_innodb.rpl_mts_spco_alter_table_analyze_partition_nobinlog`, and rebased.

laurynas-biveinis:
@luqun and others, I have now follow-up changes to InnoDB, which, if force-pushed to this PR, might complicate any in-progress code reviews. Should I wait and make a 2nd PR or a force push would not be a problem?

storage/innobase/handler/ha_innodb.cc
```diff
@@ -16003,6 +16069,9 @@ static int innodb_drop_tablespace(handlerton *hton, THD *thd,
   trx_t *trx = check_trx_exists(thd);
   TrxInInnoDB trx_in_innodb(trx);
   trx_start_if_not_started(trx, true);
+  if (!innobase_is_ddse()) {
```
luqun, line 16072: for competence, do we also need to do for alter_tabespaces? or all cmd in innobase_alter_tablespace()

storage/innobase/srv/srv0srv.cc
```diff
@@ -3081,6 +3081,7 @@ static ulint srv_do_purge(ulint *n_total_purged) {
 
     /* Take a snapshot of the history list before purge. */
     if ((rseg_history_len = trx_sys->rseg_history_len.load()) == 0) {
+      if (!innobase_is_ddse()) trx_purge_truncate_undo_spaces();
```
luqun, line 3084: could you add a comment that this is only needed for drop undo tablespace?

luqun:
As discussed, committed this PR first.

luqun:
@laurynas-biveinis, looks like this PR cause a lot(or all) rocksdb MRR failure when myrocks as DDSE. 
such as rocksdb.1st, maybe in ha_innobase::get_se_private_data(), we should also check whether DD is upgrading. if yes, still use dd_properties? 

Could you take a look https://github.com/facebook/mysql-5.6/pull/1378? or ?

storage/innobase/handler/ha_innodb.cc
laurynas-biveinis: Apparently not, they get registered with the server 2PC. What some of them do need though, is marking them as RW transactions

laurynas-biveinis:
> @laurynas-biveinis, looks like this PR cause a lot(or all) rocksdb MRR failure when myrocks as DDSE. such as rocksdb.1st, maybe in ha_innobase::get_se_private_data(), we should also check whether DD is upgrading. if yes, still use dd_properties?
> 
> Could you take a look #1378? or ?

Can you provide more details about the regressions? The DDSE MyRocks tests could not have been running without this PR in the first place

PR #1369, opened 2023-09-21
Fb mysql 8.0.28 clone plugin autotune

sunshine-Chun:
This is a fix for the overshooting upstream bug. 
The active_task is calculated in a wrong way. In the current calculation, the active_task will always be 1 and the task_target will be always equal to target_speed, which will lead up to the case that all threads use their max bandwidth. (No throttling at all). 
The fix is to change the active_task to the number of threads, since all threads should be running and active.

### DIFF
--- a/plugin/clone/src/clone_client.cc
+++ b/plugin/clone/src/clone_client.cc
@@ -129,7 +129,7 @@ void Client_Stat::update(bool reset, const Thread_Vector &threads,
     m_start_time = cur_time;
     m_initialized = true;
     reset_history(true);
-    set_target_bandwidth(num_workers, true, 0, 0);
+    set_target_bandwidth(num_workers, true);
     return;
   }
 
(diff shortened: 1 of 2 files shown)


plugin/clone/src/clone_client.cc
```diff
@@ -212,31 +212,18 @@ void Client_Stat::update(bool reset, const Thread_Vector &threads,
...
+uint64_t Client_Stat::task_target(uint64_t target_speed, uint32_t num_tasks) {
   assert(num_tasks > 0);
 
   /* Zero is special value indicating unlimited bandwidth. */
   if (target_speed == 0) {
     return (0);
   }
 
-  /* Estimate number of active tasks based on current performance. If target is
-  not set yet, start by assuming all active thread. */
```
laurynas-biveinis, line 228: Can you explain the "target not set yet" part?

sunshine-Chun: I think the target refers to the current_target variable in this function. It is passed from m_target_data_speed from the set_target_bandwidth() function call. This variable is data bytes transferred per sec per thread. The initial value is 0. I believe this comment is referring to the initial case where m_target_data_speed = 0. This function(task_target) is calculating the new target speed per thread. After the first run,  m_target_data_speed will no longer be 0.

The bug here is that it calculates the wrong target speed for each task. The code will let the  target speed for each thread be the clone target speed. (current_speed / current_target = 1 and active_tasks = 1).

laurynas-biveinis (commented):
Algorithmically this looks good, thus if it tests good too, it's good to merge.

Please report upstream with the patch attached if you have the OCA

laurynas-biveinis:
@sunshine-Chun, please close the PR

PR #1371, opened 2023-10-02
Update or skip some MTR tests for MyRocks DDSE, part 3

laurynas-biveinis:
The perfschema tests concern performance_schema sum_rows_updated columns in various tables. InnoDB detects when a row update does not actually need to update anything and returns HA_ERR_RECORD_IS_THE_SAME, and MyRocks does not, resulting in sum_rows_updated differences on DDL with different DDSE setting. Adjust esms_by_all and perfschema_periodic_snapshot_debug not to query sum_rows_updated at all, as neither replace_value nor replace_regex could handle different values well enough. For statements_row_updated, accept either value depending on what is the DDSE.

Skip innodb.innodb and innodb.innodb_misc1 because both test construct CREATE TABLE ... SELECT ... FROM a locked table, executed twice in parallel, returning different locked table errors.

### DIFF
--- a/mysql-test/suite/innodb/t/innodb.test
+++ b/mysql-test/suite/innodb/t/innodb.test
@@ -1,3 +1,6 @@
+# CREATE TABLE ... SELECT ... FROM a locked table returns different errors with
+# different DDSEs
+--source include/have_innodb_ddse.inc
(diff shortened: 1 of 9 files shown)


luqun:
committed

PR #1372, opened 2023-10-02
Use the right handlerton to register a DD table ID

laurynas-biveinis:
When DD tables can be in multiple SEs, this prevents the non-DDSE tables from being registered in the wrong storage engine.


luqun:
committed

PR #1374, opened 2023-10-12
multiple fix for ddse change

luqun:
Summary:

During DDSE change, for each DD table, it should always call ha_get_se_private_data() to fetch its table id and index ids. INNODB DD Tables and MyRocks DD tables has different table ids and index ids.

also During DDSE change, these DD table are temporarily created in <upgrade> schema and then rename it to "mysql".  Due to rename call in myrocks is a meta-data only change and DD table should always created in __system CF, We will always treat <upgrade> table as dd table if the corresponding mysql table is a DD table.

also During DDSE change, the old dd table should be dropped, thus we allow dd table to be dropped if current SE isn't DDSE.

During DDSE change, commit changes before rename tables to avoid use-after-free issue. This may need more crash recovery testing. 

Test Plan:

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/sql/dd/dd_table.cc
+++ b/sql/dd/dd_table.cc
@@ -2324,7 +2324,7 @@ static std::unique_ptr<dd::Table> create_dd_system_table(
       System_tables::instance()->find_type(system_schema.name(), table_name);
   if (opt_initialize ||
       (table_type != nullptr && *table_type == System_tables::Types::INERT) ||
-      dd::bootstrap::DD_bootstrap_ctx::instance().is_dd_engine_change()) {
+      bootstrap::DD_bootstrap_ctx::instance().is_dd_engine_change()) {
     if (file->ha_get_se_private_data(
             tab_obj.get(), (table_type != nullptr &&
                             *table_type == System_tables::Types::INERT)))
@@ -2392,12 +2392,12 @@ std::unique_ptr<dd::Table> create_table(
   const dd::Object_table *dd_table =
       dict->get_dd_table(sch_obj.name(), table_name);
 
-  // During DDSE change, when create a myrocks table
-  // treat <upgrade>.<table> as dd table if mysql.<table> is a dd table
-  if (dd_table == nullptr && sch_obj.name() != MYSQL_SCHEMA_NAME.str &&
-      thd->is_dd_system_thread() &&
+  // During DDSE change, when creating a MyRocks table
+  // treat <upgrade>.<table> as a DD table if mysql.<table> is a DD table
+  if (dd_table == nullptr && thd->is_dd_system_thread() &&
       dd::bootstrap::DD_bootstrap_ctx::instance().is_dd_engine_change() &&
       create_info->db_type->db_type == DB_TYPE_ROCKSDB &&
+      sch_obj.name() != MYSQL_SCHEMA_NAME.str &&
       dd::get_dictionary()->is_dd_table_name(MYSQL_SCHEMA_NAME.str,
                                              table_name)) {
     dd_table = dict->get_dd_table(MYSQL_SCHEMA_NAME.str, table_name);
--- a/sql/dd/dd_table.cc
+++ b/sql/dd/dd_table.cc
@@ -54,8 +54,9 @@
 #include "sql/dd/performance_schema/init.h"    // performance_schema::
                                                //   set_PS_version_for_table
 #include "sql/create_field.h"
-#include "sql/dd/dd_version.h"  // DD_VERSION
-#include "sql/dd/properties.h"  // dd::Properties
+#include "sql/dd/dd_version.h"                    // DD_VERSION
+#include "sql/dd/impl/bootstrap/bootstrap_ctx.h"  // DD_bootstrap_ctx
+#include "sql/dd/properties.h"                    // dd::Properties
 #include "sql/dd/string_type.h"
 #include "sql/dd/tablespace_id_owner_visitor.h"  // visit_tablespace_id_owners
 #include "sql/dd/types/abstract_table.h"
@@ -2315,14 +2316,15 @@ static std::unique_ptr<dd::Table> create_dd_system_table(
     return nullptr;
 
   /*
-    During --initialize, and for inert tables, get the SE private data
-    from the SE, and store it in the dd_properties table at a later stage.
+    During --initialize, DDSE change and for inert tables, get the SE private
+    data from the SE, and store it in the dd_properties table at a later stage.
     Otherwise, get the SE private data from the 'dd_properties' table.
   */
   const System_tables::Types *table_type =
       System_tables::instance()->find_type(system_schema.name(), table_name);
   if (opt_initialize ||
-      (table_type != nullptr && *table_type == System_tables::Types::INERT)) {
+      (table_type != nullptr && *table_type == System_tables::Types::INERT) ||
+      dd::bootstrap::DD_bootstrap_ctx::instance().is_dd_engine_change()) {
     if (file->ha_get_se_private_data(
             tab_obj.get(), (table_type != nullptr &&
                             *table_type == System_tables::Types::INERT)))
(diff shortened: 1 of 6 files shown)


sql/dd/dd_table.cc
```diff
@@ -2315,14 +2316,15 @@ static std::unique_ptr<dd::Table> create_dd_system_table(
...
+    During --initialize, DDSE change and for inert tables, get the SE private
+    data from the SE, and store it in the dd_properties table at a later stage.
     Otherwise, get the SE private data from the 'dd_properties' table.
   */
   const System_tables::Types *table_type =
       System_tables::instance()->find_type(system_schema.name(), table_name);
   if (opt_initialize ||
-      (table_type != nullptr && *table_type == System_tables::Types::INERT)) {
+      (table_type != nullptr && *table_type == System_tables::Types::INERT) ||
+      dd::bootstrap::DD_bootstrap_ctx::instance().is_dd_engine_change()) {
```
laurynas-biveinis: We are already in the `dd::` namespace here:
```suggestion
      bootstrap::DD_bootstrap_ctx::instance().is_dd_engine_change()) {
```

sql/dd/dd_table.cc
```diff
@@ -2390,6 +2392,16 @@ std::unique_ptr<dd::Table> create_table(
   const dd::Object_table *dd_table =
       dict->get_dd_table(sch_obj.name(), table_name);
 
+  // During DDSE change, when create a myrocks table
```
laurynas-biveinis:
```suggestion
  // During DDSE change, when creating a MyRocks table
```

sql/dd/dd_table.cc
```diff
@@ -2390,6 +2392,16 @@ std::unique_ptr<dd::Table> create_table(
...
+  // treat <upgrade>.<table> as dd table if mysql.<table> is a dd table
```
laurynas-biveinis:
```suggestion
  // treat <upgrade>.<table> as a DD table if mysql.<table> is a DD table
```

sql/dd/dd_table.cc
```diff
@@ -2390,6 +2392,16 @@ std::unique_ptr<dd::Table> create_table(
...
+  if (dd_table == nullptr && sch_obj.name() != MYSQL_SCHEMA_NAME.str &&
```
laurynas-biveinis: Consider moving `sch_obj.name() != ...` just before `is_dd_table_name`, so that the cheaper checks go first

sql/dd/impl/upgrade/dd.cc
```diff
@@ -1138,7 +1138,12 @@ bool upgrade_dd_properties_table(THD *thd, Object_id mysql_schema_id,
   }
 
   // update myrocks own data dictionary
+  // for myrocks, previous changes involve <target_table> tbl_def,
```
laurynas-biveinis:
```suggestion
  // For MyRocks, previous changes involve <target_table> tbl_def,
```

sql/dd/impl/upgrade/dd.cc
```diff
@@ -1138,7 +1138,12 @@ bool upgrade_dd_properties_table(THD *thd, Object_id mysql_schema_id,
...
+  // update_myrocks_table_names will delete <target_table> tbl_def
+  // if commit changes after update_myrocks_table_names, will cause
```
laurynas-biveinis:
```suggestion
  // If the changes are committed after update_myrocks_table_names, it will cause a
```

storage/innobase/handler/ha_innodb.cc
```diff
@@ -15296,8 +15296,11 @@ be dropped
 @return	error number
 @retval 0 on success */
 int ha_innobase::delete_table(const char *name, const dd::Table *table_def) {
+  // during ddse change, s_dd_table_ids may contain old dd table ids, thus
+  // allow drop/rename if current se isn't target ddse
   if (table_def != nullptr &&
-      dict_sys_t::is_dd_table_id(table_def->se_private_id())) {
+      dict_sys_t::is_dd_table_id(table_def->se_private_id()) &&
+      (default_dd_storage_engine == DEFAULT_DD_INNODB)) {
```
laurynas-biveinis: This does not look right unless I'm missing something. It looks like it will allow user `DROP TABLE mysql.tables` when InnoDB is the DDSE. Can the DDSE change check be added here?

https://github.com/facebook/mysql-5.6/pull/1367 adds `innobase_is_ddse` helper, which should be used here if that PR lands first, otherwise I'll clean up in one of the later PRs.

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -10288,8 +10288,15 @@ int ha_rocksdb::create_table(const std::string &table_name,
   DBUG_ENTER_FUNC();
 
   int err;
+  // DD tables should create in __system CF.
```
```suggestion
  // DD tables should be created in the __system CF.
```

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -10288,8 +10288,15 @@ int ha_rocksdb::create_table(const std::string &table_name,
...
+  // During DDSE change, sql layer will create DD tables as <upgrade>.<table>
```
```suggestion
  // During the DDSE change, the SQL layer will create DD tables as <upgrade>.<table>
```

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -10288,8 +10288,15 @@ int ha_rocksdb::create_table(const std::string &table_name,
...
+  // first, then rename it to mysql.<table>.
```
```suggestion
  // first, then rename them to mysql.<table>.
```

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -10288,8 +10288,15 @@ int ha_rocksdb::create_table(const std::string &table_name,
...
+  // rename is meta-data only change, thus treat these <upgrade>.<table>
```
```suggestion
  // The rename is a meta-data only change, thus treat these <upgrade>.<table> tables
```

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -10288,8 +10288,15 @@ int ha_rocksdb::create_table(const std::string &table_name,
...
+  // as DD table.
```
```suggestion
  // as DD tables.
```

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -10288,8 +10288,15 @@ int ha_rocksdb::create_table(const std::string &table_name,
...
+  auto db_name =
```
```suggestion
  const auto db_name =
```

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -10288,8 +10288,15 @@ int ha_rocksdb::create_table(const std::string &table_name,
...
+      ha_thd()->is_dd_system_thread() ? "mysql" : table_arg.s->db.str;
```
laurynas-biveinis, line 10297: Is the `is_dd_system_thread` check granular enough? I am a bit worried about this thread creating e.g. system not DD tables at a later point.

storage/rocksdb/rdb_native_dd.cc
```diff
@@ -34,7 +34,10 @@ bool native_dd::is_dd_table_id(dd::Object_id id) {
 }
 
 int native_dd::reject_if_dd_table(const dd::Table *table_def) {
-  if (table_def != nullptr && is_dd_table_id(table_def->se_private_id())) {
```
laurynas-biveinis, line 37: Same comments as for the parallel change in InnoDB (except for `innobase_is_ddse`)

storage/rocksdb/ha_rocksdb.cc
luqun: if it is system table, then the later code --- dd::get_dictionary()->is_dd_table_name() will return false, since it isn't DD table even with "mysql" schema name.

luqun: Yeah, the best way is to check whether it is dd engine change, but that requires a dependency between SE and DD layer..

storage/innobase/handler/ha_innodb.cc
luqun: how about add a DD system thread check.. Currently DD_bootstrap_ctx isn't exposed to SE and is_dd_system_thread() is widely used in DD layer..

laurynas-biveinis: Let's use `is_dd_system_thread`

storage/rocksdb/ha_rocksdb.cc
laurynas-biveinis: OK

storage/rocksdb/rdb_native_dd.cc
luqun: same as innodb one: add a dd_system_thread check to avoid user drop these DD tables.

PR #1375, opened 2023-10-16
Avoid memory leak during Rdb_ddl_manager::reset_map

luqun:
Summary:
These tbl_def pointers in m_ddl_map must be released before clear() during reset_map

Test Plan:

Reviewers:

Subscribers:

Tasks:

Tags:


laurynas-biveinis (commented):
LGTM

PR #1378, opened 2023-10-23
add is_dd_engine_change variable

luqun:
Summary: Innodb SE need to know whether DD engine change is in progress for its handler API to return correct data. such as get_se_private_data() should return mysql.dd_properties table data is DD engine change in in progress and also DDSE is rocksdb. 

Expose is_dd_engine_change information in dd.h. also update other places to use is_dd_engine_change variable to check whether DD engine change is in progress. 

Test Plan:

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/sql/dd/impl/upgrade/dd.cc
+++ b/sql/dd/impl/upgrade/dd.cc
@@ -1048,7 +1048,8 @@ bool update_object_ids(THD *thd, const std::set<String_type> &create_set,
 */
 bool update_myrocks_table_names(THD *thd, const String_type &old_schema_name,
                                 const String_type &new_schema_name,
-                                const std::set<String_type> &table_names) {
+    const std::set<String_type> &table_names,
+    const std::set<String_type> &filter_table_names) {
   assert(default_dd_storage_engine == DEFAULT_DD_ROCKSDB);
   assert(bootstrap::DD_bootstrap_ctx::instance().is_dd_engine_change());
 
@@ -1062,8 +1063,12 @@ bool update_myrocks_table_names(THD *thd, const String_type &old_schema_name,
     assert(false);
     return true;
   }
+
   bool err = false;
   for (const auto &table_name : table_names) {
+    // Skip some tables, such as InnoDB tables
+    if (filter_table_names.find(table_name) != filter_table_names.cend())
+      continue;
     char old_table[FN_REFLEN + 1];
     char new_table[FN_REFLEN + 1];
     // old table full name under old schema name
@@ -1146,7 +1151,7 @@ bool upgrade_dd_properties_table(THD *thd, Object_id mysql_schema_id,
       !dd::end_transaction(thd, false) &&
       update_myrocks_table_names(thd, target_table_schema_name,
                                  MYSQL_SCHEMA_NAME.str,
-                                 {dd_property_table_name})) {
+                                 {dd_property_table_name}, {})) {
     LogErr(ERROR_LEVEL, ER_LOG_PRINTF_MSG,
            "Failed to update myrocks table name for dd_properties table");
     return true;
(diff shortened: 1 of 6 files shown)


storage/innobase/handler/ha_innodb.cc
```diff
@@ -12923,7 +12923,7 @@ static bool innobase_ddse_dict_init(
     return true;
   }
 
-  if (!innobase_is_ddse()) {
+  if (!innobase_is_ddse() && !is_dd_engine_change) {
```
laurynas-biveinis: Some of the previous PRs used a different check here,  i.e. `ha_thd()->is_dd_system_thread()`. Would it work here?

If not, then could this variable
- be moved to a more specific place? Maybe `dd/dd.h`, which is already included by `ha_innodb.cc` (and the actual definition to `dd/dd.cc`)
- consider converting it to a checking function instead of a variable

luqun: > Some of the previous PRs used a different check here, i.e. ha_thd()->is_dd_system_thread(). Would it work here?

I think it won't work. since in both case(innodb ddse, myrocks ddse), it will call these ddse methods in is_dd_system_thread().. 

Let me also replace is_dd_system_thread() with the new method

luqun: also disable rocksdb ddse combination first..

luqun: @laurynas-biveinis , still see some failures such as rocksdb.tmp_table mtr in myrocks DDSE
still investigating

luqun:
updated: fixed

luqun:
@laurynas-biveinis , could you please take a look again?

sql/dd/impl/upgrade/dd.cc
```diff
@@ -1062,8 +1063,12 @@ bool update_myrocks_table_names(THD *thd, const String_type &old_schema_name,
     assert(false);
     return true;
   }
+
   bool err = false;
   for (const auto &table_name : table_names) {
+    // skip some tables, such as innodb table
```
laurynas-biveinis:
```suggestion
    // Skip some tables, such as InnoDB tables
```

sql/dd/impl/upgrade/dd.cc
```diff
@@ -1062,8 +1063,12 @@ bool update_myrocks_table_names(THD *thd, const String_type &old_schema_name,
...
+    if (filter_table_names.find(table_name) != filter_table_names.end())
```
laurynas-biveinis:
```suggestion
    if (filter_table_names.find(table_name) != filter_table_names.cend())
```

storage/innobase/handler/ha_innodb.cc
```diff
@@ -15015,7 +15015,8 @@ bool ha_innobase::get_se_private_data(dd::Table *dd_table, bool reset) {
     dict_sys_t::s_dd_table_ids.clear();
   }
 
-  if (!innobase_is_ddse()) {
+  // During DDSE change, SQL layer may ask data for "dd_properties" table
```
laurynas-biveinis: This comment needs a bit of clarification because the code in the if statement does not handle "dd_properties" table, only "dd_properties_placeholder"

laurynas-biveinis (commented):
Please update the commit message to reflect the commit contents
LGTM with these changes

laurynas-biveinis:
@luqun , I think I understand what happened with https://github.com/facebook/mysql-5.6/pull/1367 regressions, please hold merging this for short time

laurynas-biveinis:
@luqun, so at least the part of what's going on is that my work branch has RocksDB as the default DDSE value. For tests, that had the effect of initializing the instance with RocksDB, and running RocksDB DDSE tests. Now on trunk, the test instance is initialized with InnoDB DDSE, and when the test tries to run with RocksDB DDSE, the first thing that happens is the crossgrade. Please go ahead with your PR, I'll adjust the tests to avoid this behavior too.

laurynas-biveinis:
Another issue that the mysql_system_table.sql bit that prevented mysql.password being created in RocksDB got lost in its PR, while the code that was not, meaning that MyRocks DDSE instances will fail to initialize: https://github.com/facebook/mysql-5.6/pull/1381

PR #1380, opened 2023-10-27
Fix memory leak in i_s_innodb_tablespaces_fill_table

luqun:
Summary:

Looks like when space is 0, we forgot to release memory.

Following innodb code, after finish works, then call release heap memory at end

Test Plan: MTR under asan build

Reviewers:

Subscribers:

Tasks:

Tags:


laurynas-biveinis (commented):
LGTM

PR #1381, opened 2023-10-27
Do not create mysql.password_history in MyRocks if MyRocks is the DDSE

laurynas-biveinis:
This table is a system table, not a DD table. Eventually we'll create it in MyRocks, but for now the mix-up between system and DD tables is confusing, and the source code counterpart allowing the creation of password_history in MyRocks was already reverted in https://github.com/facebook/mysql-5.6/pull/1333


laurynas-biveinis:
This was already a part of https://github.com/facebook/mysql-5.6/pull/1333, but was not pushed with the rest of the patch

PR #1382, opened 2023-10-29
Fixes for InnoDB when it's not the DDSE

laurynas-biveinis:
- Mark CREATE TABLE, DROP TABLE, RENAME TABLE, TRUNCATE, CREATE TABLESPACE, & ALTER TABLESPACE RENAME transactions RW. When InnoDB is not the DDSE, the InnoDB transactions for these commands were not marked RW, consequently InnoDB prepare was skipped for them, breaking crash recovery for these commands. Add tests.
- Fix GTID persistence in InnoDB in multiple-engine transactions if binlog is disabled. This requires checking for this condition and starting the internal InnoDB transaction in innobase_xa_prepare, similar to existing code in innobase_commit. Add tests.
- Fix error handling for a CREATE TABLESPACE failure. Register the InnoDB transaction in innodb_create_tablespace with the server 2PC, add a regression testcase based on innodb.create_tablespace_debug.
- Fix the check in innobase_commit that tried to start any non-prepared InnoDB transactions in case GTID needed to be persisted. The previous checks used trx_state_eq, which is not suitable for negative checks due to internal asserts. This fixes several test failures in existing tests.
- Add a comment why srv_do_purge needs to call trx_purge_truncate_undo_spaces if InnoDB is not the DDSE.
- After crash recovery runs, if the DDSE is not InnoDB, reset dictionary tables and tablespaces in InnoDB too. This fixes the issue where the crash recovery changes the data dictionary, but InnoDB has the pre-recovery view.
- Adapt innodb_undo.undo_tablespace_encrypt to work with either DDSE, as needed by rocksdb_dd_innodb.undo_tablespace_encrypt.
- Run the InnoDB regression tests in rocksdb_dd_innodb suite with MyRocks DDSE, as originally intended.


laurynas-biveinis:
This PR contains the entire https://github.com/facebook/mysql-5.6/pull/1381 and the `mysql-test/extra/rpl_tests/rpl_mts_spco_commands.test` bit of https://github.com/facebook/mysql-5.6/pull/1379, not mentioned in the commit message, to be self-contained. Thus should be merged after these two requests

laurynas-biveinis:
@luqun , rebased

sql/dd/impl/dictionary_impl.cc
```diff
@@ -705,6 +705,13 @@ bool reset_tables_and_tablespaces() {
   if (ddse->dict_cache_reset_tables_and_tablespaces != nullptr)
     ddse->dict_cache_reset_tables_and_tablespaces();
 
+  if (default_dd_storage_engine != DEFAULT_DD_INNODB) {
+    auto *const innodb_se = ha_resolve_by_legacy_type(thd.thd, DB_TYPE_INNODB);
```
luqun, line 709: could you add a innodb isn't disabled?

storage/innobase/handler/ha_innodb.cc
```diff
@@ -20094,6 +20126,13 @@ static int innobase_xa_prepare(handlerton *hton, /*!< in: InnoDB handlerton */
 
   if (prepare_trx ||
       (!thd_test_options(thd, OPTION_NOT_AUTOCOMMIT | OPTION_BEGIN))) {
+
```
luqun, line 20130: could we add a check that trx is rw..

storage/innobase/handler/ha_innodb.cc
```diff
@@ -15651,6 +15675,10 @@ static int innodb_create_tablespace(handlerton *hton, THD *thd,
   trx_t *trx = check_trx_exists(thd);
   TrxInInnoDB trx_in_innodb(trx);
   trx_start_if_not_started(trx, true);
+  if (!innobase_is_ddse()) {
+    innobase_register_trx(hton, thd, trx);
+    mark_trx_rw(*thd);
```
luqun, line 15680: do you need similar change(mark_trx_rw) for innodb_drop_tablespace()

laurynas-biveinis: `DROP TABLESPACE` transactions were already RW

storage/innobase/handler/ha_innodb.cc
laurynas-biveinis: I have added a top-level `assert(!trx->read_only)`. It is an aggressive assert that checks that no RO-at-InnoDB level transactions were registered as RW in the server layer, the tests pass with it.

laurynas-biveinis:
@luqun, rebased, ready for review for again

PR #1383, opened 2023-10-30
Add MTR tests for MySQL data dictionary in MyRocks

laurynas-biveinis:
These are regression tests for MyRocks only. The regression tests involving InnoDB in any way are in rocksdb_dd_innodb suite.

PR #1384, opened 2023-10-30
Allow unsupported transaction isolation levels for DD operations

laurynas-biveinis:
If the current transaction isolation level is lower than READ COMMITTED, or higher than REPEATABLE READ, allow it silently if it's a DD operation. The rest of MyRocks code will work by silently clamping to the nearest supported level, and the clamped level is good enough for DD operations. InnoDB has similar code, even though it supports the other isolation levels.

### DIFF
--- a/storage/rocksdb/ha_rocksdb.cc
+++ b/storage/rocksdb/ha_rocksdb.cc
@@ -14032,6 +14032,7 @@ void ha_rocksdb::read_thd_vars(THD *const thd) {
 int ha_rocksdb::external_lock(THD *const thd, int lock_type) {
   DBUG_ENTER_FUNC();
 
+  assert(table_share != nullptr);
   assert(thd != nullptr);
 
   int res = HA_EXIT_SUCCESS;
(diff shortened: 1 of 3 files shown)


laurynas-biveinis:
This PR contains ~~https://github.com/facebook/mysql-5.6/pull/1381 and~~ `ddse.test` from https://github.com/facebook/mysql-5.6/pull/1383, to which it appends. These PRs should be merged first

laurynas-biveinis:
@luqun, rebased. #1381 dependency is dropped but #1383 dependency remains

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -14066,8 +14066,12 @@ int ha_rocksdb::external_lock(THD *const thd, int lock_type) {
       }
     }
   } else {
-    if (my_core::thd_tx_isolation(thd) < ISO_READ_COMMITTED ||
-        my_core::thd_tx_isolation(thd) > ISO_REPEATABLE_READ) {
+    // Check for unsupported transaction isolation levels but allow them for DD
+    // tables. In the latter case they get silenty clamped to RC or RR, which is
+    // good enough for DD operations.
+    if (table_share->table_category != TABLE_CATEGORY_DICTIONARY &&
```
luqun, line 14065: could we check whether table name is DD table?

laurynas-biveinis: `table_category == TABLE_CATEGORY_DICTIONARY` is a cached result of `dd::get_dictionary()->is_dd_table_name(schema_name, table_name)`, so in that way it is already that check, just cheaper, in fact, if possible, I'd convert other string-based checks to this one in the SE layer, if possible.

luqun: okay. for external_lock, use table_category may make more sense for perf.

could you add a check for table_share is nullptr or not.

luqun:
> @luqun, rebased. #1381 dependency is dropped but #1383 dependency remains

#1383 also committed.

laurynas-biveinis:
@luqun rebased, added a `table_share != nullptr` assert. All dependencies dropped now.

PR #1386, opened 2023-11-02
Apply sst file manager in myrocks clone to add slow-rm during checkpoints removal

sunshine-Chun:
This diff makes the following two changes:
1. Clone use rocksdb::DestroyDB() to clean the checkpoint. However, we didn't pass the sst file manager so didn't make use of the slow-rm feature provided by sst file manager. This will lead up to a discard spike when rolling checkpoints. This change pass the sst file manager when cleaning checkpoints.
2. During slow-rm sst files, the checkpoint directory may not be deleted immediately. In the rolling checkpoint function call, we will first delete the checkpoint directory, then create a new one with the same name. Since we are slow removing the sst files, it's possible that when we create the new checkpoints, the old directory hasn't been deleted yet. This will cause a 'directory exist' error when creating a new checkpoint directory. In this change, we add an additional suffix to differentiate rolling checkpoints directory to avoid this error.

### DIFF
--- a/storage/rocksdb/clone/donor.cc
+++ b/storage/rocksdb/clone/donor.cc
@@ -56,8 +56,8 @@ namespace {
 class [[nodiscard]] rdb_checkpoint final {
  public:
   rdb_checkpoint()
-      : m_dir{
-            make_dir_name(m_next_id.fetch_add(1, std::memory_order_relaxed))} {}
+      : m_prefix_dir{make_dir_prefix_name(
+            m_next_id.fetch_add(1, std::memory_order_relaxed))} {}
 
   ~rdb_checkpoint() {
     // Ignore the return value - at this point the clone operation is completing
@@ -68,9 +68,8 @@ class [[nodiscard]] rdb_checkpoint final {
   // Returns MySQL error code
   [[nodiscard]] int init() {
     assert(!m_active);
-    m_full_dir = make_dir_full_name(
-        m_dir.c_str(), m_next_sub_id.fetch_add(1, std::memory_order_relaxed));
-    const auto result = myrocks::rocksdb_create_checkpoint(m_full_dir.c_str());
+    m_dir = make_dir_name(m_prefix_dir, m_next_sub_id++);
+    const auto result = myrocks::rocksdb_create_checkpoint(m_dir.c_str());
     m_active = (result == HA_EXIT_SUCCESS);
     return m_active ? 0 : ER_INTERNAL_ERROR;
   }
@@ -79,22 +78,21 @@ class [[nodiscard]] rdb_checkpoint final {
   [[nodiscard]] int cleanup() {
     if (!m_active) return 0;
     m_active = false;
-    return myrocks::rocksdb_remove_checkpoint(m_full_dir.c_str()) ==
-                   HA_EXIT_SUCCESS
+    return myrocks::rocksdb_remove_checkpoint(m_dir.c_str()) == HA_EXIT_SUCCESS
                ? 0
                : ER_INTERNAL_ERROR;
   }
 
   [[nodiscard]] constexpr const std::string &get_dir() const noexcept {
     assert(m_active);
-    return m_full_dir;
+    return m_dir;
   }
 
   [[nodiscard]] std::string path(const std::string &file_name) const {
     // We might be calling this for inactive checkpoint too, if the donor is in
     // the middle of a checkpoint roll. The caller will handle any ENOENTs as
     // needed.
-    return myrocks::rdb_concat_paths(m_full_dir, file_name);
+    return myrocks::rdb_concat_paths(m_dir, file_name);
   }
 
   rdb_checkpoint(const rdb_checkpoint &) = delete;
@@ -103,17 +101,17 @@ class [[nodiscard]] rdb_checkpoint final {
   rdb_checkpoint &operator=(rdb_checkpoint &&) = delete;
 
  private:
-  const std::string m_dir;
+  const std::string m_prefix_dir;
 
-  std::string m_full_dir;
+  std::string m_dir;
 
   bool m_active = false;
 
   static std::atomic<std::uint64_t> m_next_id;
 
-  std::atomic<std::uint64_t> m_next_sub_id{1};
+  std::uint64_t m_next_sub_id = 1;
 
-  [[nodiscard]] static std::string make_dir_name(std::uint64_t id) {
+  [[nodiscard]] static std::string make_dir_prefix_name(std::uint64_t id) {
     const auto base_str = myrocks::clone::checkpoint_base_dir();
     const auto id_str = std::to_string(id);
     std::string result;
@@ -128,14 +126,14 @@ class [[nodiscard]] rdb_checkpoint final {
     return result;
   }
 
-  [[nodiscard]] static std::string make_dir_full_name(std::string base_dir,
-                                                      std::uint64_t id) {
+  [[nodiscard]] static std::string make_dir_name(
+      const std::string &dir_name_prefix, std::uint64_t id) {
     const auto id_str = std::to_string(id);
     std::string result;
-    result.reserve(base_dir.length() + id_str.length() +
+    result.reserve(dir_name_prefix.length() + id_str.length() +
                    1);  // +1 for '-', the trailing
                         // '\0' is accounted by the sizeof.
-    result = base_dir;
+    result = dir_name_prefix;
     result += '-';
     result += id_str;
     return result;
--- a/storage/rocksdb/clone/donor.cc
+++ b/storage/rocksdb/clone/donor.cc
@@ -68,7 +68,9 @@ class [[nodiscard]] rdb_checkpoint final {
   // Returns MySQL error code
   [[nodiscard]] int init() {
     assert(!m_active);
-    const auto result = myrocks::rocksdb_create_checkpoint(m_dir.c_str());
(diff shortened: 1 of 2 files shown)


laurynas-biveinis:
@sunshine-Chun, I am getting linker errors now:
```
ld: Undefined symbols:
  myrocks::rocksdb_remove_checkpoint(char const*), referenced from:
      myrocks::rocksdb_close_connection(handlerton*, THD*) in librocksdb_se.a[2](ha_rocksdb.cc.o)
      myrocks::rocksdb_create_temporary_checkpoint_validate(THD*, SYS_VAR*, void*, st_mysql_value*) in librocksdb_se.a[2](ha_rocksdb.cc.o)
      (anonymous namespace)::rdb_checkpoint::cleanup() in librocksdb_se.a[25](donor.cc.o)

```

sunshine-Chun:
@laurynas-biveinis Did you hit it during clone? Can you elaborate how you hit this error?

laurynas-biveinis:
> @laurynas-biveinis Did you hit it during clone? Can you elaborate how you hit this error?

It's a linker error during server build

storage/rocksdb/clone/donor.cc
```diff
@@ -102,10 +105,14 @@ class [[nodiscard]] rdb_checkpoint final {
  private:
   const std::string m_dir;
 
+  std::string m_full_dir;
+
   bool m_active = false;
 
   static std::atomic<std::uint64_t> m_next_id;
 
+  std::atomic<std::uint64_t> m_next_sub_id{1};
```
laurynas-biveinis: This should not be atomic. It is private to a single clone session, and all its accesses happen with `m_donor_mutex` locked.

storage/rocksdb/clone/donor.cc
```diff
@@ -68,7 +68,9 @@ class [[nodiscard]] rdb_checkpoint final {
   // Returns MySQL error code
   [[nodiscard]] int init() {
     assert(!m_active);
-    const auto result = myrocks::rocksdb_create_checkpoint(m_dir.c_str());
+    m_full_dir = make_dir_full_name(
```
laurynas-biveinis: It might be confusing to have `m_dir` and `m_full_dir` and `make_dir_name` and `make_dir_full_name`. After this patch, the dir name is the one that has all the suffixes, and the previous dir name is a prefix. Please rename `make_dir_name` to `make_dir_prefix_name`, `make_dir_full_name` to `make_dir_name`, and member variables too.

storage/rocksdb/clone/donor.cc
```diff
@@ -120,6 +127,19 @@ class [[nodiscard]] rdb_checkpoint final {
     result += id_str;
     return result;
   }
+
+  [[nodiscard]] static std::string make_dir_full_name(std::string base_dir,
```
laurynas-biveinis:
```suggestion
  [[nodiscard]] static std::string make_dir_full_name(const std::string &dir_name_prefix,
```

storage/rocksdb/clone/donor.cc
```diff
@@ -68,7 +68,9 @@ class [[nodiscard]] rdb_checkpoint final {
   // Returns MySQL error code
   [[nodiscard]] int init() {
     assert(!m_active);
-    const auto result = myrocks::rocksdb_create_checkpoint(m_dir.c_str());
+    m_full_dir = make_dir_full_name(
+        m_dir.c_str(), m_next_sub_id.fetch_add(1, std::memory_order_relaxed));
```
laurynas-biveinis: No need to take a C string out of a C++ string to convert it to a C++ string again:
```suggestion
        m_dir, m_next_sub_id.fetch_add(1, std::memory_order_relaxed));
```

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -1325,6 +1311,27 @@ static void rocksdb_set_io_write_timeout(
...
+  const auto status = rocksdb::DestroyDB(checkpoint_dir, op);
+
+  if (status.ok()) {
+    return HA_EXIT_SUCCESS;
+  }
+  my_error(ER_GET_ERRMSG, MYF(0), status.code(), status.ToString().c_str(),
+           rocksdb_hton_name);
+  return HA_EXIT_FAILURE;
+}
+
```
laurynas-biveinis: So the reason for my linker error is that this function now fell into a `#ifndef __APPLE__` block, please pull outside

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -1325,6 +1311,27 @@ static void rocksdb_set_io_write_timeout(
   RDB_MUTEX_UNLOCK_CHECK(rdb_sysvars_mutex);
 }
 
+int rocksdb_remove_checkpoint(const char *checkpoint_dir_raw) {
+  const auto checkpoint_dir = rdb_normalize_dir(checkpoint_dir_raw);
+  LogPluginErrMsg(INFORMATION_LEVEL, ER_LOG_PRINTF_MSG,
+                  "deleting temporary checkpoint in directory : %s\n",
+                  checkpoint_dir.c_str());
+
+  std::string trash_dir = std::string(rocksdb_datadir) + "/trash";
```
laurynas-biveinis: IMHO this is important to deduplicate with the other trash dir occurence in `rocksdb_init_internal`, please extract a helper to get the trash dir.

```suggestion
  const auto trash_dir = std::string(rocksdb_datadir) + "/trash";
```

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -1325,6 +1311,27 @@ static void rocksdb_set_io_write_timeout(
...
+  rocksdb::Options op = rocksdb::Options();
```
```suggestion
  auto op = rocksdb::Options();
```

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -1325,6 +1311,27 @@ static void rocksdb_set_io_write_timeout(
...
+  op.sst_file_manager.reset(NewSstFileManager(
```
laurynas-biveinis: Not an expert in SST file manager so open questions:
1) can we save reuse the existing manager from `rocksdb_init_internal` as singleton?
2) if no, is it really correct to pass `delete_existing_trash = true` to this one? Wouldn't this be stepping on the toes of the other SST file manager instances?

storage/rocksdb/clone/donor.cc
sunshine-Chun: resolved

storage/rocksdb/clone/donor.cc
sunshine-Chun: resolved.

storage/rocksdb/clone/donor.cc
sunshine-Chun: resloved

storage/rocksdb/ha_rocksdb.cc
sunshine-Chun: resolved

storage/rocksdb/clone/donor.cc
sunshine-Chun: resolved

storage/rocksdb/ha_rocksdb.cc
sunshine-Chun: resolved

storage/rocksdb/ha_rocksdb.cc
sunshine-Chun: Checked with Rocksdb team. The trash_dir has actually been deprecated. It's the old way of managing trash files. Nowadays Rocksdb team manage them by renaming the sst files. I will pass nullptr to trash dir and set the delete_existing_trash to false since we are not using them.

laurynas-biveinis: > Checked with Rocksdb team. The trash_dir has actually been deprecated. It's the old way of managing trash files. Nowadays Rocksdb team manage them by renaming the sst files. I will pass nullptr to trash dir and set the delete_existing_trash to false since we are not using them.

I see, thank you. Should we convert the existing `rocksdb_init_internal` usage as well, in this or a follow-up PR? And can we reuse that same manager object?

laurynas-biveinis:
@sunshine-Chun , since the original PR has been merged already, I believe a new PR has to be made instead of pushing new commits to this one

sunshine-Chun:
create a new pr: Address comments for 'Apply sst file manager in myrocks clone' #1388

storage/rocksdb/ha_rocksdb.cc
sunshine-Chun: Yes we can convert the existing rocksdb_init_internal usage in a follow-up PR. 

 >And can we reuse that same manager object?

I am open for suggestion. My original thought is that we only have two places use sst_file_manager. Does it worth to maintain a static variable for the usage of these two places?

laurynas-biveinis: > I am open for suggestion. My original thought is that we only have two places use sst_file_manager. Does it worth to maintain a static variable for the usage of these two places?

I am not sure neither. Clone can run many times, and rotate the checkpoint many times too, so dynamically it could be much more than two places. Also the init options are repeated, and probably should be in sync after the `rocksdb_init_internal` follow-up. A static variable for this would be closely related to `rdb` static var, which looks fine to me.

laurynas-biveinis: Forgot one more thing, is reusing the same object even safe in the first place?

PR #1387, opened 2023-11-07
Add a testcase for MyRocks DDSE statistics write fix

laurynas-biveinis:
The fix was PR'ed in https://github.com/facebook/mysql-5.6/pull/1363 without a testcase.

Squash with f176e2b9abc84a605f137a65c55f997cf81ef67f


luqun:
nit: could you rename testcase to use "ddse_" instead of "ddse-", to align with MTR in PR1383

laurynas-biveinis:
@luqun , renamed & rebased

PR #1390, opened 2023-11-13
Upgrade all system tables(exclude mysql.gtid_executed) from INNODB to MyRocks

luqun:
Summary:

During mysqld version changes, it call upgrade_system_schemas() to upgrade SYSTEM tables. During DDSE change, It will also call upgrade_system_schemas() to upgrade these SYSTEM tables.

One question is how to "reliability" upgrade SYSTEM tables during DDSE change(upgrade SYSTEM tables occurs after DD_ENGINE properties has been changed to target DDSE):
- [Current] add a new persistent variable DD_ENGINE_UPGRADED in mysql.dd_properities table, whose value is last dd engine name during dd engine change. For example, when DDSE is changed from INNODB to ROCKSDB, the DD_ENGINE_UPGRADED value will be set to "INNODB" during DD Tables upgrade. Later, mysqld will check whether  DD_ENGINE_UPGRADED is same as target DDSE in no_server_upgrade_required(), if not, upgrade these SYSTEM tables to target DDSE if required.
- add a new global variables, such as DD_ENGINE_UPGRADING. During crash recovery, we need to re-initiate this values.

Details:
- add a new Properties DD_ENGINE_UPGRADED into DD_Properties table 
- mysql_system_tables_fix.sql: update system tables if necessary. 
- main.dd_schema_definition_after_upgrade_57_debug, main.dd_schema_definition_after_upgrade_80_debug: skip DD_ENGINE_UPGRADED compare, similar to DD_ENGINE
- rocksdb.rocksdb_read_free_rpl, rocksdb.singledelete: add delta for rocksdb stats due to system table involved during commit
- rocksdb.cancel_mc, rocksdb.rocksdb_fault_injection: force to use INNODB DDSE

The change has several assumptions:
- SYSTEM Tables and DD Tables always has same SE(DDSE).
- When force upgrade is specified(https://dev.mysql.com/doc/refman/8.0/en/upgrading-what-is-upgraded.html), it won't run ALTER SYSTEM tables to target DDSE again and it will run similar as mysqld version change

Test Plan:

Reviewers:

Subscribers:

Tasks:

Tags:

Differential Revision: https://phabricator.intern.facebook.com/D51085626

### DIFF
--- a/mysql-test/suite/rocksdb/t/cancel_mc-master.opt
+++ b/mysql-test/suite/rocksdb/t/cancel_mc-master.opt
@@ -1,1 +1,4 @@
---rocksdb_rate_limiter_bytes_per_sec=256k --rocksdb_default_cf_options=write_buffer_size=64k;target_file_size_base=64k;max_bytes_for_level_base=1m;compression_per_level=kNoCompression;max_compaction_bytes=256k
+# Specify DDSE as INNODB, since during DDSE upgrade system table,
+# with MC is enabled, the instance startup will timeout
+--initialize
+--rocksdb_rate_limiter_bytes_per_sec=256k --rocksdb_default_cf_options=write_buffer_size=64k;target_file_size_base=64k;max_bytes_for_level_base=1m;compression_per_level=kNoCompression;max_compaction_bytes=256k --default-dd-system-storage-engine=InnoDB
--- a/mysql-test/suite/rocksdb/t/compact_deletes2.test
+++ b/mysql-test/suite/rocksdb/t/compact_deletes2.test
@@ -1,4 +1,7 @@
 --source include/have_rocksdb.inc
+# current test expects no entry during compaction
+# with rocksdb DDSE, this couldn't happen
+--source include/have_innodb_ddse.inc
 
 --disable_warnings
 DROP TABLE IF EXISTS r1;
--- a/mysql-test/suite/rocksdb/t/rocksdb_fault_injection-master.opt
+++ b/mysql-test/suite/rocksdb/t/rocksdb_fault_injection-master.opt
@@ -1,1 +1,4 @@
+# Specify DDSE as INNODB, since during DDSE upgrade system table,
+# myrocks drop index thread will crash due to fault inject IO error
 --rocksdb_fault_injection_options="{"retry":true,"failure_ratio":1,"filetypes":["kTableFile"]}"
+--default-dd-system-storage-engine=InnoDB
--- a/scripts/mysql_system_tables_fix.sql
+++ b/scripts/mysql_system_tables_fix.sql
@@ -1164,6 +1164,158 @@ EXECUTE stmt;
 DROP PREPARE stmt;
 
 --
+-- alter table engine=DDSE if necessary by check actual engine
+--
+SET @component_engine = (select ENGINE from information_schema.tables where
+    table_schema='mysql' and table_name='component');
+SET @cmd = CONCAT("ALTER TABLE component ENGINE=",
+    IF(@ddse = 'ROCKSDB', "ROCKSDB", "InnoDB"));
+SET @str = IF(@ddse = @component_engine, "SET @dummy = 0", @cmd);
+PREPARE stmt FROM @str;
+EXECUTE stmt;
+DROP PREPARE stmt;
+
+SET @default_roles_engine = (select ENGINE from information_schema.tables where
+    table_schema='mysql' and table_name='default_roles');
+SET @cmd = CONCAT("ALTER TABLE default_roles ENGINE=",
+    IF(@ddse = 'ROCKSDB', "ROCKSDB", "InnoDB"));
+SET @str = IF(@ddse = @default_roles_engine, "SET @dummy = 0", @cmd);
+PREPARE stmt FROM @str;
+EXECUTE stmt;
+DROP PREPARE stmt;
+
+SET @engine_cost_engine = (select ENGINE from information_schema.tables where
+    table_schema='mysql' and table_name='engine_cost');
+SET @cmd = CONCAT("ALTER TABLE engine_cost ENGINE=",
+    IF(@ddse = 'ROCKSDB', "ROCKSDB", "InnoDB"));
+SET @str = IF(@ddse = @engine_cost_engine, "SET @dummy = 0", @cmd);
+PREPARE stmt FROM @str;
+EXECUTE stmt;
+DROP PREPARE stmt;
+
+SET @global_grants_engine = (select ENGINE from information_schema.tables where
+    table_schema='mysql' and table_name='global_grants');
+SET @cmd = CONCAT("ALTER TABLE global_grants ENGINE=",
+    IF(@ddse = 'ROCKSDB', "ROCKSDB", "InnoDB"));
+SET @str = IF(@ddse = @global_grants_engine, "SET @dummy = 0", @cmd);
+PREPARE stmt FROM @str;
+EXECUTE stmt;
+DROP PREPARE stmt;
+
+SET @help_topic_engine = (select ENGINE from information_schema.tables where
+    table_schema='mysql' and table_name='help_topic');
+SET @cmd = CONCAT("ALTER TABLE help_topic ENGINE=",
+    IF(@ddse = 'ROCKSDB', "ROCKSDB", "InnoDB"));
+SET @str = IF(@ddse = @help_topic_engine, "SET @dummy = 0", @cmd);
+PREPARE stmt FROM @str;
+EXECUTE stmt;
+DROP PREPARE stmt;
+
+SET @replication_asynchronous_connection_failover_engine = (select ENGINE from information_schema.tables where
+    table_schema='mysql' and table_name='replication_asynchronous_connection_failover');
+SET @cmd = CONCAT("ALTER TABLE replication_asynchronous_connection_failover ENGINE=",
+    IF(@ddse = 'ROCKSDB', "ROCKSDB", "InnoDB"));
+SET @str = IF(@ddse = @replication_asynchronous_connection_failover_engine, "SET @dummy = 0", @cmd);
+PREPARE stmt FROM @str;
+EXECUTE stmt;
+DROP PREPARE stmt;
+
+SET @replication_asynchronous_connection_failover_managed_engine = (select ENGINE from information_schema.tables where
+    table_schema='mysql' and table_name='replication_asynchronous_connection_failover_managed');
+SET @cmd = CONCAT("ALTER TABLE replication_asynchronous_connection_failover_managed ENGINE=",
+    IF(@ddse = 'ROCKSDB', "ROCKSDB", "InnoDB"));
+SET @str = IF(@ddse = @replication_asynchronous_connection_failover_managed_engine, "SET @dummy = 0", @cmd);
+PREPARE stmt FROM @str;
+EXECUTE stmt;
+DROP PREPARE stmt;
+
+SET @replication_group_configuration_version_engine = (select ENGINE from
+    information_schema.tables where table_schema='mysql' and
+    table_name='replication_group_configuration_version');
+SET @cmd = CONCAT("ALTER TABLE replication_group_configuration_version ENGINE=",
+    IF(@ddse = 'ROCKSDB', "ROCKSDB", "InnoDB"));
+SET @str = IF(@ddse = @replication_group_configuration_version_engine,
+    "SET @dummy = 0", @cmd);
+PREPARE stmt FROM @str;
+EXECUTE stmt;
+DROP PREPARE stmt;
+
+SET @replication_group_member_actions_engine = (select ENGINE from information_schema.tables where
+    table_schema='mysql' and table_name='replication_group_member_actions');
+SET @cmd = CONCAT("ALTER TABLE replication_group_member_actions ENGINE=",
+    IF(@ddse = 'ROCKSDB', "ROCKSDB", "InnoDB"));
+SET @str = IF(@ddse = @replication_group_member_actions_engine, "SET @dummy = 0", @cmd);
+PREPARE stmt FROM @str;
+EXECUTE stmt;
+DROP PREPARE stmt;
+
+SET @role_edges_engine = (select ENGINE from information_schema.tables where
+    table_schema='mysql' and table_name='role_edges');
+SET @cmd = CONCAT("ALTER TABLE role_edges ENGINE=",
+    IF(@ddse = 'ROCKSDB', "ROCKSDB", "InnoDB"));
+SET @str = IF(@ddse = @role_edges_engine, "SET @dummy = 0", @cmd);
+PREPARE stmt FROM @str;
+EXECUTE stmt;
+DROP PREPARE stmt;
+SET @server_cost_engine = (select ENGINE from information_schema.tables where
+    table_schema='mysql' and table_name='server_cost');
+SET @cmd = CONCAT("ALTER TABLE server_cost ENGINE=",
+    IF(@ddse = 'ROCKSDB', "ROCKSDB", "InnoDB"));
+SET @str = IF(@ddse = @server_cost_engine, "SET @dummy = 0", @cmd);
+PREPARE stmt FROM @str;
+EXECUTE stmt;
+DROP PREPARE stmt;
+
+SET @slave_master_info_engine = (select ENGINE from information_schema.tables where
+    table_schema='mysql' and table_name='slave_master_info');
+SET @cmd = CONCAT("ALTER TABLE slave_master_info ENGINE=",
+    IF(@ddse = 'ROCKSDB', "ROCKSDB", "InnoDB"));
+SET @str = IF(@ddse = @slave_master_info_engine, "SET @dummy = 0", @cmd);
+PREPARE stmt FROM @str;
+EXECUTE stmt;
+DROP PREPARE stmt;
+
+SET @slave_relay_log_info_engine = (select ENGINE from information_schema.tables where
+    table_schema='mysql' and table_name='slave_relay_log_info');
+SET @cmd = CONCAT("ALTER TABLE slave_relay_log_info ENGINE=",
+    IF(@ddse = 'ROCKSDB', "ROCKSDB", "InnoDB"));
+SET @str = IF(@ddse = @slave_relay_log_info_engine, "SET @dummy = 0", @cmd);
+PREPARE stmt FROM @str;
+EXECUTE stmt;
+DROP PREPARE stmt;
+
+SET @slave_worker_info_engine = (select ENGINE from information_schema.tables where
+    table_schema='mysql' and table_name='slave_worker_info');
+SET @cmd = CONCAT("ALTER TABLE slave_worker_info ENGINE=",
+    IF(@ddse = 'ROCKSDB', "ROCKSDB", "InnoDB"));
+SET @str = IF(@ddse = @slave_worker_info_engine, "SET @dummy = 0", @cmd);
+PREPARE stmt FROM @str;
+EXECUTE stmt;
+DROP PREPARE stmt;
+
+#
+# Drop DESC due to MyRocks not supporting DESC yet
+#
+SET @password_history_engine = (select ENGINE from information_schema.tables
+    where table_schema='mysql' and table_name='password_history');
+SET @cmd = CONCAT("ALTER TABLE password_history DROP PRIMARY KEY,",
+    IF(@ddse = 'ROCKSDB', "ADD PRIMARY KEY(Host, User, Password_timestamp)",
+        "ADD PRIMARY KEY(Host, User, Password_timestamp DESC)"));
+SET @str = IF(@ddse = @password_history_engine, "SET @dummy = 0", @cmd);
+PREPARE stmt FROM @str;
+EXECUTE stmt;
+DROP PREPARE stmt;
+
+SET @password_history_engine = (select ENGINE from information_schema.tables
+    where table_schema='mysql' and table_name='password_history');
+SET @cmd = CONCAT("ALTER TABLE password_history ENGINE=",
+    IF(@ddse = 'ROCKSDB', "ROCKSDB", "InnoDB"));
+SET @str = IF(@ddse = @password_history_engine, "SET @dummy = 0", @cmd);
+PREPARE stmt FROM @str;
+EXECUTE stmt;
+DROP PREPARE stmt;
+
+--
 -- CREATE_ROLE_ACL and DROP_ROLE_ACL
 --
 ALTER TABLE user ADD Create_role_priv enum('N','Y') COLLATE utf8mb3_general_ci DEFAULT 'N' NOT NULL AFTER account_locked;
--- a/share/messages_to_clients.txt
+++ b/share/messages_to_clients.txt
@@ -10402,6 +10402,10 @@ ER_RDB_TTL_WRITES_WITH_STALE_SNAPSHOT
 OBSOLETE_ER_RDB_DDSE_CANNOT_DISABLE_LARGE_PREFIX
   eng "Can't disable rocksdb_large_prefix when MyRocks is the MySQL data dictionary engine"
 
+ER_ALTER_SYSTEM_TABLE_WITH_NOT_DDSE
+  eng "System table [%s.%s] couldn't use specified engine %s. System tables should always use default-dd-system-storage-engine."
+
+
 #
 # End of 8.0 FB MySQL error messages.
 # (Please read comments from the header of this section before adding error
--- a/share/messages_to_error_log.txt
+++ b/share/messages_to_error_log.txt
@@ -12093,6 +12093,12 @@ ER_UNSUPPORTED_DD_ENGINE
 ER_CANT_CREATE_DUMP_THREAD
   eng "Can't create thread for DUMP TABLE (errno= %d)"
 
+ER_SYSTEM_TABLE_SERVER_UPGRADE_STATUS
+  eng "Server upgrade from '%s' to '%s' %s."
+
+ER_DD_SYSTEM_ENGINE_UPGRADED_NOT_FOUND
+  eng "The DD_SYSTEM_ENGINE_UPGRADED property for the data dictionary was not found. Starting the server using DD_ENGINE='%s'."
+
 # DO NOT add server-to-client messages here;
 # they go in messages_to_clients.txt
 # in the same directory as this file.
--- a/sql/dd/impl/bootstrap/bootstrap_ctx.h
+++ b/sql/dd/impl/bootstrap/bootstrap_ctx.h
@@ -110,6 +110,7 @@ class DD_bootstrap_ctx {
   // default_dd_system_storage_engine
   legacy_db_type m_actual_dd_engine = DB_TYPE_UNKNOWN;
   legacy_db_type m_did_dd_engine_upgrade_from = DB_TYPE_UNKNOWN;
+  legacy_db_type m_upgraded_dd_system_engine = DB_TYPE_UNKNOWN;
 
  public:
   DD_bootstrap_ctx() = default;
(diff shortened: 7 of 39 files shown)


sql/dd/impl/upgrade/server.cc
```diff
@@ -518,6 +518,48 @@ bool upgrade_help_tables(THD *thd) {
...
+  // iterate SYSTEM tables and run alter engine=target ddse
+  for (System_tables::Const_iterator it =
+           System_tables::instance()->begin(System_tables::Types::SYSTEM);
+       it != System_tables::instance()->end();
+       it = System_tables::instance()->next(it, System_tables::Types::SYSTEM)) {
+    assert((*it)->property() == System_tables::Types::SYSTEM);
+
+    dd::String_type table_name = (*it)->key().first + "." + (*it)->key().second;
+
+    if (table_name == "mysql.gtid_executed" ||
```
laurynas-biveinis: Need a comment here. It might make sense to store these names in some structure or array, depending on its size

sql/dd/impl/upgrade/server.cc
```diff
@@ -518,6 +518,48 @@ bool upgrade_help_tables(THD *thd) {
...
+      continue;
+
+    dd::String_type alter_cmd =
+        "alter table " + table_name + " engine=" + get_dd_engine_name();
+    Ed_connection con(thd);
+    LEX_STRING str;
+    lex_string_strmake(thd->mem_root, &str, alter_cmd.c_str(),
+                       alter_cmd.length());
+
+    // ignore ER_NO_SUCH_TABLE due to some tables in System_tables mayn't exists
```
laurynas-biveinis: As discussed offline, we need to have a list of the tables that are allowed to be missing and allow `ER_NO_SUCH_TABLE` only for them

sql/dd/impl/upgrade/server.cc
```diff
@@ -839,16 +881,44 @@ bool upgrade_system_schemas(THD *thd) {
         !exists_version)
       return true;
 
-  MySQL_check check;
+  uint dd_engine = DB_TYPE_UNKNOWN;
```
laurynas-biveinis:
```suggestion
  legacy_db_type dd_engine = DB_TYPE_UNKNOWN;
```

sql/dd/impl/upgrade/server.cc
```diff
@@ -839,16 +881,44 @@ bool upgrade_system_schemas(THD *thd) {
...
+      return true;
+
+  // force upgrade or either version changed or dd engine is changed
+  // when force upgrade is specified, do code as version changed
+  assert(is_force_upgrade() || ((server_version == MYSQL_VERSION_ID) ^
+                                (dd_engine == get_dd_engine_type())));
+
+  if (dd_engine != get_dd_engine_type()) {
+    LogErr(SYSTEM_LEVEL, ER_SYSTEM_TABLE_SERVER_UPGRADE_STATUS,
+           get_dd_engine_name((legacy_db_type)dd_engine), get_dd_engine_name(),
```
laurynas-biveinis:
```suggestion
           get_dd_engine_name(dd_engine), get_dd_engine_name(),
```

storage/rocksdb/rdb_native_dd.cc
```diff
@@ -70,8 +70,9 @@ bool rocksdb_dict_set_server_version() {
       ->set_server_version();
 }
 
-bool rocksdb_is_supported_system_table(const char *, const char *, bool) {
-  return false;
+bool rocksdb_is_supported_system_table(const char *, const char *,
```
laurynas-biveinis: This function, unfortunately, has to go back to `ha_rocksdb.cc`, because it's not related to DD tables at all

laurynas-biveinis (commented):
Can you explain why exactly is the new `DD_PROPERTIES` entry required?
Would the following work without it?
- Handle DD tables first, as usual
- Go through all the system tables, if engine != dd engine, do `ALTER TABLE`

To me this looks crash-safe too, unless I'm missing something. One thing that would not be supported is manual `ALTER TABLE mysql.sys_table ENGINE=not RocksDB`, because such changes would be reverted on the next restart, but is this something that needs to be supported in the first place?

luqun:
> Can you explain why exactly is the new `DD_PROPERTIES` entry required? Would the following work without it?
> 
> * Handle DD tables first, as usual
> * Go through all the system tables, if engine != dd engine, do `ALTER TABLE`
> 
> To me this looks crash-safe too, unless I'm missing something.

Yes, it is crash-safe and also quick restart.  For quick restart, in order to detect whether we need to run alter tables during restart, either check all SYSTEM tables or add a mark  to record state

laurynas-biveinis:
> Yes, it is crash-safe and also quick restart. For quick restart, in order to detect whether we need to run alter tables during restart, either check all SYSTEM tables or add a mark to record state

Checking this property for SYSTEM tables should be cheap (they will be in the DD cache). I am a bit worried that a new `DD_PROPERTIES` will complicate things (i.e. have to delete it on MyRocks -> InnoDB DDSE change), but if you prefer this approach, it's OK.

luqun:
>> (i.e. have to delete it on MyRocks -> InnoDB DDSE change)

I think we can keep these variables even for MyRocks->INNODB DDSE change. I agree that remove these variables will be tricky during that DDSE change.

luqun:
@laurynas-biveinis , would you take a look again? just rebase with your latest change. also use script to upgrade these system tables.

laurynas-biveinis:
I believe we have to fully think through user-issued `ALTER TABLE mysql.systable
ENGINE=anything` support.

With your approach, or with previously-suggested my one, or, I believe, with any
approach in the server, any `ALTER TABLE` that changes engine away from the
`default_dd_system_storage_engine` will be either ignored (good for user who
issued it but bad for the idea behind this PR), either reverted (good for this
PR, bad for the user).

I see two ways out of this:
- In the server, document that `default_dd_system_storage_engine` applies to new
  instances only, handle any InnoDB <-> MyRocks changing at user level,
  effectively dropping this PR
- In the server, block `ALTER TABLE mysql.* ENGINE = not-default-dd-system-one`.
  This should be done in this PR too then.

What do you think?

mysql-test/suite/rocksdb/t/compact_deletes2.test
```diff
@@ -1,4 +1,7 @@
 --source include/have_rocksdb.inc
+# current test expect no entry during compaction
```
laurynas-biveinis: expects

mysql-test/suite/rocksdb/t/compact_deletes2.test
```diff
@@ -1,4 +1,7 @@
...
+# with rocksdb DDSE, this cound't happen
```
laurynas-biveinis: couldn't

mysql-test/suite/rocksdb/t/rocksdb_fault_injection-master.opt
```diff
@@ -1 +1,3 @@
---rocksdb_fault_injection_options="{"retry":true,"failure_ratio":1,"filetypes":["kTableFile"]}"
+# Specify DDSE as INNODB, since during DDSE upgrade system table,
+# myrocks drop index thread will crash due to fault inject IO error
+--rocksdb_fault_injection_options="{"retry":true,"failure_ratio":1,"filetypes":["kTableFile"]}" --default-dd-system-storage-engine=InnoDB
```
laurynas-biveinis: This needs `--initialize` too in order to have the described effect (otherwise
MTR re-uses the default test instance, which gets changed on the test startup)

scripts/mysql_system_tables_fix.sql
```diff
@@ -1040,28 +1053,73 @@ ALTER TABLE help_keyword CONVERT TO CHARACTER SET utf8mb3;
...
+
 SET @cmd = CONCAT("ALTER TABLE help_relation ENGINE=",
     IF(@ddse = 'ROCKSDB', "ROCKSDB", "InnoDB"),
     " STATS_PERSISTENT=0");
 PREPARE stmt FROM @cmd;
 EXECUTE stmt;
 DROP PREPARE stmt;
 
-SET @cmd = CONCAT("ALTER TABLE help_keyword ENGINE=",
+SET @cmd = CONCAT("ALTER TABLE help_topic ENGINE=",
```
laurynas-biveinis: Intentional?

scripts/mysql_system_tables_fix.sql
```diff
@@ -1040,28 +1053,73 @@ ALTER TABLE help_keyword CONVERT TO CHARACTER SET utf8mb3;
...
+
+SET @engine_cost_engine = (select ENGINE from information_schema.tables where
+    table_schema='mysql' and table_name='engine_cost');
+SET @cmd = CONCAT("ALTER TABLE engine_cost ENGINE=",
+    IF(@ddse = 'ROCKSDB', "ROCKSDB","InnoDB"));
+SET @str = IF(@ddse = @engine_cost_engine, "SET @dummy = 0", @cmd);
+PREPARE stmt FROM @str;
+EXECUTE stmt;
+DROP PREPARE stmt;
+
```
laurynas-biveinis: This and the other chunks below mix the engine change with the existing `ALTER
TABLE ENGINE=...` statements for other tables. I don't think this is right,
because the purpose and the engine choice logic is different. The existing
script looks at the DDSE sysvar, and the added logic looks at the actual engine.

I believe the existing section should be left as-is, and a new section should be
introduced, not sure if above or below this one, and it should also have `ALTER
TABLE` for the tables mentioned in this existing section.

sql/dd/impl/bootstrap/bootstrapper.cc
```diff
@@ -1162,9 +1162,15 @@ bool create_dd_schema(THD *thd) {
...
   The only difference between "actual" and "target" is SE.
+
+ @param [IN]  thd  THD Handle
+ @param [OUT] dd_properties_engine which SE stores mysql.dd_properties table
+
+ @return mysql.dd_proerpties table definition
 */
 [[nodiscard]] const Object_table_definition *get_dd_properties_table_definition(
-    THD *thd) {
+    THD *thd, legacy_db_type *dd_properties_engine) {
```
laurynas-biveinis: nit: I'd just use a reference

sql/dd/impl/bootstrap/bootstrapper.cc
```diff
@@ -1176,9 +1182,13 @@ bool create_dd_schema(THD *thd) {
           rocksdb_hton, thd, MYSQL_SCHEMA_NAME.str,
           dd::tables::DD_properties::instance().name().c_str());
     }
+
+    if (error == HA_ERR_TABLE_EXIST)
```
laurynas-biveinis: nit: ternary op?

sql/dd/impl/bootstrap/bootstrapper.cc
```diff
@@ -1162,9 +1162,15 @@ bool create_dd_schema(THD *thd) {
   For upgrade, "actual" is gotten by checking table_exists_in_engine and then
     compute.
   The only difference between "actual" and "target" is SE.
+
+ @param [IN]  thd  THD Handle
+ @param [OUT] dd_properties_engine which SE stores mysql.dd_properties table
+
+ @return mysql.dd_proerpties table definition
```
laurynas-biveinis: properties

sql/dd/impl/bootstrap/bootstrap_ctx.h
```diff
@@ -110,6 +110,7 @@ class DD_bootstrap_ctx {
   // default_dd_system_storage_engine
   legacy_db_type m_actual_dd_engine = DB_TYPE_UNKNOWN;
   legacy_db_type m_did_dd_engine_upgrade_from = DB_TYPE_UNKNOWN;
+  legacy_db_type m_upgraded_dd_engine = DB_TYPE_UNKNOWN;
```
laurynas-biveinis: Might be misleading not to have system tables in the variable name, as DD tables
and system tables are different things. Unfortunately, this will have to churn in
neighbor variables too. `m_upgrade_dd_system_engine`?

Likewise for the methods below

sql/dd/impl/bootstrap/bootstrapper.cc
```diff
@@ -1342,6 +1356,16 @@ bool initialize_dd_properties(THD *thd) {
     }
     bootstrap::DD_bootstrap_ctx::instance().set_actual_dd_engine(
         static_cast<legacy_db_type>(actual_dd_engine));
+
+    legacy_db_type upgraded_dd_engine = DB_TYPE_UNKNOWN;
```
laurynas-biveinis: Likewise

sql/dd/impl/bootstrap/bootstrapper.cc
```diff
@@ -1342,6 +1356,16 @@ bool initialize_dd_properties(THD *thd) {
...
+    bool exists_upgraded_dd_engine = false;
+    if (dd::tables::DD_properties::instance().get(thd, "DD_ENGINE_UPGRADED",
```
laurynas-biveinis: Likewise for the property name

sql/dd/impl/bootstrap/bootstrapper.cc
```diff
@@ -1342,6 +1356,16 @@ bool initialize_dd_properties(THD *thd) {
...
+                                                  (uint *)&upgraded_dd_engine,
+                                                  &exists_upgraded_dd_engine) ||
+        !exists_upgraded_dd_engine)
+      upgraded_dd_engine = (legacy_db_type)actual_dd_engine;
```
laurynas-biveinis: nit: C++ style cast

sql/dd/impl/bootstrap/bootstrapper.cc
```diff
@@ -2000,6 +2030,18 @@ bool update_versions(THD *thd, bool is_dd_upgrade_57) {
                                                   get_dd_engine_type()))
       return dd::end_transaction(thd, true);
 
+    bool exists_upgraded_dd_engine = false;
+    if (dd::tables::DD_properties::instance().get(thd, "DD_ENGINE_UPGRADED",
+                                                  (uint *)&upgraded_dd_engine,
```
laurynas-biveinis: nit: C++ style cast

sql/dd/impl/upgrade/server.cc
```diff
@@ -876,10 +876,32 @@ bool upgrade_system_schemas(THD *thd) {
         !exists_version)
       return true;
 
+  uint dd_engine = DB_TYPE_UNKNOWN;
```
laurynas-biveinis: Likewise

Also, what happened with the previous comment re. `legacy_db_type`, removing the
cast below?

sql/dd/impl/upgrade/server.cc
```diff
@@ -876,10 +876,32 @@ bool upgrade_system_schemas(THD *thd) {
...
+
+  // force upgrade or either version changed or dd engine is changed
+  // when force upgrade is specified, do code as version changed
+  assert(is_force_upgrade() || ((server_version == MYSQL_VERSION_ID) ^
+                                (dd_engine == get_dd_engine_type())));
+
   MySQL_check check;
+  if (dd_engine != get_dd_engine_type()) {
+    LogErr(SYSTEM_LEVEL, ER_SYSTEM_TABLE_SERVER_UPGRADE_STATUS,
+           get_dd_engine_name((legacy_db_type)dd_engine), get_dd_engine_name(),
```
laurynas-biveinis: nit: if the cast will still be required, C++-style cast

share/messages_to_clients.txt
```diff
@@ -10402,6 +10402,9 @@ ER_RDB_TTL_WRITES_WITH_STALE_SNAPSHOT
 OBSOLETE_ER_RDB_DDSE_CANNOT_DISABLE_LARGE_PREFIX
   eng "Can't disable rocksdb_large_prefix when MyRocks is the MySQL data dictionary engine"
 
+ER_ALTER_TABLE_FAILED_DURING_SYSTEM_TABLE_UPGRADE
```
laurynas-biveinis: Where is it used?

share/messages_to_error_log.txt
```diff
@@ -12093,6 +12093,9 @@ ER_UNSUPPORTED_DD_ENGINE
 ER_CANT_CREATE_DUMP_THREAD
   eng "Can't create thread for DUMP TABLE (errno= %d)"
 
+ER_SYSTEM_TABLE_SERVER_UPGRADE_STATUS
+  eng "Server upgrade from '%s' to '%s' %s."
```
laurynas-biveinis, line 12097: A colon before the final %s?

luqun:
stlill working on..

scripts/mysql_system_tables_fix.sql
luqun: Move all of these into its own section.

scripts/mysql_system_tables_fix.sql
luqun: Revert this.

luqun:
> What do you think?

Discuss offline and we believe we should disable "alter table system tables into non-DDSE" for users to keep simple. All DD Tables and SYSTEMS Tables always use DDSE.

share/messages_to_error_log.txt
luqun: I will skip this. since another one for version upgrade also doesn't contains colon..

luqun:
@laurynas-biveinis , now it is ready for review.. :)

mysql-test/suite/rocksdb/t/rocksdb_fault_injection-master.opt
luqun: If I had added --initialize, the instance will fail to initialize rocksdb plugin due to fault injection.

laurynas-biveinis: Will that happen even if DDSE is InnoDB?

mysql-test/suite/rocksdb/r/ddse_alter_system_table.result
```diff
@@ -0,0 +1,64 @@
+ALTER TABLE mysql.columns_priv ENGINE="ENGINE";
```
laurynas-biveinis: Please use `sys_tables_` prefix for system table tests, and then the test name could be shortened, i.e. `sys_tables_alter`

mysql-test/suite/rocksdb/t/cancel_mc-master.opt
```diff
@@ -1 +1,3 @@
---rocksdb_rate_limiter_bytes_per_sec=256k --rocksdb_default_cf_options=write_buffer_size=64k;target_file_size_base=64k;max_bytes_for_level_base=1m;compression_per_level=kNoCompression;max_compaction_bytes=256k
+# Specify DDSE as INNODB, since during DDSE upgrade system table,
+# with MC is enabled, the instance startup will timeout
+--rocksdb_rate_limiter_bytes_per_sec=256k --rocksdb_default_cf_options=write_buffer_size=64k;target_file_size_base=64k;max_bytes_for_level_base=1m;compression_per_level=kNoCompression;max_compaction_bytes=256k --default-dd-system-storage-engine=InnoDB
```
laurynas-biveinis, line 4: `--initialize`?

mysql-test/suite/rocksdb/t/ddse_alter_system_table.test
```diff
@@ -0,0 +1,135 @@
+--let ddse=`SELECT UPPER(@@default_dd_system_storage_engine)`
+
+let engine = "INNODB";
+if($ddse == "INNODB")
+{
+  let engine = "ROCKSDB";
+}
+
+--replace_regex /innodb/ENGINE/i /rocksdb/ENGINE/i
+--ERROR ER_ALTER_SYSTEM_TABLE_WITH_NOT_DDSE
```
laurynas-biveinis: Please use lowercase for MTR commands to avoid confusion with SQL
```suggestion
--error ER_ALTER_SYSTEM_TABLE_WITH_NOT_DDSE
```

scripts/mysql_system_tables_fix.sql
```diff
@@ -189,6 +189,7 @@ alter table db comment='Database privileges';
 alter table user comment='Users and global privileges';
 alter table func comment='User defined functions';
 
+
```
laurynas-biveinis: Spurious

scripts/mysql_system_tables_fix.sql
```diff
@@ -1163,6 +1164,158 @@ PREPARE stmt FROM @cmd;
 EXECUTE stmt;
 DROP PREPARE stmt;
 
+--
+-- alter table engine=DDSE if necessary by check actual engine
+--
+SET @component_engine = (select ENGINE from information_schema.tables where
+    table_schema='mysql' and table_name='component');
+SET @cmd = CONCAT("ALTER TABLE component ENGINE=",
+    IF(@ddse = 'ROCKSDB', "ROCKSDB","InnoDB"));
```
laurynas-biveinis:
```suggestion
    IF(@ddse = 'ROCKSDB', "ROCKSDB", "InnoDB"));
```

scripts/mysql_system_tables_fix.sql
```diff
@@ -1163,6 +1164,158 @@ PREPARE stmt FROM @cmd;
...
+    table_schema='mysql' and table_name='slave_worker_info');
+SET @cmd = CONCAT("ALTER TABLE slave_worker_info ENGINE=",
+    IF(@ddse = 'ROCKSDB', "ROCKSDB","InnoDB"));
+SET @str = IF(@ddse = @slave_worker_info_engine, "SET @dummy = 0", @cmd);
+PREPARE stmt FROM @str;
+EXECUTE stmt;
+DROP PREPARE stmt;
+
+#
+# DROP DESC due to ROCKSDB doesn't support DESC YET
```
laurynas-biveinis:
```suggestion
# Drop DESC due to MyRocks not supporting DESC yet
```

share/messages_to_clients.txt
```diff
@@ -10402,6 +10402,10 @@ ER_RDB_TTL_WRITES_WITH_STALE_SNAPSHOT
 OBSOLETE_ER_RDB_DDSE_CANNOT_DISABLE_LARGE_PREFIX
   eng "Can't disable rocksdb_large_prefix when MyRocks is the MySQL data dictionary engine"
 
+ER_ALTER_SYSTEM_TABLE_WITH_NOT_DDSE
+  eng "SYSTEM TABLE [%s.%s] couldn't use specified engine %s. SYSTEM TABLES should always use DDSE."
```
laurynas-biveinis:
```suggestion
  eng "System table [%s.%s] couldn't use specified engine %s. System tables should always use default-dd-system-storage-engine."
```

sql/dd/impl/bootstrap/bootstrapper.cc
```diff
@@ -1176,9 +1182,11 @@ bool create_dd_schema(THD *thd) {
           rocksdb_hton, thd, MYSQL_SCHEMA_NAME.str,
           dd::tables::DD_properties::instance().name().c_str());
     }
+
+    (error == HA_ERR_TABLE_EXIST) ? (dd_properties_engine = DB_TYPE_ROCKSDB)
```
laurynas-biveinis: `dd_properties = (error == HA_ERR_TABLE_EXIST) ? DB_TYPE_ROCKSDB : DB_TYPE_INNODB;`

sql/dd/impl/bootstrap/bootstrapper.cc
```diff
@@ -1319,10 +1331,11 @@ bool initialize_dd_properties(THD *thd) {
       unsupported DD_ENGINE; use INNODB if couldn't find DD_ENGINE, since
       existing INNODB DDSE instance doesn't contain DD_ENGINE property
     */
-    uint actual_dd_engine = DB_TYPE_UNKNOWN;
+    legacy_db_type actual_dd_engine = DB_TYPE_UNKNOWN;
     exists = false;
-    if (dd::tables::DD_properties::instance().get(thd, "DD_ENGINE",
-                                                  &actual_dd_engine, &exists) ||
+    if (dd::tables::DD_properties::instance().get(
+            thd, "DD_ENGINE", reinterpret_cast<uint *>(&actual_dd_engine),
```
laurynas-biveinis, line 1337: `reinterpret_cast` looks suspicious for an enum (with underlying `int`) to `uint`. Does a `static_cast` result in a compilation error?

sql/dd/impl/bootstrap/bootstrapper.cc
```diff
@@ -1341,7 +1354,25 @@ bool initialize_dd_properties(THD *thd) {
...
+
+    /*
+      Read DD_SYSTEM_ENGINE_UPGRADED from dd_properties, if not exist,
+      use INNODB instead;
+    */
+    legacy_db_type upgraded_dd_system_engine = DB_TYPE_UNKNOWN;
+    bool exists_upgraded_dd_system_engine = false;
+    if (dd::tables::DD_properties::instance().get(
+            thd, "DD_SYSTEM_ENGINE_UPGRADED",
+            reinterpret_cast<uint *>(&upgraded_dd_system_engine),
```
laurynas-biveinis, line 1367: Likewise

sql/dd/impl/bootstrap/bootstrapper.cc
```diff
@@ -1341,7 +1354,25 @@ bool initialize_dd_properties(THD *thd) {
       return true;
     }
     bootstrap::DD_bootstrap_ctx::instance().set_actual_dd_engine(
-        static_cast<legacy_db_type>(actual_dd_engine));
+        actual_dd_engine);
+
+    /*
+      Read DD_SYSTEM_ENGINE_UPGRADED from dd_properties, if not exist,
```
```suggestion
      Read DD_SYSTEM_ENGINE_UPGRADED from dd_properties, if it does not exist,
```

sql/sql_table.cc
```diff
@@ -19004,6 +19005,20 @@ static bool check_engine(THD *thd, const char *db_name, const char *table_name,
     return true;
   }
 
+  /*
+    check new system table storage engine is same as
```
laurynas-biveinis:
```suggestion
    Check that the new system table storage engine is same as
```

sql/dd/impl/bootstrap/bootstrapper.cc
luqun: yep. it complains Static_cast from 'legacy_db_type *' to 'uint *' (aka 'unsigned int *') is not allowed.

sql/dd/impl/bootstrap/bootstrapper.cc
luqun: same as before, static_cast doesn't allow this.

luqun:
Resolved all comments and also add a new variables skip-sys-tables-engine-check, since some MTRs alter system table to myiasm engine.

laurynas-biveinis:
> also add a new variables skip-sys-tables-engine-check, since some MTRs alter system table to myiasm engine.

Since we will never be changing system table engine to MyISAM, can we just disable those testcases instead?

luqun:
> > also add a new variables skip-sys-tables-engine-check, since some MTRs alter system table to myiasm engine.
> 
> Since we will never be changing system table engine to MyISAM, can we just disable those testcases instead?

Problem is that these testcase also test other scenarios in same time..maybe a new global variable is better..

laurynas-biveinis:
luqun ***@***.***> writes:
> yep. it complains Static_cast from 'legacy_db_type *' to 'uint *' (aka 'unsigned int *') is not allowed.

Right, it's pointers. OK to leave as-is.

laurynas-biveinis:
luqun ***@***.***> writes:

>  also add a new variables skip-sys-tables-engine-check, since some MTRs alter system table to myiasm engine.
>
>  Since we will never be changing system table engine to MyISAM, can we just disable those testcases instead?
>
> Problem is that these testcase also test other scenarios in same time..maybe a new global variable is better..

Can you list those tests?

If they happen to be debug-only, then we can piggyback on
`+d,skip_dd_table_access_check`. The debug injection name is not a
perfect match, but it would suffice IMHO.

luqun:
> Can you list those tests? If they happen to be debug-only, then we can piggyback on `+d,skip_dd_table_access_check`. The debug injection name is not a perfect match, but it would suffice IMHO.
There is only 1 debug only test.. 

mysql-test/suite/perfschema/t/error_log.test
mysql-test/t/system_tables_myisam.test
mysql-test/t/system_tables_myisam_lctn_1.test
mysql-test/t/transactional_acl_tables.test
mysql-test/suite/auth_sec/t/atomic_rename_user.test
mysql-test/t/dd_schema_dd_properties_debug.test

laurynas-biveinis:
> mysql-test/suite/perfschema/t/error_log.test
> mysql-test/t/system_tables_myisam.test
> mysql-test/t/system_tables_myisam_lctn_1.test
> mysql-test/t/transactional_acl_tables.test
> mysql-test/suite/auth_sec/t/atomic_rename_user.test
> mysql-test/t/dd_schema_dd_properties_debug.test

I see. It would require some effort to split these tests into MyISAM and
non-MyISAM parts, feel free to go ahead with the new sysvar or split the
tests as you see best

PR #1391, opened 2023-11-13
Remove Rdb_transaction::m_ddl_transaction

laurynas-biveinis:
With DD tables in MyRocks, the property of whether it is safe to skip uniqueness checks on writes should be resolved on a table, not transaction level. Thus fold the logic of setting m_ddl_transaction to true into ha_rocksdb::skip_unique. At the same time this flag is duplicated between struct update_row_info and callstack of affected functions, thus remove it from the callstack.

### DIFF
--- a/storage/rocksdb/ha_rocksdb.cc
+++ b/storage/rocksdb/ha_rocksdb.cc
@@ -3521,8 +3521,6 @@ class Rdb_transaction {
   */
   ulonglong m_writes_at_last_savepoint;
 
-  bool m_ddl_transaction = false;
-
  protected:
   THD *m_thd = nullptr;
 
@@ -3569,8 +3567,6 @@ class Rdb_transaction {
     m_auto_incr_map.clear();
     return s;
   }
-
-  void reset_ddl_transaction() { m_ddl_transaction = false; }
 
  protected:
   /*
@@ -3745,13 +3741,6 @@ class Rdb_transaction {
       set_lock_timeout(m_timeout_sec, table_type);
     }
   }
-
-  void set_ddl_transaction() {
-    assert(!is_ac_nl_ro_rc_transaction());
-    m_ddl_transaction = true;
-  }
-
-  [[nodiscard]] bool is_ddl_transaction() const { return m_ddl_transaction; }
 
   virtual void set_lock_timeout(int timeout_sec_arg, TABLE_TYPE table_type) = 0;
 
@@ -5029,7 +5018,6 @@ class Rdb_transaction_impl : public Rdb_transaction {
     m_delete_count = 0;
     m_row_lock_count = 0;
     m_auto_incr_map.clear();
-    reset_ddl_transaction();
     if (m_rocksdb_tx[TABLE_TYPE::USER_TABLE]) {
       release_snapshot(TABLE_TYPE::USER_TABLE);
       /* This will also release all of the locks: */
@@ -5324,7 +5312,6 @@ class Rdb_transaction_impl : public Rdb_transaction {
 
       set_initial_savepoint();
 
-      reset_ddl_transaction();
       m_is_delayed_snapshot = false;
     }
   }
@@ -5445,7 +5432,6 @@ class Rdb_writebatch_impl : public Rdb_transaction {
     m_read_opts[USER_TABLE] = rocksdb::ReadOptions();
     m_read_opts[USER_TABLE].ignore_range_deletions =
         !rocksdb_enable_delete_range_for_drop_index;
-    reset_ddl_transaction();
   }
 
  private:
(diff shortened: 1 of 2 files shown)


laurynas-biveinis:
This is done in preparation for always flushing DD transactions: it will require a new flag `m_dd_transaction` or similar in `Rdb_transaction`, which  would have too similar a name yet completely different functionality from `m_ddl_transaction`. Thus removing the latter first.

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -11832,12 +11818,19 @@ bool ha_rocksdb::skip_unique_check() const {
...
-  return THDVAR(table->in_use, bulk_load) ||
-         (my_core::thd_test_options(table->in_use,
-                                    OPTION_RELAXED_UNIQUE_CHECKS) &&
+  auto *const thd = table->in_use;
+  return THDVAR(thd, bulk_load) ||
+         (my_core::thd_test_options(thd, OPTION_RELAXED_UNIQUE_CHECKS) &&
           m_tbl_def->m_key_count == 1) ||
-         use_read_free_rpl();
+      use_read_free_rpl() ||
+      ((table_share->table_category == TABLE_CATEGORY_TEMPORARY ||
```
luqun: looks like it isn't easy to distinguish between user temp table and intrinsic temp table..  if it is easy, maybe user temp table could skip unique check but intrinsic temp table still use unique check

laurynas-biveinis: By code comments the user temp tables are `TABLE_CATEGORY_USER` and internal temp tables are `TABLE_CATEGORY_TEMPORAY`. So it's easy to tell temp tables apart, but harder to tell user regular and temporary tables apart.

To address your comment, I believe it is enough to remove `TABLE_CATEGORY_TEMPORARY`.

laurynas-biveinis:
@luqun , removed `TABLE_CATEGORY_TEMPORARY` and rebased; ready for review

storage/rocksdb/ha_rocksdb.cc
luqun: sadly, that code comments maybe is misunderstanding or incorrect. Just create a temporarily table and debugger shows that user temp tables are TABLE_CATEGORY_TEMPORAY

Callstack:
init_tmp_table_share // share->table_category = TABLE_CATEGORY_TEMPORARY;
ha_create_table
rea_create_tmp_table
create_table_impl
mysql_create_table_no_lock
mysql_create_table
Sql_cmd_create_table::execute

luqun: anyway, remove TABLE_CATEGORY_TEMPORARY looks good to me.

luqun:
looks like rocksdb.rqg_runtime MTR failed. still investigating

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -11831,12 +11817,18 @@ bool ha_rocksdb::skip_unique_check() const {
          any indexes. If the table has secondary keys, then those might becomes
          inconsisted/corrupted
       3) We're using read-free replication
+      4) For CREATE INDEX, DROP INDEX, & ALTER TABLE when any duplicate key
```
luqun, line 11820: For DDL, we still need unique check(see another comment with examples) but we can use writebatch->put instead of trx->put for perf win.

maybe keep ha_rocksdb::skip_unique_check() as it is. and rename is_ddl_transaction() to use_write_batch_api?

luqun:
> looks like rocksdb.rqg_runtime MTR failed. still investigating

The issue is related to DDL skip uniq check, such as following code. Add PK should fail due to duplicate col2 value.
```
create table t1(col1 int, col2 int, col3 varchar(16));
insert into t1 values(0,0,'0');
insert into t1 values(1,0,'0');
ALTER TABLE t1 ADD KEY ( col1 );
ALTER TABLE t1 ADD PRIMARY KEY ( col2 );
check table t1;
```

storage/rocksdb/ha_rocksdb.cc
laurynas-biveinis: The use of writebatch needs more consideration: 1) it effectively forces 1PC protocol; 2) how do locks not being taken fare if i.e. concurrent writes are present?

I have a vague TODO item "use writebatch more", but I'd prefer to address it separately

luqun: yes.. please address it separately...

laurynas-biveinis:
The DDL transaction flag cannot be folded into unique check flag, the original design idea was broken. Thinking of alternatives

laurynas-biveinis:
The DD durability PR is https://github.com/facebook/mysql-5.6/pull/1403. I will scavenge non-functional `skip_unique_check` cleanups from this PR

laurynas-biveinis:
The cleanups are in https://github.com/facebook/mysql-5.6/pull/1404

PR #1393, opened 2023-11-15
Test updates for MyRocks DDSE, part 5

laurynas-biveinis:
- Update main.dd_schema_definition_debug to work with either DDSE, record its MyRocks version as rocksdb.ddse_dd_schema_definition_debug.
- Add MyRocks DDSE version of main.dd_bootstrap_debug as rocksdb.ddse_bootstrap_debug.
- Convert a couple more "--skip-rocksdb" to "--loose-skip-rocksdb". This fixes main.mysql_upgrade_block_user_table failure with MyRocks DDSE and partially fixes main.log_buffered-big.
- Remove .rocksdb from innodb.bootstrap test data directory so that listing the directory gives consistent results across DDSEs.
- For perfschema.short_option_1 and short_option_2, do not set -a option which sets SERIALIZABLE transaction isolation, preventing MTR pre-test checks from running.
- Skip upgrade-related tests when MyRocks is the DDSE, because upgrade and DDSE change in a single step is not supported.


laurynas-biveinis:
Rebased on 8.0.32

PR #1394, opened 2023-11-20
MTR updates for MyRocks DDSE, part 6

laurynas-biveinis:
Update the InnoDB portable tablespace testcases. Make sure to start new server instances with the same DDSE the test runs under, and copy the MyRocks data when it carries the DD. Skip the portability_tablespace_linux test because it relies on a pre-packaged datadir that was not prepared with MyRocks DDSE.


laurynas-biveinis:
Rebased on 8.0.32. The "This branch has conflicts that must be resolved" message appears to be bogus

PR #1399, opened 2023-11-28
Update MyRocks MTR tests for MyRocks DDSE

laurynas-biveinis:
Summary:
- Stabilize rocksdb.alter_table_commit_inplace by adding some ORDER BYs.
- Skip rocksdb_sys_vars.rocksdb_large_prefix_basic and rocksdb_sys_vars.rocksdb_write_batch_max_bytes_basic with MyRocks DDSE because of DDSE requirements.

The rocksdb_sys_vars.rocksdb_large_prefix_basic will go away in 8.0.32, but until it does, it might as well be fixed.


luqun:
maybe we don't need this change anymore, since the change checked in 8.0.28 has been ported to 8.0.32..

PR #1400, opened 2023-11-30
Minor MyRocks clone code cleanups

laurynas-biveinis:
- Fix the name of an InnoDB define to force MyRocks clone rollback in a comment
- Remove always-false bool argument `empty_allowed` from `move_temp_dir_to_destination` function.


luqun:
LGTM.. committed

PR #1407, opened 2023-12-14
Support mysql.db system table in MyRocks engine

laurynas-biveinis:
- Allow mysql.db as a system table in MyRocks storage engine.
- Allow updates to ACL tables (which mysql.db is considered to be one of) in non-ROW binlog formats.
- Implement WL#14087 (Avoid locking reads from ACL tables) by honoring HA_EXTRA_NO_READ_LOCKING handler flag. It is done by a new handler flag, m_no_read_locking, which is checked in ha_rocksdb::external_lock together with the table category to downgrade RocksDB read locks to no locking at all.
- Adjust system tables script to create / alter mysql.db in the right engine depending on the default_dd_system_storage_engine setting.
- Add MTR regression tests to rocksdb suite with a sys_tables prefix.
- Adjust several tests to run with either system table engine.
- New MTR include file have_innodb_system_tables.inc, annotate several tests with it that have unmasked output difference depending on the system table storage engine.


luqun:
Thanks for adding WL#14087.
do you happen to know why innodb doesn't need is_acl_table(table)? such as

if (m_prebuilt->table->is_dd_table || m_prebuilt->no_read_locking ||  is_acl_table(table)) {

https://github.com/facebook/mysql-5.6/blob/fb-mysql-8.0.32/storage/innobase/handler/ha_innodb.cc#L19054

laurynas-biveinis:
> do you happen to know why innodb doesn't need is_acl_table(table)? such as
> 
> `if (m_prebuilt->table->is_dd_table || m_prebuilt->no_read_locking || is_acl_table(table)) {`
> 
> https://github.com/facebook/mysql-5.6/blob/fb-mysql-8.0.32/storage/innobase/handler/ha_innodb.cc#L19054

ACL tables get `extra(HA_EXTRA_NO_READ_LOCKING)` called beforehand, this gets checked in the assert right above those lines: `ut_ad(!m_prebuilt->no_read_locking || m_prebuilt->table->is_dd_table || is_acl_table(table));`

I believe MyRocks logic here is an exact copy but written differently, MyRocks does not do the equivalent of `is_acl_table` at this point neither, except in an assert

mysql-test/t/schema_read_only_ci.test
```diff
@@ -1,3 +1,4 @@
+# --source include/have_innodb_system_tables.inc
```
luqun: nit:  remove this comment?

mysql-test/suite/auth_sec/include/acl_tables_row_locking_test.inc
```diff
@@ -0,0 +1,721 @@
...
+{
+let query_row_locks= SELECT CONCAT(OBJECT_SCHEMA,'.',OBJECT_NAME)
+  tab,INDEX_NAME,LOCK_MODE, 'Acquired row lock on u2/u2x' AS Has_lock
+  FROM performance_schema.data_locks
+  WHERE LOCK_TYPE='RECORD' AND
+  (LOCATE('u2',LOCK_DATA)>0 OR LOCATE('u2x',LOCK_DATA)>0);
+}
+if ($SYSSE == "RocksDB")
+{
+let query_row_locks= SELECT `KEY`, MODE FROM INFORMATION_SCHEMA.ROCKSDB_LOCKS
```
luqun: looks like output 'KEY' may cause testcase flicky due to index id maybe different.

 SELECT `KEY`, MODE FROM INFORMATION_SCHEMA.ROCKSDB_LOCKS
 WHERE COLUMN_FAMILY_ID=0;
 KEY	MODE
-000001040000000000000001	X
+0000010b0000000000000001	X

laurynas-biveinis: > looks like output 'KEY' may cause testcase flicky due to index id maybe different.
> 
> SELECT `KEY`, MODE FROM INFORMATION_SCHEMA.ROCKSDB_LOCKS WHERE COLUMN_FAMILY_ID=0; KEY MODE -000001040000000000000001 X +0000010b0000000000000001 X

@luqun, is the bit format documented? Can I find which bits to mask?

laurynas-biveinis:
Rebased, addressed the first comment, the 2nd comment still open

mysql-test/suite/auth_sec/include/acl_tables_row_locking_test.inc
luqun: the first 4 bytes in these key is myrocks key index id --- different key has different index id to identity in myrocks.
such as for "000001040000000000000001", its key index id(in myrocks) is 0x00000104.  When a new table is created, all key will be assigned a  index id value and whose value is mono-increased.

For this case, we can't always assume that when the user_copy table is created at exactly order.  sometime, other testcass may executed before this testcase and will take these low key index id value.

maybe mark first 4 bytes?

laurynas-biveinis:
@luqun , updated with index ID masking, ready for review

PR #1408, opened 2023-12-14
Fix GCC compilation error in rocksdb_dict_recover

laurynas-biveinis:
storage/rocksdb/rdb_native_dd.cc: In function ‘bool myrocks::rocksdb_dict_recover(dict_recovery_mode_t, uint)’:
storage/rocksdb/rdb_native_dd.cc:105:1: error: control reaches end of non-void function [-Werror=return-type]

### DIFF
--- a/storage/rocksdb/rdb_native_dd.cc
+++ b/storage/rocksdb/rdb_native_dd.cc
@@ -102,6 +102,8 @@ bool rocksdb_dict_recover(dict_recovery_mode_t dict_recovery_mode, uint) {
     case DICT_RECOVERY_RESTART_SERVER:
       return false;
   }
+  MY_ASSERT_UNREACHABLE();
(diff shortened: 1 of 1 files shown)


storage/rocksdb/rdb_native_dd.cc
```diff
@@ -102,6 +102,7 @@ bool rocksdb_dict_recover(dict_recovery_mode_t dict_recovery_mode, uint) {
     case DICT_RECOVERY_RESTART_SERVER:
       return false;
   }
+  MY_ASSERT_UNREACHABLE();
```
luqun, line 105: could you also add a return true?

laurynas-biveinis: @luqun , done

PR #1410, opened 2023-12-14
Pass actual format string instead of nullptr in Parser_oom_handler::h…

laurynas-biveinis:
…andle_condition

Should the code path be actually exercised, memory errors were likely to have occurred instead of diagnostics.

This fixes a GCC build error:

sql/sql_parse.cc:7909:17: error: too many arguments for format [-Werror=format-extra-args]
 7909 |                 static_cast<ulonglong>(thd->variables.parser_max_mem_size),

Squash with 67fcc60a43687c3e116c7f57e13b52732ad25231

### DIFF
--- a/sql/sql_parse.cc
+++ b/sql/sql_parse.cc
@@ -7905,7 +7905,8 @@ class Parser_oom_handler : public Internal_error_handler {
           if (thd->variables.parser_exceeded_max_mem_capacity_action ==
               PARSER_EXCEEDED_MAX_MEM_CAPACITY_ACTION_WARN) {
             push_warning_printf(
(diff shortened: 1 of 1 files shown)


luqun:
From code(https://github.com/facebook/mysql-5.6/blob/fb-mysql-8.0.32/sql/sql_error.cc#L731), passing nullptr should be same as passing ER_THD(thd, ER_CAPACITY_EXCEEDED).

laurynas-biveinis:
> From code(https://github.com/facebook/mysql-5.6/blob/fb-mysql-8.0.32/sql/sql_error.cc#L731), passing nullptr should be same as passing ER_THD(thd, ER_CAPACITY_EXCEEDED).

I see, then `format(printf(...))` attribute annotation is not entirely correct. I still propose my patch as a GCC warning silencer

PR #1412, opened 2023-12-15
Fix build error on std::pair comparison

laurynas-biveinis:
This fixes

sql/binlog.cc:6626:52: error: invalid operands to binary expression ('std::pair<int64_t, int64_t>' (aka 'pair<long long, long long>') and 'std::pair<long, long>' (aka 'pair<long, long>'))
        if (raft_rotate_info->ingestion_checkpoint != std::pair(-1L, -1L)) {
            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ ^  ~~~~~~~~~~~~~~~~~~~

and

sql/binlog.cc:8809:30: error: invalid operands to binary expression ('std::pair<int64_t, int64_t>' (aka 'pair<long long, long long>') and 'std::pair<long, long>' (aka 'pair<long, long>'))
        if (param.checkpoint != std::pair(-1L, -1L)) {
            ~~~~~~~~~~~~~~~~ ^  ~~~~~~~~~~~~~~~~~~~

Done in a separate commit from 2ad105fc364672361a6228238dd9523c835bf399 because of a different squash patch target.

Squash with e0c8fde56ec9777307d609a69acb5c5a92c21907

### DIFF
--- a/sql/binlog.cc
+++ b/sql/binlog.cc
@@ -6623,7 +6623,8 @@ bool MYSQL_BIN_LOG::open_binlog(
         update_prev_gtid_and_opid(
             previous_logged_gtids, raft_rotate_info->rotate_opid.first,
             raft_rotate_info->rotate_opid.second, need_sid_lock);
-        if (raft_rotate_info->ingestion_checkpoint != std::pair(-1L, -1L)) {
+        if (raft_rotate_info->ingestion_checkpoint !=
+            std::pair<int64_t, int64_t>(-1, -1)) {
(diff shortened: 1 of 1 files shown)


luqun:
The only difference is squash commit? maybe I can force update 8.0.28 branch..

laurynas-biveinis:
> The only difference is squash commit? maybe I can force update 8.0.28 branch..

Yes, this one squashes to a different commit. Not sure how your trunk branches work, but a force push to a trunk is a bad idea in general case at least

PR #1416, opened 2023-12-19
Add extra asserts to prevent incorrect moved-from Rdb_tbl_def uses

laurynas-biveinis:
Rdb_tbl_def class has a move constructor-like two-arg constructor that takes the ownership of passed object's m_key_descr_arr field. Signify this moved-from state in the object by resetting m_pk_index-which is a related-field to MAX_INDEXES + 1, and add extra asserts that these and other related fields are not accessed in moved-from objects.

At the same time remove redundant "explicit" specifiers from two-argument constructors and clean them up to use initializer lists consistently.


laurynas-biveinis:
Those incorrect moved-from uses I had seen in passing during MyRocks DDSE debugging

PR #1420, opened 2024-01-11
apply slowrm to innodb clone recovery

sunshine-Chun:
Innodb clone doesn't apply slowrm when discarding files during recovery, which will potentially lock up I/O. Apply the existing slowrm feature to innodb clone from this PR: https://github.com/facebook/mysql-5.6/pull/1099?fbclid=IwAR25_2tPoiZsh0QTkyUzKnH5QgLk7Rl8dpFIG4kePgupYDkgXZfbGNhu_VA


laurynas-biveinis:
LGTM

PR #1424, opened 2024-01-22
Convert mysql.user system table to MyRocks

laurynas-biveinis:
- rocksdb_is_supported_system_table: allow mysql.user in MyRocks
- scripts/mysql_system_tables.sql and mysql_system_tables_fix.sql: use MyRocks as the engine for mysql.user depending on the default-dd-system-storage-engine setting. At the same time pull up "ROW_FORMAT=DYNAMIC" to the SE-independent part of CREATE TABLE mysql.db text too.
- Re-record affected MyRocks system tables tests.
- Update acl_tables_row_locking_test.inc to conditionally skip FK-requiring test parts.
- Fix several pre-existing regressions from running the default testsuite with --mysqld=--default-dd-system-storage-engine=ROCKSDB testing. The fixes include passing the right sysvar value to manually-started test instances, and conditionally disabling InnoDB DDSE-specific tests.


luqun:
Looks like this PR has a conflict with your previous PR for following files..
mysql-test/suite/rocksdb/r/sys_tables_mysql_db.result

laurynas-biveinis:
Interesting, I'm pretty sure I branched after the previous PR was merged. In any case, rebased

luqun:
one typo cause auth_sec.acl_tables_row_locking failure. after fixing typo, the rocksdb.sys_tables_acl_tables_row_locking start to fail


[ 50%] rocksdb.sys_tables_acl_tables_row_locking 'innodb_ddse'  [ fail ]
        Test ended at 2024-01-27 23:27:32

CURRENT_TEST: rocksdb.sys_tables_acl_tables_row_locking
mysqltest: At line 579: Query 'REPLACE INTO mysql.user SELECT * FROM user_bk' succeeded, should have failed with error '1205'
In included file mysql-test/suite/auth_sec/include/acl_tables_row_locking_test.inc: 705
included from mysql-test/suite/rocksdb/t/sys_tables_acl_tables_row_locking.test: 5

The result from queries just before the failure was:
DELETE FROM mysql.user WHERE User='u2';
ERROR HY000: Lock wait timeout exceeded; try restarting transaction: Timeout on index: mysql.user.PRIMARY
ROLLBACK;

mysql-test/suite/auth_sec/include/acl_tables_row_locking_test.inc
```diff
@@ -555,6 +557,7 @@ while($c <= 4)
   --source include/wait_innodb_all_purged.inc
   connection update_con2;
 
+  if ($suppoted_isolation) {
```
luqun: typo.. supported_isolation

laurynas-biveinis:
@luqun , ready for review.

Differences from the previous version:
- `ha_rocksdb::store_lock`: rearrange local variable declarations and split a long if
  condition to make it easier to understand
- `ha_rocksdb::external_lock`: only downgrade DD table locks to none after no read
  locking was requested. System and ACL tables are handled in `store_lock`, and
  `external_lock` should not change their locking, add a comment to clarify that.
- `auth_sec/include/acl_tables_row_locking_test.inc`:
  - all the tables are created in the default engine instead of InnoDB.
  - added `ORDER BY` to the MyRocks lock info query in attempt to stabilize test output
  - disabled case 16.2 under MyRocks that previously was silently ignored due to a
    typo, because the locks taken are too different from InnoDB for the test to work
  - set `rocksdb_lock_wait_timeout` to a low value to make test run faster because it
    is testing expected lock timeouts.

luqun:
committed

PR #1427, opened 2024-02-05
Convert MyRocks clone and associated code to use std::string_view

laurynas-biveinis:
std::string_view has several advantages over std::string char char *:
- it's a non-owning view that cheaply converts from std::string and converts from char * with a cost of a strlen call (which the patch takes care to remove where possible)
- is more efficient (ones less level of indirection to string contents, no SSO checking) than std::string while supporting similar (read-only) API
- it's cheaper to pass std::string_view by value than std::string by const reference;
- in some cases it avoids the cost of constructing new std::strings.

In converting clone code, also convert some utility functions that also called from elsewhere. In particular, convert rocksdb_datadir and rocksdb_wal_dir sysvars to be std::string_views, keeping the pointer-to-char variables only for interfacing with the MYSQL_SYSVAR macros.

### DIFF
--- a/storage/rocksdb/clone/client.cc
+++ b/storage/rocksdb/clone/client.cc
@@ -19,6 +19,7 @@
 #include <sys/types.h>
 #include <cstdint>
 #include <forward_list>
+#include <string_view>
 #include <unordered_map>
 
 #include "my_compiler.h"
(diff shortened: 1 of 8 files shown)


luqun:
Looks like there are some compilation error, maybe different gcc version?

storage/rocksdb/clone/common.h:403:12: error: call to non-‘constexpr’ function ‘std::__cxx11::basic_string<_CharT, _Traits, _Alloc>::operator __sv_type() const [with _CharT = char; _Traits = std::char_traits<char>; _Alloc = std::allocator<char>; __sv_type = std::basic_string_view<char>]’
  403 |     return m_name;
  
  storage/rocksdb/clone/donor.cc:90:12: error: call to non-‘constexpr’ function ‘std::__cxx11::basic_string<_CharT, _Traits, _Alloc>::operator __sv_type() const [with _CharT = char; _Traits = std::char_traits<char>; _Alloc = std::allocator<char>; __sv_type = std::basic_string_view<char>]’
   90 |     return m_dir;
   
   storage/rocksdb/clone/client.cc:119:12: error: call to non-‘constexpr’ function ‘std::__cxx11::basic_string<_CharT, _Traits, _Alloc>::operator __sv_type() const [with _CharT = char; _Traits = std::char_traits<char>; _Alloc = std::allocator<char>; __sv_type = std::basic_string_view<char>]’
  119 |     return m_name;

laurynas-biveinis:
@luqun, rebased and updated the PR, should be fixed now

laurynas-biveinis:
The PR passes calls data() on string_views and passes such pointers where null-terminated strings are expected. Since string_views are length-based, they are not guaranteed to be null-terminated. All string views are created from C or C++ strings here are, in fact, null terminated, but this is not a good practice to rely on that.

laurynas-biveinis:
Removed the incorrect uses of `string_view::data` in the contexts where null-terminated strings were expected. The patch is smaller now and ready for review

PR #1428, opened 2024-02-09
Support all of the system tables (except mysql.gtid_executed) in MyRocks

laurynas-biveinis:
- Move rocksdb_is_supported_system_table from rdb_native_dd.cc to ha_rocksdb.cc
  because system tables are not DD tables
- Dropped the ability to create the replication status tables in MyISAM
- Adapt main.wl14965, main.component_debug, & binlog.binlog_half_crash_safe_ddl tests
  to run with either system table engine
- Convert main.roles-upgrade to an include file, and create a MyRocks version of the
  test (rocksdb.sys_tables_roles-upgrade)
- Likewise for main.plugin_auth, whose MyRocks version is
  rocksdb.sys_tables_plugin_auth.
- Likewise for main.mysql_upgrade_slave_master_info, whose MyRocks version is
  rocksdb.sys_tables_mysql_upgrade_slave_master_info.
- Only run rocksdb_rpl.rpl_rocksdb_max_row_locks test with InnoDB as the sys table
  engine due to row locks being taken by the system tables, making their number in
  MyRocks less predictable. This might be adjusted later by increasing the row lock
  limit and the test workload size.
- Re-record rocksdb_dd_innodb.known_directory, rocksdb_dd_innodb.innodb_zip_8k,
  rocksdb_dd_innodb.innodb_tablespace, rocksdb_dd_innodb.create_tablespace_debug,
  rocksdb_dd_innodb.create_tablespace_64k, rocksdb_dd_innodb.create_tablespace_32k,
  rocksdb.sys_tables_mysqlcheck, rocksdb.sys_tables_is_statistics_mysql,
  rocksdb.sys_tables_acl_tables_row_locking, & rocksdb.sys_tables tests.


laurynas-biveinis:
@luqun , this is a draft PR before I leave for the PTO. There are test failures which I haven't analyzed yet: rocksdb_dd_innodb.virtual_basic
rocksdb_clone.remote_dml_gtid
rocksdb.sys_tables_transactional_acl_tables

laurynas-biveinis:
@luqun , ready for review

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -6034,6 +6036,45 @@ static bool rocksdb_user_table_blocked(legacy_db_type db_type) {
   return db_type == DB_TYPE_INNODB;
 }
 
+static bool rocksdb_is_supported_system_table(const char *db_name,
```
luqun, line 6039: how about these "missing" system table(such as mysql.temp_user), still use innodb instead?

scripts/mysql_system_tables_fix.sql
```diff
@@ -1028,23 +1035,87 @@ ALTER TABLE help_relation CONVERT TO CHARACTER SET utf8mb3;
...
-ALTER TABLE help_relation ENGINE=InnoDB STATS_PERSISTENT=0;
-ALTER TABLE help_keyword ENGINE=InnoDB STATS_PERSISTENT=0;
-ALTER TABLE plugin ENGINE=InnoDB STATS_PERSISTENT=0;
-ALTER TABLE servers ENGINE=InnoDB STATS_PERSISTENT=0;
-ALTER TABLE time_zone ENGINE=InnoDB STATS_PERSISTENT=0;
-ALTER TABLE time_zone_leap_second ENGINE=InnoDB STATS_PERSISTENT=0;
-ALTER TABLE time_zone_name ENGINE=InnoDB STATS_PERSISTENT=0;
-ALTER TABLE time_zone_transition ENGINE=InnoDB STATS_PERSISTENT=0;
-ALTER TABLE time_zone_transition_type ENGINE=InnoDB STATS_PERSISTENT=0;
+SET @cmd = CONCAT("ALTER TABLE func ENGINE=",
```
luqun, line 1043: why missing some of system tables  to change SE? such as mysql.component, mysql.default_roles..?

luqun:
also what's difference between rocksdb MTR test suites and rocksdb_dd_innodb MTR test suites? Previous I think rocksdb_dd_innodb MTR test suites is for innodb testcases runs under rocksdb DD, but now looks like some MTRs under rocksdb test suites also has similar pattern.

storage/rocksdb/ha_rocksdb.cc
laurynas-biveinis: Any table not listed here will use InnoDB, but, unless something fell through the cracks, that should be only the Enterprise Edition tables (MEB, tp plugin), which should not apply. What is the `mysql.temp_user` table?

scripts/mysql_system_tables_fix.sql
laurynas-biveinis: I believe this section deals with the old instance upgrade where those tables have to be converted from MyISAM. For our use case, the InnoDB -> MyRocks conversion is not handled here.

laurynas-biveinis:
> also what's difference between rocksdb MTR test suites and rocksdb_dd_innodb MTR test suites? Previous I think rocksdb_dd_innodb MTR test suites is for innodb testcases runs under rocksdb DD, 

This is correct.

>but now looks like some MTRs under rocksdb test suites also has similar pattern.

Which ones?

scripts/mysql_system_tables_fix.sql
luqun: okay.. sounds good.. I plan to also use this script(instead of code) to handle InnoDB -> MyRocks system tables. I am updating my PR based on your current PR.

luqun:
> but now looks like some MTRs under rocksdb test suites also has similar pattern.

>> Which ones?

such as mysql-test/suite/rocksdb/t/sys_tables_mysql_upgrade_slave_master_info.test

storage/rocksdb/ha_rocksdb.cc
luqun: > What is the mysql.temp_user table?

https://github.com/facebook/mysql-5.6/blob/fb-mysql-8.0.32/sql/dd/impl/system_registry.cc#L254

The reason I had asked this is because the list is different from https://github.com/facebook/mysql-5.6/pull/1390.. I am changing my PR to also use exactly same list..

luqun: just rebased my PR based on this PR(will rebase again after this PR is checked in)

https://github.com/facebook/mysql-5.6/pull/1390/commits/7db177936be7af489b530c758f9ae6650cb3376a#diff-773930212c131b69bdf9e5944d12142364419730bb32c19a1266dedd77aa72f8

laurynas-biveinis:
> > but now looks like some MTRs under rocksdb test suites also has similar pattern.
> 
> > > Which ones?
> 
> such as mysql-test/suite/rocksdb/t/sys_tables_mysql_upgrade_slave_master_info.test

This is not an InnoDB-specific test, rather it was one of the general tests that regressed with a different DDSE

storage/rocksdb/ha_rocksdb.cc
laurynas-biveinis: > > What is the mysql.temp_user table?
> 
> https://github.com/facebook/mysql-5.6/blob/fb-mysql-8.0.32/sql/dd/impl/system_registry.cc#L254

Oh, I see. As far I can tell these are temporary test tables created by MTR tests in mysql schema. I think it's fine to convert them once i.e. InnoDB is disabled, because at that point we'll be able to confirm that there are indeed no other unconverted system tables left, and the conversion of these should be trivial. If I'm right that they are temporary then there's no need to handle their upgrade neither.

luqun: yep. sounds good. we can convert them during disable innodb time

mysql-test/suite/binlog/t/binlog_half_crash_safe_ddl.test
```diff
@@ -50,7 +50,14 @@
 --source include/have_binlog_format_mixed.inc
 
 --disable_query_log
+--let $SYSSE=`SELECT @@default_dd_system_storage_engine`
+if ($SYSSE == "RocksDB") {
+  SET SESSION binlog_format=ROW;
+}
 call mtr.add_suppression("Could not find a valid tablespace file for");
+if ($SYSSE == "RocksDB") {
```
luqun, line 58: nit: looks like duplicate as line 54-line 56

luqun: anyway, commit first

laurynas-biveinis: Will do a follow-up PR to fix

PR #1429, opened 2024-02-09
update main.mdl_sync for myrocks data dictioanry due to miss xa_detach_on_prepare support

luqun:
Summary:
main.mdl_sync will crash(read after free) while running under myrocks data dictionary. 

The root issue is that myrocks doesn't support xa_detach_on_prepare while the default value for variable xa_detach_on_prepare is true ---- SQL layer think myrocks support xa_detach_on_prepare(due to variable value) while myrocks SE didn't implement related handler functions. This mismatch cause rocksdb_commit_by_xid free trx but THD still has a pointer to trx.

Temporarily set xa_detach_on_prepare =0 while adding xa_detach_on_prepare support in myrocks

Test Plan:

./mtr main.mdl_sync --mysqld=--default-dd-system-storage-engine=ROCKSDB  --repeat 10
./mtr main.mdl_sync --repeat 10

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/mysql-test/suite/rocksdb/r/ddse_mdl_sync.result
+++ b/mysql-test/suite/rocksdb/r/ddse_mdl_sync.result
@@ -1,0 +1,4056 @@
+SET DEBUG_SYNC= 'RESET';
+drop table if exists t1,t2,t3;
+create table t1 (i int);
+create table t2 (i int);
+connection: default
+lock tables t2 read;
+connection: con1
+set debug_sync='mdl_upgrade_lock SIGNAL parked WAIT_FOR go';
+alter table t1 rename t3;
+connection: default
+set debug_sync= 'now WAIT_FOR parked';
+connection: con2
+set debug_sync='mdl_acquire_lock_wait SIGNAL go';
+drop table t1,t2;
+connection: con1
+connection: default
+unlock tables;
+connection: con2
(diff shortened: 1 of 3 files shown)


laurynas-biveinis:
@luqun, I believe we still have to address the crasher in the code level in the current interaction between MyRocks and `xa_detach_on_prepare`, if viable. If we are not implementing the missing handlerton methods in the short-term, then is it possible to hack something force old behavior `xa_detach_on_prepare` = 0 if the current transaction has MyRocks as one of its engines? Or would that be worse than the current situation?

In any case, please add a new testcase to rocksdb suite, i.e. `ddse_mdl_sync`, which just has one line to include `main.mdl_sync` and has the necessary `-master.opt` which is the sum of `--initialize --default-dd-system-storage-engine=ROCKSDB`  and the existing `mdl_sync-master.opt` contents

luqun:
@laurynas-biveinis , we are adding missing handlerton methods.. I will add another testcase into rocksdb suites instead.

luqun:
added a new MTR and force pushed

laurynas-biveinis:
LGTM, thank you! Let's ensure that removing `--xa_detach_on_prepare=off` from
the option file is not forgotten once the handlerton methods are implemented.

PR #1185, opened 2022-02-14
Range Locking support, MyRocks part

spetrunia:
(Range Locking pull request,filed against fb-mysql-8.0.23)

This adds a my.cnf parameter, rocksdb_use_range_locking.

When it is ON, MyRocks will:
- initialize RocksDB to use range-locking lock manager
- for all DML operations (including SELECT .. FOR UPDATE) will lock
   the scanned range before reading/modifying rows.
- In range locking mode, there is no snapshot checking (cannot do that
  for ranges). Instead, MyRocks will read and modify latest committed
  data, just like InnoDB does (in the code, grep for (start|end)_ignore_
  snapshot)
- Queries that do not have a finite range to scan, like
    UPDATE t1 .... ORDER BY t1.key LIMIT n
  will use a  "Locking iterator" which will read rows, lock the range,
  and re-read the rows. See class LockingIterator.

PR #1430, opened 2024-02-21
[mysql-5.6][PR] Range Locking support, MyRocks part, updated

hermanlee:
Summary:
(Range Locking pull request,filed against fb-mysql-8.0.23)

This adds a my.cnf parameter, rocksdb_use_range_locking.

When it is ON, MyRocks will:
- initialize RocksDB to use range-locking lock manager
- for all DML operations (including SELECT .. FOR UPDATE) will lock the scanned range before reading/modifying rows.
- In range locking mode, there is no snapshot checking (cannot do that for ranges). Instead, MyRocks will read and modify latest committed data, just like InnoDB does (in the code, grep for (start|end)_ignore_ snapshot)
- Queries that do not have a finite range to scan, like UPDATE t1 .... ORDER BY t1.key LIMIT n will use a  "Locking iterator" which will read rows, lock the range, and re-read the rows. See class LockingIterator.

Pull Request resolved: https://github.com/facebook/mysql-5.6/pull/1185
GitHub Author: Sergei Petrunia <sergey@mariadb.com>

### DIFF
--- a/storage/rocksdb/ha_rocksdb.cc
+++ b/storage/rocksdb/ha_rocksdb.cc
@@ -104,6 +104,8 @@
 #include "./ObjectFactory.h"
 #endif
 
+#include "./rdb_locking_iter.h"
+
 // Internal MySQL APIs not exposed in any header.
 extern "C" {
 /**
@@ -667,6 +669,9 @@ static void rocksdb_set_delayed_write_rate(THD *thd, struct SYS_VAR *var,
 static void rocksdb_set_max_latest_deadlocks(THD *thd, struct SYS_VAR *var,
                                              void *var_ptr, const void *save);
 
+static void rocksdb_set_max_lock_memory(THD *thd, struct SYS_VAR *var,
+                                        void *var_ptr, const void *save);
+
 static void rdb_set_collation_exception_list(const char *exception_list);
 static void rocksdb_set_collation_exception_list(THD *thd, struct SYS_VAR *var,
                                                  void *var_ptr,
@@ -822,6 +827,16 @@ static bool rocksdb_alter_table_comment_inplace = false;
 static bool rocksdb_disable_instant_ddl = false;
 bool rocksdb_enable_tmp_table = false;
 bool rocksdb_enable_delete_range_for_drop_index = false;
+
+// Range Locking: how much memory can be used used for the lock data structure
+//  (which hold the locks acquired by all clients).
+static ulonglong rocksdb_max_lock_memory;
+
+bool rocksdb_use_range_locking = 0;
+static bool rocksdb_use_range_lock_manager_as_point = 0;
+std::shared_ptr<rocksdb::RangeLockManagerHandle> range_lock_mgr;
+
+std::shared_ptr<rocksdb::RangeLockManagerHandle> range_lock_mgr_used_as_point;
 
 static std::atomic<uint64_t> rocksdb_row_lock_deadlocks(0);
 static std::atomic<uint64_t> rocksdb_row_lock_wait_timeouts(0);
@@ -1523,6 +1538,13 @@ static MYSQL_SYSVAR_UINT(max_latest_deadlocks, rocksdb_max_latest_deadlocks,
                          "deadlocks to store",
                          nullptr, rocksdb_set_max_latest_deadlocks,
                          rocksdb::kInitialMaxDeadlocks, 0, UINT32_MAX, 0);
+
+static MYSQL_SYSVAR_ULONGLONG(
+    max_lock_memory, rocksdb_max_lock_memory, PLUGIN_VAR_RQCMDARG,
+    "Range-locking mode: Maximum amount of memory "
+    "that locks from all transactions can use at a time",
+    nullptr, rocksdb_set_max_lock_memory,
+    /*initial*/ 1073741824, 0, UINT64_MAX, 0);
 
 static MYSQL_SYSVAR_ENUM(
     info_log_level, rocksdb_info_log_level, PLUGIN_VAR_RQCMDARG,
@@ -2365,6 +2387,17 @@ static MYSQL_SYSVAR_BOOL(table_stats_use_table_scan,
                          rocksdb_update_table_stats_use_table_scan,
                          rocksdb_table_stats_use_table_scan);
 
+static MYSQL_SYSVAR_BOOL(use_range_locking, rocksdb_use_range_locking,
+                         PLUGIN_VAR_OPCMDARG | PLUGIN_VAR_READONLY,
+                         "Use Range Locking", nullptr, nullptr,
+                         rocksdb_use_range_locking);
+
+static MYSQL_SYSVAR_BOOL(use_range_lock_manager_as_point,
+                         rocksdb_use_range_lock_manager_as_point,
+                         PLUGIN_VAR_OPCMDARG | PLUGIN_VAR_READONLY,
+                         "Use Range Lock Manager as point", nullptr, nullptr,
+                         rocksdb_use_range_lock_manager_as_point);
+
 static MYSQL_SYSVAR_BOOL(
     large_prefix, rocksdb_large_prefix, PLUGIN_VAR_RQCMDARG,
     "Support large index prefix length of 3072 bytes. If off, the maximum "
@@ -2706,7 +2739,9 @@ static struct SYS_VAR *rocksdb_system_variables[] = {
     MYSQL_SYSVAR(manual_compaction_threads),
     MYSQL_SYSVAR(manual_compaction_bottommost_level),
     MYSQL_SYSVAR(rollback_on_timeout),
-
+    MYSQL_SYSVAR(use_range_locking),
+    MYSQL_SYSVAR(use_range_lock_manager_as_point),
+    MYSQL_SYSVAR(max_lock_memory),
     MYSQL_SYSVAR(enable_insert_with_update_caching),
     MYSQL_SYSVAR(trace_block_cache_access),
     MYSQL_SYSVAR(trace_queries),
@@ -2935,6 +2970,7 @@ class Rdb_transaction {
   ulonglong m_update_count = 0;
   ulonglong m_delete_count = 0;
   // per row data
+  // (with range locking, each locked range is counted here, too)
   ulonglong m_row_lock_count = 0;
   std::unordered_map<GL_INDEX_ID, ulonglong> m_auto_incr_map;
 
@@ -3008,8 +3044,50 @@ class Rdb_transaction {
   virtual rocksdb::Status do_pop_savepoint() = 0;
   virtual void do_rollback_to_savepoint() = 0;
 
+ private:
+  /*
+    If true, the current statement should not use a snapshot for reading.
+    Note that in a multi-statement transaction, the snapshot may have been
+    allocated by another statement.
+  */
+  bool m_stmt_ignores_snapshot = false;
+
+  /* Snapshot-ignore mode will put away m_reads_opts.snapshot here: */
+  const rocksdb::Snapshot *m_saved_snapshot = nullptr;
+
  public:
+  void start_ignore_snapshot() {
+    // note: this may be called several times for the same statement
+    if (!m_stmt_ignores_snapshot) {
+      m_saved_snapshot = m_read_opts.snapshot;
+      m_read_opts.snapshot = nullptr;
+      m_stmt_ignores_snapshot = true;
+
+      // For repeatable-read AUTOCOMMIT statements on systems using range
+      // locking and ttl, there is a possibility no snapshot and
+      // m_snapshot_time assigned. Compaction relies on the oldest snapshot
+      // time found in rocksdb to determine what is visibile to transactions,
+      // so it's possible for compaction to remove rows that were already read
+      // by an autocomit statement (e.g. insert into .. select union select)
+      // since no snapshot was created to hold them.
+      //
+      // There will be further fixes for TTL which should resolve this case.
+      if (!m_snapshot_timestamp)
+        rdb->GetEnv()->GetCurrentTime(&m_snapshot_timestamp);
+    }
+  }
+
+  void end_ignore_snapshot_if_needed() {
+    if (m_stmt_ignores_snapshot) {
+      m_stmt_ignores_snapshot = false;
+      m_read_opts.snapshot = m_saved_snapshot;
+      m_saved_snapshot = nullptr;
+    }
+  }
+  bool in_snapshot_ignore_mode() const { return m_stmt_ignores_snapshot; }
+
   rocksdb::ReadOptions m_read_opts;
+
   const char *m_mysql_log_file_name;
   my_off_t m_mysql_log_offset;
   const char *m_mysql_gtid;
@@ -3190,6 +3268,18 @@ class Rdb_transaction {
   virtual void release_lock(const Rdb_key_def &key_descr,
                             const std::string &rowkey, bool force = false) = 0;
 
+  virtual rocksdb::Status lock_range(rocksdb::ColumnFamilyHandle *const cf,
+                                     const rocksdb::Endpoint &start,
+                                     const rocksdb::Endpoint &end) = 0;
+
+  rocksdb::Status lock_singlepoint_range(rocksdb::ColumnFamilyHandle *const cf,
+                                         const rocksdb::Slice &point) {
+    // Normally, one needs to "flip" the endpoint type for reverse-ordered CFs.
+    // But here we are locking just one point so this is not necessary.
+    rocksdb::Endpoint endp(point, false);
+    return lock_range(cf, endp, endp);
+  }
+
   virtual bool prepare() = 0;
 
   bool commit_or_rollback() {
@@ -3248,10 +3338,17 @@ class Rdb_transaction {
     m_is_delayed_snapshot = false;
   }
 
+  void locking_iter_created() {
+    if (!m_snapshot_timestamp)
+      rdb->GetEnv()->GetCurrentTime(&m_snapshot_timestamp);
+  }
+
   virtual void acquire_snapshot(bool acquire_now) = 0;
   virtual void release_snapshot() = 0;
 
-  bool has_snapshot() const { return m_read_opts.snapshot != nullptr; }
+  bool has_snapshot() const {
+    return m_read_opts.snapshot != nullptr || m_saved_snapshot;
+  }
 
  private:
   // The Rdb_sst_info structures we are currently loading.  In a partitioned
@@ -3901,7 +3998,9 @@ class Rdb_transaction {
 
   virtual rocksdb::Iterator *get_iterator(
       const rocksdb::ReadOptions &options,
-      rocksdb::ColumnFamilyHandle *column_family) = 0;
+      rocksdb::ColumnFamilyHandle *column_family,
+      const std::shared_ptr<Rdb_key_def> &,
+      bool use_locking_iterator = false) = 0;
 
   virtual void multi_get(rocksdb::ColumnFamilyHandle *const column_family,
                          const size_t num_keys, const rocksdb::Slice *keys,
@@ -3910,10 +4009,11 @@ class Rdb_transaction {
                          const bool sorted_input) const = 0;
 
   rocksdb::Iterator *get_iterator(
-      rocksdb::ColumnFamilyHandle *const column_family, bool skip_bloom_filter,
+      rocksdb::ColumnFamilyHandle *const column_family,
+      const std::shared_ptr<Rdb_key_def> &kd, bool skip_bloom_filter,
       const rocksdb::Slice &eq_cond_lower_bound,
       const rocksdb::Slice &eq_cond_upper_bound, bool read_current = false,
-      bool create_snapshot = true) {
+      bool create_snapshot = true, bool use_locking_iterator = false) {
     // Make sure we are not doing both read_current (which implies we don't
     // want a snapshot) and create_snapshot which makes sure we create
     // a snapshot
@@ -3943,12 +4043,13 @@ class Rdb_transaction {
     if (read_current) {
       options.snapshot = nullptr;
     }
-    return get_iterator(options, column_family);
+    return get_iterator(options, column_family, kd, use_locking_iterator);
   }
 
   virtual bool is_tx_started() const = 0;
   virtual void start_tx() = 0;
-  virtual void start_stmt() = 0;
+  virtual void start_stmt(bool is_dml_statement) = 0;
+  virtual void start_autocommit_stmt(bool /*is_dml_statement*/) {}
   virtual void set_name() = 0;
 
  protected:
@@ -4146,6 +4247,14 @@ class Rdb_transaction_impl : public Rdb_transaction {
 
   virtual bool is_writebatch_trx() const override { return false; }
 
+  // Lock the range between two specified endpoints
+  rocksdb::Status lock_range(rocksdb::ColumnFamilyHandle *const cf,
+                             const rocksdb::Endpoint &start_endp,
+                             const rocksdb::Endpoint &end_endp) override {
+    ++m_row_lock_count;
+    return m_rocksdb_tx->GetRangeLock(cf, start_endp, end_endp);
+  }
+
  private:
   void release_tx(void) {
     // We are done with the current active transaction object.  Preserve it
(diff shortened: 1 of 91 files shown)


storage/rocksdb/ha_rocksdb.h
```diff
@@ -266,6 +269,8 @@ class ha_rocksdb : public my_core::handler, public blob_buffer {
   /* Type of locking to apply to rows */
   Rdb_lock_type m_lock_rows;
 
+  bool m_use_range_locking;
```
hermanlee, line 272: It seems to me like this belongs more on the `Rdb_transaction` object than on `ha_rocksdb` (whereas `m_lock_rows` may be different per table in the same transaction).

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -3008,8 +3044,50 @@ class Rdb_transaction {
   virtual rocksdb::Status do_pop_savepoint() = 0;
   virtual void do_rollback_to_savepoint() = 0;
 
+ private:
+  /*
+    If true, the current statement should not use a snapshot for reading.
+    Note that in a multi-statement transaction, the snapshot may have been
+    allocated by another statement.
+  */
+  bool m_stmt_ignores_snapshot = false;
```
hermanlee, line 3053: I wonder if this works well with TTL tables, since it relies on rocksdb tracking oldest snapshot timestamp for us so that old transactions won't see rows expiring while they are still in progress. We won't get this if we don't create rocksdb snapshots though.

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -3190,6 +3268,18 @@ class Rdb_transaction {
...
                             const std::string &rowkey, bool force = false) = 0;
 
+  virtual rocksdb::Status lock_range(rocksdb::ColumnFamilyHandle *const cf,
+                                     const rocksdb::Endpoint &start,
+                                     const rocksdb::Endpoint &end) = 0;
+
+  rocksdb::Status lock_singlepoint_range(rocksdb::ColumnFamilyHandle *const cf,
+                                         const rocksdb::Slice &point) {
+    // Normally, one needs to "flip" the endpoint type for reverse-ordered CFs.
+    // But here we are locking just one point so this is not necessary.
```
hermanlee, line 3278: I wonder if this comment is necessarily true. For rev cfs, the locking iterator will always lock adjacent keys using `inf_suffix=true.` However, locking using `inf_suffix=false` here would mean that the locking iterator would not block modifications from happening.

For example, if the locking iterator iterates from 6->8 and then it locks the range `(8, inf_suffix=true)` to `(6, inf_suffix=true)`. A second query could delete 8 since it would use the point lock `(8, inf_suffix=false)` which falls outside of the locked range.

The following shows this is possible:
```
CONNECTION 1:
mysql> show create table t;
+-------+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| Table | Create Table                                                                                                                                                                                      |
+-------+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| t     | CREATE TABLE `t` (
  `i` int NOT NULL,
  `j` int NOT NULL,
  PRIMARY KEY (`i`) COMMENT 'rev:cf',
  KEY `j` (`j`) COMMENT 'cf'
) ENGINE=ROCKSDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_0900_ai_ci |
+-------+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
1 row in set (0.00 sec)
mysql> begin; select * from t force index (PRIMARY) where i > 5 order by i asc limit 3 for update ;
+---+---+
| i | j |
+---+---+
| 6 | 6 |
| 7 | 7 |
| 8 | 8 |
+---+---+
3 rows in set (2.60 sec)
mysql> select * From information_schema.rocksdb_locks;
+------------------+-----------------+---------------------------------------+------+
| COLUMN_FAMILY_ID | TRANSACTION_ID  | KEY                                   | MODE |
+------------------+-----------------+---------------------------------------+------+
|                3 | 140534119312896 | 0000010780000008:1-0000010780000006:1 | X    |
+------------------+-----------------+---------------------------------------+------+
1 row in set (0.00 sec)
CONNECTION 2:
mysql> delete from t where i = 8;
Query OK, 1 row affected (6.58 sec)
CONNECTION 1:
mysql> select * from t force index (PRIMARY) where i > 5 order by i asc limit 3 for update ;
+---+---+
| i | j |
+---+---+
| 6 | 6 |
| 7 | 7 |
| 9 | 9 |
+---+---+
3 rows in set (3.21 sec)
```

Incidentally, `RangeLockManagerBase::TryLock` also has the same behaviour of always setting `inf_suffix=false` and it's possible that also introduces similar bugs (This is called when we call `get_for_update`, I wonder if we should call `get_for_update` with nullptr as value instead of adding a `lock_singlepoint_range` function here).

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -5516,7 +5686,7 @@ class Rdb_snapshot_status : public Rdb_tx_list_walker {
 
     /* Calculate the duration the snapshot has existed */
     int64_t snapshot_timestamp = tx->m_snapshot_timestamp;
-    if (snapshot_timestamp != 0) {
+    if (snapshot_timestamp != 0 && tx->has_snapshot()) {
```
hermanlee, line 5689: Some of this information still seems useful even if there's no actual snapshot on the transaction. It might be useful to also indicate whether this transaction has an active snapshot too.
I guess some of this information is in rocksdb_trx, but dogpiles only has show engine status.

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -6837,6 +7050,28 @@ static int rocksdb_init_internal(void *const p) {
   tx_db_options.write_policy =
       static_cast<rocksdb::TxnDBWritePolicy>(rocksdb_write_policy);
 
+  if (rocksdb_use_range_locking && rocksdb_use_range_lock_manager_as_point) {
+    // rdb_log_status_error(
```
hermanlee, line 7054: Is it useful to keep this commented out?

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -9556,6 +9809,247 @@ int ha_rocksdb::index_read_intern(uchar *const buf, const uchar *const key,
...
+    The only way to achieve the required ordering is to flip the endpoint
+    flag value before passing class Endpoint to RocksDB. This function does
+    it before the lock_range() call.
+
+  @return
+     0      Ok
+     Other  Error acquiring the lock (wait timeout, deadlock, etc)
+*/
+
+int ha_rocksdb::set_range_lock(Rdb_transaction *tx, const Rdb_key_def &kd,
```
hermanlee, line 9860: It feels like this function could generally be better written to reduce duplicate code, since the current code seems error prone.

If we want to keep things simple, we can also just lock the iterator bounds which is a superset of what we're locking here (like we do in the locking iterator), and not think too hard about all this.

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -9556,6 +9809,247 @@ int ha_rocksdb::index_read_intern(uchar *const buf, const uchar *const key,
...
+                               const enum ha_rkey_function &find_flag,
+                               const rocksdb::Slice &slice_arg,
+                               const key_range *const end_key,
+                               bool *use_locking_iterator) {
+  rocksdb::Slice end_slice;
+  uchar end_slice_buf[MAX_KEY_LENGTH];
```
hermanlee, line 9866: Is `MAX_KEY_LENGTH` safe to use? From what I know, it refers to the max length of sql layer keys, but packed rocksdb keys can be longer due to things like varchar encoding etc.

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -9556,6 +9809,247 @@ int ha_rocksdb::index_read_intern(uchar *const buf, const uchar *const key,
...
+            select * from t1 where pk1=const order by pk1 desc for update
+
+          assuming this uses an index on (pk1, ...).
+          We get end_key=nullptr.
+      */
+      start_has_inf_suffix = false;
+      end_has_inf_suffix = true;
+      end_slice = slice;
+    }
+  } else if (find_flag == HA_READ_PREFIX_LAST_OR_PREV) {
```
hermanlee, line 9904: I think we're missing the nullptr check for `end_key` like we do in the `HA_READ_BEFORE_KEY` case. I was able to reproduce a segfault with non-exclusive end range, so I wonder if there's a gap in test coverage here.

As a side note, I find it a little odd that we decided to merge the `HA_READ_KEY_OR_NEXT`, `HA_READ_AFTER_KEY` cases (for the forwards direction), but not for `HA_READ_PREFIX_LAST_OR_PREV`, `HA_READ_BEFORE_KEY` (the backwards direction). The only difference between `HA_READ_PREFIX_LAST_OR_PREV` and `HA_READ_BEFORE_KEY` should just be whether end_has_inf_suffix is true/false.

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -9556,6 +9809,247 @@ int ha_rocksdb::index_read_intern(uchar *const buf, const uchar *const key,
...
+    /*
+       We get here for queries like:
+
+         select * from t1 where pk1=const1 and pk2 between const2 and const3
+         order by pk1 desc
+         for update
+
+       assuming this uses an index on (pk1, pk2).
+       The slice has the right endpoint: {const1, const3}
+       the end_key has the left endpoint: {const1, const2}.
```
hermanlee, line 9914: I think we should also be looking at `end_key->flag `to determine `start_has_inf_suffix` (similar to the forwards case). From what I can tell, there's no correctness implication, so maybe just add a TODO?

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -9556,6 +9809,247 @@ int ha_rocksdb::index_read_intern(uchar *const buf, const uchar *const key,
...
+    */
+
+    // Move the right endpoint from slice to end_slice
+    end_slice = slice;
+
+    // Pack the left endpoint and make "slice" point to it
+    uchar pack_buffer[MAX_KEY_LENGTH];
```
hermanlee, line 9921: Similarly with `pack_buffer` here, not sure if `MAX_KEY_LENGTH` is correct, although we should be able to just reuse `m_pack_buffer` on the `ha_rocksdb` object.

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -9556,6 +9809,247 @@ int ha_rocksdb::index_read_intern(uchar *const buf, const uchar *const key,
...
+    uint end_slice_size = kd.pack_index_tuple(
+        table, pack_buffer, end_slice_buf, end_key->key, end_key->keypart_map);
```
hermanlee, line 9923: I think we can avoid calling `pack_index_tuple` again here if we pass down the `end_slice` object in the caller of this function. We seem to be computing the exact same thing?

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -9556,6 +9809,247 @@ int ha_rocksdb::index_read_intern(uchar *const buf, const uchar *const key,
...
+                              end_key->keypart_map);
+
+      slice = rocksdb::Slice(reinterpret_cast<char *>(end_slice_buf),
+                             end_slice_size);
+
+      // end_has_inf_suffix is false, because we're looking key<const
+    } else {
+      uint end_slice_size;
+      kd.get_infimum_key(end_slice_buf, &end_slice_size);
+      slice = rocksdb::Slice((char *)end_slice_buf, end_slice_size);
```
hermanlee, line 9953: Is this work useful if `big_range == true`?

Also, I wonder if it makes sense to pull out the cases where big_range is true at the very beginning. It seems like `big_range` is true iff `end_key` is nullptr, and would reduce all these if checks here. There are some places where we set `big_range` for unknown range types, but arguably those should just error out.

Actually, I think `end_key` is also nullptr for the "prefix" flags, so it's not true that `big_range` iff `end_key == nullptr`. It may still be simpler to pull it out to the beginning though.

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -9556,6 +9809,247 @@ int ha_rocksdb::index_read_intern(uchar *const buf, const uchar *const key,
...
+    else if (find_flag == HA_READ_AFTER_KEY)
+      start_has_inf_suffix= true;
+    else
+      DBUG_ASSERT(0);
+
+    uint end_slice_size;
+    kd.get_infimum_key(end_slice_buf, &end_slice_size);
+    end_slice= rocksdb::Slice((char*)end_slice_buf, end_slice_size);
+    end_has_inf_suffix= true;
+#endif
```
hermanlee, line 10005: Is there value in keeping this code commented out?

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -9556,6 +9809,247 @@ int ha_rocksdb::index_read_intern(uchar *const buf, const uchar *const key,
...
+
+    RocksDB has Iterator::Refresh() method, but alas, it is not implemented for
+    the iterator returned by Transaction object (Transaction object returns
+    BaseDeltaIterator which allows one to see the transactions's own changes).
+
+    Our solution to this is to release the iterator and create the new one.
+    We release it here, it will be created as soon as there's a need to read
+    records.
+  */
+  m_iterator->reset();
```
hermanlee, line 10043: One option that we can consider is to use the tailing option in rocksdb iterators. It may be worth capturing this in a comment so that we don't lose this idea.

Actually, I thinking tailing won't work because it will only read from files from its superversion. I guess we do just need iterator refresh after all.

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -10775,6 +11271,8 @@ int ha_rocksdb::check_and_lock_sk(
     lock for this key.
   */
   if (!(key_info->flags & HA_NOSAME)) {
+    if (rocksdb_use_range_locking)
+      return check_and_lock_non_unique_sk(key_id, row_info);
```
hermanlee, line 11275: Maybe add a comment about why we don't use m_use_range_locking instead? I assume it's because we still want RC transactions to be blocked by gap locks held by RR transactions.

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -10891,6 +11389,53 @@ int ha_rocksdb::check_and_lock_sk(
...
+  const Rdb_key_def &kd = *m_key_descr_arr[key_id];
+  bool store_row_debug_checksums = should_store_row_debug_checksums();
+
+  if (row_info.old_data != nullptr) {
+    rocksdb::Slice old_key_slice;
+    int old_packed_size;
+
+    old_packed_size = kd.pack_record(
+        table, m_pack_buffer, row_info.old_data, m_sk_packed_tuple_old,
+        &m_sk_tails_old, store_row_debug_checksums, row_info.hidden_pk_id, 0,
```
hermanlee, line 11408: I don't think we care about the value, so we can just pass nullptr for `m_sk_tails_old` (similarly below for the new key).

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -15078,6 +15642,34 @@ static int show_rocksdb_stall_vars(THD *thd MY_ATTRIBUTE((unused)),
   return 0;
 }
 
+//
+// Lock Tree Status variables
+//
+static longlong rocksdb_locktree_escalation_count = 1234;
```
hermanlee, line 15648: Initializing to 0 is probably less confusing.

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -15078,6 +15642,34 @@ static int show_rocksdb_stall_vars(THD *thd MY_ATTRIBUTE((unused)),
...
+    // end of the array marker
+    {NullS, NullS, SHOW_LONG, SHOW_SCOPE_GLOBAL}};
+
+static SHOW_VAR rocksdb_empty_status_variables[] = {
+    {NullS, NullS, SHOW_LONG, SHOW_SCOPE_GLOBAL}};
+
+static void show_rocksdb_locktree_vars(THD *, SHOW_VAR *var, char *) {
+  var->type = SHOW_ARRAY;
+  if (range_lock_mgr) {
+    auto status = range_lock_mgr->GetStatus();
```
hermanlee, line 15665: Do we want to expose `status.lock_wait_count` as well?

storage/rocksdb/nosql_access.cc
```diff
@@ -2220,6 +2221,7 @@ bool INLINE_ATTR select_exec::setup_iterator(THD *thd) {
   } else {
     m_iterator.reset(
         new Rdb_iterator_base(thd, m_key_def, m_pk_def, m_tbl_def));
+    //      m_lower_bound_slice, m_upper_bound_slice);
```
hermanlee, line 2224: Can be removed.

storage/rocksdb/rdb_iterator.h
```diff
@@ -123,6 +130,18 @@ class Rdb_iterator_base : public Rdb_iterator {
...
 
+  /* Whether m_scan_it is a locking iterator */
+  bool m_iter_uses_locking;
+
+  /*
+    Whether the iterator should use locking. The intended workflow is:
+
+      iter.set_use_locking() // sets m_iter_should_use_locking=true
+      iter.seek() // this will re-create m_scan_it to be the right kind
+                  // of iterator
```
hermanlee, line 141: It's not clear to me whether this workflow is ever useful where we change locking mode between seeks. It seems like the code would be simpler if we just specified if we want a locking iterator or not at construction time, and we wouldn't have to track whether we need to recreate the iterator or not.

storage/rocksdb/rdb_iterator.cc
```diff
@@ -208,6 +215,18 @@ int Rdb_iterator_base::calc_eq_cond_len(enum ha_rkey_function find_flag,
...
+    int not_found_code) {
+  if (it->Valid()) return HA_EXIT_SUCCESS;
+
+  rocksdb::Status s = it->status();
+  if (s.ok() || s.IsNotFound()) return not_found_code;
+
+  Rdb_transaction *tx = get_tx_from_thd(m_thd);
+  return rdb_tx_set_status_error(tx, s, *kd, m_tbl_def);
+}
+
```
hermanlee, line 229: We currently have most of our error handling code for iterators in `is_valid_iterator`. Would be nice to keep it in one place.

storage/rocksdb/rdb_locking_iter.cc
```diff
@@ -0,0 +1,278 @@
...
+  }
+  // This will set m_status accordingly
+  lock_up_to(scan_forward, target, end);
+}
+
+/*
+  Lock the range between [target, (current m_iter position)] and position
+  the iterator on the first record in it.
+
+  @param call_next false means current iterator position is achieved by
```
hermanlee, line 163: We have `ha_rocksdb::index_next_with_direction_intern` already with `skip_next` that does something similar so I wonder if we should stick to one naming convention.

storage/rocksdb/rdb_locking_iter.cc
```diff
@@ -0,0 +1,278 @@
...
+      lock_till_iterator_end(scan_forward, target);
+    }
+    return;
+  }
+
+  while (1) {
+    DEBUG_SYNC(my_core::thd_get_current_thd(), "rocksdb.locking_iter_scan");
+
+    if (my_core::thd_killed(current_thd)) {
+      m_status = rocksdb::Status::Aborted();
```
hermanlee, line 183: It's a little strange that we end up returning `HA_ERR_ROCKSDB_STATUS_ABORTED` instead of the usual `HA_ERR_QUERY_INTERRUPTED` here.

storage/rocksdb/rdb_locking_iter.cc
```diff
@@ -0,0 +1,278 @@
...
+}
+
+/*
+  @detail
+  Ideally, this function should
+   - find the first key $first_key
+   - lock the range [-inf; $first_key]
+   - return, the iterator is positioned on $first_key
+
+  The problem here is that we cannot have "-infinity" bound.
```
hermanlee, line 254: Looking at the comparator functions, I would assume that "-infinity" would be `Endpoint(slice="", inf_suffix=false)`, though I agree that this function is basically unused for us. Might be worth clarifying what the comment means.

hermanlee (commented):
Adding Manuel's comments.

storage/rocksdb/ha_rocksdb.h
laurynas-biveinis: > It seems to me like this belongs more on the `Rdb_transaction` object than on `ha_rocksdb` (whereas `m_lock_rows` may be different per table in the same transaction).

My current best understanding is 50-50 re. the best home for this field. In its current form, its value is decided by `ha_rocksdb::m_lock_rows` too. If it's changed to be independent (as it already half-way is in the checks), then it can be stored in `Rdb_transaction`, but then some of the code paths currently avoid looking at the transaction object at all, so they would get extra `get_or_create_tx` calls.

storage/rocksdb/ha_rocksdb.cc
laurynas-biveinis: > Some of this information still seems useful even if there's no actual snapshot on the transaction. 

Are there any constraints on the output should look like in the case of no snapshot, if tooling parses it?

storage/rocksdb/ha_rocksdb.cc
laurynas-biveinis: We should be able to drop this and use `ha_rocksdb::m_pack_buffer` instead

storage/rocksdb/ha_rocksdb.cc
laurynas-biveinis: Discussed offline

storage/rocksdb/rdb_iterator.h
laurynas-biveinis: It seems this would result in cleaner iterator code, but it would require postponing the creation of iterator from `ha_rocksdb::index_init` right to the point of attempting to take the range lock, and at that point the iterator is already in use. I will look some more.

storage/rocksdb/rdb_iterator.cc
laurynas-biveinis: The functions are a bit different:
- `is_valid_iterator`: works on a RDB-level iterator, has no THD or MyRocks transaction
- `iter_status_to_retval`: works on a MyRocks-level iterator, has THD, transaction, key definition.

Almost all uses follow the pattern of calling `is_valid_iterator` first and then `iter_status_to_retval`. But there is also `calculate_cardinality_table_scan`, which works with an RDB-level iterator directly.

Unless we find a different way to address that last use, the best I can think of is add a call to is_valid_iterator to `iter_status_to_retval`, perhaps with a rename, so that it's more natural to use in iteration loops.

storage/rocksdb/rdb_locking_iter.cc
laurynas-biveinis: I don't see a straightforward way to do the regular THD kill flag handling here, will need to revisit.

storage/rocksdb/ha_rocksdb.cc
laurynas-biveinis: Rewrote to simplify with the current algorithm. If it will be deemed still too complex, will do the iterator bound locking.

storage/rocksdb/ha_rocksdb.cc
laurynas-biveinis: Testcases confirm that this is indeed the reason. Will need to check whether the plain `rocksdb_use_range_locking` check is not too broad regarding temporary, DD etc. tables.

storage/rocksdb/rdb_iterator.h
laurynas-biveinis: Looks possible. It might result in a separate iterator reorg PR

laurynas-biveinis: Reimplemented this without much refactoring, essentially merged ~reset~ and ~set_use_locking~.

storage/rocksdb/rdb_iterator.cc
laurynas-biveinis: https://github.com/facebook/mysql-5.6/pull/1444

laurynas-biveinis:
Continued in https://github.com/facebook/mysql-5.6/pull/1449

PR #1442, opened 2024-03-22
Cleanup rdb_hexdump

laurynas-biveinis:
Currently, `rdb_hexdump` has a third argument `maxsize` with a default value of 0,
meaning unlimited output length, and there is also a `RDB_MAX_HEXDUMP` constant
for the callers to use. However, only a part of the callers use it, the rest use
the default or something else altogether (i.e. `FN_REFLEN`)

Change as follows:
- Default to `RDB_MAX_HEXDUMP` `maxsize` instead of unlimited. This should cover
  all the reasonable cases, and prevent runaway dumping in the case of corrupted
  `data_len` values.
- Bump `RDB_MAX_HEXDUMP_LEN` from 1000 to 7000. This way it can dump any MyRocks
  key (3072 byte maximum). Make it `constexpr`.

Minor cleanups:
- Where `std::string` is used as an input for `rdb_hexdump`, pass its buffer
  through `data()`, not through `c_str()`, because it's used as a sized buffer
  and not as a zero-terminated C string
- Mark `rdb_hexdump` as `nodiscard`, make its constants `constexpr`, remove
  redundant `const` from by-value parameters.


luqun:
>  This way it can dump any MyRocks key (3072 byte maximum).

This sounds incorrect.  max_supported_key_part_length is 3072. max_supported_key_length is  16 * 1024 = 16k.

luqun:
in Rdb_deadlock_info::Rdb_dl_trx_info get_dl_txn_info(), would you prefer  calling rdb_hexdump with 0?

laurynas-biveinis:
> This sounds incorrect. max_supported_key_part_length is 3072. max_supported_key_length is 16 * 1024 = 16k.

Right, I missed this. Is it OK to bump the `rdb_hexdump` default to
cover 16K keys?

laurynas-biveinis:
> in Rdb_deadlock_info::Rdb_dl_trx_info get_dl_txn_info(), would you prefer calling rdb_hexdump with 0?

Yes, updating

laurynas-biveinis:
Rebased, addressed both comments (the rdb_hexdump one speculatively by bumping the default to 32K). Let me know if you prefer another default.

storage/rocksdb/rdb_utils.h
```diff
@@ -149,10 +149,9 @@ namespace myrocks {
   rdb_check_mutex_call_result(__PRETTY_FUNCTION__, false, \
                               mysql_mutex_unlock(&m))
 
-/*
-  Generic constant.
-*/
-const size_t RDB_MAX_HEXDUMP_LEN = 1000;
+// The default limit for rdb_hexdump output length, happens to fit the longest
+// possible keys (16K)
+constexpr size_t RDB_MAX_HEXDUMP_LEN = 2 * 16 * 1024;
```
luqun, line 154: The comment mention 16K, but the value is 32K?

laurynas-biveinis:
> +// The default limit for rdb_hexdump output length, happens to fit the longest
> +// possible keys (16K)
> +constexpr size_t RDB_MAX_HEXDUMP_LEN = 2 * 16 * 1024;
>
> The comment mention 16K, but the value is 32K?

hexdump: two output bytes per one input byte

PR #1444, opened 2024-03-27
Refactor MyRocks iterator status handling

laurynas-biveinis:
- Move is_valid_iterator from ha_rocksdb.cc to rdb_iterator.cc, rename to
  is_valid_rdb_iterator, make its argument const reference. Remove redundant
  rdb_persist_corruption_marker call because rdb_handle_io_error calls it too.
- Introduce Rdb_iterator_base::convert_iterator_status similar to existing
  Rdb_iterator_base::convert_get_status, move HA_ERR_END_OF_FILE-returning logic
  there and additionally handle any other non-I/O errors by saving them in the
  transaction for a later return.
- Make Rdb_iterator_base::convert_get_status take transaction by reference, make
  it const method. Propagate down the callstack pointer-to-reference changes:
  rdb_should_hide_ttl_rec, rdb_tx_set_status_error.
- Mark touched functions and methods [[nodiscard]] where appropriate.
- Annotate rdb_persist_corruption_marker with cold and noinline attributes.
- Remove dbug_change_status_to_corrupted, use DBUG_EVALUATE_IF so that the
  variable gets its final value right at the declaration and could be made const
  too.

This is split out from the range locking feature.

### DIFF
--- a/storage/rocksdb/ha_rocksdb.cc
+++ b/storage/rocksdb/ha_rocksdb.cc
@@ -4938,10 +4938,6 @@ class Rdb_transaction {
 };
 
 #ifndef NDEBUG
-// simulate that RocksDB has reported corrupted data
-static void dbug_change_status_to_corrupted(rocksdb::Status *status) {
-  *status = rocksdb::Status::Corruption();
-}
 static void dbug_change_status_to_io_error(rocksdb::Status *status) {
   *status = rocksdb::Status::IOError();
 }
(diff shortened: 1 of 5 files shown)


laurynas-biveinis:
As discussed offline, will add comments explaining why two different ways to check RDB iterator status are needed

laurynas-biveinis:
Unfortunately there is a small conflict with https://github.com/facebook/mysql-5.6/pull/1445, the second PR to go in will have to be rebased

PR #1445, opened 2024-04-05
Pass ColumnFamilyHandle around by reference; wrap Iterator in unique_ptr

laurynas-biveinis:
In preparation for the range locking patch, clean up some code related to
iterators:
- Pass rocksdb::ColumnFamilyHandle around and store in classes by references
  instead of pointers. It is never nullptr, nor is ever reseated in the classes
  using it.
- Wrap the rocksdb::Iterator objects in std::unique_ptr: change the factory
  method return types and class fields.
- In both cases above mark the classes containing reference fields as
  non-copyable and non-moveable as needed, matching their current usage.
- Mark touched methods [[nodiscard]] as applicable. In rdb_index_merge.h, remove
  many instances of redundant MY_ATTRIBUTE((__nonnull__)) too.
- Make touched local vars auto, but with reference or pointer pulled out,
  sometimes avoiding redundant returned object copies.
- Remove redundant instances of const from passed-by-value arguments.

### DIFF
--- a/storage/rocksdb/ha_rocksdb.cc
+++ b/storage/rocksdb/ha_rocksdb.cc
@@ -4060,19 +4060,21 @@ class Rdb_transaction {
   }
 
  public:
-  int get_key_merge(GL_INDEX_ID kd_gl_id, rocksdb::ColumnFamilyHandle *cf,
-                    Rdb_index_merge **key_merge) {
+  [[nodiscard]] int get_key_merge(GL_INDEX_ID kd_gl_id,
+                                  rocksdb::ColumnFamilyHandle &cf,
+                                  Rdb_index_merge **key_merge) {
     assert(!is_ac_nl_ro_rc_transaction());
 
     int res;
     auto it = m_key_merge.find(kd_gl_id);
     if (it == m_key_merge.end()) {
+      auto *const thd = get_thd();
       m_key_merge.emplace(
           std::piecewise_construct, std::make_tuple(kd_gl_id),
-          std::make_tuple(
-              get_rocksdb_tmpdir(), THDVAR(get_thd(), merge_buf_size),
-              THDVAR(get_thd(), merge_combine_read_size),
-              THDVAR(get_thd(), merge_tmp_file_removal_delay_ms), cf));
+          std::make_tuple(get_rocksdb_tmpdir(), THDVAR(thd, merge_buf_size),
+                          THDVAR(thd, merge_combine_read_size),
+                          THDVAR(thd, merge_tmp_file_removal_delay_ms),
+                          std::ref(cf)));
       it = m_key_merge.find(kd_gl_id);
       if ((res = it->second.init()) != 0) {
         return res;
@@ -4553,7 +4555,7 @@ class Rdb_transaction {
     assert(!is_ac_nl_ro_rc_transaction());
 
     uint n_null_fields = 0;
-    const rocksdb::Comparator *index_comp = key_def.get_cf()->GetComparator();
+    const auto *const index_comp = key_def.get_cf().GetComparator();
 
     /* Get proper SK buffer. */
     uchar *sk_buf = sk_info->swap_and_get_sk_buf();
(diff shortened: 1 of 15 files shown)


laurynas-biveinis:
Unfortunately there is a small conflict with https://github.com/facebook/mysql-5.6/pull/1444, the second PR to go in will have to be rebased

storage/rocksdb/rdb_sst_info.h
```diff
@@ -35,12 +35,13 @@ namespace myrocks {
 class Rdb_sst_file_ordered {
  private:
   class Rdb_sst_file {
-   private:
-    Rdb_sst_file(const Rdb_sst_file &p) = delete;
-    Rdb_sst_file &operator=(const Rdb_sst_file &p) = delete;
+    Rdb_sst_file(const Rdb_sst_file &) = delete;
+    Rdb_sst_file &operator=(const Rdb_sst_file &) = delete;
+    Rdb_sst_file(const Rdb_sst_file &&) = delete;
```
laurynas-biveinis:
```suggestion
    Rdb_sst_file(Rdb_sst_file &&) = delete;
```

storage/rocksdb/rdb_sst_info.h
```diff
@@ -35,12 +35,13 @@ namespace myrocks {
...
+    Rdb_sst_file &operator=(const Rdb_sst_file &&) = delete;
```
laurynas-biveinis:
```suggestion
    Rdb_sst_file &operator=(Rdb_sst_file &&) = delete;
```

laurynas-biveinis:
Fixed incorrect move ctor/assignment operator signatures and rebased

luqun:
could you also update storage/rocksdb/rdb_vector_db.cc

storage/rocksdb/rdb_vector_db.cc:192:22: error: no matching function for call to 'rdb_tx_get_iterator'
    m_iterator.reset(rdb_tx_get_iterator(

laurynas-biveinis:
> could you also update storage/rocksdb/rdb_vector_db.cc
>
> storage/rocksdb/rdb_vector_db.cc:192:22: error: no matching function for call to 'rdb_tx_get_iterator'
> m_iterator.reset(rdb_tx_get_iterator(

Today I learned about -DWITH_FB_VECTORDB=ON build, and it looks like
I'll have to do some PRs for it as well. I was able to make the changes,
but it's possible I missed something, please let me know.

Rebased too

laurynas-biveinis:
@luqun , extended the MEM_* macros to cover ASan, no regressions there, & moved the TODO comment. Ready for review

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -4060,19 +4060,21 @@ class Rdb_transaction {
...
       m_key_merge.emplace(
           std::piecewise_construct, std::make_tuple(kd_gl_id),
-          std::make_tuple(
-              get_rocksdb_tmpdir(), THDVAR(get_thd(), merge_buf_size),
-              THDVAR(get_thd(), merge_combine_read_size),
-              THDVAR(get_thd(), merge_tmp_file_removal_delay_ms), cf));
+          std::make_tuple(get_rocksdb_tmpdir(), THDVAR(thd, merge_buf_size),
+                          THDVAR(thd, merge_combine_read_size),
+                          THDVAR(thd, merge_tmp_file_removal_delay_ms),
+                          std::ref(cf)));
```
luqun, line 4077: why do we need std::ref? since cf is already a reference type

laurynas-biveinis:
> In storage/rocksdb/ha_rocksdb.cc:
>
>>        m_key_merge.emplace(
>            std::piecewise_construct, std::make_tuple(kd_gl_id),
> -          std::make_tuple(
> -              get_rocksdb_tmpdir(), THDVAR(get_thd(), merge_buf_size),
> -              THDVAR(get_thd(), merge_combine_read_size),
> -              THDVAR(get_thd(), merge_tmp_file_removal_delay_ms), cf));
> +          std::make_tuple(get_rocksdb_tmpdir(), THDVAR(thd, merge_buf_size),
> +                          THDVAR(thd, merge_combine_read_size),
> +                          THDVAR(thd, merge_tmp_file_removal_delay_ms),
> +                          std::ref(cf)));
>
> why do we need std::ref? since cf is already a reference type

Without std::ref, the compiler apparently trys to instantiate
std::tuple<..., ColumnFamilyHandle>, and gets an error that
ColumnFamilyHandle is an abstract class

PR #1446, opened 2024-04-09
MyRocks: merge the key packing buffer allocation, annotate m_pack_buffer lifetime

laurynas-biveinis:
All the key packing buffers in the MyRocks handler have exactly the same
lifetime between alloc_key_buffers and free_key_buffers. Thus allocate a single
large buffer and point the key packing buffers to offsets inside it, reducing
number of calls to the heap allocator on table opening/close code path. Annotate
the padding bytes for Valgrind as inaccessible. Add additional handling for
m_pack_buffer: since its use it's fully contained in Rdb_key_def::pack_record,
annotate it as such.

### DIFF
--- a/include/memory_debugging.h
+++ b/include/memory_debugging.h
@@ -41,10 +41,26 @@
 #include <valgrind/memcheck.h>
 
 #define MEM_UNDEFINED(a, len) VALGRIND_MAKE_MEM_UNDEFINED(a, len)
-#define MEM_DEFINED_IF_ADDRESSABLE(a, len) \
-  VALGRIND_MAKE_MEM_DEFINED_IF_ADDRESSABLE(a, len)
+// #define MEM_DEFINED_IF_ADDRESSABLE(a, len) \
+//   VALGRIND_MAKE_MEM_DEFINED_IF_ADDRESSABLE(a, len)
 #define MEM_NOACCESS(a, len) VALGRIND_MAKE_MEM_NOACCESS(a, len)
 #define MEM_CHECK_ADDRESSABLE(a, len) VALGRIND_CHECK_MEM_IS_ADDRESSABLE(a, len)
+
+#elif defined(HAVE_ASAN)
+
+#include <cassert>
+#include <sanitizer/asan_interface.h>
+
+#define MEM_MALLOCLIKE_BLOCK(p1, p2, p3, p4) ASAN_UNPOISON_MEMORY_REGION(p1, p2)
+#define MEM_FREELIKE_BLOCK(p1, p2, p3, p4) ASAN_POISON_MEMORY_REGION(p1, p2)
+
+#define MEM_UNDEFINED(a, len) ASAN_UNPOISON_MEMORY_REGION(a, len)
+#define MEM_NOACCESS(a, len) ASAN_POISON_MEMORY_REGION(a, len)
+
+// In the case of error, this will immediatelly terminate the process instead of
+// printing an error and continuing, which is more common for ASan errors.
+// Change it to log and continue if that becomes an issue.
+#define MEM_CHECK_ADDRESSABLE(a, len) assert(!__asan_region_is_poisoned(a, len))
 
 #else /* HAVE_VALGRIND */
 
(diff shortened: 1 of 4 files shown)


laurynas-biveinis:
This resulted from me proving to myself the lifetime of `ha_rocksdb::m_pack_buffer` during range locking debugging.
Valgrind testing still ongoing

laurynas-biveinis:
Valgrind testing timed out after a few days, with no errors up to that point (45% of `rocksdb,rocksdb_clone,rocksdb_rpl,rocksdb_sys_vars,rocksdb_dd_innodb`, no `--big-test`)

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -9418,48 +9424,57 @@ int ha_rocksdb::alloc_key_buffers(const TABLE &table_arg,
...
-      my_malloc(PSI_NOT_INSTRUMENTED, max_packed_sk_len, MYF(0)));
-  m_sk_packed_tuple_old = reinterpret_cast<uchar *>(
-      my_malloc(PSI_NOT_INSTRUMENTED, max_packed_sk_len, MYF(0)));
-  m_sk_packed_tuple_updated = reinterpret_cast<uchar *>(
-      my_malloc(PSI_NOT_INSTRUMENTED, max_packed_sk_len, MYF(0)));
-  m_end_key_packed_tuple = reinterpret_cast<uchar *>(
-      my_malloc(PSI_NOT_INSTRUMENTED, max_packed_sk_len, MYF(0)));
-  m_pack_buffer = reinterpret_cast<uchar *>(
-      my_malloc(PSI_NOT_INSTRUMENTED, max_packed_sk_len, MYF(0)));
+  const auto pad1 [[maybe_unused]] = get_next_aligned(buf_size);
```
luqun, line 9427: This might be incorrect. Before your change, the size will be max(PK, SKs). after your chagne, the size will be PK.

storage/rocksdb/ha_rocksdb.h
```diff
@@ -194,8 +194,14 @@ class ha_rocksdb : public my_core::handler, public blob_buffer {
   */
   mutable bool m_pk_can_be_decoded;
 
+  // The common buffer for m_pk_packed_tuple, m_sk_packed_tuple,
+  // m_sk_packed_tuple_old, m_sk_packed_tuple_updated, m_end_key_packed_tuple,
+  // & m_pack_buffer.
+  unique_ptr_my_free<uchar[]> buffers;
+  // ^^ todo: change it to 'char[]'?
```
luqun: nit: maybe todo for m_pk_packed_tuple instead of buffers? see "^^"

storage/rocksdb/ha_rocksdb.cc
laurynas-biveinis: Could you elaborate? `m_pk_packed_tuple` length should be PK length, and all the others should be max(PK, SK). I believe this is satisfied both before the change and now.

storage/rocksdb/ha_rocksdb.h
laurynas-biveinis: You mean like this?
```c++
  // The common buffer for m_pk_packed_tuple, m_sk_packed_tuple,
  // m_sk_packed_tuple_old, m_sk_packed_tuple_updated, m_end_key_packed_tuple,
  // & m_pack_buffer.
  unique_ptr_my_free<uchar[]> buffers;

  uchar *m_pk_packed_tuple; /* Buffer for storing PK in StorageFormat */
  // TODO: ^ can we join this with last_rowkey?
  // ^^ todo: change it to 'char[]'?
```

luqun: yep

storage/rocksdb/ha_rocksdb.cc
luqun: my misunderstanding.. the code is doing same as before.

laurynas-biveinis:
Ack, I am waiting for MEM_* macro for ASan test results and will push again

laurynas-biveinis:
@luqun , extended the MEM_* macros to cover ASan, no regressions there, & moved the TODO comment. Ready for review

include/memory_debugging.h
```diff
@@ -41,11 +41,27 @@
 #include <valgrind/memcheck.h>
 
 #define MEM_UNDEFINED(a, len) VALGRIND_MAKE_MEM_UNDEFINED(a, len)
-#define MEM_DEFINED_IF_ADDRESSABLE(a, len) \
-  VALGRIND_MAKE_MEM_DEFINED_IF_ADDRESSABLE(a, len)
+// #define MEM_DEFINED_IF_ADDRESSABLE(a, len) \
```
luqun, line 44: nit: why do you comment these?

laurynas-biveinis: See the `#pragma GCC poison` below and the comment there: it's not called in the sources

PR #1448, opened 2024-04-22
Disable binlog.binlog_mysqlbinlog_source_gipk_info

laurynas-biveinis:
Meta's patch has changed the table map log event and this test uses
pre-generated upstream ones, which are incompatible. It would be possible to
recreate pre-generated events in Meta format, but there is little value in
restoring the functionality of this test anyway.

Squash with e13fd4103d3c9347aedf9cad02a2b7e445288413


laurynas-biveinis:
Please re-check the squash commit reference, as it was only a guess on my part

PR #1450, opened 2024-04-30
[clone] Persist synchronization gtid from P_S.log_status

sunxiayi:
Summary:

### Summary
When synchronizing engines, get the synchronization coordinates from P_S.log_status table, send from the server plugin to client plugin. Upon receiving, client plugin writes to a file `#clone/#synchronization_coordinates`. 

### Approach
**Protocol**
Add a new response type `COM_RES_GTID_V4` and update latest protocol to `CLONE_PROTOCOL_VERSION_V4`.

**Common**
`synchronize_engines()` call was in `Ha_clone_common_cbk`. But it actually is only called from server or local, not client. So I move this function into the child class(client, server, local, with client implementation is a no-op). The reason I make this change is because, I want to send the gtid from the plugin layer without calling into engine again, and moving `synchronize_engines()` into `Server_Cbk` has the advantage that I can get the server handler, while in `Ha_clone_common_cbk` I cannot. 

The log_status query and set_log_stop steps are moved into a new function `synchronize_logs` under `Ha_clone_common_cbk`. 

`populate_synchronization_coordinates` would populate a Key_Values data structure of:
- gtid(from log_status table)
- binlog_file
- binlog_offset
- gtid(from binlog_file and binlog_offset)

We want to record both gtid from log_status and from binlog_file/offset because they seem to be out of sync in prod, and need a way to confirm this.

**Client**
`synchronize_engines()` is a no-op and errors out upon calling.

Upon receiving `COM_RES_GTID_V4`, deserialize the coordinates and persist that in `#clone/#synchronization_coordinates`.

**Server**
`synchronize_engines()` would call `synchronize_logs` and get the server handle to send the coordinates one by one, utilizing existing helper functions.

**Local**
`synchronize_engines()` would call `synchronize_logs`, then get the client handle to persist coordinates in `#clone/#synchronization_coordinates`.

**Handle version mismatch**
In server, only send coordinate if negotiated version >= V4.
In client, #synchronization_coordinates file is cleaned upon start of clone.

Test Plan:

### MTR
`local_create_synchronization_coordinates`
`remote_create_synchronization_coordinates`

### Local clone
Test on my debug build in devserver after `install plugin clone SONAME 'mysql_clone.so';`:
1/ `mysql> CLONE LOCAL DATA DIRECTORY = '/home/sunxiayi/mysql/mysql-fork/_build-8.0-Debug/mysql-test/var/tmp/mysqld.1/data_new';`
2/ Check synchronization coordinate from log_status is the same as in file:
```
mysql> select local from performance_schema.log_status;
+------------------------------------------------------------------------------------------------------------------------------------+
| local                                                                                                                              |
+------------------------------------------------------------------------------------------------------------------------------------+
| {"gtid_executed": "e0e4654f-00fb-11ef-815f-95ead878d1b0:1-3", "binary_log_file": "master-bin.000001", "binary_log_position": 1019} |
+------------------------------------------------------------------------------------------------------------------------------------+
1 row in set (0.01 sec)

2024-03-29T21:35:31.654256Z 13 [Note] [MY-013273] [Clone] Plugin Clone reported: 'Server: w_local: {"gtid_executed": "279dd150-f06c-11ee-abb5-60fac8fc94f1:1-4", "binary_log_file": "master-bin.000001", "binary_log_position": 1019}.'
```
```
# with binlog
[sunxiayi@devvm7592.atn0 ~/mysql/mysql-fork/_build-8.0-Debug/mysql-test/var/tmp/mysqld.1/data_new_1/#clone (128f0405)]$ cat '#synchronization_coordinates'
gtid_from_log_status
2287bbe2-02be-11ef-aef5-9d0d68e44093:1-2
binary_log_file
master-bin.000001
binary_log_position
736
gtid_from_binlog_file_offset
2287bbe2-02be-11ef-aef5-9d0d68e44093:1-2

# without binlog
[sunxiayi@devvm7592.atn0 ~/mysql/mysql-fork/_build-8.0-Debug/mysql-test/var/tmp/mysqld.1/data_new_2/#clone (128f0405)]$ cat '#synchronization_coordinates'
gtid_from_log_status
2287bbe2-02be-11ef-aef5-9d0d68e44093:1-2
binary_log_file
master-bin.000001
binary_log_position
736
```

### Remote clone, normal
1/ Take udb35350.ftw5:3301 as the test server, install my debug build. Take udb12221.atn5:3301 as the client server, install my debug build. Install plugin on both servers.
2/ On udb12221.atn5:3301, do 
```
(admin:sys.database@udb12221.atn5 3301) [(none)]> SET GLOBAL clone_valid_donor_list = 'udb35350.ftw5.facebook.com:3301';
(admin:sys.database@udb12221.atn5 3301) [(none)]> set global clone_enable_compression="on";
(admin:sys.database@udb12221.atn5 3301) [(none)]> set enable_block_stale_hlc_read=0;
(admin:sys.database@udb12221.atn5 3301) [(none)]> set allow_noncurrent_db_rw=on;
(admin:sys.database@udb12221.atn5 3301) [(none)]> CLONE INSTANCE FROM 'dba_scripts:sys.database'@'udb35350.ftw5.facebook.com':3301 IDENTIFIED BY '' REQUIRE SSL;
```
3/ Check the new file is written correctly
4/ Repeat the clone, file is overwritten correctly.

### Remote clone, sev scenario, apply-logs
Issue a clone command using instance whose mysql.gtid_executed table has hole as donor in the raft world. Also this is copying from secondary to secondary meaning binlog is an apply-logs. Check file on recipient is correct.

### Remote clone, donor and client version mismatch
Only update recipient. Clone finishes, file is not there.
```
[root@udb12221.atn5 /data/mysql/3304/#clone]# ls
'#old_files'  '#replace_files'  '#status_fix'  '#view_progress'  '#view_status'
```
Only update donor. Clone finishes, file is not there.
```
[root@udb12221.atn5 /data/mysql/3305/#clone]# ls
'#old_files'  '#replace_files'  '#status_fix'  '#view_progress'  '#view_status'
```
Update both and do a successful clone. Then only update client. File is not there.

Reviewers:

Subscribers:

Tasks:

Tags:

Differential Revision: https://phabricator.intern.facebook.com/D55614528


mysql-test/suite/clone/t/local_create_synchronization_coordinates.test
```diff
@@ -0,0 +1,104 @@
+# Test after local clone command, synchronization_coordinates file is created
```
laurynas-biveinis:
```suggestion
# Test that after a local clone command, synchronization_coordinates file is created
```

mysql-test/suite/clone/t/local_create_synchronization_coordinates.test
```diff
@@ -0,0 +1,104 @@
...
+# synchronization_coordinates file contains 4 key/val pairs, only examine 3 Here
```
laurynas-biveinis:
```suggestion
# containing 4 key/val pairs, we only examine 3 here
```
(refill the paragraph with the above line )

mysql-test/suite/clone/t/local_create_synchronization_coordinates.test
```diff
@@ -0,0 +1,104 @@
...
+# excluding gtid get from binlog_file/offset
+
+--source include/have_example_plugin.inc
```
laurynas-biveinis: Why?

mysql-test/suite/clone/t/local_create_synchronization_coordinates.test
laurynas-biveinis: Placing the tests in `rocksdb` suite tests synchronization between binlog and InnoDB. Consider moving it to `rocksdb_clone` and adding MyRocks tables too.

mysql-test/suite/clone/t/local_create_synchronization_coordinates.test
```diff
@@ -0,0 +1,104 @@
...
+
+# Cleanup
+SET @@GLOBAL.GTID_MODE=ON_PERMISSIVE;
+SET @@GLOBAL.GTID_MODE=OFF_PERMISSIVE;
+SET @@GLOBAL.GTID_MODE=OFF;
+SET @@GLOBAL.ENFORCE_GTID_CONSISTENCY=OFF;
+DROP TABLE t1;
+UNINSTALL PLUGIN clone;
+--force-rmdir $CLONE_DATADIR
+--exec rm -f $MYSQLTEST_VARDIR/tmp/v_local.json;
```
laurynas-biveinis: Use the MTR command, IIIRC `remove-file`

plugin/clone/include/clone.h
```diff
@@ -135,8 +135,11 @@ const uint32_t CLONE_PROTOCOL_VERSION_V2 = 0x0101;
 /** Send more configurations required by recipient. */
 const uint32_t CLONE_PROTOCOL_VERSION_V3 = 0x0102;
 
+/** Send GTID executed set after engine syncronization. */
+const uint32_t CLONE_PROTOCOL_VERSION_V4 = 0x0103;
```
laurynas-biveinis, line 139: This will conflict if/when upstream bumps the protocol version on its own. No perfect way to solve this, but if you take a higher bit, say `V128`, perhaps they will not be able to burn through that many protocol versions.

laurynas-biveinis, line 139:
```suggestion
constexpr uint32_t CLONE_PROTOCOL_VERSION_V4 = 0x0103;
```

plugin/clone/include/clone.h
```diff
@@ -135,8 +135,11 @@ const uint32_t CLONE_PROTOCOL_VERSION_V2 = 0x0101;
...
+
 /**  Clone protocol latest version */
-const uint32_t CLONE_PROTOCOL_VERSION = CLONE_PROTOCOL_VERSION_V3;
+const uint32_t CLONE_PROTOCOL_VERSION = CLONE_PROTOCOL_VERSION_V4;
```
laurynas-biveinis, line 142:
```suggestion
constexpr uint32_t CLONE_PROTOCOL_VERSION = CLONE_PROTOCOL_VERSION_V4;
```

plugin/clone/include/clone.h
```diff
@@ -194,6 +197,10 @@ typedef enum Type_Command_Response : uchar {
   /** Additional configuration : introduced in version 0x0102 */
   COM_RES_CONFIG_V3,
 
+  /** GTID executed set after engine synchronization: introduced in version
+     0x0103 */
```
laurynas-biveinis: Indentation of the second comment line

plugin/clone/include/clone_client.h
```diff
@@ -580,6 +581,11 @@ class Client {
   /** Destroy PFS mutex for table. */
   static void uninit_pfs();
 
+  /** Write syncronization coordinate to a new file.
```
laurynas-biveinis:
```suggestion
  /** Write synchronization coordinate to a new file.
```

plugin/clone/include/clone_client.h
```diff
@@ -580,6 +581,11 @@ class Client {
...
+  @param[in]	syncronization_coordinate	syncronization coordinate */
+  void persist_syncronization_coordinate(
```
laurynas-biveinis:
```suggestion
  void persist_synchronization_coordinate(
```

plugin/clone/include/clone_client.h
```diff
@@ -701,6 +707,13 @@ class Client {
   @return error code */
   int set_descriptor(const uchar *buffer, size_t length);
 
+  /** Set syncronization coordinate by recording that in a new file,
+  after extracting key and value from the buffer.
+  @param[in]	buffer	serialized data descriptor
+  @param[in]	length	length of serialized data
+  @return error code */
+  int set_syncronization_coordinate(const uchar *buffer, size_t length);
```
laurynas-biveinis:
```suggestion
  [[nodiscard]] int set_synchronization_coordinate(const uchar *buffer, size_t length);
```

plugin/clone/include/clone_client.h
```diff
@@ -29,6 +29,7 @@ Clone Plugin: Client Interface
 #ifndef CLONE_CLIENT_H
 #define CLONE_CLIENT_H
 
+#include "clone.h"
```
laurynas-biveinis: Why?

plugin/clone/include/clone_common.h
```diff
@@ -17,6 +17,7 @@
 #ifndef CLONE_COMMON_H
 #define CLONE_COMMON_H
 
+#include "clone.h"
```
laurynas-biveinis, line 20: Why?

plugin/clone/include/clone_common.h
```diff
@@ -25,7 +26,37 @@ class Ha_clone_common_cbk : public Ha_clone_cbk {
...
-  [[nodiscard]] int synchronize_engines() override;
+ protected:
+  /** Perform the cross-engine synchronization for logs: execute a
+    performance_schema.log_status query, and call set_log_stop for each storage
+    engine with its part of STORAGE_ENGINES column JSON object from that query.
+    Also assign synchronization coordinates get from log_status table to the
+    parameter synchronization_coordinates.
+    @param[in]	synchronization_coordinates	syncronization coordinates
+    @return error code */
+  int synchronize_logs(Key_Values &synchronization_coordinates);
```
laurynas-biveinis:
```suggestion
  [[nodiscard]] int synchronize_logs(Key_Values &synchronization_coordinates);
```

plugin/clone/include/clone_common.h
```diff
@@ -25,7 +26,37 @@ class Ha_clone_common_cbk : public Ha_clone_cbk {
...
+
+ private:
+  /** Get json object from the json wrapper after doing json validations
+    @param[in]	object_name the object name, used for debug printing
+    @param[in]	json_wrapper json wrapper to be converted to be json object
+    @return converted json object */
+  const Json_object *get_json_object(const char *object_name,
```
laurynas-biveinis:
```suggestion
  [[nodiscard]] const Json_object *get_json_object(const char *object_name,
```

plugin/clone/include/clone_common.h
```diff
@@ -25,7 +26,37 @@ class Ha_clone_common_cbk : public Ha_clone_cbk {
...
+      local_repl_info has 3 pieces of information: gtid, binlog_file, offset.
+      We will also get gtid corresponding to the binlog_file and offset and
+    record that gtid as well. This is to sanity check that this gtid we get from
+    binlog_file and offset is same as what we have in local_repl_info.
+    @param[in]	local_repl_info the "local" column we get from log_status table
+    @param[in]	synchronization_coordinates Key_Values of gtid(from log_status
+    table), binlog_file, binlog_offset and gtid(from binlog_file and
+    binlog_offset)
+    @return error code */
+  int populate_synchronization_coordinates(
```
laurynas-biveinis:
```suggestion
  [[nodiscard]] int populate_synchronization_coordinates(
```

plugin/clone/include/clone_server.h
```diff
@@ -127,6 +127,12 @@ class Server {
     return m_protocol_version < CLONE_PROTOCOL_VERSION_V3;
   }
 
+  /** @return true iff sending synchronization coordinates(gtid & binlog info)
+   */
+  bool should_send_synchronization_coordinates() const {
```
laurynas-biveinis:
```suggestion
  [[nodiscard]] bool should_send_synchronization_coordinates() const {
```

plugin/clone/include/clone_status.h
```diff
@@ -31,6 +31,8 @@ Clone Plugin: Client Status Interface
 
 #include <mysql/components/services/pfs_plugin_table_service.h>
 #include <array>
+#include <random>
+#include "clone.h"
```
laurynas-biveinis: Why?

plugin/clone/include/clone_status.h
```diff
@@ -316,6 +326,9 @@ class Status_pfs : public Table_pfs {
 
     /** Clone GTID set */
     std::string m_gtid_string;
+
+    /** Synchronization coordinates. */
```
laurynas-biveinis: No need for a Doxygen comment that just repeats the variable name

plugin/clone/src/clone_client.cc
```diff
@@ -1660,6 +1664,22 @@ int Client::set_locators(const uchar *buffer, size_t length) {
   return (err);
 }
 
+void Client::persist_syncronization_coordinate(
+    const Key_Value &syncronization_coordinate) {
```
laurynas-biveinis:
```suggestion
    const Key_Value &synchronization_coordinate) {
```

plugin/clone/src/clone_client.cc
```diff
@@ -1962,4 +1982,9 @@ int Client_Cbk::apply_cbk(Ha_clone_file to_file, bool apply_file,
   return (err);
 }
 
+[[nodiscard]] int Client_Cbk::synchronize_engines() {
+  my_error(ER_NOT_SUPPORTED_YET, MYF(0), "Remote Clone Client");
```
laurynas-biveinis: Add something to crash the debug build, i.e. `assert(0)`

laurynas-biveinis: The "YET" part might be slightly confusing. Maybe `ER_FEATURE_UNSUPPORTED`?

plugin/clone/src/clone_status.cc
```diff
@@ -465,6 +470,34 @@ void Status_pfs::Data::write(bool write_error) {
...
+  }
+  if (remove) {
+    std::filesystem::remove(file_name);
+    return;
+  }
+  m_synchronization_coordinates.push_back(coordinate);
+  std::ofstream status_file;
+  status_file.open(file_name, std::ofstream::out | std::ofstream::trunc);
+  if (!status_file.is_open()) {
+    return;
```
laurynas-biveinis, line 490: Error log diagnostics here?

plugin/clone/src/clone_status.cc
```diff
@@ -465,6 +470,34 @@ void Status_pfs::Data::write(bool write_error) {
...
+  }
+  for (auto &coordinate : m_synchronization_coordinates) {
+    status_file << coordinate.first << std::endl;
```
laurynas-biveinis: Never use `std::endl`
```suggestion
    status_file << coordinate.first << '\n';
```

plugin/clone/src/clone_status.cc
```diff
@@ -53,6 +54,10 @@ SERVICE_TYPE_NO_CONST(pfs_plugin_column_text_v1) *mysql_pfscol_text = nullptr;
 const char CLONE_RECOVERY_FILE[] =
     CLONE_FILES_DIR FILE_PREFIX "status_recovery";
 
+/** Clone synchronization coordinate file. */
+const char CLONE_SYNCHRONIZATION_COORDINATES_FILE[] =
```
laurynas-biveinis, line 58: Consider `constexpr std::string_view`

plugin/clone/include/clone_common.h
laurynas-biveinis: Consider 
```suggestion
  const Json_object *get_json_object(std::string_view object_name,
```

plugin/clone/include/clone_common.h
```diff
@@ -25,7 +26,37 @@ class Ha_clone_common_cbk : public Ha_clone_cbk {
...
+      const Json_object *local_repl_info,
```
laurynas-biveinis: Here and everywhere else: replace pointer args with reference args whenever possible (i.e. the pointer cannot be nullptr, no arithmetics done with it). This applies to out parameters too

plugin/clone/src/clone_client.cc
```diff
@@ -1660,6 +1664,22 @@ int Client::set_locators(const uchar *buffer, size_t length) {
...
+void Client::persist_syncronization_coordinate(
+    const Key_Value &syncronization_coordinate) {
+  mysql_mutex_lock(&s_table_mutex);
+  s_status_data.write_synchronization_coordinate(syncronization_coordinate);
+  mysql_mutex_unlock(&s_table_mutex);
+}
+
+int Client::set_syncronization_coordinate(const uchar *packet, size_t length) {
+  Key_Value syncronization_coordinate;
+  auto err = extract_key_value(packet, length, syncronization_coordinate);
```
laurynas-biveinis:
```suggestion
  const auto err = extract_key_value(packet, length, syncronization_coordinate);
```

plugin/clone/src/clone_client.cc
```diff
@@ -1660,6 +1664,22 @@ int Client::set_locators(const uchar *buffer, size_t length) {
...
+  if (err == 0) {
+    persist_syncronization_coordinate(syncronization_coordinate);
+  }
+  return (err);
```
laurynas-biveinis:
```suggestion
  return err;
```

plugin/clone/src/clone_client.cc
```diff
@@ -1962,4 +1982,9 @@ int Client_Cbk::apply_cbk(Ha_clone_file to_file, bool apply_file,
   return (err);
 }
 
+[[nodiscard]] int Client_Cbk::synchronize_engines() {
+  my_error(ER_NOT_SUPPORTED_YET, MYF(0), "Remote Clone Client");
+  return (ER_NOT_SUPPORTED_YET);
```
laurynas-biveinis:
```suggestion
  return ER_NOT_SUPPORTED_YET;
```

plugin/clone/src/clone_common.cc
```diff
@@ -16,8 +16,10 @@
 
 #include "plugin/clone/include/clone_common.h"
 
+#include <filesystem>
```
laurynas-biveinis, line 19: Good time as any to start using C++17 filesystem :)

plugin/clone/src/clone_server.cc
```diff
@@ -730,6 +730,39 @@ int Server_Cbk::buffer_cbk(uchar *from_buffer, uint buf_len) {
...
+// Also send the synchronization coordinates to client.
+int Server_Cbk::synchronize_engines() {
+  Key_Values synchronization_coordinates;
+  auto err = synchronize_logs(synchronization_coordinates);
+  if (err != 0) {
+    return err;
+  }
+  auto server = get_clone_server();
+  if (server->should_send_synchronization_coordinates()) {
+    for (auto &coordinate : synchronization_coordinates) {
```
laurynas-biveinis:
```suggestion
    for (const auto &coordinate : synchronization_coordinates) {
```

laurynas-biveinis (commented):
I am sorry for a partial review, but a full review depends on how some of the current comments are going to be resolved.

What is the plan with the two sets of binlog positions, once the mismatch is fixed or its absence is confirmed?

sunxiayi:
> I am sorry for a partial review, but a full review depends on how some of the current comments are going to be resolved.
> 
> What is the plan with the two sets of binlog positions, once the mismatch is fixed or its absence is confirmed?

in the client side, we can reset replica and set global gtid_purged so it can resume replication from that gtid.

plugin/clone/include/clone.h
sunxiayi: Given we are going to move away from upstream soon I think it is okay..

sunxiayi: Given we have all the CLONE_PROTOCOL_VERSION values defined as const, and also I don't see the need for it to be evaluated at compile time, I think it's better to keep that as const.

plugin/clone/src/clone_client.cc
sunxiayi: was referring to int Client_Cbk::file_cbk which has the similar structure.

plugin/clone/src/clone_status.cc
sunxiayi: keeping consistent with the other const declarations for now. maybe later change them all.

plugin/clone/include/clone_common.h
sunxiayi: to use Key_Values type declaration.

laurynas-biveinis:
sunxiayi ***@***.***> writes:

>  What is the plan with the two sets of binlog positions, once the mismatch is fixed or its absence is confirmed?
>
> in the client side, we can reset replica and set global gtid_purged so it can resume replication from that gtid.
>

Right, but I am asking specifically about the *two sets* of binlog
positions that should be equal: one from pfs.log_status and one from the
binlog itself.

plugin/clone/include/clone.h
laurynas-biveinis: I don't have enough context here, can we discuss offline?

plugin/clone/src/clone_client.cc
laurynas-biveinis: OK then

plugin/clone/include/clone.h
laurynas-biveinis: That's fine, and OK to leave `const`, but note there is no need to allocate for it storage neither.

sunxiayi:
> sunxiayi ***@***.***> writes:
> What is the plan with the two sets of binlog positions, once the mismatch is fixed or its absence is confirmed? in the client side, we can reset replica and set global gtid_purged so it can resume replication from that gtid.
> Right, but I am asking specifically about the *two sets* of binlog positions that should be equal: one from pfs.log_status and one from the binlog itself.

If the two sets match, we would use that gtid to resume replication in client.
If they are not the same, we would abort the clone, log the mismatch details, then make plans to fix the bug.

plugin/clone/include/clone_client.h
```diff
@@ -701,6 +706,13 @@ class Client {
   @return error code */
   int set_descriptor(const uchar *buffer, size_t length);
 
+  /** Set synchronization coordinate by recording that in a new file,
+  after extracting key and value from the buffer.
+  @param[in]	buffer	serialized data descriptor
+  @param[in]	length	length of serialized data
+  @return error code */
+  int set_synchronization_coordinate(const uchar *buffer, size_t length);
```
laurynas-biveinis:
```suggestion
  [[nodiscard]] int set_synchronization_coordinate(const uchar *buffer, size_t length);
```

plugin/clone/include/clone_common.h
```diff
@@ -25,7 +26,37 @@ class Ha_clone_common_cbk : public Ha_clone_cbk {
...
+    @param[in]	object_name the object name, used for debug printing
+    @param[in]	json_wrapper json wrapper to be converted to be json object
+    @return converted json object */
+  [[nodiscard]] const Json_object *get_json_object(std::string_view object_name,
+                                                   Json_wrapper &json_wrapper);
+
+  /** Extract key and values from local_repl_info.
+      local_repl_info has 3 pieces of information: gtid, binlog_file, offset.
+      We will also get gtid corresponding to the binlog_file and offset and
+    record that gtid as well. This is to sanity check that this gtid we get from
```
laurynas-biveinis, line 50: Formatting

plugin/clone/src/clone_common.cc
```diff
@@ -98,16 +81,121 @@ int Ha_clone_common_cbk::synchronize_engines() {
   // Size reverse-engineered from ER_CLONE_SERVER_TRACE used by log_error. No
   // way to track it automatically, but very unlikely it would silently shrink.
   char msg_buf[512];
-  snprintf(msg_buf, sizeof(msg_buf), "engine positions: %.*s",
+  snprintf(msg_buf, sizeof(msg_buf), "%s: %.*s", object_name.data(),
```
laurynas-biveinis: To printf-format an `std::string_view`, you always have to use sized formatting: `%.s*`, similar to `json_str_buf` in the same call here. The reason is that string views do not have to be nul-terminated

plugin/clone/include/clone_common.h
```diff
@@ -25,7 +26,37 @@ class Ha_clone_common_cbk : public Ha_clone_cbk {
...
+    @return error code */
+  [[nodiscard]] int synchronize_logs(Key_Values &synchronization_coordinates);
+
+ private:
+  /** Get json object from the json wrapper after doing json validations
+    @param[in]	object_name the object name, used for debug printing
+    @param[in]	json_wrapper json wrapper to be converted to be json object
+    @return converted json object */
+  [[nodiscard]] const Json_object *get_json_object(std::string_view object_name,
+                                                   Json_wrapper &json_wrapper);
```
laurynas-biveinis:
```suggestion
                                                   const Json_wrapper &json_wrapper);
```

plugin/clone/include/clone_common.h
```diff
@@ -25,7 +26,37 @@ class Ha_clone_common_cbk : public Ha_clone_cbk {
...
+    @param[in]	synchronization_coordinates	synchronization coordinates
+    @return error code */
+  [[nodiscard]] int synchronize_logs(Key_Values &synchronization_coordinates);
+
+ private:
+  /** Get json object from the json wrapper after doing json validations
+    @param[in]	object_name the object name, used for debug printing
+    @param[in]	json_wrapper json wrapper to be converted to be json object
+    @return converted json object */
+  [[nodiscard]] const Json_object *get_json_object(std::string_view object_name,
```
laurynas-biveinis:
```suggestion
  [[nodiscard]] static const Json_object *get_json_object(std::string_view object_name,
```

laurynas-biveinis: It looks like the result of this method can never be a nullptr - if true, return const reference

plugin/clone/src/clone_common.cc
```diff
@@ -98,16 +81,121 @@ int Ha_clone_common_cbk::synchronize_engines() {
...
+  std::filesystem::path file_path(full_file_name);
+  if (!std::filesystem::exists(file_path)) {
+    return 0;
+  }
+  char info_mesg[512];
+  snprintf(info_mesg, sizeof(info_mesg),
+           "Reading gtid from binlog %s offset %s", full_file_name,
+           std::to_string(binary_log_position_int).c_str());
+  log_error(nullptr, false, 0, info_mesg);
+  MYSQL_BIN_LOG::enum_read_gtids_from_binlog_status ret =
```
laurynas-biveinis:
```suggestion
  const auto ret =
```

plugin/clone/src/clone_common.cc
```diff
@@ -98,16 +81,121 @@ int Ha_clone_common_cbk::synchronize_engines() {
...
+      mysql_bin_log.read_gtids_from_binlog(full_file_name, &gtid_executed, NULL,
+                                           NULL, &sid_map, false, false,
+                                           binary_log_position_int);
+  if (ret == MYSQL_BIN_LOG::ERROR || ret == MYSQL_BIN_LOG::TRUNCATED) {
+    return ER_BINLOG_FILE_OPEN_FAILED;
+  } else {
+    char *gtid_from_binlog_file_offset;
```
laurynas-biveinis, line 161: There is `unique_ptr_my_free`, but probably not worth it here

plugin/clone/src/clone_common.cc
```diff
@@ -133,5 +221,4 @@ int Ha_clone_common_cbk::synchronize_engines() {
 
   return 0;
 }
-
```
laurynas-biveinis, line 136: Spurious diff

plugin/clone/src/clone_local.cc
```diff
@@ -392,4 +392,17 @@ int Local_Callback::apply_cbk(Ha_clone_file to_file, bool apply_file,
...
   return (error);
 }
+
+[[nodiscard]] int Local_Callback::synchronize_engines() {
+  Key_Values synchronization_coordinates;
+  auto err = synchronize_logs(synchronization_coordinates);
+  if (err != 0) {
+    return err;
+  }
+  auto client = get_clone_client();
```
laurynas-biveinis, line 402: References and (non-smart) pointers should be pulled up from auto:
```suggestion
  auto *client = get_clone_client();
```

plugin/clone/src/clone_local.cc
```diff
@@ -392,4 +392,17 @@ int Local_Callback::apply_cbk(Ha_clone_file to_file, bool apply_file,
...
+  for (auto &coordinate : synchronization_coordinates) {
```
laurynas-biveinis:
```suggestion
  for (const auto &coordinate : synchronization_coordinates) {
```

plugin/clone/src/clone_server.cc
```diff
@@ -730,6 +730,39 @@ int Server_Cbk::buffer_cbk(uchar *from_buffer, uint buf_len) {
   return (err);
 }
 
+int Server_Cbk::send_synchronization_coordinate(
+    Server *server, const Key_Value &synchronization_coordinate) {
+  String_Key key(synchronization_coordinate.first);
+  String_Key val(synchronization_coordinate.second);
+  auto err = server->send_key_value(COM_RES_GTID_V4, key, val);
+  return (err);
```
laurynas-biveinis:
```suggestion
  return err;
```

plugin/clone/src/clone_server.cc
```diff
@@ -730,6 +730,39 @@ int Server_Cbk::buffer_cbk(uchar *from_buffer, uint buf_len) {
...
+// stages. When MyRocks is the sole storage engine, it should be called after
+// creating the final checkpoint.
+// Also send the synchronization coordinates to client.
+int Server_Cbk::synchronize_engines() {
+  Key_Values synchronization_coordinates;
+  auto err = synchronize_logs(synchronization_coordinates);
+  if (err != 0) {
+    return err;
+  }
+  auto server = get_clone_server();
```
laurynas-biveinis, line 754:
```suggestion
  auto *server = get_clone_server();
```

plugin/clone/src/clone_status.cc
```diff
@@ -465,6 +470,38 @@ void Status_pfs::Data::write(bool write_error) {
...
+  /* Append data directory if cloning to different place. */
+  if (!is_local()) {
+    file_name.assign(m_destination);
+    file_name.append(FN_DIRSEP);
+    file_name.append(CLONE_SYNCHRONIZATION_COORDINATES_FILE);
+  } else {
+    file_name.assign(CLONE_SYNCHRONIZATION_COORDINATES_FILE);
+  }
+  if (remove) {
+    std::filesystem::remove(file_name);
```
laurynas-biveinis: There could be a reason not to use `std::filesystem`: do we need to issue diagnostics in the case of file not found here? If yes, we need to print to error log, then use `my_delete` from mysys

plugin/clone/src/clone_status.cc
```diff
@@ -465,6 +470,38 @@ void Status_pfs::Data::write(bool write_error) {
...
+    return;
+  }
+  m_synchronization_coordinates.push_back(coordinate);
+  std::ofstream status_file;
+  status_file.open(file_name, std::ofstream::out | std::ofstream::trunc);
+  if (!status_file.is_open()) {
+    char msg_buf[512];
+    snprintf(msg_buf, sizeof(msg_buf), "Cannot open %s for read",
```
laurynas-biveinis: Definitely not for read here:
```suggestion
    snprintf(msg_buf, sizeof(msg_buf), "Cannot create %s",
```
... or similar

plugin/clone/src/clone_status.cc
```diff
@@ -465,6 +470,38 @@ void Status_pfs::Data::write(bool write_error) {
...
+             CLONE_SYNCHRONIZATION_COORDINATES_FILE);
+    log_error(thd_get_current_thd(), true, ER_CANT_OPEN_FILE, msg_buf);
+    return;
+  }
+  for (auto &coordinate : m_synchronization_coordinates) {
```
laurynas-biveinis:
```suggestion
  for (const auto &coordinate : m_synchronization_coordinates) {
```

plugin/clone/include/clone_common.h
laurynas-biveinis: The header file does not need to see the full declaration, forward declare it instead

plugin/clone/include/clone_status.h
```diff
@@ -212,6 +212,12 @@ class Status_pfs : public Table_pfs {
     @param[in]	write_error	write error information. */
     void write(bool write_error);
 
+    /** Write to synchronization gtid file.
+    @param[in]	coordinate	synchronization coordinate sent from server.
+    @param[in]  remove      remove the file if exists */
+    void write_synchronization_coordinate(const Key_Value &coordinate,
```
laurynas-biveinis: The API is overloaded. "remove" is not "write", and the 1st argument gets ignored then. Please split into two. They might need to share a helper to construct the file path.

laurynas-biveinis (commented):
Overall looks good, minor comments only

plugin/clone/src/clone_status.cc
sunxiayi: if file not found, it could be because it is a new instance or during version update etc, which I think is not something we particularly care about.

plugin/clone/include/clone_common.h
sunxiayi: Key_Values is a type alias. How to forward declare that?

plugin/clone/src/clone_local.cc
sunxiayi: Across this file, pointers are not pulled up... keeping it consistent might be better for understanding.

plugin/clone/include/clone_common.h
laurynas-biveinis: Right, in this case the "forward declaration" would be the duplication of the `using` directives. Please decide if that or an include looks better

plugin/clone/src/clone_local.cc
laurynas-biveinis: I believe that making local code better trumps consistency, especially if the surrounding code is not owned (at the moment) by us, otherwise it risks never getting improved, but it's a matter of opinion

plugin/clone/include/clone_common.h
sunxiayi: what's your opinion on returning by value and let RVO take care of the copy? here it is not returning a class variable but json_wrapper.get_dom(); Not sure if returning by reference here is the best practice

laurynas-biveinis: I like RVO and returning the value, but here it wouldn't apply: `json_wrapper.get_dom()` returns a pointer to its internals. If we dereference it and return by value, we copy, nothing to optimize away.

mysql-test/suite/clone/t/local_create_synchronization_coordinates.test
laurynas-biveinis: Resolved

mysql-test/suite/clone/t/local_create_synchronization_coordinates.test
laurynas-biveinis: Comment still relevant

mysql-test/suite/clone/t/local_create_synchronization_coordinates.test
```diff
@@ -0,0 +1,103 @@
+# Test that after a local clone command, synchronization_coordinates file is created
+# containing 4 key/val pairs, we only examine 3 here
+# excluding gtid get from binlog_file/offset
+
+--source ../include/clone_connection_begin.inc
+
+# Populate gtid
+SET @@GLOBAL.ENFORCE_GTID_CONSISTENCY=ON;
```
laurynas-biveinis: Ideally the testcases have to support arbitrary `--mysqld=--any-server-option=value` run arguments. This testcase requires GTID, so in general case we do the pattern of saving the start global value and restoring at the end. Since `gtid_mode` is a special in that setting it goes through `_permissive` stages, in this case it is simpler to gate the testcase with `--source have_gtid.inc`, and adding a testcase option file with `--gtid-mode=ON`

mysql-test/suite/clone/t/local_create_synchronization_coordinates.test
```diff
@@ -0,0 +1,103 @@
+# Test that after a local clone command, synchronization_coordinates file is created
```
laurynas-biveinis, line 1: Add `--source have_clone_plugin.inc` in both testcases

plugin/clone/include/clone_common.h
```diff
@@ -17,6 +17,7 @@
 #ifndef CLONE_COMMON_H
 #define CLONE_COMMON_H
 
+#include "clone.h"
 #include "sql/handler.h"
```
laurynas-biveinis, line 21: `#include <string_view>`

plugin/clone/src/clone_client.cc
```diff
@@ -26,6 +26,7 @@ Clone Plugin: Client implementation
 
 */
 #include <inttypes.h>
+#include <cassert>
```
laurynas-biveinis, line 29: Suspicious formatting (clang-format sorts headers alphabetically in each section)

plugin/clone/src/clone_common.cc
```diff
@@ -16,8 +16,10 @@
 
 #include "plugin/clone/include/clone_common.h"
 
+#include <filesystem>
 #include <string>
 #include <utility>
+#include "sql/binlog.h"
```
laurynas-biveinis: Move to the next section

plugin/clone/src/clone_common.cc
```diff
@@ -98,16 +81,118 @@ int Ha_clone_common_cbk::synchronize_engines() {
...
+  }
+  const auto &binary_log_position_json = *binary_log_position_json_ptr;
+  assert(binary_log_position_json.json_type() == enum_json_type::J_INT);
+  const auto &binary_log_position_int_json =
+      static_cast<const Json_int &>(binary_log_position_json);
+  const auto &binary_log_position_int = binary_log_position_int_json.value();
+  synchronization_coordinates.push_back(
+      {binary_log_position_key_str, std::to_string(binary_log_position_int)});
+
+  // get gtid from binlog file/pos
```
laurynas-biveinis, line 135: On the second thought, please add a comment here why we are getting the same information second time in a different way (hunting for a bug, will update once debugged)

plugin/clone/include/clone_status.h
```diff
@@ -212,6 +212,17 @@ class Status_pfs : public Table_pfs {
...
+    /** Write to synchronization gtid file.
+    @param[in]	coordinate	synchronization coordinate sent from server. */
+    void write_synchronization_coordinate(const Key_Value &coordinate);
+
+    /** Delete synchronization coordinate file, if the file does not exist, will
+     * not report error.  */
+    void delete_synchronization_coordinate_file();
+
+    /** Get full file name from the given file_name. */
+    std::string get_full_file_name(const char *file_name);
```
laurynas-biveinis: Does the argument convert to `std::string_view` easily?

plugin/clone/src/clone_status.cc
```diff
@@ -465,6 +474,33 @@ void Status_pfs::Data::write(bool write_error) {
   status_file.close();
 }
 
+void Status_pfs::Data::write_synchronization_coordinate(
+    const Key_Value &coordinate) {
+  std::string file_name =
```
laurynas-biveinis:
```suggestion
  const auto file_name =
```

plugin/clone/src/clone_status.cc
```diff
@@ -465,6 +474,33 @@ void Status_pfs::Data::write(bool write_error) {
...
+  }
+  for (const auto &coordinate : m_synchronization_coordinates) {
+    status_file << coordinate.first << '\n';
+    status_file << coordinate.second << '\n';
+  }
+  status_file.close();
+}
+
+void Status_pfs::Data::delete_synchronization_coordinate_file() {
+  std::string file_name =
```
laurynas-biveinis:
```suggestion
  const auto file_name =
```

mysql-test/suite/clone/t/local_create_synchronization_coordinates.test
laurynas-biveinis: Resolved

plugin/clone/src/clone_client.cc
sunxiayi: it is weird. I used internal standard clang-format tooling for formatting.

plugin/clone/src/clone_status.cc
sunxiayi: would const std::string suffice? why is declaring the type directly so frowned upon? for types that are not long I personally like to read the direct type for better understanding.

laurynas-biveinis: TL;DR: yes "const std::string" is OK too. 
Having said that, I believe AAA (Almost Always Auto) is the right C++ style: usually we don't care about the exact types of local variables, while we do care about the parameter and return types of methods (incidentally it is how Rust looks by default). Should we ever refactor and change the type, auto would reduce the diff churn. In rare cases where the type of a local variable matters and is not obvious from its initializer, then usually IDE support can show it.

plugin/clone/include/clone_common.h
sunxiayi: what is it used for?

plugin/clone/src/clone_client.cc
sunxiayi: the internal linter forces me to do this. I am guessing it differentiates on lib that has the .h suffix when formatting.

plugin/clone/include/clone_common.h
laurynas-biveinis: `std::string_view` for a `get_json_object` parameter

plugin/clone/src/clone_client.cc
laurynas-biveinis: OK then

plugin/clone/include/clone_common.h
sunxiayi: Weird I can use string_view without declaring the header still?

laurynas-biveinis: Probably one of the other headers pulls it in, but
1) transitive includes are brittle as the other headers (including standard library ones) may stop including it at any time
2) source files and headers should include their dependencies directly - this is also something that tooling like include-what-you-use and clang-tidy enforce

PR #1452, opened 2024-05-03
Cleanup Rdb_transaction::m_is_tx_flag

laurynas-biveinis:
- The flag does not mark the failure of the whole transaction, but rather
  whether the previous statement was rolled back. Rename the field to
  m_was_last_stmt_rolledback accordingly, rename its setter set_tx_failed to
  set_last_stmt_rollback, make the field private, introduce a getter
  was_last_stmt_rolledback, add it to the class invariant as much as possible.
- Rename commit_or_rollback method to commit_or_rollback_autocommit_trx
- Extract a private helper on_finish, which is called from on_commit and
  on_rollback.
- Introduce an abstract virtual method start_autocommit_stmt
- Remove an outdated comment from ha_rocksdb::external_lock

### DIFF
--- a/storage/rocksdb/ha_rocksdb.cc
+++ b/storage/rocksdb/ha_rocksdb.cc
@@ -3610,6 +3610,15 @@ class Rdb_transaction {
   bool m_bulk_index_transaction = false;
   bool m_dd_transaction = false;
 
+  /** Set in rocksdb_rollback whenever a single-statement autocommit transaction
+  or a statement in a multistatement transaction is being rolled back. Its value
+  is used by ha_rocksdb::external_lock to commit or rollback the autocommit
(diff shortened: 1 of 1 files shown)


laurynas-biveinis:
This fell out from debugging range locking snapshot ignore start and stop points for DML statements, and will be a dependency PR for the range locking.

laurynas-biveinis:
Havin slept on this, I don't like that the flag is set for two cases but read only for one of them. Let's try reducing its setting to the one case, and maybe the reset code will become simpler too.

laurynas-biveinis:
Even better, it should be possible to remove the flag altogether

laurynas-biveinis:
Superseded by https://github.com/facebook/mysql-5.6/pull/1453

PR #1454, opened 2024-05-10
Merge all the no-op MyRocks system variable update functions

laurynas-biveinis:
Code removal, they differed only by their names.

### DIFF
--- a/storage/rocksdb/ha_rocksdb.cc
+++ b/storage/rocksdb/ha_rocksdb.cc
@@ -372,10 +372,10 @@ static void rocksdb_flush_all_memtables() {
   }
 }
 
-static void rocksdb_delete_column_family_stub(THD *const /* thd */,
-                                              struct SYS_VAR *const /* var */,
-                                              void *const /* var_ptr */,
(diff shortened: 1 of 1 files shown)


storage/rocksdb/ha_rocksdb.cc
```diff
@@ -372,10 +372,10 @@ static void rocksdb_flush_all_memtables() {
   }
 }
 
-static void rocksdb_delete_column_family_stub(THD *const /* thd */,
-                                              struct SYS_VAR *const /* var */,
-                                              void *const /* var_ptr */,
-                                              const void *const /* save */) {}
+// A stub sysvar update function that does nothing. Used instead of nullptr to
+// indicate that the sysvar is not read-only.
+static void rw_sysvar_update_noop(THD *const, SYS_VAR *const, void *const,
```
luqun: nit: maybe start with rocksdb_?

laurynas-biveinis: Done
(I mildly prefer not having prefixes when we have namespaces and in this instance static functions)

PR #1455, opened 2024-05-13
update sys.sys_config table for RocksDB DDSE

luqun:
Summary:
Currently there are few tables still running INNDOB. and sys.sys_config table is one of them. although it doesn't belong to DD table or SYSTEM TABLE, also convert this table into RocksDB table

Test Plan:

./mtr --suite=sysschema --mysqld=--default_dd_system_storage_engine=rocksdb


mysql-test/suite/sysschema/t/mysqldump.test
```diff
@@ -1,5 +1,7 @@
 -- source include/no_valgrind_without_big.inc
 
+SET @@global.SKIP_SYS_TABLES_ENGINE_CHECK = 1;
```
laurynas-biveinis, line 3: What happens without this?

luqun: I should add some comments about this. it will fail to create mysql.gtid_executed table due to its engine is innodb while DDSE is rocksdb.

PR #1458, opened 2024-05-23
Make MyRocks transactions never expire additional rows with TTL while running

laurynas-biveinis:
The TTL row filtering is governed by Rdb_transaction::m_snapshot_timestamp
field. This field sometimes get reset to zero when a snapshot is released,
indicating that it should be reinitialized at the next snapshot, or to the
current timestamp (when a single statement in a multi-statement transaction gets
rolled back). The timestamp going forward means that additional rows may be
hidden in the same transaction if they happened to expire between the two
snapshots. Change it so that additional rows are never expired:

- Only initialize the field once per transaction, when the first snapshot is
  created.
- Do not touch it for single-statement rollbacks in multi-statement
  transactions.
- Rename the field to m_earliest_snapshot_ts to better reflect what it does.
- Change the global transaction list to be sorted by the oldest snapshot
  timestamp, so that the oldest global one can be found by reading the
  transaction at the list head. The transactions without snapshot are inserted
  to the its tail, and reinserted as needed upon acquiring the first snapshot.
  To make reinsertions fast, a third list pointer is maintained that points to
  the newest transaction with a snapshot.
- Extend the SHOW ENGINE ROCKSDB TRANSACTION STATUS output with transactions
  that currently do not have a snapshot, for both cases of having a snapshot
  before and never having had one.
- Add a test to rocksdb.ttl_primary_read_filtering for the behavior
  difference introduced by this commit.

At the same time:
- Initialize m_earliest_snapshot_ts from the binlog HLC instead of the current
  system time.

### DIFF
--- a/storage/rocksdb/ha_rocksdb.cc
+++ b/storage/rocksdb/ha_rocksdb.cc
@@ -110,6 +110,7 @@
 #include "./rdb_threads.h"
 #include "./sql_dd.h"
 #include "my_rapidjson_size_t.h"
+#include "rdb_compact_filter.h"
 
 using namespace std::string_view_literals;
 
@@ -3508,26 +3509,47 @@ static int rdb_dbug_set_ttl_read_filter_ts();
 }
 
 // The global list of currently-active MyRocks transactions, implemented as a
-// doubly-linked intrusive linked list.
+// doubly-linked intrusive linked list. The list is sorted with the oldest
+// snapshot timestamp transaction at the head. The tail of the list contains
+// transactions that did not have yet a snapshot opened and there is a pointer
+// to the newest transaction with a snapshot in the middle of the list.
+// Transactions without snapshots are inserted to the tail, and reinserted to
+// the middle as they acquire their first snapshots.
 class [[nodiscard]] Rdb_transaction_list final {
  private:
   static Rds_mysql_mutex mutex;
   static Rdb_transaction *head;
+  static Rdb_transaction *tail;
+  static Rdb_transaction *newest_tx_with_snapshot;
+
+  static void do_erase(Rdb_transaction *tx);
+  static void fixup_newest_snapshot_after_erase(
+      const Rdb_transaction *erased_tx);
+  static void do_insert_without_snapshot(Rdb_transaction *tx);
+  [[nodiscard]] static std::int64_t oldest_timestamp_locked();
 
  public:
   static void init() { mutex.init(key_mutex_tx_list, MY_MUTEX_INIT_FAST); }
 
   static void shutdown() {
     assert(head == nullptr);
+    assert(tail == nullptr);
+    assert(newest_tx_with_snapshot == nullptr);
     mutex.destroy();
   }
 
-  static void insert(Rdb_transaction *tx);
+  static void insert_without_snapshot(Rdb_transaction *tx);
   static void erase(Rdb_transaction *tx);
+  static void reinsert_with_snapshot(Rdb_transaction *tx, std::int64_t ts);
+  static void reinsert_without_snapshot(Rdb_transaction *tx);
+
   static void for_each(Rdb_tx_list_walker &walker);
 
+  [[nodiscard]] static std::int64_t oldest_timestamp();
+
 #ifndef NDEBUG
-  [[nodiscard]] static bool contains(Rdb_transaction *tx);
+  static std::int64_t check_node(const Rdb_transaction *tx, std::int64_t ts);
+  [[nodiscard]] static bool contains(const Rdb_transaction *tx);
 #endif
 
   Rdb_transaction_list() = delete;
@@ -3540,6 +3562,8 @@ class [[nodiscard]] Rdb_transaction_list final {
 
 Rds_mysql_mutex Rdb_transaction_list::mutex;
 Rdb_transaction *Rdb_transaction_list::head{nullptr};
+Rdb_transaction *Rdb_transaction_list::tail{nullptr};
+Rdb_transaction *Rdb_transaction_list::newest_tx_with_snapshot{nullptr};
 
 /* This is the base class for transactions when interacting with rocksdb.
  */
@@ -3575,6 +3599,14 @@ class Rdb_transaction {
   bool m_bulk_index_transaction = false;
   bool m_dd_transaction = false;
 
+  /** The timestamp of the earliest snapshot in the transaction, which does not
+  change even if the snapshot gets renewed or released. It is used to ensure
+  that the TTL filtering returns the same set of the rows throughout the
+  transaction lifetime. Zero if the transaction hasn't acquired a snapshot
+  yet. Writes and reads from other threads are protected by
+  Rdb_transaction_list::mutex. */
+  std::int64_t m_earliest_snapshot_ts = 0;
+
  protected:
   THD *m_thd = nullptr;
 
@@ -3592,8 +3624,6 @@ class Rdb_transaction {
   std::shared_ptr<Rdb_snapshot_notifier> m_notifier;
 
   rocksdb::ReadOptions m_read_opts[2];
-
-  std::int64_t m_snapshot_timestamp = 0;
 
   // This should be used only when updating binlog information.
   virtual rocksdb::WriteBatchBase *get_write_batch() = 0;
@@ -3796,8 +3826,23 @@ class Rdb_transaction {
 
   [[nodiscard]] bool is_dd_transaction() const { return m_dd_transaction; }
 
-  [[nodiscard]] std::int64_t get_snapshot_timestamp() const noexcept {
-    return m_snapshot_timestamp;
+  [[nodiscard]] std::int64_t get_earliest_snapshot_ts() const noexcept {
+    const auto result = m_earliest_snapshot_ts;
+    // Once it is safe to call get_snapshot_ts from different threads, assert
+    // that its value is not less than the earliest timestamp.
+    assert(result >= 0);
+    return result;
+  }
+
+  [[nodiscard]] std::int64_t get_snapshot_ts() const {
+    const auto *const snapshot = m_read_opts[USER_TABLE].snapshot;
+    const auto result = snapshot != nullptr ? snapshot->GetUnixTime() : 0;
+
+    assert(result >= 0);
+    assert(m_earliest_snapshot_ts >= 0);
+    assert(m_earliest_snapshot_ts <= result || result == 0);
+
+    return result;
   }
 
   virtual void set_lock_timeout(int timeout_sec_arg, TABLE_TYPE table_type) = 0;
@@ -3817,24 +3862,30 @@ class Rdb_transaction {
 
   void incr_insert_count(TABLE_TYPE table_type) {
     assert(!is_ac_nl_ro_rc_transaction());
+    // Unlike with incr_update_count and incr_delete_count below, inserts may
+    // happen without a snapshot, i.e. during a bulk load.
 
     if (table_type == TABLE_TYPE::USER_TABLE) ++m_insert_count;
   }
 
   void incr_update_count(TABLE_TYPE table_type) {
     assert(!is_ac_nl_ro_rc_transaction());
+    assert(table_type != TABLE_TYPE::USER_TABLE || get_snapshot_ts() != 0);
 
     if (table_type == TABLE_TYPE::USER_TABLE) ++m_update_count;
   }
 
   void incr_delete_count(TABLE_TYPE table_type) {
     assert(!is_ac_nl_ro_rc_transaction());
+    assert(table_type != TABLE_TYPE::USER_TABLE || get_snapshot_ts() != 0);
 
     if (table_type == TABLE_TYPE::USER_TABLE) ++m_delete_count;
   }
 
   void incr_row_lock_count(TABLE_TYPE table_type) {
     assert(!is_ac_nl_ro_rc_transaction());
+    assert(table_type == TABLE_TYPE::USER_TABLE);
+    assert(get_snapshot_ts() != 0);
 
     if (table_type == TABLE_TYPE::USER_TABLE) ++m_row_lock_count;
   }
@@ -3920,20 +3971,27 @@ class Rdb_transaction {
     return my_core::thd_tx_isolation(m_thd) <= ISO_READ_COMMITTED;
   }
 
+  [[nodiscard]] static std::int64_t binlog_hlc_or_snapshot_ts(
+      const rocksdb::Snapshot &snapshot) {
+    if (!rdb_is_binlog_ttl_enabled()) return snapshot.GetUnixTime();
+    const auto binlog_max_write_hlc_ns = mysql_bin_log.get_max_write_hlc_nsec();
+    return (binlog_max_write_hlc_ns != 0)
+               ? binlog_max_write_hlc_ns / 1'000'000'000
+               : snapshot.GetUnixTime();
+  }
+
   void snapshot_created(const rocksdb::Snapshot *const snapshot) {
     assert(snapshot != nullptr);
 
     m_read_opts[USER_TABLE].snapshot = snapshot;
 
-    // TODO: Use snapshot timestamp from rocksdb Snapshot object itself. This
-    // saves the extra call to fetch current time, and allows TTL compaction
-    // (which uses rocksdb timestamp) to be consistent with TTL read filtering
-    // (which uses this timestamp).
-    //
-    // There is no correctness problem though since m_snapshot_timestamp is
-    // generally set after the snapshot has been created, so compaction is
-    // not dropping anything that should have been visible.
-    rdb->GetEnv()->GetCurrentTime(&m_snapshot_timestamp);
+    const auto earliest_ts = get_earliest_snapshot_ts();
+    if (earliest_ts == 0) {
+      const auto new_earliest_ts = binlog_hlc_or_snapshot_ts(*snapshot);
+      Rdb_transaction_list::reinsert_with_snapshot(this, new_earliest_ts);
+    } else {
+      assert(earliest_ts <= binlog_hlc_or_snapshot_ts(*snapshot));
+    }
     m_is_delayed_snapshot = false;
   }
 
@@ -4752,6 +4810,8 @@ class Rdb_transaction {
     modified_tables.clear();
 
     m_binlog_ttl_read_filtering_ts = m_thd->binlog_ttl_read_filtering_ts = 0;
+
+    Rdb_transaction_list::reinsert_without_snapshot(this);
   }
 
  protected:
@@ -4786,7 +4846,7 @@ class Rdb_transaction {
 
   uint64_t get_or_create_ttl_read_filtering_ts() {
     if (!rdb_is_binlog_ttl_enabled()) {
-      return static_cast<uint64_t>(m_snapshot_timestamp);
+      return static_cast<uint64_t>(get_earliest_snapshot_ts());
     }
     if (m_binlog_ttl_read_filtering_ts.load()) {
       return m_binlog_ttl_read_filtering_ts;
@@ -4806,7 +4866,7 @@ class Rdb_transaction {
 
   uint64_t get_ttl_read_filtering_ts() const {
     if (!rdb_is_binlog_ttl_enabled()) {
-      return static_cast<uint64_t>(m_snapshot_timestamp);
+      return static_cast<uint64_t>(get_earliest_snapshot_ts());
     }
     return m_binlog_ttl_read_filtering_ts.load();
   }
@@ -5149,7 +5209,6 @@ class Rdb_transaction_impl : public Rdb_transaction {
     bool need_clear = m_is_delayed_snapshot;
 
     if (has_snapshot(table_type)) {
-      m_snapshot_timestamp = 0;
       if (m_explicit_snapshot) {
         m_explicit_snapshot.reset();
         need_clear = false;
@@ -5458,12 +5517,8 @@ class Rdb_transaction_impl : public Rdb_transaction {
       const rocksdb::Snapshot *const cur_snapshot =
           m_rocksdb_tx[TABLE_TYPE::USER_TABLE]->GetSnapshot();
       if (org_snapshot != cur_snapshot) {
-        if (org_snapshot != nullptr) m_snapshot_timestamp = 0;
-
         m_read_opts[TABLE_TYPE::USER_TABLE].snapshot = cur_snapshot;
-        if (cur_snapshot != nullptr) {
-          rdb->GetEnv()->GetCurrentTime(&m_snapshot_timestamp);
-        } else {
+        if (cur_snapshot == nullptr) {
           m_rocksdb_tx[TABLE_TYPE::USER_TABLE]->SetSnapshotOnNextOperation(
               m_notifier);
           m_is_delayed_snapshot = true;
@@ -5479,11 +5534,11 @@ class Rdb_transaction_impl : public Rdb_transaction {
   }
 
   virtual ~Rdb_transaction_impl() override {
+    rollback();
+
     // Remove from the global list before all other processing is started.
     // Otherwise, information_schema.rocksdb_trx can crash on this object.
     Rdb_transaction::remove_from_global_trx_list();
-
-    rollback();
 
     // Theoretically the notifier could outlive the Rdb_transaction_impl
     // (because of the shared_ptr), so let it know it can't reference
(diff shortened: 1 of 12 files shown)


hermanlee:
I think there is a potential issue in rdb_compact_filter.h with the  `calculate_expiration_timestamp()` function. This function relies on RocksDB to return back the oldest snapshot time to be used for compaction. Even though each transaction keeps it own snapshot time, the compaction code would also need to figure out the oldest snapshot time across all transactions.

MyRocks currently relies on RocksDB to maintain this heap.

I wonder if this diff should be based on the RR/RC change to allow the read-only snapshot in RR to be created from the first read-only select. In that case, the snapshot never goes away, and the snapshot time never changes.

For RC, if we want the same behavior, we could also keep the read-only snapshot around just to maintain the timestamp for compaction, but each new read-only query would still acquire its own snapshot.

laurynas-biveinis:
@hermanlee If I understand your comment correctly, currently there is a race condition between the compaction filter and the oldest transaction temporarily releasing its snapshot, thus maintaining only a per-transaction timestamp for the per-transaction filtering is not enough, and a transaction snapshot must be kept open constantly once it has been acquired.

laurynas-biveinis:
@hermanlee , it seems that each transaction has to have three snapshot fields, even if it does not use them all in all cases:
- the snapshot for the currently-executed statement;
- the RR RO snapshot;
- the snapshot that prevents expiration.

The 3rd one and the 1st one cannot be merged obviously.
Likewise the 2nd one and the 1st one.
The 3rd one and the 2nd one cannot be merged because an RR transaction may execute writes before its first read.

It could be possible to reduce complexity of transaction object maintaining three snapshots by moving the oldest timestamp tracking from RocksDB to MyRocks (have a heap/priority queue of timestamps in Rdb_compact_filter, insert/remove from it on transaction 1st snapshot create/commit&rollback), but a shared data structure that duplicates RocksDB internals might not be the best trade-off here.

laurynas-biveinis:
As discussed offline, will use the global transaction list instead. Step 1 of this approach is at https://github.com/facebook/mysql-5.6/pull/1471

storage/rocksdb/rdb_compact_filter.h
```diff
@@ -50,15 +110,13 @@ class Rdb_compact_filter : public rocksdb::CompactionFilter {
...
-        oldest_snapshot_timestamp == 0) {
-      oldest_snapshot_timestamp = static_cast<uint64_t>(std::time(nullptr));
+    uint64_t oldest_transaction_timestamp =
+        Rdb_transaction_timestamps::instance().oldest();
+    if (oldest_transaction_timestamp == 0) {
+      oldest_transaction_timestamp = static_cast<uint64_t>(std::time(nullptr));
     }
 
-    m_expiration_timestamp = oldest_snapshot_timestamp;
+    m_expiration_timestamp = oldest_transaction_timestamp;
```
sunshine-Chun: Hey Laurynas, we recently notice a new issue that will bring inconsistency to primary and secondary in long running transactions. The oldest_transaction_timestamp here is retrieved from snapshot->GetUnixTime(). This timestamp will be different between primary and secondary. Secondary will have a slightly larger timestamp, hence compact more data away than primary. Can we change the oldest_transaction_timestamp from snapshot->GetUnixTime() to mysql_bin_log.get_max_write_hlc_nsec()? This can ensure the consistency between primary and secondary because mysql_bin_log.get_max_write_hlc_nsec() returns the same value if primary and secondary has the same GTID set when taking snapshot.
cc @abhinav04sharma

sunshine-Chun: #1471 has been closed. Can we have a rebase to include that change? Thanks!

laurynas-biveinis: @sunshine-Chun, OK, I will make it use `mysql_bin_log.get_max_write_hlc_nsec()`. I am rebasing (and rewriting) this PR right now.

laurynas-biveinis:
Non-functional changes in this PR split off to https://github.com/facebook/mysql-5.6/pull/1487

sunshine-Chun:
> Non-functional changes in this PR split off to #1487

The change has been committed to the codebase.

storage/rocksdb/rdb_compact_filter.h
laurynas-biveinis: @sunshine-Chun , I am not too familiar with the binlog HLC patch. What value should we use if the HLC value is zero? Should we use `snapshot->GetUnixTime()` then?

laurynas-biveinis: @sunshine-Chun, the relationship between binlog and RocksDB timestamps is not clear to me. We initialize the earliest timestamp from binlog, but then the current snapshot timestamps come from the RocksDB snapshots directly. I'd expect the invariant "the current timestamp can be earlier than the earliest transaction timestamp" to hold, but, i.e. `rocksdb_rpl.rocksdb_binlog_ttl` test shows (`result` is the current snapshot timestamp)
```
(lldb) p/x m_earliest_snapshot_ts
(int64_t) 0x00047fc94dc5cd7c
(lldb) p/x result
(const int64_t) 0x0000000066c7143b
```

laurynas-biveinis:
@sunshine-Chun , I have pushed my current version, which is as finished as possible, except for the binlog/RDB timestamp question above. Keeping in draft status until that is resolved.

storage/rocksdb/rdb_compact_filter.h
sunshine-Chun: Here is what these two timestamp represents:
1. Each write transaction has a HLC timestamp assigned to it. The `mysql_bin_log.get_max_write_hlc_nsec()` tells us the lastest HLC that is written to this instance. 
2. The RocksDB timestamp represents the unix time when creating the snapshot. 

In most cases, these two timestamp should be very close in value. In the long running transaction case, when creating snapshot,  `mysql_bin_log.get_max_write_hlc_nsec()` should always be smaller than the RocksDB timestamp, cause no new writes will update the max_write_hlc_nsec at that time.

The reason why we use  `mysql_bin_log.get_max_write_hlc_nsec()` here is that it can help the primary and secondary be consistent at the this value. When we create snapshot in long running transactions in primary and secondary, we can make sure they are on the same GTID set. This also means they will have the same `mysql_bin_log.get_max_write_hlc_nsec()` value. Noted primary and secondary maintains their own `mysql_bin_log.get_max_write_hlc_nsec()` value. But since they have the same committed transactions, they share the same `mysql_bin_log.get_max_write_hlc_nsec()` value. Rocksdb timestamp can not do this cause it will get the unix now in primary and secondary separately. Secondary takes snapshot later than primary.

sunshine-Chun: `mysql_bin_log.get_max_write_hlc_nsec()` is unlikely to be zero. But yeah, `snapshot->GetUnixTime()` seems the only thing we can do to guard the `new_earliest_ts` from zero.

sunshine-Chun: MTR test seems to be failing. Have you rebased it to the lastest branch?

laurynas-biveinis:
> MTR test seems to be failing. Have you rebased it to the lastest 
> branch?

Which tests? Two debug MTR tests are failing with assert RDB 
timestamp > binlog timestamp, as I wrote before, and are the 
reason this PR is still draft.

sunshine-Chun:
> MTR test seems to be failing. Have you rebased it to the lastest branch?
> Which tests? Two debug MTR tests are failing with assert RDB timestamp > binlog timestamp, as I wrote before, and are the reason this PR is still draft.

Other than the assertions:
1. `drop_cf_before_show_deadlock_info` needs to be re-recorded.
2. `performance_schema` length mismatch: `+Performance_schema_digest_lost	2374
`
3. `rocksdb_concurrent_delete` result content mismatch
4. sanitize_report, will sent the log via email.

storage/rocksdb/rdb_compact_filter.h
```diff
@@ -50,15 +50,12 @@ class Rdb_compact_filter : public rocksdb::CompactionFilter {
...
-      oldest_snapshot_timestamp = static_cast<uint64_t>(std::time(nullptr));
+    auto oldest_transaction_ts = oldest_transaction_timestamp();
+    if (oldest_transaction_ts == 0) {
+      oldest_transaction_ts = static_cast<uint64_t>(std::time(nullptr));
     }
 
-    m_expiration_timestamp = oldest_snapshot_timestamp;
+    m_expiration_timestamp = oldest_transaction_ts;
 
     if (rdb_is_binlog_ttl_enabled()) {
```
sunshine-Chun, line 60: I wonder if we can directly use the rocksdb_binlog_ttl_compaction_timestamp here instead of getting the min between two values.

sunshine-Chun:
Just think aloud.  Feel free to correct me if I miss anything here.

`m_binlog_ttl_read_filtering_ts` will only be only assigned once per transaction, so it basically behaves like the earliest_snapshot_ts in this change.  If we refresh the snapshot during the long running transaction, the `m_binlog_ttl_read_filtering_ts` won't change its value. The `rocksdb_binlog_ttl_compaction_timestamp` is based on `m_binlog_ttl_read_filtering_ts`, so it won't change too. So it seems with binlog_ttl enabled, we already won't filter out more row when refreshing snapshot.

storage/rocksdb/rdb_compact_filter.h
laurynas-biveinis: @sunshine-Chun, thank you, this makes sense. It seems that the "L" in "HLC" is not stored explicitly, thus the binlog HLC is directly comparable to RocksDB timestamps, different resolutions aside. In which case I still cannot understand what is going on with `rocksdb_rpl.rocksdb_binlog_ttl`. The testcase does not inject any artificial timestamps, and yet:
```
$ p result # The current RDB snapshot timestamp
(const int64_t) 1724752351 # OK, resolves to $NOWish timestamp at the second resolution
(lldb) p m_earliest_snapshot_ts # Binlog HLC
(int64_t) 1695877370630958 # Does not resolve to anything reasonable at any resolution
```

laurynas-biveinis:
> 2. `performance_schema` length mismatch: `+Performance_schema_digest_lost	2374 `
> 3. `rocksdb_concurrent_delete` result content mismatch

Cannot reproduce these two. For the latter, I intermittently get a pre-existing failure:
```
[ 71%] rocksdb.rocksdb_concurrent_delete 'innodb_ddse'  [ fail ]
        Test ended at 2024-08-27 12:24:08

CURRENT_TEST: rocksdb.rocksdb_concurrent_delete
--- /home/laurynas/vilniusdb/fb-mysql/mysql-test/suite/rocksdb/r/rocksdb_concurrent_delete.result	2024-03-19 14:53:38.570572735 +0300
+++ /home/laurynas/vilniusdb/fb-mysql/_build-llvm-debug/mysql-test/var/log/rocksdb_concurrent_delete.reject	2024-08-27 15:24:08.695347028 +0300
@@ -453,9 +453,9 @@
 set debug_sync='now SIGNAL go';
 select * from t1 where id1=1;
 id1	id2	value
+1	2	200
 1	4	1
 1	5	1
-1	2	200
 --End row delete with PRIMARY
 SET SESSION TRANSACTION ISOLATION LEVEL REPEATABLE READ;
 set debug_sync='rocksdb.get_row_by_rowid SIGNAL parked WAIT_FOR go';

mysqltest: Result content mismatch
```

storage/rocksdb/rdb_compact_filter.h
sunshine-Chun: > I wonder if we can directly use the rocksdb_binlog_ttl_compaction_timestamp here instead of getting the min between two values.

We can't. Because if we have a long running transaction in secondary, the oldest_snapshot_timestamp can be older than the rocksdb_binlog_ttl_compaction_timestamp. We nee the oldest_snapshot_timestamp to prevent compacting away records needed in this long running transcation.

sunshine-Chun:
> Just think aloud. Feel free to correct me if I miss anything here.
> 
> `m_binlog_ttl_read_filtering_ts` will only be only assigned once per transaction, so it basically behaves like the earliest_snapshot_ts in this change. If we refresh the snapshot during the long running transaction, the `m_binlog_ttl_read_filtering_ts` won't change its value. The `rocksdb_binlog_ttl_compaction_timestamp` is based on `m_binlog_ttl_read_filtering_ts`, so it won't change too. So it seems with binlog_ttl enabled, we already won't filter out more row when refreshing snapshot.

The current change and binlog ttl feature covers different scenarios. Both are needed. 
`m_expiration_timestamp = min(oldest_transaction_ts, rocksdb_binlog_ttl_compaction_timestamp)`

- When m_expiration_timestamp = oldest_transaction_ts, the current change works. This happens when we have a long running transaction in secondary. 
- When m_expiration_timestamp = rocksdb_binlog_ttl_compaction_timestamp, the binlog ttl feature works. This happens when we have a long running transaction in both primary and secondary.

storage/rocksdb/rdb_compact_filter.h
sunshine-Chun: I run the test in my local server and got the following number. The `m_earliest_snapshot_ts` looks correct to me. Note, this is the value that hit the assertion `assert(m_earliest_snapshot_ts <= result || result == 0);`
```
(gdb) p m_earliest_snapshot_ts
$1 = 1724827776952667189
(gdb) p result
$2 = 1724827777
```
Let me post the logic to get HLC here.
```
uint64_t HybridLogicalClock::get_next() {
  uint64_t current_hlc, next_hlc = 0;
  bool done = false;

  while (!done) {
    // Get the current wall clock in nanosecond precision
    uint64_t current_wall_clock =
        std::chrono::duration_cast<std::chrono::nanoseconds>(
            std::chrono::high_resolution_clock::now().time_since_epoch())
            .count();

    // get the 'current' internal HLC
    current_hlc = current_.load();

    // Next HLC timestamp is max of current-hlc and current-wall-clock
    next_hlc = std::max(current_hlc + 1, current_wall_clock);

    // Conditionally update the internal hlc
    done = current_.compare_exchange_strong(current_hlc, next_hlc);
  }

  return next_hlc;
}
```

Can try to print the `current_wall_clock` to see if the value is abnormal? This determines the baseline value of the HLC. My guess is that the abnormal value starts from here.

sunshine-Chun:
> > 2. `performance_schema` length mismatch: `+Performance_schema_digest_lost	2374 `
> > 3. `rocksdb_concurrent_delete` result content mismatch
> 
> Cannot reproduce these two. For the latter, I intermittently get a pre-existing failure:
> 
> ```
> [ 71%] rocksdb.rocksdb_concurrent_delete 'innodb_ddse'  [ fail ]
>         Test ended at 2024-08-27 12:24:08
> 
> CURRENT_TEST: rocksdb.rocksdb_concurrent_delete
> --- /home/laurynas/vilniusdb/fb-mysql/mysql-test/suite/rocksdb/r/rocksdb_concurrent_delete.result	2024-03-19 14:53:38.570572735 +0300
> +++ /home/laurynas/vilniusdb/fb-mysql/_build-llvm-debug/mysql-test/var/log/rocksdb_concurrent_delete.reject	2024-08-27 15:24:08.695347028 +0300
> @@ -453,9 +453,9 @@
>  set debug_sync='now SIGNAL go';
>  select * from t1 where id1=1;
>  id1	id2	value
> +1	2	200
>  1	4	1
>  1	5	1
> -1	2	200
>  --End row delete with PRIMARY
>  SET SESSION TRANSACTION ISOLATION LEVEL REPEATABLE READ;
>  set debug_sync='rocksdb.get_row_by_rowid SIGNAL parked WAIT_FOR go';
> 
> mysqltest: Result content mismatch
> ```

Then they are not related to this change.

laurynas-biveinis:
> 4. sanitize_report, will sent the log via email.

This was caused by concurrent accesses to `m_read_opts[TABLE_TYPE::USER_TABLE].snapshot`, which appears to be a pre-existing issue and it seems that the current code may crash on `SHOW ENGINE ROCKSDB STATUS` concurrent to i.e. statement rollbacks. I am fixing it as follows, but overall the accesses do not appear to be synchronized properly and I'm making a TODO note to review it:
```
diff --git a/storage/rocksdb/ha_rocksdb.cc b/storage/rocksdb/ha_rocksdb.cc
index 3b5cf2da90c..9ba4860b03a 100644
@@ -5503,6 +5503,10 @@ class Rdb_transaction_impl : public Rdb_transaction {
     if (m_rocksdb_tx[TABLE_TYPE::USER_TABLE]) {
       const rocksdb::Snapshot *const org_snapshot =
           m_rocksdb_tx[TABLE_TYPE::USER_TABLE]->GetSnapshot();
+
+      assert(org_snapshot == m_read_opts[TABLE_TYPE::USER_TABLE].snapshot);
+      m_read_opts[TABLE_TYPE::USER_TABLE].snapshot = nullptr;
+
       rollback_to_stmt_savepoint();
 
       const rocksdb::Snapshot *const cur_snapshot =
```

laurynas-biveinis:
This patch is not enough. I'll remove the extra added accesses from other threads until the `m_read_opts[USER_TABLE].snapshot` concurrency is fixed

storage/rocksdb/rdb_compact_filter.h
laurynas-biveinis: `current_wall_clock` value is indeed correct, thus the only issue was comparing nanoseconds to seconds directly. Adding the required conversion fixed the tests.

laurynas-biveinis:
@sunshine-Chun, addressed all comments as discussed above, fixed one linked list middle pointer manipulation logic error, and nanosecond to second conversion. Ready for review!

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -5479,12 +5525,12 @@ class Rdb_transaction_impl : public Rdb_transaction {
   }
 
   virtual ~Rdb_transaction_impl() override {
+    rollback();
```
sunshine-Chun, line 5537: What will happen when we finish a transaction? My understanding is this transaction will be first reinsert without snapshot to the end of the list in on_commit(). Then when the transaction is deconstructed, it will be moved to the end again in the rollback(). Lastly, it will be removed from the list via erase(). Did I miss something, or we do have duplicated/redundant operations?

laurynas-biveinis: The destructor is only executed on a session disconnect, where it may be called with or without going through commit/rollback first. While a session is alive, the transaction object does not go through destruction but is reused after completing rollback/commit.

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -3823,18 +3868,22 @@ class Rdb_transaction {
 
   void incr_update_count(TABLE_TYPE table_type) {
     assert(!is_ac_nl_ro_rc_transaction());
+    assert(table_type != TABLE_TYPE::USER_TABLE || get_snapshot_ts() != 0);
```
luqun, line 3873: maybe add similar check into incr_insert_count()?

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -5794,55 +5840,179 @@ class Rdb_writebatch_impl : public Rdb_transaction {
...
+}
+
+void Rdb_transaction_list::reinsert_without_snapshot(Rdb_transaction *tx) {
+  if (tx->get_earliest_snapshot_ts() == 0) return;
+
+  RDB_MUTEX_LOCK_CHECK(mutex);
+  do_erase(tx);
+  fixup_newest_snapshot_after_erase(tx);
+  tx->m_earliest_snapshot_ts = 0;
+  tx->next = nullptr;
```
luqun, line 6007: why only clean tx->next? how about tx->prev?

laurynas-biveinis:
>    void incr_update_count(TABLE_TYPE table_type) {
>      assert(!is_ac_nl_ro_rc_transaction());
> +    assert(table_type != TABLE_TYPE::USER_TABLE || 
> get_snapshot_ts() != 0);
>
> maybe add similar check into incr_insert_count()?

That one might be called without snapshot during bulkloading. I 
have added a comment there.

storage/rocksdb/ha_rocksdb.cc
laurynas-biveinis: That one might be called without snapshot during bulkloading. I
have added a comment there.

storage/rocksdb/ha_rocksdb.cc
laurynas-biveinis: `do_insert_without_snapshot` handles `tx->prev`. It does not handle `tx->next` because the other caller sets it outside the mutex critical section, to have it as short as possible.

laurynas-biveinis:
Pushed the branch again. The only changes are an extra comment at `incr_insert_count` and an additional note in the commit message, no actual code changes, any reviews in progress should not be invalidated by this push.

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -3925,15 +3976,15 @@ class Rdb_transaction {
...
-    // not dropping anything that should have been visible.
-    rdb->GetEnv()->GetCurrentTime(&m_snapshot_timestamp);
+    const auto earliest_ts = get_earliest_snapshot_ts();
+    if (earliest_ts == 0) {
+      auto new_earliest_ts =
+          mysql_bin_log.get_max_write_hlc_nsec() / 1'000'000'000;
+      if (new_earliest_ts == 0) new_earliest_ts = snapshot->GetUnixTime();
+      Rdb_transaction_list::reinsert_with_snapshot(this, new_earliest_ts);
+    } else {
+      assert(earliest_ts <= snapshot->GetUnixTime());
```
sunshine-Chun: Abhinav left some comment here:
1. Can we use the mysql_bin_log.get_max_write_hlc_nsec() only when binlog ttl is enabled?
2. For the assertion, when binlog ttl is enabled, we are comparing hlc with wall clock, that doesn't seem correct.
Can we do something like
```
if (earliest_ts == 0) {
      auto new_earliest_ts =
          rdb_is_binlog_ttl_enabled() && mysql_bin_log.get_max_write_hlc_nsec()
              ? mysql_bin_log.get_max_write_hlc_nsec() / 1'000'000'000
              : snapshot->GetUnixTime();
      Rdb_transaction_list::reinsert_with_snapshot(this, new_earliest_ts);
    } else {
      assert(earliest_ts <= rdb_is_binlog_ttl_enabled() &&
                     mysql_bin_log.get_max_write_hlc_nsec()
                 ? mysql_bin_log.get_max_write_hlc_nsec() / 1'000'000'000
                 : snapshot->GetUnixTime());
    }
```

sunshine-Chun: This is the last comment needs to be addressed before commit if Luqun and Herman don't have more comments.

laurynas-biveinis: I am testing the following:
```c++
  [[nodiscard]] static std::int64_t binlog_hlc_or_snapshot_ts(
      const rocksdb::Snapshot &snapshot) {
    if (!rdb_is_binlog_ttl_enabled()) return snapshot.GetUnixTime();
    const auto binlog_max_write_hlc_ns = mysql_bin_log.get_max_write_hlc_nsec();
    return (binlog_max_write_hlc_ns != 0) ? binlog_max_write_hlc_ns
                                          : snapshot.GetUnixTime();
  }

  void snapshot_created(const rocksdb::Snapshot *const snapshot) {
    assert(snapshot != nullptr);

    m_read_opts[USER_TABLE].snapshot = snapshot;

    const auto earliest_ts = get_earliest_snapshot_ts();
    if (earliest_ts == 0) {
      const auto new_earliest_ts = binlog_hlc_or_snapshot_ts(*snapshot);
      Rdb_transaction_list::reinsert_with_snapshot(this, new_earliest_ts);
    } else {
      assert(earliest_ts <= binlog_hlc_or_snapshot_ts(*snapshot));
    }
    m_is_delayed_snapshot = false;
  }
```

laurynas-biveinis: Ugh, of course with a division by a billion

laurynas-biveinis:
Addressed last review comments, ready for review again

sunshine-Chun:
rocksdb_binlog_ttl.test is failing for ```std::int64_t myrocks::Rdb_transaction::get_snapshot_ts() const: Assertion `m_earliest_snapshot_ts <= result || result == 0' failed.```

laurynas-biveinis:
> rocksdb_binlog_ttl.test is failing for `` std::int64_t myrocks::Rdb_transaction::get_snapshot_ts() const: Assertion `m_earliest_snapshot_ts <= result || result == 0' failed. ``

@sunshine-Chun , division by a billion got lost again :( Fixed

PR #1459, opened 2024-05-29
Remove rdb_sysvars_mutex

laurynas-biveinis:
The mutex is used by sysvar update functions and synchronizes updates to a
subset of global system variables. But sysvar update functions are called with
LOCK_global_system_variables server lock taken, making the extra locking in the
SE redundant, thus just remove it.

However it was also partially (writes but not reads) protecting concurrent
accesses to properties_collector_factory variable, thus introduce
synchronization for its class Rdb_tbl_prop_coll_factory by adding an rwlock.

At the same time convert Rdb_tbl_prop_coll_factory pointer fields to const
references, remove redundant explicit from two-argument constructor, add
[[nodiscard]] to touched methods, make the class non-movable. Remove redundant
const from passed by-value arguments.

To enable converting m_ddl_manager field to a const reference, mark
Rdb_ddl_manager::m_rwlock field as mutable, which then enables marking many
logically-const methods as such. Also mark touched methods [[nodiscard]], remove
redundant const from passed by-value arguments, convert pass by const reference
to pass by value for a small object arguments of GL_INDEX_ID type.

Add return value handling for one call site of Rdb_ddl_manager::find_indexes as
pointed out by [[nodiscard]].

### DIFF
--- a/storage/rocksdb/ha_rocksdb.cc
+++ b/storage/rocksdb/ha_rocksdb.cc
@@ -708,7 +708,6 @@ static void rocksdb_drop_index_wakeup_thread(
     void *const var_ptr MY_ATTRIBUTE((__unused__)), const void *const save);
 
 static bool rocksdb_pause_background_work = 0;
-static Rds_mysql_mutex rdb_sysvars_mutex;
 static Rds_mysql_mutex rdb_block_cache_resize_mutex;
 static Rds_mysql_mutex rdb_bottom_pri_background_compactions_resize_mutex;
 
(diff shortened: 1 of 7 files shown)


laurynas-biveinis:
`rdb_sysvars_mutex` was renamed from `pause_background_mutex` by 3c19c3d7aae542508d59fc34e01a9dfe5be8a43c 
`pause_background_mutex` introduced by c5acee2651a1

laurynas-biveinis:
No sysvar check function was taking this mutex.

PR #1466, opened 2024-06-10
Use python3 from $PATH instead of absolute paths

laurynas-biveinis:
This allows using Python virtualenv directly. This partially reverts changes in
9d0dda0c6565f1a7d926e5c82628cb134d4b1ac9, but its commit message is not stating
anything about absolute paths.

### DIFF
--- a/mysql-test/include/raft_config.py
+++ b/mysql-test/include/raft_config.py
@@ -1,4 +1,4 @@
-#!/usr/local/bin/python3
+#!/usr/bin/env python3
 
 import os
 import sys
--- a/mysql-test/suite/innodb_stress/include/innodb_stress.inc
+++ b/mysql-test/suite/innodb_stress/include/innodb_stress.inc
@@ -39,7 +39,7 @@ if ($do_checksum)
 {
     # populate the table and store its checksum before any load.
     let $exec =
-/usr/bin/python3 $MYSQL_TEST_DIR/suite/innodb_stress/t/load_generator.py $pid_file $kill_db_after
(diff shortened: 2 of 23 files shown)


luqun:
after change from /usr/bin/python3 to pythohn3, some of MTR failed to find due to "ModuleNotFoundError: No module named 'MySQLdb'".

laurynas-biveinis:
Do you know what python3 are the scripts picking up? Is it possible to adjust PATH to put /usr/bin before that?

luqun:
> Do you know what python3 are the scripts picking up? Is it possible to adjust PATH to put /usr/bin before that?

the pyhon3 in /usr/local/sbin instead of /usr/bin/python3.. need to do some investigation since this PATH variable comes from  SYSTEM.

laurynas-biveinis:
Ack, let me know. I believe the right thing is for the script to run in
a correctly-set up environment instead of hardcoding paths, but let me
know if that's impossible for some reason

laurynas-biveinis:
As discussed offline, will conditionalize on the OS

laurynas-biveinis:
@luqun , @hermanlee , could you tell which tests break with this PR in its current form? I am wondering if I could leave script shebangs using env

luqun:
> @luqun , @hermanlee , could you tell which tests break with this PR in its current form? I am wondering if I could leave script shebangs using env

such as main.admission_control_multi_query_wait_events

laurynas-biveinis:
> > @luqun , @hermanlee , could you tell which tests break with this PR in its current form? I am wondering if I could leave script shebangs using env
> 
> such as main.admission_control_multi_query_wait_events

I have updated the three tests in the main suite to choose $PATH or hardcoded python conditionally. Let me know if any other tests need updating

PR #1471, opened 2024-06-25
Replace the global MyRocks transaction multiset with an intrusive doubly linked list

laurynas-biveinis:
Previously, the data structure used for the global MyRocks transaction list was
std::multiset. It was not optimal, for instance, it did not need to be sorted
nor multi-. Replace it with an intrusive doubly linked list. The list pointers
are stored directly in the transaction object, which has the advantage that
adding and removing the transaction from the global list does not need to
maintain any extra nodes on heap, shortening the code path and improving the
data locality even if the latter still remains poor due to linked list data
structure. This change also improves the complexity of insertion and removal
operations from logarithmic to constant.

At the same time split out the global list methods and fields from
Rdb_transaction class to a new Rdb_transaction_list class.

### DIFF
--- a/storage/rocksdb/ha_rocksdb.cc
+++ b/storage/rocksdb/ha_rocksdb.cc
@@ -3503,9 +3503,49 @@ static int rdb_dbug_set_ttl_read_filter_ts();
   return true;
 }
 
+// The global list of currently-active MyRocks transactions, implemented as a
+// doubly-linked intrusive linked list.
+class [[nodiscard]] Rdb_transaction_list final {
+ private:
+  static Rds_mysql_mutex mutex;
+  static Rdb_transaction *head;
+
+ public:
+  static void init() { mutex.init(key_mutex_tx_list, MY_MUTEX_INIT_FAST); }
+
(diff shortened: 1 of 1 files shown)


hermanlee (commented):
The changes look good overall. Do we want to incorporate something like storage/innobase/include/ut0lst.h for the link list manipulation rather than re-implement it?

laurynas-biveinis:
> Do we want to incorporate something like storage/innobase/include/ut0lst.h for the link list manipulation rather than re-implement it?

In the next PR I am planning to make this list sorted by the oldest timestamp, and it is possible to do it cheaply (except for pathological cases) without a background thread: maintain a pointer to the middle of the global transaction list, separating transactions with the timestamp from the tail of transactions without one. Once a transaction acquires a timestamp, remove it from the tail and re-insert in the right place, starting the search from that middle pointer, most of the time no further walking will be needed. I am not sure ut0lst.h will be a good fit for another global list pointer, thus I'd prefer to review it afterwards. (if Boost dependency is permanent, then Boost.Intrusive is also in the running)

PR #1472, opened 2024-07-04
MyRocks: skip row lock during DDSE upgrade

luqun:
There are some instance contain huge number of DBs and each DB contains a lot tables/columns(such as 500K).
During DDSE upgrade, It use single transaction to insert select statement to copy data from old SE to new SE. if new DDSE is rocksdb, it may hit max row lock error.

There are 3 places during DD upgrade that may involve a lot row locks:
- Insert select statement: This involves copy data from one storage engine to
       another storage engine. This only happens for DD SE upgrade, and with
       this change, no row locks are hold.
- Drop DD tables: when dropping old DD tables, the SQL layer reads one rows
        and deletes it. This happens for both DD SE upgrade and DD version
        upgrade. The required max number of row lock is the max number of rows
        among these DD tables.
 - SYSTEM table update: there are many SQL scripts involving insertion of data
        into system tables. This happens for both DD SE upgrade and DD version
        upgrade. Currently, some tables involve 15-20K rows locks during
        insertion. Consider to use bulk load in future if more row lock required     
        
The change is to skip row lock during DDSE upgrade for Insert select statement. Future change may required to skip other two places(Drop DD tables and SYSTEM table update)

### DIFF
--- a/mysql-test/suite/rocksdb/t/ddse_upgrade_row_lock.test
+++ b/mysql-test/suite/rocksdb/t/ddse_upgrade_row_lock.test
@@ -1,0 +1,41 @@
+--source include/have_rocksdb.inc
+--source include/have_innodb_ddse.inc
+
+--let $MYSQLD_DATADIR= `select @@datadir`
+
+# shutdown for backup
+--source include/shutdown_mysqld.inc
+
+# backup data
+let $tmp_dir = $MYSQLTEST_VARDIR/tmp;
+let $saved_datadir = $tmp_dir/saved_datadir;
+force-cpdir $MYSQLD_DATADIR $saved_datadir;
+
+# There are 3 places during DD upgrade that may involve a lot row locks:
+# - Insert select statement: This involves copy data from one storage engine to
+#       another storage engine. This only happens for DD SE upgrade, and with
+#       this change, no row locks are hold.
+# - Drop DD tables: when dropping old DD tables, the SQL layer reads one rows
+#       and deletes it. This happens for both DD SE upgrade and DD version
+#       upgrade. The required max number of row lock is the max number of rows
+#       among these DD tables.
+# - SYSTEM table update: there are many SQL scripts involving insertion of data
+#       into system tables. This happens for both DD SE upgrade and DD version
+#       upgrade. Currently, some tables involve 15-20K rows locks during
+#       insertion. Consider to use bulk load in future if more row lock required
+--echo upgrade from INNODB->ROCKSDB DDSE
+let restart_parameters = restart: --default_dd_system_storage_engine=rocksdb;
+let $restart_parameters = $restart_parameters --datadir=$saved_datadir;
+let $restart_parameters = $restart_parameters --rocksdb_max_row_locks=20000;
+--replace_result $saved_datadir DATADIR
+--source include/start_mysqld.inc
+
+select @@rocksdb_max_row_locks, @@default_dd_system_storage_engine;
+
+# restart
+let $restart_parameters = "restart";
+--replace_result $MYSQLD_DATADIR DATADIR
+--source include/restart_mysqld.inc
+
+# cleanup
+--force-rmdir $saved_datadir
--- a/mysql-test/suite/rocksdb/r/ddse_upgrade_row_lock.result
+++ b/mysql-test/suite/rocksdb/r/ddse_upgrade_row_lock.result
@@ -1,0 +1,6 @@
+upgrade from INNODB->ROCKSDB DDSE
+# restart: --default_dd_system_storage_engine=rocksdb --datadir=DATADIR --rocksdb_max_row_locks=20000
+select @@rocksdb_max_row_locks, @@default_dd_system_storage_engine;
+@@rocksdb_max_row_locks	@@default_dd_system_storage_engine
+20000	RocksDB
+# restart
--- a/storage/rocksdb/ha_rocksdb.cc
+++ b/storage/rocksdb/ha_rocksdb.cc
@@ -13264,10 +13264,11 @@ int ha_rocksdb::update_write_row(const uchar *const old_data,
   }
 
   // Case: We skip both unique checks and rows locks only when bulk load is
-  // enabled or if rocksdb_skip_locks_if_skip_unique_check is ON
+  // enabled or if rocksdb_skip_locks_if_skip_unique_check is ON or DDSE upgrade
   if (!THDVAR(table->in_use, bulk_load) &&
       (!rocksdb_skip_locks_if_skip_unique_check ||
-       !row_info.skip_unique_check)) {
+       !row_info.skip_unique_check) &&
+      !dd::is_dd_engine_change_in_progress()) {
     /*
       Check to see if we are going to have failures because of unique
       keys.  Also lock the appropriate key values.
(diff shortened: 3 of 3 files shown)


mysql-test/suite/rocksdb/t/ddse_upgrade_row_lock.test
```diff
@@ -0,0 +1,33 @@
+--source include/have_rocksdb.inc
+--source include/have_innodb_ddse.inc
+
+
+--let $MYSQLD_DATADIR= `select @@datadir`
+
+--let $shutdown_server_timeout= 10
```
laurynas-biveinis: Why?

mysql-test/suite/rocksdb/t/ddse_upgrade_row_lock.test
```diff
@@ -0,0 +1,33 @@
...
+--source include/shutdown_mysqld.inc
+
+# save data
+let $tmp_dir = $MYSQLTEST_VARDIR/tmp;
+let $saved_datadir = $tmp_dir/saved_datadir;
+force-cpdir $MYSQLD_DATADIR $saved_datadir;
+
+# There are 3 place during DD upgrade which may involves a lot row locks
```
laurynas-biveinis:
```suggestion
# There are 3 places during DD upgrade which may involve a lot row locks
```

mysql-test/suite/rocksdb/t/ddse_upgrade_row_lock.test
```diff
@@ -0,0 +1,33 @@
...
+# - insert select statement: copy data from one storage engine to another storage engine, this only happens for DD SE upgrade, with this change, no row lock is hold
```
laurynas-biveinis: The lines appear to be too long, wrap around column 80?

mysql-test/suite/rocksdb/t/ddse_upgrade_row_lock.test
```diff
@@ -0,0 +1,33 @@
...
+# - drop DD tables: when drop DD table, SQL layer read each row and delete row. This happens for both DD SE change and DD version upgrade. max row lock is max number of rows in these DD tables
+# - SYSTEM table update: there are a lot sql scripts involves insert data into system table. This happens for both DD SE change and DD version upgrade. Currently some tables may involves 15-20K rows locks 
+--echo upgrade from INNODB->ROCKSDB DDSE
+let restart_parameters="restart: --default_dd_system_storage_engine=rocksdb --datadir=$saved_datadir --rocksdb_max_row_locks=20000";
```
laurynas-biveinis: Why 20000? Can you go lower, i.e. 0 (if that means actual zero and not unlimited) or 1?

mysql-test/suite/rocksdb/t/ddse_upgrade_row_lock.test
```diff
@@ -0,0 +1,33 @@
...
+--replace_result $saved_datadir DATADIR
+--source include/start_mysqld.inc
+
+select @@rocksdb_max_row_locks, @@default_dd_system_storage_engine;
+
+--let $shutdown_server_timeout= 10
```
laurynas-biveinis: Why?

laurynas-biveinis (commented):
Minor testcase comments, code OK

mysql-test/suite/rocksdb/t/ddse_upgrade_row_lock.test
luqun: Removed.. previously thinking is that maybe it is more reliable to shutdown instance and copy data. but looks like it is okay for MTR to copy while instance is running.

mysql-test/suite/rocksdb/t/ddse_upgrade_row_lock.test
luqun: remove shutdown.. just restart..

mysql-test/suite/rocksdb/t/ddse_upgrade_row_lock.test
luqun: Fixed..

mysql-test/suite/rocksdb/t/ddse_upgrade_row_lock.test
luqun: currently no. due to system table upgrade(these sql script) still hold locks. may need another change to use bulk load for these sql scripts.

mysql-test/suite/rocksdb/t/ddse_upgrade_row_lock.test
laurynas-biveinis: I think it's still better to shutdown first, then copy, or we may end up inadvertently testing it as a "hot backup" by cp, which is not always consistent. It's just the shutdown does not need a specific timeout value.

mysql-test/suite/rocksdb/t/ddse_upgrade_row_lock.test
```diff
@@ -0,0 +1,36 @@
...
+
+# backup data
+let $tmp_dir = $MYSQLTEST_VARDIR/tmp;
+let $saved_datadir = $tmp_dir/saved_datadir;
+force-cpdir $MYSQLD_DATADIR $saved_datadir;
+
+# There are 3 places during DD upgrade which may involve a lot row locks
+# - insert select statement: copy data from one storage engine to another
+#       storage engine, this only happens for DD SE upgrade. with this change,
+#       no row lock is hold
```
laurynas-biveinis:
```suggestion
#       no row locks are held
```

mysql-test/suite/rocksdb/t/ddse_upgrade_row_lock.test
```diff
@@ -0,0 +1,36 @@
...
+# - drop DD tables: when drop old DD table, SQL layer read one row and delete it
```
laurynas-biveinis:
```suggestion
# - drop DD tables: when dropping old DD tables, the SQL layer reads one row and deletes it
```

mysql-test/suite/rocksdb/t/ddse_upgrade_row_lock.test
```diff
@@ -0,0 +1,36 @@
...
+#       This happens for both DD SE upgrade and DD version upgrade. max row lock
+#       required is the max number of rows among these DD tables
+# - SYSTEM table update: there are a lot sql scripts involves insert data into
```
laurynas-biveinis:
```suggestion
# - SYSTEM table update: there are a lot sql scripts involving insertion of data into
```

mysql-test/suite/rocksdb/t/ddse_upgrade_row_lock.test
```diff
@@ -0,0 +1,36 @@
...
+#       system table. This happens for both DD SE upgrade and DD version upgrade
```
laurynas-biveinis:
```suggestion
#       system tables. This happens for both DD SE upgrade and DD version upgrade
```

mysql-test/suite/rocksdb/t/ddse_upgrade_row_lock.test
```diff
@@ -0,0 +1,36 @@
...
+#       Currently some tables may involves 15-20K rows locks during insert data
```
laurynas-biveinis:
```suggestion
#       Currently some tables involve 15-20K rows locks during data insertion
```

laurynas-biveinis: Since I replied to an "obsolete" comment, it may not be visible, I'll repeat here: I think it's still better to shutdown first, then copy, or we may end up inadvertently testing it as a "hot backup" by cp, which is not always consistent. It's just the shutdown does not need a specific timeout value.

luqun:
next time, will use chatgpt to check.. :)

PR #1487, opened 2024-08-09
Snapshot time-related cleanups

laurynas-biveinis:
- Make m_earliest_snapshot_ts and m_read_opts protected fields in
  Rdb_transaction class. To replace
  their previous public uses, introduce new methods get_snapshot_timestamp &
  create_explicit_snapshot; use has_snapshot in more places.
  Remove unused return value for rdb_tx_acquire_snapshot.
- Factor out Rdb_transaction::on_finish for the common code in on_commit and
  on_rollback methods.
- Delete move constructor and move assignment operator for
  Rdb_compact_filter_factory class, replace "{}" with "= default" for other
  special members.
- Extract the gnarly regexes to make SHOW ENGINE ROCKSDB TRANSACTION STATUS
  output deterministic to a separate include file
  mysql-test/suite/rocksdb/include/show_transaction_status.inc

### DIFF
--- a/storage/rocksdb/ha_rocksdb.cc
+++ b/storage/rocksdb/ha_rocksdb.cc
@@ -3573,6 +3573,10 @@ class Rdb_transaction {
 
   std::shared_ptr<Rdb_snapshot_notifier> m_notifier;
 
+  rocksdb::ReadOptions m_read_opts[2];
+
+  std::int64_t m_snapshot_timestamp = 0;
+
   // This should be used only when updating binlog information.
   virtual rocksdb::WriteBatchBase *get_write_batch() = 0;
   virtual bool commit_no_binlog(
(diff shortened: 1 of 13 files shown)


storage/rocksdb/ha_rocksdb.cc
```diff
@@ -3904,10 +3910,16 @@ class Rdb_transaction {
   virtual void release_snapshot(TABLE_TYPE table_type) = 0;
 
   bool has_snapshot(TABLE_TYPE table_type) const {
+    assert(m_read_opts[INTRINSIC_TMP].snapshot == nullptr);
```
sunshine-Chun, line 3913: Hey Laurynas, I wonder why the snapshot for intrinsic tmp table is always null?

laurynas-biveinis: Temp tables are always private to the owning connection and don't need to participate in isolation concerns

PR #1488, opened 2024-08-19
Do not assume that std::int64_t is identical to long int

laurynas-biveinis:
- printf int64_t values with PRId64
- Specify std::make_pair template arguments. Alternatively we could introduce a
  constexpr variable for uninitialized opids.

This fixes new macOS build regressions, only one example per error class:

sql/xa.cc:382:11: error: format specifies type 'long' but the argument has type 'long long' [-Werror,-Wformat]
          info.smallest_lwm_opid.first, info.smallest_lwm_opid.second,
          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~

storage/innobase/trx/trx0sys.cc:271:16: error: invalid operands to binary expression ('const std::pair<int64_t, int64_t>' (aka 'const pair<long long, long long>') and 'pair<typename __unwrap_ref_decay<long>::type, typename __unwrap_ref_decay<long>::type>' (aka 'pair<long, long>'))
  if (lwm_opid != std::make_pair(-1L, -1L)) {
      ~~~~~~~~ ^  ~~~~~~~~~~~~~~~~~~~~~~~~

Squash with 5f045b49ed3a2b451773dcebdf7b179fad768045

### DIFF
--- a/sql/xa.cc
+++ b/sql/xa.cc
@@ -22,6 +22,7 @@
 
 #include "sql/xa.h"
 
+#include <cstdint>
 #include <memory>
 #include <new>
 #include <string>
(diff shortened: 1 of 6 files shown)


luqun:
@laurynas-biveinis , could you rebase this PR?

laurynas-biveinis:
@luqun , rebased

PR #1489, opened 2024-09-04
Enable file checksum for clone startup

sunshine-Chun:
Summary: This change is to enable file checksum check only for recipients of clone. For regular startup, we don't do the sst file checksum check.

Test Plan:

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/storage/rocksdb/clone/common.cc
+++ b/storage/rocksdb/clone/common.cc
@@ -235,6 +235,16 @@ void fixup_on_startup() {
   if (path_exists(current_datadir_in_progress_marker_path))
     rdb_fatal_error("In-progress clone marker found in the MyRocks datadir");
 
+  // Enable sst file checksum verification if
+  // CHECKSUMS_WRITE_AND_VERIFY_ON_CLONE flag is enabled and this restart is
+  // after a clone.
+  if (myrocks::rocksdb_file_checksums ==
+          myrocks::file_checksums_type::CHECKSUMS_WRITE_AND_VERIFY_ON_CLONE &&
+      temp_dir_exists_abort_if_not_dir(in_place_temp_datadir)) {
+    myrocks::rocksdb_file_checksums =
+        myrocks::file_checksums_type::CHECKSUMS_WRITE_AND_VERIFY;
+  }
+
   move_temp_dir_to_destination(in_place_temp_datadir, in_place_old_datadir,
                                rocksdb_datadir);
 
--- a/storage/rocksdb/ha_rocksdb.cc
+++ b/storage/rocksdb/ha_rocksdb.cc
@@ -919,12 +919,7 @@ static bool rocksdb_debug_skip_bloom_filter_check_on_iterator_bounds = 0;
 bool rocksdb_enable_autoinc_compact_mode = false;
 char max_timestamp_uint64[ROCKSDB_SIZEOF_TTL_RECORD];
 
-enum file_checksums_type {
-  CHECKSUMS_OFF = 0,
-  CHECKSUMS_WRITE_ONLY,
-  CHECKSUMS_WRITE_AND_VERIFY
-};
-static ulong rocksdb_file_checksums = file_checksums_type::CHECKSUMS_OFF;
+ulong rocksdb_file_checksums = file_checksums_type::CHECKSUMS_OFF;
 static std::time_t last_binlog_ttl_compaction_ts = std::time(nullptr);
 
 static std::atomic<uint64_t> rocksdb_row_lock_deadlocks(0);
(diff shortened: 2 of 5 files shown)


storage/rocksdb/clone/common.cc
```diff
@@ -101,6 +101,9 @@ void move_temp_dir_to_destination(const std::string &temp,
                              temp.c_str());
   }
 
+  myrocks::rocksdb_file_checksums =
```
laurynas-biveinis: This does not appear to be the best place for this. Why not in `fixup_on_startup` in the same file?

laurynas-biveinis (commented):
Is it possible to test this by adding a clone MTR test that corrupts a few bytes before the cloned instance startup?

storage/rocksdb/clone/common.cc
sunshine-Chun: Moved it to the fixup_on_startup before renaming temp datadir to datadir.

sunshine-Chun: Tested in test server. In the restart during clone, it triggers the checksum verification.

> 2024-09-04T14:05:04.364182-07:00 0 [Note] [MY-011071] [Server] Plugin rocksdb reported: 'Verifying file checksums...'
2024-09-04T14:05:04.364244-07:00 0 [Note] [MY-011071] [Server] Plugin rocksdb reported: '...done'

Restart the recipient instance after clone finished, no checksum verification is triggered cause there is no related log reported by plugin rocksdb.

storage/rocksdb/clone/common.cc
```diff
@@ -235,6 +235,16 @@ void fixup_on_startup() {
   if (path_exists(current_datadir_in_progress_marker_path))
     rdb_fatal_error("In-progress clone marker found in the MyRocks datadir");
 
+  // Enable sst file checksum verification if
+  // CHECKSUMS_WRITE_AND_VERIFY_ON_CLONE flag is enabled and this restart is
+  // after a clone.
+  if (myrocks::rocksdb_file_checksums ==
+          myrocks::file_checksums_type::CHECKSUMS_WRITE_AND_VERIFY_ON_CLONE &&
+      temp_dir_exists_abort_if_not_dir(in_place_temp_datadir)) {
```
laurynas-biveinis, line 243: Thank you, this is a better location IMHO.I I'd also remove `temp_dir_exists_abort_if_not_dir` check for clarity, but up to you

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -1171,7 +1166,7 @@ static TYPELIB read_free_rpl_typelib = {array_elements(read_free_rpl_names) - 1,
  * myrocks::rocksdb_file_checksum_type */
 static const char *file_checksums_names[] = {
     "CHECKSUMS_OFF", "CHECKSUMS_WRITE_ONLY", "CHECKSUMS_WRITE_AND_VERIFY",
-    NullS};
+    "CHECKSUMS_WRITE_AND_VERIFY_ON_CLONE", NullS};
```
laurynas-biveinis, line 1169: Are there tests that need re-recording? `main.mysqld--help-notwin`, maybe others?

storage/rocksdb/clone/common.cc
sunshine-Chun: If ```temp_dir_exists_abort_if_not_dir``` is removed, will the restart happens on every mysqld restart, not just the restart right after clone? The intended behavior is the checksum verification only happens for clone.

storage/rocksdb/ha_rocksdb.cc
sunshine-Chun: Doesn't seem so. The related MTR test passes. But I add one more MTR test for this change. It's included in the latest iteration.

storage/rocksdb/clone/common.cc
laurynas-biveinis: You are right, the current code is OK

mysql-test/suite/rocksdb_clone/t/checksum_verification_after_clone-master.opt
```diff
@@ -0,0 +1 @@
+--rocksdb-file-checksums=CHECKSUMS_WRITE_AND_VERIFY_ON_CLONE
```
laurynas-biveinis: Can this be moved to the testcase restart_parameter settings?

sunshine-Chun: Moved to restart parameter.

PR #1494, opened 2024-09-26
Assert that explicit transaction snapshot never overwrites an older one

laurynas-biveinis:
Explicit transaction snapshots are maintained by std::shared_ptr objects. When
the last shared pointer is destructed, the shared snapshot is released. Add
asserts that these snapshots are not held too long by checking that, whenever a
new explicit snapshot is assigned to a transaction, that it does not overwrite
an older existing one.

To be able to cover all the places and to clean up the code, make the
m_explicit_snapshot field protected instead of public, and add a few new methods
instead.

At the same time convert pointer to reference arguments, add [[nodiscard]],
shorten Rdb_explicit_snapshot::explicit_snapshot_mutex critical sections.

### DIFF
--- a/storage/rocksdb/ha_rocksdb.cc
+++ b/storage/rocksdb/ha_rocksdb.cc
@@ -3236,41 +3236,40 @@ static struct SYS_VAR *rocksdb_system_variables[] = {
 
 class Rdb_explicit_snapshot : public explicit_snapshot {
  public:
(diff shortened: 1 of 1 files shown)


luqun:
@laurynas-biveinis , could you do a rebase?

laurynas-biveinis:
@luqun , rebased

PR #1496, opened 2024-10-11
Fix LLVM 18 compilation errors

laurynas-biveinis:
- my_charset_file_reset_charset: placement-new construct an new default object
  instead of memset'ing it to zero, fixing [1]
- strings/conf_to_src.cc: do not memset an object that is already
  default-initialized to zero, fixing [2]

Squash with 6637b51c7df3fb8e2c4994dd0358acb92f7ba017

[1]:

strings/ctype.cc: In function ‘void my_charset_file_reset_charset(MY_CHARSET_FILE*)’:
strings/ctype.cc:303:9: error: ‘void* memset(void*, int, size_t)’ clearing an object of type ‘struct CHARSET_INFO’ with no trivial copy-assignment; use value-initialization instead [-Werror=class-memaccess]
  303 |   memset(&i->cs, 0, sizeof(i->cs));
      |   ~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from strings/ctype.cc:36:
include/m_ctype.h:383:8: note: ‘struct CHARSET_INFO’ declared here
  383 | struct CHARSET_INFO {
      |        ^~~~~~~~~~~~

[2]:

strings/conf_to_src.cc: In function ‘int main(int, char**)’:
strings/conf_to_src.cc:292:9: error: ‘void* memset(void*, int, size_t)’ clearing an object of type ‘struct CHARSET_INFO’ with no trivial copy-assignment; use value-initialization instead [-Werror=class-memaccess]
  292 |   memset(&ncs, 0, sizeof(ncs));
      |   ~~~~~~^~~~~~~~~~~~~~~~~~~~~~
In file included from strings/conf_to_src.cc:32:
include/m_ctype.h:383:8: note: ‘struct CHARSET_INFO’ declared here
  383 | struct CHARSET_INFO {
      |        ^~~~~~~~~~~~

### DIFF
--- a/mysys/charset.cc
+++ b/mysys/charset.cc
@@ -226,9 +226,10 @@ int MY_CHARSET_LOADER::add_collation(CHARSET_INFO *cs) {
        (cs->number = get_collation_number_internal(cs->m_coll_name))) &&
       cs->number < array_elements(all_charsets)) {
     if (!all_charsets[cs->number]) {
-      if (!(all_charsets[cs->number] = (CHARSET_INFO *)my_once_alloc(
-                sizeof(CHARSET_INFO), MYF(MY_ZEROFILL))))
+      if (!(all_charsets[cs->number] =
+                (CHARSET_INFO *)my_once_alloc(sizeof(CHARSET_INFO), MYF(0))))
         return MY_XML_ERROR;
+      new (all_charsets[cs->number]) CHARSET_INFO;
     } else if (all_charsets[cs->number]->state & MY_CS_COMPILED) {
       // Disallow overwriting compiled character sets
       clear_cs_info(cs);
--- a/strings/conf_to_src.cc
+++ b/strings/conf_to_src.cc
@@ -289,7 +289,6 @@ int main(int argc, char **argv [[maybe_unused]]) {
     exit(EXIT_FAILURE);
   }
 
-  memset(&ncs, 0, sizeof(ncs));
   memset(&all_charsets, 0, sizeof(all_charsets));
 
   sprintf(filename, "%s/%s", argv[1], "Index.xml");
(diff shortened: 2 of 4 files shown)


strings/conf_to_src.cc
```diff
@@ -289,7 +289,6 @@ int main(int argc, char **argv [[maybe_unused]]) {
     exit(EXIT_FAILURE);
   }
 
-  memset(&ncs, 0, sizeof(ncs));
```
bladepan, line 292: should also change CHARSET_INFO to explicitly set everything to 0 in field initialization?

laurynas-biveinis: Yeah, the current patch is incorrect due to doing default-initialization leaving fields uninitialized, not zero-initialized.

laurynas-biveinis:
Forgot to reset the Draft state of this PR :( It is ready for review

mysys/charset.cc
```diff
@@ -226,9 +226,10 @@ int MY_CHARSET_LOADER::add_collation(CHARSET_INFO *cs) {
        (cs->number = get_collation_number_internal(cs->m_coll_name))) &&
       cs->number < array_elements(all_charsets)) {
     if (!all_charsets[cs->number]) {
-      if (!(all_charsets[cs->number] = (CHARSET_INFO *)my_once_alloc(
-                sizeof(CHARSET_INFO), MYF(MY_ZEROFILL))))
+      if (!(all_charsets[cs->number] =
```
bladepan, line 229: my_once_alloc does not seem to do any initialization

laurynas-biveinis: I added a placement new call below

PR #1497, opened 2024-10-14
Introduce statement snapshot type for Rdb_transaction

laurynas-biveinis:
Previously, the snapshot state was tracked by a combination of
flags (m_is_delayed_snapshot, m_tx_read_only), and by certain snapshot pointers
being null or non-null. (i.e. m_read_opts[USER_TABLE].snapshot,
m_explicit_snapshot). The transaction state specified by different combinations
of the above was not immediately obvious and the purpose of the read only flag
was not the RO-ness as such but rather using the DB instead of Transaction API
to acquire snapshots.

Consolidate the states into an enum class snapshot_type { ... }
statement_snapshot_type field, consuming the two flags above. Introduce and
assert invariant across the state values and snapshot pointers being null or
non-null. Add to the invariant of AC-NL-RO-RC transactions. Rename
snapshot_created to assign_snapshot to better reflect what the method does (as
it does not, in fact, take a newly-created snapshot i.e. in the case of existing
explicit snapshot). Make share_explicit_snapshot and create_explicit_snapshot
set the m_read_opts snapshot and remove the redundant code from the callers. For
the latter, make it create the snapshot instead of taking the m_read_opts one.

Two of the possible states are EXPLICIT and READ_ONLY_TRX, and the existing code
implicitly had another state which was the combination of the two. Avoid
introducing such new state, and use EXPLICIT in it too. This results in a minor
user-visible change that START TRANSACTION WITH SHARED|EXISTING SNAPSHOT
transactions are grouped with CREATE|ATTACH EXPLICIT SNAPSHOT ones instead of
START TRANSACTION WITH CONSISTENT SNAPSHOT for the purposes of
INFORMATION_SCHEMA read-only flag and diagnostics on write attempts. Update the
diagnostic messages to reflect this.

### DIFF
--- a/storage/rocksdb/ha_rocksdb.cc
+++ b/storage/rocksdb/ha_rocksdb.cc
@@ -3658,7 +3658,6 @@ class Rdb_transaction {
   ulonglong m_row_lock_count = 0;
   std::unordered_map<GL_INDEX_ID, ulonglong> m_auto_incr_map;
 
-  bool m_is_delayed_snapshot = false;
   bool m_is_two_phase = false;
 
  private:
@@ -3687,14 +3686,63 @@ class Rdb_transaction {
 
   Rdb_io_perf *m_tbl_io_perf;
 
-  bool m_tx_read_only = false;
-
   int m_timeout_sec = 0; /* Cached value of @@rocksdb_lock_wait_timeout */
 
   /* Maximum number of locks the transaction can have */
   ulonglong m_max_row_locks;
 
   bool m_rollback_only = false;
+
+  enum class snapshot_type {
+    NONE,
+    // A snapshot created through Tranaction API for regular transactions and by
+    // DB::GetSnapshot() for WB ones
+    CURRENT,
+    CURRENT_DELAYED,
+    // This is used by transactions started with "START TRANSACTION WITH
+    // CONSISTENT [ROCKSDB] SNAPSHOT". The snapshot has to be created via
+    // DB::GetSnapshot(), not via Transaction API.
+    READ_ONLY_TRX,
+    // Explicit snapshot requests trump everything else and are also created via
+    // DB::GetSnapshot()
+    EXPLICIT
+  } statement_snapshot_type{snapshot_type::NONE};
+
+  void assert_snapshot_invariants() const noexcept {
+#ifndef NDEBUG
+    assert(m_read_opts[INTRINSIC_TMP].snapshot == nullptr);
+
+    switch (statement_snapshot_type) {
+      case snapshot_type::NONE:
+      case snapshot_type::CURRENT_DELAYED:
+        assert(m_explicit_snapshot == nullptr);
+        assert(m_read_opts[USER_TABLE].snapshot == nullptr);
+        break;
+      case snapshot_type::CURRENT:
+        assert(m_explicit_snapshot == nullptr);
+        assert(m_read_opts[USER_TABLE].snapshot != nullptr);
+        break;
+      case snapshot_type::READ_ONLY_TRX:
+        assert(m_explicit_snapshot == nullptr);
+        break;
+      case snapshot_type::EXPLICIT:
+        assert(m_explicit_snapshot != nullptr);
+        assert(m_read_opts[USER_TABLE].snapshot ==
+               m_explicit_snapshot->get_snapshot()->snapshot());
+        break;
+    }
+#endif
+  }
+
+#ifndef NDEBUG
+ public:
+  [[nodiscard]] bool is_delayed_snapshot() const noexcept {
+    assert_snapshot_invariants();
+    return statement_snapshot_type == snapshot_type::CURRENT_DELAYED;
+  }
+
+ protected:
+#endif
 
   std::shared_ptr<Rdb_snapshot_notifier> m_notifier;
 
@@ -3717,6 +3765,11 @@ class Rdb_transaction {
    */
   [[nodiscard]] rocksdb::Status merge_auto_incr_map(
       rocksdb::WriteBatchBase &wb) {
+    assert(statement_snapshot_type == snapshot_type::NONE ||
+           statement_snapshot_type == snapshot_type::CURRENT ||
+           statement_snapshot_type == snapshot_type::CURRENT_DELAYED ||
+           statement_snapshot_type == snapshot_type::EXPLICIT);
+
     // Iterate through the merge map merging all keys into data dictionary.
     rocksdb::Status s;
     for (auto &it : m_auto_incr_map) {
@@ -3731,6 +3784,8 @@ class Rdb_transaction {
   }
 
   void reset_flags() {
+    assert(statement_snapshot_type == snapshot_type::NONE);
+
     m_bulk_index_transaction = false;
     m_dd_transaction = false;
   }
@@ -3914,6 +3969,8 @@ class Rdb_transaction {
   }
 
   [[nodiscard]] std::int64_t get_snapshot_ts() const {
+    assert_snapshot_invariants();
+
     const auto *const snapshot = m_read_opts[USER_TABLE].snapshot;
     const auto result = snapshot != nullptr ? snapshot->GetUnixTime() : 0;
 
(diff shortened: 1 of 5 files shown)


luqun:
could you take a look main.subselect_debug 'rocksdb_intrinsic_table' MTR? The following assert failed

storage/rocksdb/ha_rocksdb.cc:4976: std::unique_ptr<rocksdb::Iterator> myrocks::Rdb_transaction::get_iterator(rocksdb::ColumnFamilyHandle &, bool, const rocksdb::Slice &, const rocksdb::Slice &, myrocks::TABLE_TYPE, bool, bool): Assertion `statement_snapshot_type != snapshot_type::NONE || create_snapshot' failed.

laurynas-biveinis:
@luqun , fixed, the assert did not accept temp tables. Also rebased, ready for review

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -3687,15 +3686,66 @@ class Rdb_transaction {
...
 
+  enum class snapshot_type {
+    NONE,
+    // A snapshot created through Tranaction API for regular transactions and by
+    // DB::GetSnapshot() for WB ones
+    CURRENT,
+    CURRENT_DELAYED,
+    // This is used by transactions started with "START TRANSACTION WITH
+    // CONSISTENT [ROCKSDB] SNAPSHOT". The snapshot has to be created via
+    // DB::GetSnapshot(), not via Transaction API. */
```
luqun: nit: remove */

laurynas-biveinis: @luqun, fixed, ready for review

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -3687,15 +3686,66 @@ class Rdb_transaction {
...
+  enum class snapshot_type {
+    NONE,
+    // A snapshot created through Tranaction API for regular transactions and by
+    // DB::GetSnapshot() for WB ones
+    CURRENT,
+    CURRENT_DELAYED,
+    // This is used by transactions started with "START TRANSACTION WITH
+    // CONSISTENT [ROCKSDB] SNAPSHOT". The snapshot has to be created via
+    // DB::GetSnapshot(), not via Transaction API.
+    READ_ONLY_TRX,
```
luqun, line 3705: From rocksdb_explicit_snapshot/create_explicit_snapshot, Looks like EXPLICIT snapshot also create via DB::GetSnapshot(). maybe add a comment for EXPLICIT

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -5060,26 +5171,35 @@ class Rdb_transaction {
...
   }
 
-  /*
-    This is used by transactions started with "START TRANSACTION WITH "
-    "CONSISTENT [ROCKSDB] SNAPSHOT". When tx_read_only is turned on,
-    snapshot has to be created via DB::GetSnapshot(), not via Transaction
-    API.
-  */
-  bool is_tx_read_only() const { return m_tx_read_only; }
+  [[nodiscard]] bool is_tx_read_only() const noexcept {
```
luqun, line 5182: how about add _snapshot into method name?

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -15163,7 +15362,8 @@ int ha_rocksdb::external_lock(THD *const thd, int lock_type) {
         DBUG_RETURN(HA_ERR_UNSUPPORTED);
       }
 
-      if (thd->get_explicit_snapshot()) {
+      if (unlikely(tx->has_explicit_snapshot() ||
```
luqun, line 15368: why do we need check both(tx and thd)?

laurynas-biveinis: The state lifetimes don't fully match but only overlap. For instance,
```
CREATE EXPLICIT ROCKSDB SNAPSHOT;
INSERT INTO T1 VALUES(); # tx state not explicit, thd flag = true
```
And the opposite case:
```
START TRANSACTION WITH SHARED ROCKSDB SNAPSHOT;
INSERT INTO T1 VALUES(); # tx state explicit, thd flag = false
```

The tx state is set during transaction execution. The thd snapshot is set on
CREATE/ATTACH ROCKSDB SNAPSHOT, when there is no tx object.

I'm adding a clarifying comment.

storage/rocksdb/ha_rocksdb.cc
laurynas-biveinis: It is the whole transaction that is read only

laurynas-biveinis:
@luqun , ready for review

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -3687,15 +3686,67 @@ class Rdb_transaction {
...
+        assert(m_read_opts[USER_TABLE].snapshot == nullptr);
+        break;
+      case snapshot_type::CURRENT:
+        assert(m_explicit_snapshot == nullptr);
+        assert(m_read_opts[USER_TABLE].snapshot != nullptr);
+        break;
+      case snapshot_type::CURRENT_DELAYED:
+        assert(m_explicit_snapshot == nullptr);
+        assert(m_read_opts[USER_TABLE].snapshot == nullptr);
+        break;
```
hermanlee: Since these are all the same, it might be simpler to combine them all. Logically, it means that even though these are slightly different snapshot cases, they follow the same invariantsonce the snapshot is allocated.

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -4059,9 +4122,30 @@ class Rdb_transaction {
...
+      case snapshot_type::EXPLICIT:
+        assert(snapshot ==
+               m_explicit_snapshot->get_snapshot()->snapshot());
+        break;
+      case snapshot_type::NONE:
+        assert(false);
+        __builtin_unreachable();
+        break;
+    }
+
```
hermanlee: Maybe have this whole block be under NDEBUG, and pull out the case for changing the snapshot type from  CURRENT_DELAYED to CURRENT?

storage/rocksdb/ha_rocksdb.cc
```diff
@@ -4082,28 +4167,43 @@ class Rdb_transaction {
   virtual rocksdb::Slice *get_tx_read_timestamp_slice() = 0;
 
   bool has_snapshot(TABLE_TYPE table_type) const {
-    assert(m_read_opts[INTRINSIC_TMP].snapshot == nullptr);
+    assert_snapshot_invariants();
     if (table_type == INTRINSIC_TMP) return false;
-    return m_read_opts[table_type].snapshot != nullptr;
+    return m_read_opts[USER_TABLE].snapshot != nullptr;
```
hermanlee, line 4169: Probably assert `table_type == USER_TABLE`

hermanlee (commented):
Overall looks good to me.

storage/rocksdb/ha_rocksdb.cc
laurynas-biveinis: Will do (but only for the 1st and the 3rd: the 2nd has the 2nd condition check inverted)

storage/rocksdb/ha_rocksdb.cc
laurynas-biveinis: OK, also merging the case branches like for the previous comment

laurynas-biveinis: And removing `__builtin_unreachable()` because it no longer has effect in debug-only code

laurynas-biveinis:
Ready for review again

luqun:
rocksdb_rpl.rpl_skip_trx_api_binlog_format MTR failed with 
  virtual void myrocks::Rdb_writebatch_impl::acquire_snapshot(bool, TABLE_TYPE): Assertion `statement_snapshot_type == snapshot_type::CURRENT' failed.

laurynas-biveinis:
> rocksdb_rpl.rpl_skip_trx_api_binlog_format MTR failed with virtual void myrocks::Rdb_writebatch_impl::acquire_snapshot(bool, TABLE_TYPE): Assertion `statement_snapshot_type == snapshot_type::CURRENT' failed.

@luqun, I am unable to reproduce. Do you have more details?

PR #1501, opened 2024-10-24
Skip drop system tables during clone drop user data

luqun:
For MyRocks DDSE, its system table doesn't have valid tablespace id, thus these system tables(such as mysql.user) in the recipient instance will be droped during drop user data stage. if at same time,  some automations may check the instance and found out the instance doesn't have these important tables. 

The change is to use table name to check whether these table are DD/systems tables

### DIFF
--- a/mysql-test/suite/rocksdb_clone/t/ddse_debug.test
+++ b/mysql-test/suite/rocksdb_clone/t/ddse_debug.test
@@ -1,0 +1,47 @@
+--source include/have_rocksdb.inc
+--source include/have_debug.inc
+--source include/have_debug_sync.inc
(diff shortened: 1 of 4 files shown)


mysql-test/suite/rocksdb_clone/t/ddse_debug.test
```diff
@@ -0,0 +1,47 @@
+--source include/have_rocksdb.inc
```
laurynas-biveinis, line 1: Since the test has remote clone-specific cnf file, suggest to rename to remote_ddse_debug.

Otherwise LGTM!

PR #1503, opened 2024-10-24
MyRocks: cleanup Rdb_convert_to_record_key_decoder class

laurynas-biveinis:
- Since this class is a non-instantiable collection of static methods, delete
  the rest of special methods (default constructor, move constructor, move
  assignment operator, destructor) in addition to the currently deleted ones.
- Convert method arguments from pointers to references and propagate this up and
  down the call hierarchy. Most notably this converts many "TABLE *" uses to
  "const TABLE &" throughout MyRocks, deletes a bit of dead code, converts some
  constructor assignments to initializer list items.
- For touched methods, add [[nodiscard]] as applicable and remove redundant
  const from parameters, do other minor cleanups.
- Move struct Rdb_unpack_func_context definition from rdb_datadic.h to
  rdb_datadic.cc leaving a forward declaration in the header file.

### DIFF
--- a/storage/rocksdb/ha_rocksdb.cc
+++ b/storage/rocksdb/ha_rocksdb.cc
@@ -4382,7 +4382,7 @@ class Rdb_transaction {
 
   int finish_bulk_load(bool *is_critical_error = nullptr,
                        bool print_client_error = true,
-                       TABLE *table_arg = nullptr) {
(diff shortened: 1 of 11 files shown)


laurynas-biveinis:
@luqun , @hermanlee , rebased and ready for review again

PR #1509, opened 2024-12-17
Introduce rdb_snapshot_unique_ptr & make rdb_get_rocksdb_db return reference

laurynas-biveinis:
This is a second batch of cleanups from the repeatable read snapshot work.

- Make rdb_get_rocksdb_db return reference instead of pointer, update callers
  and callees. Inline this function in ha_rocksdb_proto.h, move rdb declaration
  there to support inlining in detail namespace as documentation that it should
  not be accessed directly. Update ha_rocksdb.cc itself to use
  rdb_get_rocksdb_db instead of rdb. Remove many redudant rdb != nullptr checks
  and asserts.
- Introduce rdb_snapshot_unique_ptr with a custom deleter and a factory function
  get_rdb_snapshot to manager global RocksDB snapshots
- Make Rdb_explicit_snapshot create the snapshot itself, by calling
  get_rdb_snapshot, instead of receiving one through parameters, simplify its
  signature.
- Replace rocksdb::ManagedSnapshot uses with rdb_snapshot_unique_ptr ones.
- Remove rarely-used explicit transaction snapshot assignment code from
  acquire_snapshot and move it to the few callers where it may happen.
  Add new method Rdb_transaction::has_explicit_or_read_only_snapshot to support
  this.
- For Rdb_transaction, make m_insert_count, m_update_count, m_delete_count,
  m_auto_incr_map, & m_rollback_only fields private instead of protected. Move
  their reset at the end of committed or rolledback transactions to on_finish
  method in the base class. Remove redundant m_writes_at_last_savepoint reset in
  set_initial_savepoint.
- Add several asserts.

### DIFF
--- a/storage/rocksdb/clone/donor.cc
+++ b/storage/rocksdb/clone/donor.cc
@@ -614,8 +614,8 @@ donor::donor(const myrocks::clone::locator &l, const uchar *&loc,
 
 donor::~donor() {
   if (m_rdb_file_deletes_disabled) {
-    auto *const rdb = myrocks::rdb_get_rocksdb_db();
-    const auto result = rdb->EnableFileDeletions();
+    auto &rdb = myrocks::rdb_get_rocksdb_db();
+    const auto result = rdb.EnableFileDeletions();
     if (!result.ok()) {
       myrocks::rdb_log_status_error(result,
                                     "RocksDB file deletion re-enable failed");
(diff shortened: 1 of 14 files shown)


laurynas-biveinis:
Needs rebasing

laurynas-biveinis:
@luqun , rebased  & ready for review

PR #1510, opened 2025-01-06
Rr snapshot

laurynas-biveinis:


laurynas-biveinis:
Creating a draft PR because it contains a prerequisite commits, which is PR'ed separately. Will be rebased after the prerequisite is merged

PR #1511, opened 2025-01-10
Make parts of Rdb_transaction objects safe to access from other threads

laurynas-biveinis:
Almost all accesses to Rdb_transaction objects are from the owning
query-executing thread, but some fields are read from other threads, for
example, to execute SHOW ENGINE ROCKSDB STATUS, causing intermittent crashes due
to e.g. wild snapshot pointer read to get its timestamp.

Fix by moving all shared fields to the beginning of the class, documenting their
protection and making atomic as needed. For atomic fields, use only relaxed
memory order accesses, which should result in the same compiled code as before.
For the cases where something is read through a pointer, cache that data instead
in an atomic field and don't dereference the pointer from other threads:
Rdb_transaction::m_snapshot_ts instead of
m_read_opts[USER_TABLE].snapshot->GetUnixTime() and m_num_ongoing_bulk_load
instead of m_bulk_load_ctx->num_bulk_load().

### DIFF
--- a/storage/rocksdb/ha_rocksdb.cc
+++ b/storage/rocksdb/ha_rocksdb.cc
@@ -3653,22 +3653,78 @@ Rdb_transaction *Rdb_transaction_list::head{nullptr};
 Rdb_transaction *Rdb_transaction_list::tail{nullptr};
 Rdb_transaction *Rdb_transaction_list::newest_tx_with_snapshot{nullptr};
 
+// Bump an atomic counter that can be read from multiple threads but written
+// only by one thread. More efficient than
+// fetch_add(1, std::memory_order_relaxed);
+template <typename T>
+void private_ctr_inc(T &ctr) noexcept {
+  ctr.store(ctr.load(std::memory_order_relaxed) + 1, std::memory_order_relaxed);
+}
+
+// Decrement an atomic counter that can be read from multiple threads but
+// written only by one thread. More efficient than
+// fetch_sub(1, std::memory_order_relaxed);
+template <typename T>
+void private_ctr_dec(T &ctr) noexcept {
+  ctr.store(ctr.load(std::memory_order_relaxed) - 1, std::memory_order_relaxed);
+}
+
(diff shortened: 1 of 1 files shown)


storage/rocksdb/ha_rocksdb.cc
```diff
@@ -6134,10 +6225,12 @@ class Rdb_writebatch_impl : public Rdb_transaction {
     }
     assert_snapshot_invariants();
     if (has_snapshot(table_type)) {
-      assert(statement_snapshot_type == snapshot_type::CURRENT);
-      statement_snapshot_type = snapshot_type::NONE;
+      assert(get_statement_snapshot_type() == snapshot_type::CURRENT);
+      statement_snapshot_type.store(snapshot_type::NONE,
+                                    std::memory_order_relaxed);
       rdb->ReleaseSnapshot(m_read_opts[table_type].snapshot);
       m_read_opts[table_type].snapshot = nullptr;
```
luqun, line 6232: For line 6231 and line 6232, may also not safe for other threads(use after delete issue)

current thread finished execute "rdb->ReleaseSnapshot()" but before execute "m_read_opts[table_type].snapshot = nullptr", while another thread is calling tx->get_snapshot_ts();(https://github.com/facebook/mysql-5.6/blob/fb-mysql-8.0.32/storage/rocksdb/ha_rocksdb.cc#L3987). the snapshot has been cleared but m_read_opts[USER_TABLE].snapshot may still reference old memory


maybe change to

```
      m_read_opts[table_type].snapshot = nullptr; 
      rdb->ReleaseSnapshot(m_read_opts[table_type].snapshot);
```

laurynas-biveinis: `get_snapshot_ts` does not race here, because this PR introduces a caching field exactly for this scenario.

Having said that, `m_read_opts[table_type].snapshot` is indeed accessed incorrectly in other contexts (see my e-mail) and will be fixed one way or another in a follow-up

